PR #125, opened 2026-04-01
test: add RateLimiter boundary tests

kipra19821810-cloud:
## What does this PR do?

Adds boundary value tests for `RateLimiter` in `tests/unit/test_rate_limiter.py`. This serves as a regression test for the v0.3.3 ZeroDivisionError fix and covers edge cases.

## Test cases (11 tests)

- `requests_per_second=0` — must not crash or sleep (ZeroDivisionError regression)
- - `requests_per_second=-1` — negative values disable limiting
- - `requests_per_second=1` — basic throttling: first call no sleep, second call sleeps ~1s
- - No sleep after interval has naturally elapsed
- - Very large rps (1,000,000) — sub-millisecond intervals
- - Rapid successive calls — each respects the interval
- - `_last` state management — updated correctly, starts at 0.0, excluded from repr
## Related Issue

Closes #101

PR #144, opened 2026-04-01
test: add fractional requests_per_second test for RateLimiter

Pranavv157:
## What does this PR do?

Adds a test case for fractional `requests_per_second` values in RateLimiter.

## Why?

Existing tests cover 0, 1, and large values, but fractional values like 2.5 are not tested. This ensures correct behavior for non-integer rates.

## Changes

- Added test for `requests_per_second=2.5`
- Verifies correct sleep interval (~0.4 seconds)

### DIFF
--- a/tests/unit/test_rate_limiter.py
+++ b/tests/unit/test_rate_limiter.py
@@ -107,6 +107,18 @@ class TestRapidSuccessiveCalls:
             assert abs(call[0][0] - 0.5) < 1e-9
 
 
+class TestFractionalRps:
+    @patch("drt.destinations.rate_limiter.time.sleep")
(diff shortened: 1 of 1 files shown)


masukai (changes requested):
Thanks for adding the fractional rps test — great complement to #125! 🙌

A couple of things to address before merging:

1. **Rebase onto main** — #125 has been merged, so this PR now conflicts on `tests/unit/test_rate_limiter.py`. Please rebase and add your test to the existing file instead of creating a new one.
2. **Revert `conftest.py` change** — The `#type:ignore` addition on the pytest import isn't related to this PR, please remove it.
3. **Add newline at end of file** — Minor, but missing trailing newline.

The test itself looks correct 👍 Should be quick to fix!

Issue #275, opened 2026-04-13
feat: structured JSON logging for sync execution


masukai:
## Summary

Add a \`--log-format json\` flag to \`drt run\` that outputs structured JSON logs.

## Why

For production deployments, structured logs are essential for integration with log aggregation tools (Datadog, CloudWatch, etc.). Currently drt outputs plain text logs.

## What to do

1. Add a \`--log-format\` flag to the \`drt run\` command in \`drt/cli/main.py\`
   - Options: \`text\` (default, current behavior) and \`json\`
2. When \`json\` is selected, configure Python's \`logging\` to output JSON lines
3. Each log line should include: \`ts\`, \`level\`, \`sync\`, \`msg\`
4. On sync completion, include: \`rows\`, \`duration_ms\`, \`status\`

## Example output

\`\`\`json
{"ts": "2026-04-13T10:00:00Z", "level": "INFO", "sync": "hubspot_contacts", "msg": "sync_started"}
{"ts": "2026-04-13T10:00:03Z", "level": "INFO", "sync": "hubspot_contacts", "msg": "sync_complete", "rows": 142, "duration_ms": 3200}
\`\`\`

## Acceptance criteria

- [ ] \`drt run my_sync.yaml --log-format json\` outputs JSON lines
- [ ] \`drt run my_sync.yaml\` (default) outputs text as before
- [ ] Unit test covering JSON log output

## Difficulty

⭐⭐ Easy — CLI flag + logging config, 1-3 files

## Reference

- Current CLI code: \`drt/cli/main.py\`
- Python \`logging\` JSON formatter: [python-json-logger](https://github.com/madzak/python-json-logger) or a simple custom formatter
- Similar pattern: how dbt handles \`--log-format json\`

PR #330, opened 2026-04-15
feat: add --log-format json flag to drt run

Khush-domadia:
## Summary
Implements #275 — adds `--log-format json` flag to `drt run` for structured 
JSON logging, enabling integration with log aggregators like Datadog and CloudWatch.

## Changes
- `drt/cli/main.py` — added `--log-format` flag and `_configure_json_logging()` 
  using stdlib `logging` only (no new dependencies)
- `tests/unit/test_json_logging.py` — 4 unit tests covering JSON output format

## Acceptance Criteria
- [x] `drt run my_sync.yaml --log-format json` outputs JSON lines
- [x] `drt run my_sync.yaml` (default) outputs text as before  
- [x] Unit tests covering JSON log output (4/4 passing)

## Testing
uv run pytest tests/unit/test_json_logging.py -v
# 4 passed

### DIFF
--- a/drt/cli/main.py
+++ b/drt/cli/main.py
@@ -88,9 +88,7 @@ class _JsonFormatter(logging.Formatter):
     """Emit each log record as a single JSON object (JSON Lines format)."""
 
     def format(self, record: logging.LogRecord) -> str:
-        ts = datetime.fromtimestamp(record.created, tz=timezone.utc).strftime(
-            "%Y-%m-%dT%H:%M:%SZ"
-        )
+        ts = datetime.fromtimestamp(record.created, tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
         payload: dict[str, object] = {
             "ts": ts,
             "level": record.levelname,
@@ -108,7 +106,7 @@ def _configure_json_logging() -> None:
     handler = logging.StreamHandler()
     handler.setFormatter(_JsonFormatter())
     logging.root.handlers = [handler]
-    logging.root.setLevel(logging.DEBUG)
+    logging.root.setLevel(logging.INFO)
 
 
 def _resolve_profile_name(cli_flag: str | None, project_profile: str) -> str:
@@ -258,8 +256,13 @@ def run(
         None, "--profile", "-p", help="Override profile (default: drt_project.yml or DRT_PROFILE)."
     ),
     log_format: str = typer.Option(
-        "text", "--log-format", help="Log format: text or json (structured JSON lines)."
+    "text",
+    "--log-format",
+    help=(
+        "Log format: text or json (structured JSON lines for each sync event"
+        " — separate from --output json which controls the final result)."
     ),
+),
 ) -> None:
     """Run sync(s) defined in the project."""
     import json as json_mod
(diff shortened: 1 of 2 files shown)


masukai (changes requested):
Thanks @Khush-domadia! 🎉 The \`_JsonFormatter\` approach is clean, and JSONLines is the right format for log aggregators.

A few things to address before merging:

1. **Ruff errors** — 3 fixes needed (I001 import order, F401 unused pytest, E501 long line). Run locally:
   \`\`\`bash
   ruff check --fix drt/ tests/
   ruff format drt/ tests/
   \`\`\`

2. **\`logging.root.setLevel(logging.DEBUG)\` is too broad** — this will flood output with debug logs from \`httpx\`, \`google-cloud-*\`, etc. when JSON mode is on. Please change to \`logging.INFO\` (or make it configurable).

3. **Flag naming clarity** — \`drt run\` now has both \`--output json\` (final result output) and \`--log-format json\` (per-event logs). To reduce user confusion, please expand the \`--log-format\` help text to clarify the distinction, e.g.:
   > "Log format: text or json (structured JSON lines for each sync event — separate from --output json which controls the final result)."

Happy to merge once those are addressed!

masukai (changes requested):
Thanks @Khush-domadia — the three items from my first review are all addressed, nice work! ✅

One more thing I caught on this pass that would make the JSON output unusable in practice:

### 🔴 Extra-field extraction includes standard LogRecord attributes

```python
for key, value in record.__dict__.items():
    if key not in logging.LogRecord.__dict__ and not key.startswith("_"):
```

\`logging.LogRecord.__dict__\` only has **class-level** attributes (methods, descriptors). Standard instance attributes set by \`__init__\` — \`name\`, \`msg\`, \`pathname\`, \`lineno\`, \`args\`, \`exc_info\`, \`created\`, etc. — aren't in the class dict, so they all leak into the JSON payload as false "extras."

Result: every log line looks like:
```json
{"ts":"...","level":"INFO","msg":"sync_complete","sync":"foo",
 "name":"root","pathname":"main.py","lineno":42,"args":null,...}
```

Fix — build an exclusion set from a default LogRecord instance:

```python
_STANDARD_LOG_FIELDS = frozenset(
    vars(logging.LogRecord("", 0, "", 0, "", (), None))
)
```

Then:
```python
if key not in _STANDARD_LOG_FIELDS and not key.startswith("_"):
```

### 🟡 Minor (not blocking, but nice to have)

- Constrain \`--log-format\` to \`["text", "json"]\` via \`typer.Option(...)\` choices — currently any string silently falls back to text
- Test cleanup: snapshot/restore \`logging.root.handlers\` and \`logging.root.level\` in each test to avoid leaking state

Once the LogRecord fix lands, happy to merge 🚀

masukai (approved):
🎉 Clean implementation of structured JSON logging — the `_JsonFormatter` with `_STANDARD_LOG_FIELDS` exclusion set is the right approach, and the test cleanup (snapshot/restore of logging state) shows real attention to detail.

Appreciated the quick turnaround on all three review rounds!

Thanks for the contribution @Khush-domadia! If you enjoy using drt, a ⭐ on the repo would be appreciated 🙏

Looking for more to work on? Check out our good first issues:
https://github.com/drt-hub/drt/issues?q=is%3Aopen+label%3A%22good+first+issue%22

PR #335, opened 2026-04-16
feat: add --all, --select tag:, --threads for parallel sync execution

### LINKED ISSUE #288: feat: multi-sync orchestration — drt run --all
## Summary

Add \`drt run --all\` to execute all syncs in a project directory, with optional dependency ordering.

## Motivation

Production deployments typically have multiple syncs. Running them one by one is tedious. \`drt run --all\` is the equivalent of \`dbt run\` — it's expected functionality for production use.

## Scope

- \`drt run --all\`: discover and run all \`*.yaml\` syncs in the project
- \`drt run --select tag:crm\`: run syncs matching a tag/selector
- Parallel execution with configurable concurrency (\`--threads N\`)
- Summary report: N syncs succeeded, M failed, total duration
- Exit code: non-zero if any sync failed

Pawansingh3889:
## Summary

Closes #288. Adds multi-sync orchestration to `drt run`.

## Changes

**CLI (`drt/cli/main.py`):**
- `--all` flag for explicit "run everything" intent
- `--select tag:crm` filters syncs by the new `tags` field
- `--threads N` runs syncs in parallel via `ThreadPoolExecutor`
- Summary report after multi-sync runs: "N succeeded, M failed, Xs total"
- Non-zero exit if any sync fails

**Model (`drt/config/models.py`):**
- Added `tags: list[str]` to `SyncConfig` (default empty, backward compatible)

## Usage

```bash
# Run all syncs (existing behaviour, unchanged)
drt run

# Explicit --all
drt run --all

# Filter by tag
drt run --select tag:crm

# Parallel execution
drt run --threads 4

# Combined
drt run --select tag:crm --threads 2 --verbose
```

## YAML example

```yaml
name: sync_users
tags: [crm, daily]
model: ref('users')
destination:
  type: postgres
  ...
```

## Backward compatibility

- `tags` defaults to empty list — existing syncs without tags work unchanged
- Without `--select` or `--all`, behaviour is identical to before
- `--threads 1` (default) runs sequentially, same as before

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Parallel sync execution** (#335): New `drt run --threads N` runs configured syncs in a `ThreadPoolExecutor`. `--select tag:<name>` filters by sync tag; `--select *` or `--select all` are explicit "run every sync" sentinels. `StateManager` now takes a process-local `threading.Lock` so concurrent `save_sync` calls from worker threads don't lose each other's writes.
 - **SQL Server source connector** (#91): Extract data from Microsoft SQL Server using pure-Python `pymssql`. Supports host, port, database, user, password_env, schema. Install: `pip install drt-core[sqlserver]`.
 - **Databricks source connector** (#88): Extract data from Databricks SQL Warehouse using `databricks-sql-connector`. Supports Unity Catalog, access token auth. Install: `pip install drt-core[databricks]`.
 - **Webhook trigger endpoint** (#218): New `drt serve` command starts a lightweight HTTP server (stdlib `http.server`) so you can trigger syncs via `POST /sync/<name>`. Includes health check, bearer token auth, and single-sync concurrency control (423 on parallel requests). Guide: `docs/guides/using-webhook-trigger.md`.
--- a/drt/cli/main.py
+++ b/drt/cli/main.py
@@ -333,7 +333,7 @@ def run(
     succeeded = 0
     failed = 0
 
-    def _run_one(sync):
+    def _run_one(sync: SyncConfig) -> tuple[str, dict[str, object], bool]:
         """Execute a single sync and return (name, result_dict, had_error)."""
         dest = _get_destination(sync)
         wm_storage = _get_watermark_storage(sync, Path("."))
--- a/drt/cli/main.py
+++ b/drt/cli/main.py
@@ -256,8 +256,12 @@ def _init_from_dbt(manifest_path: Path) -> None:
 
 @app.command()
 def run(
-    select: str = typer.Option(None, "--select", "-s", help="Run sync by name or tag (tag:crm)."),
-    all_syncs: bool = typer.Option(False, "--all", help="Explicitly run all syncs."),
+    select: str = typer.Option(
+        None,
+        "--select",
+        "-s",
+        help='Run sync by name, tag (tag:crm), or "*" / "all" for every sync.',
+    ),
     threads: int = typer.Option(1, "--threads", "-t", help="Parallel execution threads."),
     dry_run: bool = typer.Option(False, "--dry-run", help="Preview without writing data."),
     verbose: bool = typer.Option(False, "--verbose", help="Show row-level error details."),
@@ -277,8 +281,9 @@ def run(
 ) -> None:
     """Run sync(s) defined in the project.
 
-    Without --select or --all, runs all syncs sequentially (existing behaviour).
+    Without --select, runs all syncs sequentially (existing behaviour).
     Use --select to filter by name or tag (e.g. --select tag:crm).
+    Use --select "*" or --select all to be explicit about running every sync.
     Use --threads N for parallel execution.
     """
     import json as json_mod
@@ -314,7 +319,10 @@ def run(
         raise typer.Exit()
 
     if select:
-        if select.startswith("tag:"):
+        if select in ("*", "all"):
+            # Explicit "run every sync" sentinel — no filtering.
+            pass
+        elif select.startswith("tag:"):
             tag = select[4:]
             syncs = [s for s in syncs if tag in getattr(s, "tags", [])]
             if not syncs:
--- a/drt/cli/main.py
+++ b/drt/cli/main.py
@@ -256,7 +256,9 @@ def _init_from_dbt(manifest_path: Path) -> None:
 
 @app.command()
 def run(
-    select: str = typer.Option(None, "--select", "-s", help="Run a specific sync by name."),
+    select: str = typer.Option(None, "--select", "-s", help="Run sync by name or tag (tag:crm)."),
+    all_syncs: bool = typer.Option(False, "--all", help="Explicitly run all syncs."),
+    threads: int = typer.Option(1, "--threads", "-t", help="Parallel execution threads."),
     dry_run: bool = typer.Option(False, "--dry-run", help="Preview without writing data."),
     verbose: bool = typer.Option(False, "--verbose", help="Show row-level error details."),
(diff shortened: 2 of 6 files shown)


masukai (changes requested):
Thanks for the solid work here @Pawansingh3889! 🙌 The `_run_one` extraction is clean, backward compatibility is preserved, and the JSON output enhancements are nice. A few things blocking merge, and some suggestions:

---

### 🔴 Blocking

**1. `StateManager.save_sync()` is not thread-safe**

This is the biggest concern. Looking at `drt/state/manager.py`:

```python
def save_sync(self, state: SyncState) -> None:
    data = self._load_all()           # read state.json
    data[state.sync_name] = asdict(state)
    self._save_all(data)              # write state.json
```

It's a read-modify-write without any locking. Under `--threads > 1`, two syncs finishing concurrently will race:

```
T1: _load_all() → {A, B}
T2: _load_all() → {A, B}
T1: _save_all({A, B, sync1_new})
T2: _save_all({A, B, sync2_new})   ← sync1_new is lost
```

Options:
- Add a `threading.Lock` around `save_sync()` in `StateManager` (simplest)
- Or serialize state writes by doing them on the main thread after all futures complete

I'd lean toward the lock — `StateManager` is the natural owner of state.json's invariants, and a lock makes it concurrency-safe for any caller.

**2. No tests for the new functionality**

`--threads`, `--select tag:`, `--all`, and `SyncConfig.tags` are untested. For a threading feature especially, at minimum I'd want:

- A parallel-execution test: e.g. 3 syncs with `--threads=3` all complete, `succeeded` count is correct, state.json has all 3 entries (this would also catch #1 above)
- A tag-filter test: `--select tag:crm` picks up only tagged syncs
- A tags field test: YAML with `tags: [crm, daily]` parses into `SyncConfig.tags`

`tests/unit/` has precedent for CLI tests via Typer's `CliRunner`.

**3. `--all` flag has no functional effect**

`all_syncs` is defined on the signature but never referenced in the body — without `--select`, we already run all syncs, so `--all` is cosmetic. Either:

- Remove it (simplest)
- Or give it teeth: require `--select` _or_ `--all` so `drt run` alone can't accidentally fire every sync (opt-in pattern, matches dbt `--select` + `--selector` semantics)

Happy with either — just don't want a dead flag in the CLI surface.

**4. CHANGELOG.md entry missing**

One line under `## [Unreleased]` → `### Added`, similar in style to existing entries.

---

### 🟡 Suggestions

**5. Output interleaving in parallel mode**

With `--threads > 1`, `print_sync_start` / `print_sync_result` from different threads will interleave in the terminal — the output will be noisy. Not blocking, but worth either:

- Buffering per-sync output and flushing in completion order, or
- Noting the caveat in docs (parallel output is unordered by design)

**6. Source instance shared across threads**

All threads share a single `source` object (e.g. `BigQuerySource`). For BigQuery the underlying client is generally thread-safe, but this isn't universally true (DuckDB connections, for example, are per-connection). Worth either:

- Documenting that parallel execution requires a thread-safe source, or
- Constructing a source per thread inside `_run_one`

**7. `docs/llm/CONTEXT.md` CLI commands section**

The LLM context doc lists `drt run` variants at `## CLI Commands` — should include `--all`, `--threads`, and `tag:` filtering for MCP/agent consumers to discover them.

---

Happy to pair on the `StateManager` lock if any of the above is unclear. Nice feature overall — once thread safety and tests land, this is a shippable v0.6 piece 🎯

Pawansingh3889:
Three commits pushed:

001ae7f — added a threading.Lock to StateManager. The load-modify-save
in save_sync runs under the lock so parallel workers can't overwrite
each other. Reads take it too so you never see a partial state.

6eaf8c6 — dropped --all. It was a no-op anyway — declared on the
command but never referenced by the filter code below. --select * and
--select all now both mean "every sync", so you can be explicit when
you want.

9b7bb80 — 11 tests in tests/unit/test_cli_run_parallel.py covering
the selector variants, --threads dispatch (verified via
threading.get_ident() so we know syncs actually spread across
workers), and a StateManager concurrency test with 40 writers.
CHANGELOG entry under [Unreleased].

469/475 pass locally on Python 3.14. The 6 failures are all in
test_sqlite.py on Windows and pre-existing — sqlite not releasing
the file handle before os.unlink in tearDown. Linux CI isn't affected.

Pawansingh3889:
Ran into a conflict on drt/cli/main.py pulling main in. Our _run_one
overlaps with the JSON logging from #275 and the rows_extracted
changes from #345/#347.

Haven't pushed anything. Didn't want to pick an approach without
checking first. A few things I'd like your steer on:

Merge commit or rebase?

For _run_one, keep building an entry dict and returning it, or
restructure to match the inline json_results.append on main?

For rows_extracted in the parallel path — getattr-default, add it to
the Result dataclass, or something else?

If you'd rather take it from here, that's completely fine — just say.

masukai:
Nice work @Pawansingh3889 — all three blocking items addressed 🎉

On the merge conflict: either rebase or merge commit is fine — we squash-merge PRs so the final history is clean either way. I'd lean toward rebase (`git rebase origin/main`) since it makes the PR diff easier to review, but no strong preference.

The conflict area is `drt/cli/main.py` around the `run()` function. The main changes you'll need to reconcile:

- **#330 JSON logging**: `--log-format` flag + `_configure_json_logging()` call at the top of `run()`
- **#345/#347 rows_extracted**: minor additions to the result reporting

Your `_run_one` refactor should wrap around both cleanly. If you get stuck on the resolution, feel free to push what you have and I can help sort it out.

Pawansingh3889:
Rebased onto current `main` and force-pushed. CI green on 3.10 / 3.11 / 3.12 / 3.13.

**Conflict resolution in `drt/cli/main.py`** (both branches of `_run_one`):

- Kept the factored-out `entry = {...}` dict from this PR.
- Wrapped the `log_format == "json"` blocks from #330 around it — they run alongside the entry-dict construction using the same `status_str`, not instead of it.
- Added `"rows_extracted": result.rows_extracted` from #345/#347 into the success branch's entry.

No changes needed to the selector logic, `--threads` dispatch, or `StateManager` locking — those rebased cleanly.

**Two test-fake updates landed with the rebase:**

- `_FakeResult` in `test_cli_run_parallel.py` (my fake) — gained a `rows_extracted` attribute. Needed because moving the entry-dict construction out of the `if json_mode:` branch made `result.rows_extracted` read on every path, not just the JSON-output one.
- `FakeResult` in `test_dry_run_summary.py` (pre-existing) — same fix for the same reason. On `main` today this test passes by accident: `rows_extracted` only gets read under `--output json`, and the test doesn't pass that flag. Once my refactor reads the attribute unconditionally, the fake's staleness surfaces.

**`check-i18n` warning** about `CONTRIBUTING.ja.md` / `README.ja.md` also fires — happy to leave that for a separate translation PR since the CI step is gated with `|| echo` and doesn't fail the build. The hash markers point at pre-#344 commits; upstream `396f47a` and `0babbe5` moved the English bases without bumping the Japanese ones.

Ready when you are.

masukai (approved):
🎉 Excellent work @Pawansingh3889 — the StateManager lock is cleanly implemented with proper docstrings, and the test coverage (274 lines!) is exactly what a threading feature needs. The rebase was handled perfectly too.

I'll clean up any minor formatting after merge. No action needed from your side.

Great to have you as our first Triage Collaborator — this PR shows exactly why 🙌

masukai:
@Pawansingh3889 — should have acknowledged this before merging, sorry for the gap. Your rebase report was thorough and appreciated:

- Conflict resolution approach (entry dict + JSON logging blocks alongside, rows_extracted in success branch) was exactly right
- Test fake updates for rows_extracted make sense — good catch on the accidental pass in test_dry_run_summary

On the i18n warning: agreed, separate translation PR is the right call. I'll open a tracking issue for the Japanese file sync.

Thanks for the clear communication throughout — it made the review much smoother.

Issue #315, opened 2026-04-14
fix(postgres): crash on dict values bound for JSONB columns


masukai:
## Summary

The PostgreSQL destination passes record values directly to `psycopg2.cursor.execute()`. psycopg2 has **no default adapter for `dict`**, so any `dict` value in a record (e.g. a BigQuery `JSON` column read as a Python dict, then synced to a Postgres `JSONB` column) causes the row to fail with `ProgrammingError: can't adapt type 'dict'`.

This is the Postgres analogue of #311 (MySQL), discovered while shipping v0.5.1.

## Reproduction

```yaml
# sync.yml
source:
  type: bigquery
  query: |
    SELECT user_id, JSON '{"lang": "ja"}' AS profile FROM ...
destination:
  type: postgres
  table: user_profiles      -- profile column is JSONB
  upsert_key: [user_id]
```

→ row fails on `cur.execute()` with adapter error.

## Current behavior

[`drt/destinations/postgres.py:54`](https://github.com/drt-hub/drt/blob/main/drt/destinations/postgres.py#L54):

```python
values = [record.get(c) for c in columns]
cur.execute(sql, values)
```

## Expected behavior

Wrap `dict` values with `psycopg2.extras.Json` before binding (idiomatic psycopg2 pattern, JSONB-compatible):

```python
from psycopg2.extras import Json

def _serialize_value(value):
    if isinstance(value, dict):
        return Json(value)
    return value

values = [_serialize_value(record.get(c)) for c in columns]
```

## Scope of this fix

- ✅ `dict` → JSONB column: fixed
- ⏸️ `list` is left untouched (psycopg2's default `List` adapter converts to PostgreSQL `ARRAY`, which is correct for `int[] / text[]` columns). `list → JSONB` is still unsupported in this fix.

For a more principled, schema-aware solution see #316 and #317 (config-driven and introspection-based). This issue covers the **immediate crash fix** only — same shape as #311 for MySQL.

## Acceptance criteria

- [ ] `_serialize_value` helper added to `drt/destinations/postgres.py`
- [ ] Unit test in `tests/unit/test_postgres_destination.py` verifying `dict` value is wrapped with `Json`
- [ ] CHANGELOG entry under `[Unreleased] ### Fixed`
- [ ] Backward compatible: list / str / int / etc. pass through unchanged

## Why `good first issue`

- Mechanical fix following the established #311 / v0.5.1 pattern
- Single file change + single test
- Clear acceptance criteria

## References

[...]

Issue #311, opened 2026-04-14
feat(mysql): auto json.dumps() for dict/list values in MySQL destination


K-Masuda-SL:
## Summary

When BigQuery JSON columns are read as Python `dict`/`list`, the MySQL destination passes them directly to pymysql, which fails because pymysql doesn't serialize complex types automatically.

## Current behavior

```python
# mysql.py load()
values = [record.get(c) for c in columns]
cur.execute(sql, values)  # fails if value is dict/list
```

## Expected behavior

Automatically serialize `dict`/`list` values to JSON strings before passing to pymysql:

```python
values = []
for c in columns:
    v = record.get(c)
    if isinstance(v, (dict, list)):
        v = json.dumps(v, ensure_ascii=False)
    values.append(v)
```

## Use case

BigQuery → MySQL reverse ETL where BQ has JSON columns (e.g., LLM analysis results stored as JSON in BQ Gold, synced to CloudSQL for product consumption).

## Impact

- Small change, backward compatible
- `json` is already imported in `mysql.py`
- Only affects dict/list values; strings, ints, etc. pass through unchanged

PR #382, opened 2026-04-19
feat(destinations): add json_columns config for explicit JSON serialization

### LINKED ISSUE #316: feat(destinations): add json_columns config for explicit JSON serialization across SQL destinations
## Motivation

Both #311 (MySQL) and #315 (Postgres) are symptoms of a deeper issue: **drt does not know which destination columns are JSON/JSONB, so it cannot reliably serialize `dict`/`list` values**. The current per-destination patches (`json.dumps()` for MySQL, `Json()` wrapper for Postgres) are pragmatic but inconsistent and don't cover all cases (e.g. `list → JSONB`, `dict` going to a non-JSON column with a custom type adapter).

This issue proposes a **shared, declarative configuration mechanism** for SQL-style destinations to express which columns hold JSON.

## Proposal

Add a `json_columns` field to SQL destination configs:

```yaml
destination:
  type: postgres   # or mysql, snowflake, clickhouse, ...
  table: user_profiles
  upsert_key: [user_id]
  json_columns:
    - profile        # JSONB column — drt will JSON-serialize dict/list values
    - llm_analysis   # JSONB column
```

### Behavior
- Values for columns listed in `json_columns` are JSON-serialized before binding (`json.dumps(..., ensure_ascii=False)` for MySQL; `Json(...)` for Postgres; native equivalents for other backends).
- Columns NOT in `json_columns` are passed through with the destination driver's default adapter (so Postgres `list → ARRAY` keeps working).
- Empty / unset `json_columns` → current default behavior preserved (backward compatible).

### Why config-driven (Layer 2) over auto-detect (Layer 3 / #317)
- Zero round-trips at sync start (no `information_schema` query)
- Works without `INFORMATION_SCHEMA` read permissions
- Predictable: behavior visible in YAML, no surprises
- Fits drt's declarative-YAML philosophy
- Stepping stone toward Layer 3 — the same internal interface (column → "is JSON") can later be populated by introspection

## Affected destinations

- `mysql` — replace `_serialize_value` from #311 to consult `json_columns`
- `postgres` — replace `_serialize_value` from #315 to consult `json_columns`
- Future SQL destinations (Snowflake/BigQuery as destination — v0.7 milestone, ClickHouse) should adopt the same field

## Open questions for design

1. **Default behavior**: when `json_columns` is unset, should we keep the current per-destination heuristic (dict → JSON for compat with #311/#315 fixes), or require explicit declaration? (Backward compat suggests: keep heuristic, document `json_columns` as the recommended explicit form.)
2. **Naming**: `json_columns`? `json_fields`? `serialize_as_json`? Aligning with how other tools name this (dlt uses `complex_columns` IIRC).
3. **Per-destination override vs shared mixin**: extract into a `JsonColumnsMixin` or similar in `config/models.py`?
4. **Validation**: should drt validate that listed columns actually exist in the target table? (Requires introspection — defer to #317.)

## Acceptance criteria

- [ ] Design discussion resolved (open questions above)
- [ ] `json_columns: list[str] | None` field added to `MySQLDestinationConfig` and `PostgresDestinationConfig`
- [ ] Both destinations honor the field
- [ ] Unit tests covering: column in list → JSON-serialized; column not in list → passthrough; mixed records
- [ ] Docs updated: destination YAML reference, README connectors section
- [ ] CHANGELOG entry

## References

- #311 — MySQL `dict`/`list` crash (fixed in v0.5.1, Layer 1)
- #315 — Postgres `dict` crash (Layer 1)
- See also: forthcoming #317 — schema introspection for SQL destinations (Layer 3, fundamental)

armorbreak001:
## Summary

Closes #316 — adds a **declarative `json_columns` config** for SQL destinations to express which columns hold JSON/JSONB data.

### Problem
Both #311 (MySQL) and #315 (Postgres) are symptoms of a deeper issue: **drt does not know which destination columns are JSON/JSONB**, so it cannot reliably serialize `dict`/`list` values. The current per-destination patches are pragmatic but inconsistent.

### Solution
Add `json_columns: list[str] | None` to both `PostgresDestinationConfig` and `MySQLDestinationConfig`.

**Behavior:**
| json_columns | Column in list | Column NOT in list |
|---|---|---|
| set (e.g. `["profile"]`) | JSON-serialized (`Json()` / `json.dumps`) | Passed through as native Python type |
| `None` (unset) | JSON-serialized (backward compat heuristic) | JSON-serialized (backward compat) |

### Example YAML
```yaml
destination:
  type: postgres
  table: user_profiles
  upsert_key: [user_id]
  json_columns:
    - profile        # JSONB column → dict/list auto-wrapped with Json()
    - llm_analysis   # JSONB column
```

### Files changed
- `drt/config/models.py` — +2 lines (field on both configs)
- `drt/destinations/postgres.py` — `_maybe_json()` helper, Json() wrapping
- `drt/destinations/mysql.py` — `_serialize_value()` now respects json_columns
- `tests/unit/test_mysql_destination.py` — +7 new tests

### Tests
26/26 passing ✅ (19 existing + 7 new json_columns tests)

### DIFF
--- a/drt/config/models.py
+++ b/drt/config/models.py
@@ -312,6 +312,7 @@ class PostgresDestinationConfig(BaseModel):
     upsert_key: list[str]  # columns for ON CONFLICT
     ssl: SslConfig | None = None
     lookups: dict[str, LookupConfig] | None = None
+    json_columns: list[str] | None = None  # columns that hold JSON/JSONB data
 
     def describe(self) -> str:
         return f"{self.type} ({self.table})"
@@ -343,6 +344,7 @@ class MySQLDestinationConfig(BaseModel):
     upsert_key: list[str]  # columns for ON DUPLICATE KEY
     ssl: SslConfig | None = None
     lookups: dict[str, LookupConfig] | None = None
+    json_columns: list[str] | None = None  # columns that hold JSON data
 
     def describe(self) -> str:
         return f"{self.type} ({self.table})"
--- a/drt/destinations/mysql.py
+++ b/drt/destinations/mysql.py
@@ -36,6 +36,10 @@ def _serialize_value(value: Any, column: str | None = None, json_columns: list[s
 
     When json_columns is None (backward compat), all dict/list values are
     serialized — matching the pre-#316 heuristic behavior.
+
+    Raises:
+        ValueError: If *json_columns* is set and an unlisted column receives
+            a dict or list value.
     """
     if not isinstance(value, (dict, list)):  # noqa: UP038
         return value
@@ -43,7 +47,12 @@ def _serialize_value(value: Any, column: str | None = None, json_columns: list[s
     if json_columns is not None:
         if column and column in json_columns:
             return json.dumps(value, ensure_ascii=False)
-        return value
+        # Unlisted dict/list column with explicit json_columns → fail early
+        raise ValueError(
+            f"Column '{column}' contains a {type(value).__name__} value but "
+            f"is not listed in json_columns={json_columns}. "
+            f"Add '{column}' to json_columns or remove the value."
+        )
     # Backward compat: no config → serialize all complex types
     return json.dumps(value, ensure_ascii=False)
 
--- a/drt/destinations/mysql.py
+++ b/drt/destinations/mysql.py
@@ -27,17 +27,25 @@ from drt.destinations.base import SyncResult
 from drt.destinations.row_errors import RowError
 
 
-def _serialize_value(value: Any) -> Any:
+def _serialize_value(value: Any, column: str | None = None, json_columns: list[str] | None = None) -> Any:
     """Serialize dict/list values to JSON strings for pymysql.
 
-    pymysql does not auto-serialize complex Python types, so values
-    bound for JSON columns (common when sourcing from BigQuery) must
-    be converted to strings before execute().
-    """
-    if isinstance(value, dict | list):
-        return json.dumps(value, ensure_ascii=False)
-    return value
+    If json_columns is specified, only columns in that list are JSON-serialized.
+    This allows non-JSON columns to receive native Python types (e.g. list →
+    ARRAY) when the driver supports it.
 
+    When json_columns is None (backward compat), all dict/list values are
+    serialized — matching the pre-#316 heuristic behavior.
+    """
+    if not isinstance(value, (dict, list)):  # noqa: UP038
+        return value
+    # Explicit config: only serialize listed columns
+    if json_columns is not None:
+        if column and column in json_columns:
+            return json.dumps(value, ensure_ascii=False)
+        return value
+    # Backward compat: no config → serialize all complex types
+    return json.dumps(value, ensure_ascii=False)
 
 
 class MySQLDestination:
@@ -71,6 +79,7 @@ class MySQLDestination:
                     columns,
                     config.table,
                     sync_options,
+                    config,
                 )
             else:
                 result = self._load_upsert(
@@ -122,6 +131,7 @@ class MySQLDestination:
         columns: list[str],
         table: str,
         sync_options: SyncOptions,
+        config: MySQLDestinationConfig,
     ) -> SyncResult:
         """TRUNCATE (once) → INSERT within a transaction."""
         result = SyncResult()
@@ -134,7 +144,7 @@ class MySQLDestination:
 
         for i, record in enumerate(records):
             try:
-                values = [_serialize_value(record.get(c)) for c in columns]
+                values = [_serialize_value(record.get(c), c, config.json_columns) for c in columns]
                 cur.execute(sql, values)
                 result.success += 1
             except Exception as e:
@@ -172,7 +182,7 @@ class MySQLDestination:
 
         for i, record in enumerate(records):
             try:
-                values = [_serialize_value(record.get(c)) for c in columns]
+                values = [_serialize_value(record.get(c), c, config.json_columns) for c in columns]
                 cur.execute(sql, values)
                 result.success += 1
             except Exception as e:
@@ -279,7 +289,7 @@ class MySQLDestination:
             ca = resolve_env(None, config.ssl.ca_env)
             if ca:
                 ssl_dict["ca"] = ca
-            cert = resolve_env(None, config.ssl.cert_env)
+            cert = resolve_env(None, config.config.cert_env)
             if cert:
                 ssl_dict["cert"] = cert
             key = resolve_env(None, config.ssl.key_env)
--- a/drt/destinations/mysql.py
+++ b/drt/destinations/mysql.py
@@ -27,7 +27,11 @@ from drt.destinations.base import SyncResult
 from drt.destinations.row_errors import RowError
 
 
-def _serialize_value(value: Any, column: str | None = None, json_columns: list[str] | None = None) -> Any:
+def _serialize_value(
+    value: Any,
+    column: str | None = None,
+    json_columns: list[str] | None = None,
+) -> Any:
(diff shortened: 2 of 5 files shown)


masukai:
Thanks @armorbreak001! Great work on the json_columns config — this is the right long-term solution for #316.

**Plan:** We'll merge #389 (the #315 bug fix) first, then rebase #382 on top. After rebase, the postgres side of #382 should build on #389's `_serialize_value()` rather than introducing a separate `_maybe_json()`. The json_columns logic would extend `_serialize_value` to check the column list before wrapping.

No action needed from you right now — let's get #389 green first, then I'll ping you here for the rebase.

masukai:
Heads up @armorbreak001 — #395 (the JSONB fix, rebased from your #389) is now merged. As mentioned earlier, please rebase #382 on top of main.

The key change: `postgres.py` now has `_serialize_value()` that wraps `dict` with `Json()`. Your `json_columns` logic should extend this function rather than introducing a separate `_maybe_json()`. Something like:

```python
def _serialize_value(value: Any, column: str | None = None, json_columns: list[str] | None = None) -> Any:
    if isinstance(value, dict):
        if json_columns is not None:
            if column and column in json_columns:
                from psycopg2.extras import Json
                return Json(value)
            return value  # not in json_columns → pass through
        from psycopg2.extras import Json
        return Json(value)  # backward compat: no config → always wrap
    return value
```

Let me know if you need help with the rebase!

armorbreak001:
Thanks @masukai! Rebased on top of main (with #395 merged). 

Changes:
- **postgres.py**: Extended `_serialize_value()` to accept `column` and `json_columns` params (per your suggestion), removed the separate `_maybe_json()` method
- **mysql.py**: Same pattern, kept json_columns-aware `_serialize_value()`
- **Tests**: Updated to use new `_serialize_value()` API

All 693 tests passing ✅

Pawansingh3889:
Thanks for digging into this @armorbreak001. The opt-in shape is good and the backward-compat default keeps existing pipelines safe.

One real concern: when json_columns is None, all dicts get wrapped. When you set json_columns=["profile"], only profile gets the treatment and other dict-valued columns pass through as native Python types. So a user with two JSONB columns who sets json_columns=["profile"] thinking they're fixing one column ends up breaking the other, because psycopg2 can't adapt a bare dict (and pymysql can't serialize one). The error then surfaces deep inside the load loop as "can't adapt type 'dict'" instead of something the user can act on.

Two ways out. Either make the docstring (and the YAML example) explicit that once json_columns is set, it has to list every dict/list column or the load will fail at the driver. Or, when json_columns is set and an unlisted column receives a dict/list, raise an early validation error pointing at the missing column. The second one costs one isinstance check and removes the confusing failure mode. I'd lean that way but either's defensible.

A couple of smaller things while I'm reading.

TestPostgresJsonColumns is sitting in test_mysql_destination.py, would be easier to find in test_postgres_destination.py. And test_serialize_value_dict_in_json_columns just asserts the result is not None, which a regression returning a bare dict would still pass. Bumping that to assert isinstance(result, Json) when psycopg2 is available would actually catch the case it's testing for.

CI is showing mergeable_state: unstable. The PR body says 26/26 passing locally so it might be a flaky integration check, worth confirming what's red before merge.

The from psycopg2.extras import Json is inside _serialize_value and repeated in both branches. Hoisting to module scope with a try/except guard for the optional dependency centralises the import story. Not blocking.

Will swing back when this lands another round.

armorbreak001:
Thanks @Pawansingh3889 for the thorough review! All addressed:

1. **Early validation error** ✅ — When `json_columns` is set and an unlisted column receives a dict/list, `_serialize_value` now raises a `ValueError` with a clear message pointing at the missing column (applies to both postgres and mysql). This removes the confusing driver-level failure mode.

2. **Test placement** ✅ — Moved `TestPostgresJsonColumns` from `test_mysql_destination.py` to `test_postgres_destination.py`.

3. **Test improvement** ✅ — `test_serialize_value_dict_in_json_columns` now asserts `isinstance(result, Json)` when psycopg2 is available (with ImportError fallback). Also added tests for list validation and error message content.

4. **Import hoisting** ✅ — Hoisted `from psycopg2.extras import Json` to module scope with try/except guard.

5. **CI** — Will keep an eye on it, all 700 unit tests pass locally.

Full 700/700 tests passing 🎉

masukai:
Genuinely heartwarming to see this thread evolve — thank you @Pawansingh3889 for the careful review, the early-validation suggestion alone makes this PR meaningfully better for users hitting "can't adapt type 'dict'" 🙏

@armorbreak001 — thank you for the fast turnaround on all four points (early `ValueError`, test relocation, `isinstance(Json)` assertion, hoisted import). The implementation reads cleanly now and the error messages will save real debugging time.

Just one CI nit before final review — three ruff `E501` (line too long) errors:

```
drt/destinations/mysql.py:30    def _serialize_value(value: Any, column: str | None = None, json_columns: list[str] | None = None) -> Any:
drt/destinations/postgres.py:35 def _serialize_value(value: Any, column: str | None = None, json_columns: list[str] | None = None) -> Any:
drt/destinations/postgres.py:71 if isinstance(value, list) and json_columns is not None and column and column not in json_columns:
```

Wrap them across lines, e.g.:

```python
def _serialize_value(
    value: Any,
    column: str | None = None,
    json_columns: list[str] | None = None,
) -> Any:
```

```python
if (
    isinstance(value, list)
    and json_columns is not None
    and column
    and column not in json_columns
):
```

`make lint` will catch these locally. Once CI's green I'll do a final review pass and we can ship 🚀

masukai:
Hey @armorbreak001 — sorry, my last comment glossed over CI status. Looks like the run is actually red on a small `ruff E501` (line-too-long, 3 spots) introduced when the early-validation logic was added:

- `drt/destinations/mysql.py:30` — `_serialize_value` signature (106 chars)
- `drt/destinations/postgres.py:35` — same signature (106 chars)
- `drt/destinations/postgres.py:71` — the new `if isinstance(value, list) and ...` guard (102 chars)

Could you reformat the two signatures onto multiple lines, and break the `if` at `postgres.py:71`? Something like:

```python
def _serialize_value(
    value: Any,
    column: str | None = None,
    json_columns: list[str] | None = None,
) -> Any:
```

`make lint` locally should catch it. Once CI is green I'll approve and merge. 🙏

Thanks again for sticking with this through two review rounds — the json_columns design is a real upgrade for users hitting the JSONB edge cases.

masukai:
@armorbreak001 — pushed a tiny commit (`8a63c63`) directly to your branch to unblock CI:

- Wrapped the two `_serialize_value()` signatures and the multi-condition `if` to fit the 100-char limit (the 3 `ruff E501` errors that were red on CI).
- While I was in there, mypy also flagged a typo at [`drt/destinations/mysql.py:305`](drt/drt/destinations/mysql.py#L305) — `config.config.cert_env` → fixed to `config.ssl.cert_env`. Pre-existing nit but it would have broken cert-based SSL auth, and lint needed to be green anyway. Hope that's OK; happy to revert if you'd rather handle it separately.

CI should go green now. Once it does, this is good to land — thanks for sticking with this through such a thorough review thread. The early-validation behavior is genuinely a UX upgrade for everyone using Postgres/MySQL JSON columns. 🚀

⭐ If you find drt useful, please consider [starring the repo](https://github.com/drt-hub/drt) — it really helps with visibility!

masukai (approved):
LGTM! 🎉 The early-validation behavior with explicit `json_columns` is a real UX upgrade — error message points right at the misconfigured column instead of the user having to decode a `can't adapt type 'dict'` from psycopg2. Coverage looks solid.

Thanks for sticking with this through the long review thread @armorbreak001. 🙏

PR #388, opened 2026-04-20
fix(postgres): wrap dict values with Json adapter for JSONB columns

### LINKED ISSUE #315: fix(postgres): crash on dict values bound for JSONB columns

armorbreak001:
## Problem

Syncing records with `dict` values (e.g. from a BigQuery `JSON` column) to a PostgreSQL `JSONB` column crashes with `ProgrammingError: can't adapt type 'dict'`. psycopg2 has no default adapter for `dict`.

## Fix

Added a `_serialize_value()` helper that wraps `dict` values with `psycopg2.extras.Json` before binding them in both `_load_upsert` and `_load_replace`. Non-dict types pass through unchanged — fully backward compatible.

This is the same pattern used in the MySQL destination fix (#311).

## Changes

- **`drt/destinations/postgres.py`**: add `_serialize_value()` helper, use it in both load paths
- **`tests/unit/test_postgres_destination.py`**: new tests for dict serialization in upsert and replace modes
- **`CHANGELOG.md`**: entry under `[Unreleased] ### Fixed`

Fixes #315.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **PostgreSQL destination**: crash when dict values are bound for JSONB columns — `dict` values are now wrapped with `psycopg2.extras.Json` before binding (#315)
+
 - **`${VAR}` env substitution in sync YAML** (#385): Environment variable placeholders now work in all string fields of sync YAML (e.g. `watermark.bucket`, `destination.url`), not just `model:` SQL. Also shipped in [0.6.1](#061---2026-04-20).
 
 ## [0.6.1] - 2026-04-20
--- a/drt/destinations/postgres.py
+++ b/drt/destinations/postgres.py
@@ -21,6 +21,15 @@ from __future__ import annotations
 import json
 from typing import Any
 
+
+def _serialize_value(value: Any) -> Any:
+    """Wrap dict values so psycopg2 can adapt them for JSONB columns."""
+    if isinstance(value, dict):
+        from psycopg2.extras import Json
+
+        return Json(value)
+    return value
+
 from drt.config.credentials import resolve_env
 from drt.config.models import DestinationConfig, PostgresDestinationConfig, SyncOptions
 from drt.destinations.base import SyncResult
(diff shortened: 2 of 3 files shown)


armorbreak001:
Superseded by #389

Issue #434, opened 2026-04-29
feat: Snowflake support for replace_strategy: swap


masukai:
## Summary

Once both #338 (zero-downtime replace) and #353 (Snowflake destination) land, extend the Snowflake destination to support `replace_strategy: swap`.

## Motivation

Snowflake has native `ALTER TABLE ... SWAP WITH ...` syntax which is atomic, making this a clean fit. Production Snowflake users will want zero-downtime replace just as much as Postgres / MySQL / ClickHouse users — and skipping Snowflake here would create a gap right as the destination is gaining adoption.

## Proposed Behavior

```yaml
sync:
  mode: replace
  replace_strategy: swap
destination:
  type: snowflake
  # ...
```

Implementation pattern:

```sql
CREATE OR REPLACE TABLE {table}__drt_swap LIKE {table};
-- batch INSERT into shadow
INSERT INTO {table}__drt_swap (...) VALUES (...);
ALTER TABLE {table} SWAP WITH {table}__drt_swap;
DROP TABLE {table}__drt_swap;
```

## Implementation Notes

- `ALTER TABLE ... SWAP WITH ...` is atomic in Snowflake — this is the cleanest case among supported destinations
- `CREATE TABLE ... LIKE ...` in Snowflake copies columns and clustering keys but **not** policies/grants — verify against expected production usage
- Reuse the same `replace_strategy: Literal["truncate", "swap"]` config introduced in #338

## Depends on

- #338 — introduces the `replace_strategy: swap` config and the per-destination contract.
- #353 — Snowflake destination must be merged first.

Muawiya-contact:
This looks like a great addition, @masukai. Leveraging Snowflake's atomic `SWAP WITH` definitely makes this the cleanest implementation of the zero-downtime strategy so far.

A couple of thoughts for the implementation:

* **Permission Persistence:** Since `LIKE` won't carry over grants/policies, do we want to include a utility to fetch and re-apply grants? Otherwise, downstream BI tools might lose access after the swap.
* **First-run Logic:** If the destination table doesn't exist yet, I assume we'll just default to a standard `CREATE` and skip the swap logic for that cycle?
* **Clustering Keys:** It’s great that `LIKE` catches these, as that's a huge performance factor for Snowflake users.

Looking forward to seeing this land after #338 and #353!

masukai:
@Muawiya-contact Sorry for the slow turnaround. Great questions — answers inline:

[...]

Issue #433, opened 2026-04-29
feat: auto-cleanup of orphaned shadow tables from replace_strategy: swap


masukai:
## Summary

Add automated detection and cleanup of orphan shadow tables (`{table}__drt_swap`) left behind by interrupted `replace_strategy: swap` runs.

## Motivation

#338 introduces `replace_strategy: swap`, which creates a shadow table during sync and drops it after an atomic rename. If a sync is killed (SIGKILL, OOM, host failure between INSERT and RENAME), the shadow table is left behind.

The initial implementation in #338 handles this only by:

1. Dropping any existing same-named shadow at the start of the next swap run
2. Documenting manual cleanup (`DROP TABLE {name}__drt_swap`) in the README

For production deployments, an explicit cleanup mechanism is preferable so operators don't accumulate hidden disk usage.

## Proposed Behavior

Two possible approaches (pick one or both during design):

### Option 1: `drt clean --orphans` CLI command

```bash
drt clean --orphans                # dry-run by default — list what would be dropped
drt clean --orphans --execute      # actually drop
drt clean --orphans --older-than 24h  # only drop shadows older than N
```

Scans destinations referenced by sync configs for tables matching `*__drt_swap` and reports / drops them.

### Option 2: Automatic cleanup at swap start

When `replace_strategy: swap` runs, scan the same schema for shadow tables older than N hours and drop them before starting the new shadow.

Pros: zero-config. Cons: silently mutates state.

## Implementation Notes

- Per-destination `list_orphan_swap_tables()` helper following the existing `_load_replace` pattern
- Suffix match (`__drt_swap`) is probably strict enough for ownership detection, but consider adding a metadata table or comment marker for stronger guarantees
- Option 1 requires destinations to expose a discovery API — may be a Protocol addition

## Depends on

- #338 (zero-downtime replace via staging table swap) — this issue is conditional on #338 shipping `replace_strategy: swap` first.

Muawiya-contact:
I’ll take this one, @masukai!

Regarding the implementation, I think 
**Option 1 (`drt clean --orphans`)** is the best starting point because it gives users explicit control and visibility. However, combining it with a conservative version of 
**Option 2** (cleaning up old shadows at the start of a new run) would make the system much more self-healing.

I'll start by drafting the `list_orphan_swap_tables()` helper following the existing `_load_replace` pattern and then move on to the CLI command. I'm assigning this to myself now.

masukai:
[...]

PR #353, opened 2026-04-17
feat: add Snowflake destination

PFCAaron12:
## What does this PR do?

1. drt/destinations/snowflake.py — Destination implementation

2. drt/config/models.py — SnowflakeDestinationConfig

3. Supports MERGE (upsert) and INSERT (append) modes

4. Auth via password_env or key pair

5. Added in drt/cli/main.py

## Related Issue

Closes # 164

## Checklist

- [ ] Tests pass (`make test`)
- [ ] Linter passes (`make lint`)
- [ ] Updated `CHANGELOG.md` (if user-facing change)

PR #435, opened 2026-04-29
feat: zero-downtime replace via staging table swap

### LINKED ISSUE #338: feat: zero-downtime replace via staging table swap
## Summary

Enhance `sync.mode: replace` with an optional staging table swap pattern for zero-downtime table replacement.

## Motivation

The current `replace` mode (v0.5.2) does TRUNCATE → INSERT within a transaction. While Postgres/MySQL transactions prevent most visibility issues, there's still a brief window during TRUNCATE where the table is empty. For high-availability production tables queried by downstream applications, a staging table swap would provide truly atomic replacement.

## Proposed Behavior

```yaml
sync:
  mode: replace
  replace_strategy: swap  # optional, default: truncate
```

### Strategy: `swap` (new)
1. CREATE TEMP TABLE (or shadow table) with same schema
2. INSERT all rows into temp table
3. Swap tables atomically (e.g., `ALTER TABLE ... RENAME`)

### Per destination:
- **PostgreSQL**: `CREATE TABLE ... (LIKE)` → INSERT → `ALTER TABLE RENAME` in transaction
- **MySQL**: `CREATE TABLE ... LIKE` → INSERT → `RENAME TABLE` (atomic)
- **ClickHouse**: `EXCHANGE TABLES` (atomic since 21.8+)

## Depends on

- #337 (sync.mode: replace — shipped in v0.5.2)

masukai:
## Summary

Implements `sync.replace_strategy: swap` for atomic table replacement, closing #338.

- Adds `replace_strategy: Literal["truncate", "swap"] = "truncate"` to `SyncOptions`, with a validator ensuring `swap` is only valid with `mode=replace`. Default behavior unchanged → fully backwards compatible.
- Engine gains a duck-typed `finalize_sync()` hook (no Protocol break) called once after the load loop for non-staged destinations.
- **PostgreSQL**: `CREATE TABLE shadow (LIKE original INCLUDING ALL)` → batch INSERT → transactional `ALTER TABLE RENAME` swap → `DROP` old.
- **MySQL**: `CREATE TABLE shadow LIKE original` → batch INSERT → atomic single-statement multi-table `RENAME TABLE` → `DROP` old.
- **ClickHouse**: `CREATE TABLE shadow AS original` → batch INSERT → atomic `EXCHANGE TABLES` (21.8+) → `DROP` shadow.

## Architecture note

The Protocol `Destination` does not have an end-of-sync hook (only `StagedDestination` does, via `finalize()`). Rather than introducing a Protocol-breaking change, the engine uses a duck-typed `getattr(destination, "finalize_sync", None)` call — destinations that don't need it are unaffected.

## Out of Scope (split into follow-ups)

- Orphan shadow auto-cleanup → #433 (v0.8)
- Snowflake support → #434 (v0.8, depends on #353)
- Dependent-object handling (views / mviews / triggers) → docs caveat only, future v0.9

## Test plan

- [x] `make lint` — ruff + mypy clean
- [x] `make test` — 735 passed, 2 skipped (was 715 before this PR; 20 new tests across config, engine, and 3 destinations)
- [x] `make check-i18n` — clean
- [x] New TDD coverage:
  - Config: `replace_strategy` field + cross-field validator (4 tests)
  - Engine: duck-typed `finalize_sync` hook — called when present, skipped when absent, dry-run guard, result accumulation (4 tests)
  - Postgres: shadow create, multi-batch insert, atomic rename, finalize-noop (4 tests)
  - MySQL: same coverage with atomic multi-table `RENAME TABLE` (4 tests)
  - ClickHouse: same coverage with `EXCHANGE TABLES` (4 tests)

## Governance

Per [GOVERNANCE.md](https://github.com/drt-hub/drt/blob/main/GOVERNANCE.md): **Medium change** (config schema addition + engine internal hook, no Protocol break) — 1 Owner approval + 24h objection window.

Reviewer: @yodakanohoshi

Closes #338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Sync failure alerts** (#414): Configure `alerts.on_failure` in sync YAML to send Slack or generic HTTP webhook notifications when a sync ends with `failed > 0` or raises an exception. Two target types in v0.7: `slack` (Slack incoming webhook) and `webhook` (generic HTTP POST/PUT with optional `body_template`). Template variables: `sync_name`, `error`, `rows_processed`, `duration_s`, `started_at`. Dispatch is best-effort — alert failures are logged but never affect sync correctness or override the original exception.
 - **Snowflake destination** (#353): Write rows back to Snowflake tables. Supports `mode: insert` (append) and `mode: merge` (upsert via temp staging table + `MERGE` statement using `upsert_key`). Auth via `account_env` / `user_env` / `password_env`. Install: `pip install drt-core[snowflake]`. Contributed by @PFCAaron12.
+- **Zero-downtime replace via staging table swap** (#338): `sync.replace_strategy: swap` enables truly atomic table replacement — drt writes to a shadow table (`{table}__drt_swap`) per batch and atomically renames it to the original at the end of the sync. Supported on PostgreSQL (transactional `ALTER TABLE RENAME`), MySQL (atomic `RENAME TABLE`), and ClickHouse (atomic `EXCHANGE TABLES`, requires 21.8+). Default remains `truncate` (existing TRUNCATE → INSERT behavior). Follow-ups tracked in #433 (orphan auto-cleanup) and #434 (Snowflake support).
 
 ### Changed
 
--- a/docs/connectors/postgres.md
+++ b/docs/connectors/postgres.md
@@ -71,6 +71,23 @@ sync:
   mode: replace
 ```
 
+**Zero-downtime replace via staging swap:**
+```yaml
+sync:
+  mode: replace
+  replace_strategy: swap   # default: truncate
+```
+
+drt creates a shadow table `{table}__drt_swap`, populates it across batches, then atomically renames it to the original. Readers of the original table never see an empty state.
+
+Requirements: the destination user needs `CREATE TABLE`, `ALTER TABLE`, and `DROP TABLE` privileges.
+
+Caveats:
+- Tables with dependent views, materialized views, or triggers are **not recommended** for swap mode — the rename can break dependent objects.
+- If a sync is killed before completion, an orphan shadow may remain. Drop manually with `DROP TABLE {table}__drt_swap`. Auto-cleanup is tracked in [#433](https://github.com/drt-hub/drt/issues/433).
+
+Same `replace_strategy: swap` is supported on MySQL (atomic `RENAME TABLE`) and ClickHouse (atomic `EXCHANGE TABLES`, requires CH 21.8+).
+
 **FK resolution with destination_lookup:**
 ```yaml
 lookups:
--- a/drt/config/models.py
+++ b/drt/config/models.py
@@ -639,6 +639,7 @@ class WatermarkConfig(BaseModel):
 
 class SyncOptions(BaseModel):
     mode: Literal["full", "incremental", "upsert", "replace"] = "full"
+    replace_strategy: Literal["truncate", "swap"] = "truncate"
     cursor_field: str | None = None  # required when mode=incremental
     watermark: WatermarkConfig | None = None
     batch_size: int = 100
@@ -650,6 +651,12 @@ class SyncOptions(BaseModel):
     def _check_incremental_cursor(self) -> "SyncOptions":
         if self.mode == "incremental" and not self.cursor_field:
             raise ValueError("cursor_field is required when mode is 'incremental'.")
+        return self
+
+    @model_validator(mode="after")
+    def _check_replace_strategy(self) -> "SyncOptions":
+        if self.replace_strategy == "swap" and self.mode != "replace":
+            raise ValueError("replace_strategy='swap' requires mode='replace'.")
         return self
 
 
--- a/drt/destinations/clickhouse.py
+++ b/drt/destinations/clickhouse.py
@@ -7,7 +7,10 @@ PostgreSQL and MySQL destination pattern).
 Deduplication is handled by ClickHouse's ReplacingMergeTree engine at merge
 time — the destination performs simple INSERTs.
 
-Supports ``sync.mode: replace`` (TRUNCATE TABLE → INSERT).
+Supports ``sync.mode: replace`` (TRUNCATE TABLE → INSERT) and
+``replace_strategy: swap`` (zero-downtime: build a shadow table via
+``CREATE TABLE ... AS ...``, INSERT into the shadow, then atomically
+``EXCHANGE TABLES`` in :meth:`finalize_sync`).
 
 Requires: pip install drt-core[clickhouse]
 
@@ -38,6 +41,8 @@ class ClickHouseDestination:
 
     def __init__(self) -> None:
         self._replace_truncated: bool = False
+        self._swap_shadow_created: bool = False
+        self._swap_table: str | None = None
 
     def load(
         self,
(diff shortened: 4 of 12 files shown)


yodakanohoshi (commented):
（このレビューは取り消しました）

yodakanohoshi (commented):
The overall direction (duck-typed `finalize_sync`, no Protocol break, atomic-swap idiom per DB) is solid and the test coverage is thoughtful. However, I found **one data-loss-grade bug in the ClickHouse implementation** that needs to be fixed before merge.

---

## ⛔ Blocker — ClickHouse: `on_error=fail` swaps a partial shadow into the live table and destroys the original data

### Location

`drt/destinations/clickhouse.py` L141-143:

```python
if sync_options.on_error == "fail":
    return result
continue
```

The Postgres and MySQL `_load_replace_swap` paths handle the same condition by

- `DROP TABLE IF EXISTS shadow`
- `self._swap_shadow_created = False`
- `self._swap_table = None`

The ClickHouse path only `return`s — it leaves both the shadow table and the in-memory state intact.

### What goes wrong

With `replace_strategy: swap` + `on_error: fail` on ClickHouse, as soon as a single row-level error occurs (type mismatch, constraint violation, etc.):

1. The shadow holds **partial data** (rows up to the failure point).
2. `load()` returns with state still set (`_swap_shadow_created = True`).
3. `drt/engine/sync.py` L184-185 breaks the batch loop because `result.failed > 0` and `on_error=fail`.
4. L202-205 invokes `finalize_sync`.
5. `finalize_sync` sees `_swap_shadow_created = True` and proceeds to `EXCHANGE TABLES original AND shadow` — **the original data moves into the shadow handle**.
6. The immediate `DROP TABLE shadow` then **physically deletes the original data**.

End state:
- Live table = partial data
- Original full dataset = gone
- Recovery = restore from backup

### Contracts violated

| Contract | Actual behavior |
|---|---|
| Zero-downtime (readers always see complete old or new state) | Live table replaced with partial data |
| `on_error: fail` (failure leaves live table untouched) | Failure causes destructive replacement |
| Purpose of swap mode (safer replace) | More dangerous than truncate in this case |

### Why this is hard to notice

The sync result reports `failed: 1`, so an operator may shrug it off as "expected fail behavior — original table is safe." But unlike every other DB where `on_error=fail` truly means "no change," here it means "destructive partial swap." Discovery typically happens after the fact, when the live table is already corrupted.

### Why Postgres / MySQL aren't affected

- Postgres: DDL is transactional, so `conn.rollback()` undoes the shadow as part of the batch transaction; state is also explicitly reset.
- MySQL: explicit `DROP TABLE IF EXISTS shadow` cleanup plus state reset.
- ClickHouse: no transactions, **and** no explicit cleanup — that's the gap.

### Suggested fix

Mirror the Postgres / MySQL pattern. A few lines:

```python
if sync_options.on_error == "fail":
    try:
        client.command(f"DROP TABLE IF EXISTS {shadow}")
    finally:
        self._swap_shadow_created = False
        self._swap_table = None
    return result
```

The `try/finally` guarantees state is reset even if the cleanup `DROP` itself fails — at worst we leave an orphan shadow (already tracked in #433), but we never let `finalize_sync` run an EXCHANGE against the live table.

---

## 🔧 High priority (recommended before merge)

### 2. Missing test coverage for the `on_error="fail"` cleanup path (this is the bulk of Codecov's missing lines)

None of the three destinations have a test for the mid-batch `on_error=fail` path in swap mode. A test for this path would have caught Bug 1.

For each of Postgres / MySQL / ClickHouse, please add a test that:

- Lets batch 1 commit successfully, then triggers a row error in batch 2
- Asserts that the shadow is dropped and the swap state is reset
- Asserts that a subsequent `finalize_sync()` returns `None` and produces no side effects against the live table

### 3. Postgres `finalize_sync` — `.split('.')[-1]` is fragile

`drt/destinations/postgres.py` L248-249:

```python
cur.execute(f"ALTER TABLE {table} RENAME TO {old.split('.')[-1]}")
cur.execute(f"ALTER TABLE {shadow} RENAME TO {table.split('.')[-1]}")
```

This works for the typical `public.scores` form but breaks for identifiers containing `.`, mixed case, reserved words, or spaces.

There is precedent in the same file: L117-126 (`get_row_count`) uses `psycopg2.sql.Identifier` for proper quoting. The swap path could follow the same approach. The existing `_load_replace` also uses raw f-string interpolation, so this PR doesn't add a regression — but it does extend the inconsistency by another method.

Minimal fix: add a `_quote_ident` static method on the Postgres destination and apply it to all `ALTER TABLE` / `DROP TABLE` / shadow-name-construction sites in the swap path.

---

## 💡 Nice to have

### 4. Document ClickHouse engine assumption

The PR description says `CREATE TABLE shadow AS original` clones engine / partitioning / ORDER BY / column defs. This holds for MergeTree-family engines but not for `Memory` / `Log` / `Distributed`, etc. A one-liner in `docs/connectors/postgres.md` (the ClickHouse section) noting "verified with MergeTree-family engines" would save users a debugging session.

### 5. Make the `finalize_sync` exception-propagation policy explicit

`drt/engine/sync.py` L205 calls `finalizer(...)` without `try/except`. If `finalize_sync` raises (e.g., a `RENAME` blocked by `lock_timeout`), the exception propagates and the shadow is orphaned. This is tracked in #433, but a comment at the call site stating "exceptions intentionally propagate; orphan cleanup is out of scope (#433)" would help future readers avoid mistaking it for an oversight.

### 6. Unused `sync_options` parameter in `finalize_sync`

None of the three `finalize_sync` implementations actually reference `sync_options`. Either keep it for forward compatibility with a comment explaining why, or drop it from the signature. The current state is ambiguous.

---

## Verdict

Bug 1 is the only blocker; the rest is polish. Once Bug 1 is fixed and the related test is added, I'm happy to approve.

masukai:
Thanks for the deep review @yodakanohoshi — the ClickHouse data-loss path is exactly the kind of subtle bug that's terrifying to miss. 🙏 Force-pushed a rebased branch with fixes:

## ⛔ Bug 1 — ClickHouse `on_error=fail` data-loss — **fixed**

Commit `0efef0d`. Mirrored the Postgres/MySQL pattern using your suggested `try/finally` shape — guarantees state reset even if the cleanup `DROP` itself fails. At worst we leave an orphan shadow (already tracked by #433); we never let `finalize_sync` run an `EXCHANGE` against the live table.

```python
if sync_options.on_error == "fail":
    try:
        client.command(f"DROP TABLE IF EXISTS {shadow}")
    finally:
        self._swap_shadow_created = False
        self._swap_table = None
    return result
```

## 🔧 #2 — Missing `on_error=fail` cleanup tests — **added**

Commit `8323eb5`. Added regression tests for all three destinations:

- `test_swap_on_error_fail_drops_shadow_and_resets_state` (Postgres / MySQL / ClickHouse) — asserts shadow is dropped on hard fail, subsequent `finalize_sync` returns `None`, and no `RENAME`/`EXCHANGE` is issued against the live table.
- `test_swap_on_error_fail_resets_state_even_if_drop_fails` (ClickHouse only) — exercises the `try/finally` guarantee.

Verified locally: `make lint` clean, 765 passed (up from 735), `make check-i18n` clean.

## 🔧 #3 — Postgres `.split('.')[-1]` fragility — **deferred to follow-up**

Opened **#442** to tackle this properly. The `get_row_count` precedent itself doesn't actually handle schema-qualified names (`Identifier("public.scores")` produces `\"public.scores\"`, not `\"public\".\"scores\"`), and `_build_insert_sql` / `_load_upsert` use raw f-strings too — so a clean fix needs a `_split_qualified` + `_qualified_ident` pair applied across the whole module, plus the same audit on MySQL/ClickHouse. That's a meaningful refactor that deserves its own PR rather than scope-creeping this one. Labeled `good first issue` since the cleanup is mechanical once the helpers are in place.

## 💡 #4 / #5 / #6 — happy to take these inline

- **#4 (ClickHouse engine assumption)** — happy to add the MergeTree-family note to the docs in this PR if you want; otherwise I'll fold it into the next docs pass.
- **#5 (`finalize_sync` exception propagation comment)** — same, can add a one-liner at the call site if you'd prefer it land here.
- **#6 (unused `sync_options` in `finalize_sync` signatures)** — leaning toward keeping for forward compatibility (e.g., a future `sync_options.swap_timeout` would land here naturally), but I can add a short docstring note explaining that. Or drop it — your call.

Let me know on 4/5/6 and I'll roll the changes into a single follow-up commit. The blocker is fixed and the regression coverage is in, so I think we're close to mergeable from your end.

masukai:
Rebased onto current `main` (resolved conflicts from #382 json_columns landing in parallel). All 778 unit tests pass locally.

@yodakanohoshi — re-review please when convenient. Diff vs your last review is identical except for the rebase conflict resolutions in:

- `drt/destinations/postgres.py` — combined json_columns `config` arg + swap branching
- `drt/destinations/mysql.py` — same shape
- `tests/unit/test_postgres_destination.py` / `test_mysql_destination.py` — both test classes (`TestPostgresJsonColumns` from #382 + `TestPostgresReplaceSwap` from this PR) co-exist
- `CHANGELOG.md` — both Snowflake (#353) and zero-downtime swap entries co-exist

## 🔍 Followup discovered during rebase → #448

While merging the two code paths I noticed `_load_replace_swap` doesn't thread `config.json_columns` through to `_serialize_value`, so swap mode silently ignores the json_columns config. Filed as #448 with milestone v0.7. **Not addressed in this PR** — let me know if you'd prefer I tack the ~5-line fix onto this PR (one re-review cycle) vs landing #435 as-is and handling #448 separately.

yodakanohoshi (approved):
Thank you for the careful work on the ClickHouse cleanup path and the regression coverage — the try/finally pattern handles the cleanup-failure case nicely.

#3 → #442 is the right call (and the issue already scopes the MySQL/ClickHouse audit).

#4 / #5 are nice-to-haves; happy to defer to follow-ups rather than scope-creep this PR. #6 — agreed, keep `sync_options` for forward-compat.

#448 — let's land this as-is and handle separately, since it's a parallel-development artifact and you've already milestoned the fix for v0.7.

Approving . Thanks for shipping this — atomic-swap is a meaningful step for v0.7 🚀

PR #449, opened 2026-05-03
feat(lookup): add check_only mode for FK existence filtering (#354)

### LINKED ISSUE #354: feat: FK existence check (filter without value resolution)
## Summary

Add an explicit way to filter source rows based on whether a foreign key exists in the destination table, without needing to resolve a value.

## Context

Follow-up to #345 (destination_lookup). Currently, lookups can be used for existence check as a side effect:

```yaml
lookups:
  user_id:
    table: users
    match: { id: user_id }
    select: id      # same as source column (identity mapping)
    on_miss: skip
```

This works because:
- `target_col == source_col` → drop_match_columns keeps the column
- `on_miss: skip` → row is filtered out when FK doesn't exist

But it's semantically awkward — we're pretending to "resolve" a value that we already have.

## Motivation

Reverse ETL use case: BQ dev contains prd-like data, but destination CloudSQL stg has a subset of users/companies. Rows referencing non-existent FKs should be silently skipped, not fail the entire sync.

### Option A — Dedicated `filters` section
```yaml
destination:
  type: mysql
  table: interviewer_profiles
  upsert_key: [user_id]
  filters:
    # Drop rows where user_id doesn't exist in users.id
    user_exists:
      table: users
      match: { id: user_id }
      on_miss: skip  # or fail
```

### Option B — Extend `lookups` with a no-select flag
```yaml
lookups:
  user_id_check:
    table: users
    match: { id: user_id }
    check_only: true  # no value resolution, just existence check
    on_miss: skip
```

### Option C — Document the current idiom
Add a docs/recipe section showing how to use lookups for existence checks with identity mapping. Lower effort, no code change.

## Recommendation

**Option A** — `filters` is semantically cleanest. It separates "enrichment" (lookups) from "filtering" (filters), which matches standard reverse ETL mental models (e.g. Hightouch has "row conditions").

## Use Case

- Parent table sync: existence check on FK fields before INSERT to avoid `foreign key constraint fails`
- Staging environments: BQ has more data than destination DB; filter to only matching rows

masukai:
Closes #354

## Summary

Adds `lookups.check_only: true` for **filter-only existence checks** — no value resolution. Implements Option B from the issue (the option I leaned toward in the issue thread; no objections raised in the 4-day window).

**Use case** (from the issue): BigQuery has prd-like data but the destination DB (e.g. staging) holds only a subset of users / companies. Rows pointing at non-existent FKs should be silently dropped rather than failing the entire sync with a foreign-key constraint error.

## YAML

```yaml
destination:
  type: mysql
  table: interviewer_profiles
  upsert_key: [user_id]

  lookups:
    user_exists:                # arbitrary label, not a destination column
      table: users
      match: { id: user_id }    # destination column: source column
      check_only: true          # existence check only
      on_miss: skip             # drop rows whose user_id isn't in users
```

## Design notes

| Concern | Behavior |
|---|---|
| `select` field | Becomes optional. Validator: required iff `check_only=False`. Setting both → fail-fast `ValueError`. |
| `build_lookup_map` | When check_only, queries only match columns (omits `select`) and returns `dict[tuple, None]` — type-stable with the existing API; only key membership matters. |
| `apply_lookups` hit | Row kept, target name **not written** (it's just a label). |
| `apply_lookups` miss | `on_miss` policy: `skip`/`fail` as usual. `null` is treated as `skip` because NULL is meaningless without a target column. |
| `drop_match_columns` | Skipped for check_only lookups — source FK columns must reach the destination. Other lookups still drop their own match columns normally. |
| Mixing with regular lookups | Fully supported — see `test_check_only_combined_with_regular_lookup`. |

## Tests

- **+13 new tests** (4 validator / 3 `build_lookup_map` / 5 `apply_lookups` / 1 SyncConfig parse), all passing alongside the existing 25 lookup tests.
- Full unit suite: **767 passed, 2 skipped**.
- mypy clean (79 source files), ruff clean.

## Out of scope

- New top-level `filters:` section (Option A in the issue) — rejected for blast-radius reasons in the issue thread.
- Documentation-only solution (Option C) — preserves the semantic awkwardness this PR is meant to fix.

## Files

- `drt/config/models.py` — `LookupConfig.check_only` + validator
- `drt/destinations/lookup.py` — `build_lookup_map` + `apply_lookups` check_only paths
- `tests/unit/test_lookup.py` — 13 new tests across 3 new test classes
- `docs/guides/destination-lookup.md` — new "Existence-only Check" section + comparison table
- `CHANGELOG.md` — Unreleased / Added entry


Pawansingh3889 (commented):
Nice clean PR. The validator on `select` is exactly the right shape, and splitting `build_lookup_map` and `apply_lookups` into two `check_only` branches keeps the regular path completely untouched. A few observations from a read-through.

The headline one is around `on_miss="null"` plus `check_only=True`. The new `_check_select_consistency` validator catches the `select` case at config-load (great), but `on_miss=null` with `check_only=True` is silently coerced to `skip` at runtime in `lookup.py` L97-113, with a `Lookup miss (check_only, skipped)` RowError per missing row. From the operator's seat they wrote `on_miss: null` and got `skip` semantics, which differ on whether the row reaches the destination at all. Same family as the early-validation thread on #382: better to fail at YAML parse than to surprise the operator on the first sync.

A sibling validator on `models.py` would do it:

```python
@model_validator(mode="after")
def _check_on_miss_consistency(self) -> "LookupConfig":
    if self.check_only and self.on_miss == "null":
        raise ValueError(
            "lookups.on_miss='null' is invalid with check_only=True "
            "(no target column to set NULL on; use 'skip' or 'fail')."
        )
    return self
```

If you adopt this, the runtime branch at L97-113 becomes unreachable, drop it, and the 88.46% patch-coverage gap closes for free. (Those 3 uncovered lines are exactly that branch.)

If you'd rather keep the runtime fallback, it just wants a test:

```python
def test_check_only_on_miss_null_is_treated_as_skip_with_record(self) -> None:
    lk = LookupConfig(
        table="users", match={"id": "user_id"},
        check_only=True, on_miss="null",
    )
    records = [{"user_id": "u1"}, {"user_id": "ghost"}]
    enriched, errors = apply_lookups(
        records, {"user_exists": (lk, {("u1",): None})}, "skip"
    )
    assert len(enriched) == 1
    assert len(errors) == 1
    assert "check_only" in errors[0].error_message
```

Three smaller things while I was reading.

In `tests/unit/test_lookup.py`, `test_query_omits_select_column` does a string-strip dance that's a bit fragile:

```python
assert "SELECT user_id FROM users" in query
assert "id" not in query.replace("user_id", "")
```

The second line is trying to assert "no leftover `id` column", but it leans on the fixture happening to use `user_id` (not `id`) for the match column. Change the fixture to `match={"id": "user_id"}` and the assertion silently inverts, because `dest_match_cols=["id"]` puts `id` legitimately in the query. Cheaper to just pin the query exactly:

```python
assert query.strip() == "SELECT user_id FROM users"
```

Worth a regression test for `target_name` colliding with a source column. A user can write the obvious `{"user_exists": ...}`, but they can also write `{"user_id": {check_only: true, match: {id: user_id}}}` where the target name happens to match a source column. The current code does the right thing (`if not lk_config.check_only: record[target_col] = ...`, source column not overwritten), but a future refactor could easily start writing `None` there:

```python
def test_check_only_target_name_colliding_with_source_col_preserved(self) -> None:
    lk = LookupConfig(table="users", match={"id": "user_id"}, check_only=True)
    records = [{"user_id": "u1"}]
    enriched, _ = apply_lookups(
        records, {"user_id": (lk, {("u1",): None})}, "fail"
    )
    assert enriched[0]["user_id"] == "u1"
```

`drop_match_columns` is silently honored as ignored. Docs table has it right ("ignored, source columns are always preserved"), but the default is `True`, so most check_only configs will be carrying an ignored `True` they didn't ask for. Pinning the contract:

```python
def test_check_only_ignores_drop_match_columns_setting(self) -> None:
    lk = LookupConfig(
        table="users", match={"id": "user_id"},
        check_only=True, drop_match_columns=True,
    )
    records = [{"user_id": "u1", "extra": "x"}]
    enriched, _ = apply_lookups(
        records, {"label": (lk, {("u1",): None})}, "fail"
    )
    assert enriched[0] == {"user_id": "u1", "extra": "x"}
```

(Alternative: have the validator reject explicit `drop_match_columns=True` with `check_only=True` the same way it rejects `select`. More disruptive than test-only, I'd lean test-only.)

Happy to send the runtime-branch removal plus the tests as follow-up commits against `feat/lookup-check-only-354` if that's easier than handling them in this PR.

yodakanohoshi (commented):
Agree with @Pawansingh3889's points, particularly promoting `on_miss="null"` +
`check_only=True` to a parse-time validator error — worth folding in here.

Same family of "miss-time fate is hard to predict from the YAML alone":
**`apply_lookups` is order-sensitive across lookups.** It iterates
`lookup_maps` in insertion order (`drt/destinations/lookup.py:78`), built in
YAML key order at `drt/engine/sync.py:194-197`. The first lookup to miss
`break`s the loop and its `on_miss` wins.

For a row whose `user_id` is missing in both lookup tables:

```yaml
# A — value-resolving first
lookups:
  interviewer_profile_id: { select: id, match: {user_id: user_id}, on_miss: fail }
  user_exists:            { check_only: true, match: {id: user_id}, on_miss: skip }
```
→ `fail=True`, sync fails.

```yaml
# B — check_only first (same intent)
lookups:
  user_exists:            { check_only: true, match: {id: user_id}, on_miss: skip }
  interviewer_profile_id: { select: id, match: {user_id: user_id}, on_miss: fail }
```
→ `skip=True`, row silently dropped, sync succeeds.

Possibly out of scope for this PR depending on how you want to frame
`check_only`'s contract. A few directions if it feels worth addressing:

- A docs note recommending `check_only` filters be listed before
  value-resolving lookups
- Auto-sorting `check_only` entries to the front inside `apply_lookups`
- Validating it at parse time, in the same spirit as the `on_miss=null` fix

Happy to defer to your read on which (if any) fits — flagging mainly so the
ordering dependency is on the record.

Otherwise this looks great.

masukai:
Thanks both for the careful read 🙏

## @Pawansingh3889 — adopted in `db36649`

Promoted to a parse-time validator exactly as suggested:

```python
@model_validator(mode="after")
def _check_on_miss_consistency(self) -> "LookupConfig":
    if self.check_only and self.on_miss == "null":
        raise ValueError(
            "lookups.on_miss='null' is invalid with check_only=True "
            "(no target column to set NULL on; use 'skip' or 'fail')."
        )
    return self
```

Runtime fallback at `lookup.py` L97-113 is gone — replaced with a one-line comment pointing at the validator so the next reader knows why the `null` branch can only fire for value-resolving lookups.

Test rewrite: dropped the now-untestable runtime-coercion test; added `test_check_only_with_on_miss_null_raises` (validator error) + `test_check_only_with_on_miss_skip_or_fail_valid` (positive case). Net **+2 / -1**, total 40 lookup tests.

**Codecov gap closed** — `drt/destinations/lookup.py` now at **100%** line coverage. Welcome aboard, and thanks for the strong opening review 🚀

## @yodakanohoshi — docs note added, parse-time validation deferred to #453

The order-sensitivity is real and surprising. For this PR I added a docs note in `docs/guides/destination-lookup.md` documenting the `first-miss-wins` invariant and recommending `check_only` filters be listed before value-resolving lookups when both could miss:

> **Ordering note.** When multiple lookups can miss on the same row, the **first** miss wins ... Practical guidance: list `check_only` filters **before** value-resolving lookups when both could miss, so existence-failures take precedence over join-failures.

Implementation of parse-time validation (or auto-sort) tracked in **#453** with milestone v0.8 — agree the heuristic for "potential ambiguity" needs design before implementation, and wanted to keep this PR's surface focused on `check_only` itself. Happy to revisit the call if you'd rather take it here.

CI rerun in progress; should be green with the same 100% patch coverage.

PR #450, opened 2026-05-03
feat(cli): graceful shutdown on SIGTERM/SIGINT (#279)

### LINKED ISSUE #279: feat: graceful shutdown on SIGTERM/SIGINT
## Summary

Handle SIGTERM and SIGINT gracefully during sync execution — finish the current batch, save state, then exit cleanly.

## Motivation

In containerized/orchestrated environments (Docker, K8s, Airflow), processes receive SIGTERM before being killed. Without graceful shutdown, state can be corrupted and records may be duplicated on next run.

## Scope

- Register signal handlers for SIGTERM and SIGINT
- On signal: finish current batch, save incremental state, log shutdown reason
- Exit with appropriate code (130 for SIGINT, 143 for SIGTERM)
- Timeout: force exit after 30s if batch doesn't complete

masukai:
Closes #279

## Summary

Cooperative graceful shutdown for `drt run`. SIGTERM (container stop, K8s eviction, Airflow cancel) and SIGINT (Ctrl+C) now finish the current batch, persist state and watermark, then exit with the POSIX-conventional code (130 / 143). A 30-second watchdog force-exits if the current batch hangs.

## Why this is a Production Ready theme item

In K8s / Docker / Airflow, processes get SIGTERM'd routinely (eviction, deployment rollover, task cancellation). Without graceful shutdown, drt could:
- Leave half-completed batches → duplicate records on retry
- Skip watermark persistence → re-extract already-synced data
- Exit with default Python signal handler → `KeyboardInterrupt` / SystemExit traceback in operator dashboards

This change makes drt a well-behaved citizen in those environments.

## Design

| Layer | Change |
|---|---|
| `engine/sync.py` | New `stop_event: threading.Event \| None = None` kwarg on `run_sync` (default None = backward compat). Checked at the **top** of the batch loop → finishes current batch, marks `SyncResult.interrupted=True`, drops out so the existing state-save / watermark-persist tail still runs. |
| `destinations/base.py` | New `SyncResult.interrupted: bool = False` for integrations (Dagster / Airflow / Prefect) to distinguish clean cancellation from errors. |
| `cli/main.py` | Registers SIGINT (always) and SIGTERM (when available — guarded with `hasattr(signal, "SIGTERM")` for Windows) in the `run()` command. Tracks received signal for exit code mapping. Watchdog `threading.Timer(30.0, os._exit)` for hung-batch protection. Idempotent against double-tap signals. |
| `docs/guides/graceful-shutdown.md` | New guide: behavior, exit codes, K8s example, verification steps. |

### Exit codes (POSIX convention `128 + signum`)

| Signal | Exit code |
|---|---|
| SIGINT (2) | 130 |
| SIGTERM (15) | 143 |

These are recognized as non-zero by Airflow / Prefect / shell scripts → next run resumes from the persisted watermark.

### What does *not* happen (intentional non-goals)

- **No mid-batch interruption.** drt does not abort an in-flight HTTP call or DB transaction. Half-completed batches are exactly what this feature exists to prevent.
- **No state loss on watchdog force-exit.** Batches that completed before the signal are already persisted before the watchdog fires.
- **No re-entrancy.** A second signal is a no-op — the watchdog only starts once.

## Tests

- **Engine layer (4)** — `tests/unit/test_engine.TestGracefulShutdown`:
  - `test_stop_event_none_default_no_behavior_change` (regression for existing callers)
  - `test_stop_event_set_before_first_batch_no_loads`
  - `test_stop_event_set_mid_sync_finishes_current_batch`
  - `test_interrupted_state_saved`
- **CLI layer (8)** — `tests/unit/test_cli_graceful_shutdown`:
  - `_exit_code_for_signal` mapping (3 tests including POSIX convention)
  - `signal.signal()` registration for SIGINT + SIGTERM
  - `stop_event` propagated to engine
  - SIGTERM → exit 143 via captured-handler invocation (pytest-safe; no real signal delivery)
  - SIGINT → exit 130
  - Idempotent double-tap (only one watchdog Timer started)

Real signal delivery is intentionally avoided in tests — sending SIGTERM to the pytest process can race with the harness. Tests instead capture the registered handler via `monkeypatch.setattr(signal, "signal", capture)` and invoke it directly.

Full unit suite: **779 passed, 2 skipped**. ruff + mypy clean.

## Files

- `drt/engine/sync.py` — stop_event kwarg + cooperative break
- `drt/destinations/base.py` — `SyncResult.interrupted`
- `drt/cli/main.py` — signal handlers, exit code mapping, watchdog, `_exit_code_for_signal`
- `tests/unit/test_engine.py` — 4 new tests in `TestGracefulShutdown`
- `tests/unit/test_cli_graceful_shutdown.py` — new file, 8 tests
- `docs/guides/graceful-shutdown.md` — new guide
- `CHANGELOG.md` — Unreleased / Added entry


masukai:
Rebased onto current `main` and going ahead with self-merge to keep v0.7 moving — CI green across the board (130 / 143 exit-code paths included, signal-handler captures + idempotency tested without real signal delivery so the suite stays pytest-safe). No rush on a post-merge review pass — happy to address anything via a follow-up commit if you spot something. Thanks for the review attention this week 🙏

Issue #448, opened 2026-05-03
bug: replace_strategy=swap silently ignores json_columns config (Postgres/MySQL/ClickHouse)


masukai:
## Summary

`replace_strategy=swap` (added in #338) silently ignores the `json_columns` destination config (added in #316 / #382). Discovered while rebasing #435 onto current main — the two features were developed in parallel and the swap-mode code path doesn't thread `config.json_columns` through to `_serialize_value`.

## Affected destinations

- PostgreSQL — [drt/destinations/postgres.py L257](https://github.com/drt-hub/drt/blob/main/drt/destinations/postgres.py#L257)
- MySQL — same pattern in `_load_replace_swap`
- ClickHouse — same pattern

## What goes wrong

In `truncate` strategy (default), `_load_replace` calls:

```python
values = [_serialize_value(record.get(c), c, config.json_columns) for c in columns]
```

In `swap` strategy, `_load_replace_swap` calls:

```python
values = [_serialize_value(record.get(c)) for c in columns]
```

→ When users set both `replace_strategy: swap` AND `json_columns: [...]`, the explicit JSON column declarations are silently dropped. dict/list values fall back to the legacy auto-wrap behavior, and the strict ValueError that protects against accidental JSON serialization (\"column not in json_columns\") never fires.

## Why it matters before v0.7 ships

Both features land in v0.7. A user who configures both will see different behavior between strategies on the same table, which is a correctness regression vs. either feature in isolation. Better to fix before tag.

## Suggested fix

Add `config: <Postgres|MySQL|ClickHouse>DestinationConfig` to `_load_replace_swap` signatures and pass `config.json_columns` to `_serialize_value`. Mirror the `_load_replace` pattern exactly. ~5 lines per destination.

## Test coverage

Add a swap-mode + json_columns combo test for each affected destination, mirroring the existing `TestPostgresJsonColumns` / `TestMySQLJsonColumns` shape but with `replace_strategy=\"swap\"`.

## Context

- Discovered: rebase of #435 onto current `main` (after #382 merged on 2026-05-02)
- Original PRs: #382 (json_columns), #338 (zero-downtime swap, in-flight as #435)
- Not a regression of either PR alone — it's an interaction missed because they were developed in parallel

PR #452, opened 2026-05-04
feat: use psycopg2.sql for safe SQL composition in postgres destination

Khush-domadia:
## What
- Added `import psycopg2.sql as _pgsql` to the try/except block in `postgres.py`
- Replaced raw string formatting with `psycopg2.sql.Composed` objects for safe identifier quoting
- Fixed test assertions to use `str(sql)` instead of direct `in sql` checks on Composed objects

## Why
Raw string formatting for SQL identifiers is unsafe and can cause SQL injection vulnerabilities. psycopg2.sql properly quotes identifiers at the driver level.

## Tests
All 761 tests passing, 2 skipped.

### DIFF
--- a/drt/destinations/postgres.py
+++ b/drt/destinations/postgres.py
@@ -172,7 +172,7 @@ class PostgresDestination:
         config: PostgresDestinationConfig,
     ) -> SyncResult:
         """TRUNCATE (once) → INSERT within a transaction."""
-        from psycopg2 import sql as _pgsql   # <-- ADD THIS LINE at the top of the method
+        from psycopg2 import sql as _pgsql
         result = SyncResult()
 
         if not self._replace_truncated:
--- a/drt/destinations/postgres.py
+++ b/drt/destinations/postgres.py
@@ -27,8 +27,10 @@ from drt.destinations.base import SyncResult
 from drt.destinations.row_errors import RowError
 
 try:
+    import psycopg2.sql as _pgsql
     from psycopg2.extras import Json as _Psycopg2Json
 except ImportError:
+    _pgsql = None  # type: ignore[assignment]
     _Psycopg2Json = None  # type: ignore[assignment,misc]
 
 
@@ -175,15 +177,15 @@ class PostgresDestination:
         result = SyncResult()
 
         if not self._replace_truncated:
-            cur.execute(f"TRUNCATE TABLE {table}")
+            cur.execute(_pgsql.SQL("TRUNCATE TABLE {}").format(_pgsql.Identifier(table)))
             self._replace_truncated = True
 
-        sql = self._build_insert_sql(table, columns)
+        query = self._build_insert_sql(table, columns)
 
         for i, record in enumerate(records):
             try:
                 values = [_serialize_value(record.get(c), c, config.json_columns) for c in columns]
-                cur.execute(sql, values)
+                cur.execute(query, values)
                 result.success += 1
             except Exception as e:
                 result.failed += 1
@@ -201,7 +203,7 @@ class PostgresDestination:
                 conn.rollback()
                 cur = conn.cursor()
                 if not self._replace_truncated:
-                    cur.execute(f"TRUNCATE TABLE {table}")
+                    cur.execute(_pgsql.SQL("TRUNCATE TABLE {}").format(_pgsql.Identifier(table)))
                     self._replace_truncated = True
                 continue
 
@@ -219,7 +221,7 @@ class PostgresDestination:
     ) -> SyncResult:
         result = SyncResult()
         update_cols = [c for c in columns if c not in config.upsert_key]
-        sql = PostgresDestination._build_upsert_sql(
+        query = PostgresDestination._build_upsert_sql(
             config.table,
             columns,
             config.upsert_key,
@@ -229,7 +231,7 @@ class PostgresDestination:
         for i, record in enumerate(records):
             try:
                 values = [_serialize_value(record.get(c), c, config.json_columns) for c in columns]
-                cur.execute(sql, values)
+                cur.execute(query, values)
                 result.success += 1
             except Exception as e:
                 result.failed += 1
@@ -252,11 +254,13 @@ class PostgresDestination:
         return result
 
     @staticmethod
-    def _build_insert_sql(table: str, columns: list[str]) -> str:
+    def _build_insert_sql(table: str, columns: list[str]) -> Any:
         """Build plain INSERT SQL (no conflict handling)."""
-        cols_str = ", ".join(f'"{c}"' for c in columns)
-        placeholders = ", ".join(["%s"] * len(columns))
-        return f"INSERT INTO {table} ({cols_str}) VALUES ({placeholders})"
+        return _pgsql.SQL("INSERT INTO {} ({}) VALUES ({})").format(
+            _pgsql.Identifier(table),
+            _pgsql.SQL(", ").join(_pgsql.Identifier(c) for c in columns),
+            _pgsql.SQL(", ").join(_pgsql.Placeholder() * len(columns)),
+        )
 
     @staticmethod
     def _build_upsert_sql(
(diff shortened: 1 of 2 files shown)


masukai (commented):
Thanks @Khush-domadia for opening this — using `psycopg2.sql.Composed` for identifier quoting is the right defensive direction, and matches the precedent we already follow elsewhere (e.g. `get_row_count` at [drt/destinations/postgres.py:178-180](https://github.com/drt-hub/drt/blob/main/drt/destinations/postgres.py#L178-L180)).

A few things to address before this can land. Direction is approved; the rest is mechanical.

## ⛔ Blocker — CI failure root cause

The 13 failing tests on `test (3.10)` (and the cancelled 3.11/3.12/3.13 jobs) all surface as:

```
AttributeError: 'NoneType' object has no attribute 'SQL'
```

at `drt/destinations/postgres.py:259` (`_pgsql.SQL(...)`).

**Why:** psycopg2 is in the `[postgres]` extras, not in core or dev deps — the test environment doesn't install it, so the module-level import falls into the `except ImportError` branch and `_pgsql = None`. Reproduces locally with `python -m pytest tests/unit/test_postgres_destination.py`.

The existing code worked because `_build_insert_sql` / `_build_upsert_sql` were pure f-string composition — no psycopg2 call needed even when not installed. Your change makes the SQL builders depend on `_pgsql` at call time, which breaks the test isolation.

**Suggested fix** — two options, pick whichever fits your taste:

**(a)** Mark the affected test classes with `pytest.importorskip("psycopg2.sql")` at module top (or per-class). Mirrors how we'd treat any postgres-extras-only feature.

```python
# tests/unit/test_postgres_destination.py
import pytest
pytest.importorskip("psycopg2.sql")
```

**(b)** Make the module-level import unconditional so a missing psycopg2 is a hard error from the start (not the half-working state where the destination class loads but `_build_insert_sql` crashes). This is more invasive — would shift psycopg2 to a hard dep — and probably out of scope.

I'd go with (a).

## 🔧 Likely runtime bug — `Placeholder() * len(columns)`

[postgres.py:262](https://github.com/drt-hub/drt/blob/main/drt/destinations/postgres.py#L262) and similar in `_build_upsert_sql`:

```python
_pgsql.SQL(", ").join(_pgsql.Placeholder() * len(columns)),
```

`psycopg2.sql.Placeholder` doesn't define `__mul__`, so `Placeholder() * 3` will raise `TypeError` once you get past Bug 1. The standard idiom is:

```python
_pgsql.SQL(", ").join(_pgsql.Placeholder() for _ in columns)
# or
_pgsql.SQL(", ").join([_pgsql.Placeholder()] * len(columns))
```

Worth adding a test that actually renders the SQL (not just substring-match `str(sql)`) so future readers catch the same class of issue.

## ⚠️ Timing — `postgres.py` is hot

Heads-up that `drt/destinations/postgres.py` is in active flux right now:

- **#435** (zero-downtime swap) **merged a few hours ago** — added `_load_replace_swap` with the same `f"... {table} ..."` pattern this PR cleans up. Worth covering those new sites too if you take another pass.
- **#448** (queued, v0.7) will touch `_load_replace_swap` to thread `json_columns` through. Will conflict with this PR.

To minimise rebase pain, suggest landing #448 first (it's small, ~5 lines), then this PR can rebase and cover both `_load_replace` and `_load_replace_swap` in one consistent pass. I'm happy to coordinate the order — let me know if you'd rather hold this PR or push fixes now and we'll handle the rebase together.

## ✅ Direction is right

Once Bug 1 + Bug 2 are addressed, this is a real defense-in-depth improvement — even though `table` currently comes from trusted YAML config (not user input), `psycopg2.sql.Identifier` is the textbook-correct way to compose identifiers and brings us in line with `get_row_count`'s precedent. Welcome contribution 🙌

⭐ And if drt is useful to you, please consider [starring the repo](https://github.com/drt-hub/drt) — really helps with visibility 🙏

Khush-domadia:
Fixed both blockers:

Moved from psycopg2 import sql as _pgsql inside each method (deferred import) — removed module-level try/except for _pgsql. Tests now skip cleanly without psycopg2 via pytest.importorskip.

Fixed Placeholder() * len(columns) → generator expression Placeholder() for _ in columns.

_load_replace_swap and finalize_sync are not yet in origin/main from my side — happy to cover those once #448 lands and I rebase again.

masukai:
Thanks for the fast turnaround @Khush-domadia 🙏 Both bug-1 fixes look good in concept (deferred import + generator expression for `Placeholder`), but the CI is still red — the new failure is mechanical, not design. 30 seconds to fix.

## Single remaining blocker — stray instructional comment

[`drt/destinations/postgres.py:187`](https://github.com/drt-hub/drt/blob/fix/postgres-sql-injection/drt/destinations/postgres.py#L187):

```python
from psycopg2 import sql as _pgsql   # <-- ADD THIS LINE at the top of the method
```

That trailing `# <-- ADD THIS LINE at the top of the method` looks like leftover instructional text (probably from a tool / copy-paste of review feedback). It's now sitting in the source, and ruff's `I001` check is unhappy about the import block formatting because of it. Just remove the comment so the line reads:

```python
from psycopg2 import sql as _pgsql
```

Repeat anywhere else the same pattern appears (your latest commit message hints at "deferred imports inside each method" — likely shows up in `_load_replace`, `_load_upsert`, and any swap path you covered).

## Local verification before push

```bash
make lint
# ruff check drt tests
# All checks passed!
```

Once that runs green locally, `git push --force-with-lease` and we should have green CI.

## Heads-up on rebase scope

Your comment notes "_load_replace_swap and finalize_sync are not yet in origin/main from my side" — they actually merged yesterday in #435. Your branch should rebase on current `main` to pick them up, then your `_pgsql` migration can cover those new sites in the same PR. Two commands:

```bash
git fetch origin main
git rebase origin/main
# resolve any conflicts if they appear
git push --force-with-lease
```

## Direction is unchanged

Once lint + rebase are clean, this is a straightforward security improvement. Genuine appreciation for sticking with it through two iterations 🙌

masukai:
@Khush-domadia thanks for sticking with this — the actual code changes look great:

- ✅ **Bug 2 (Placeholder)** is fixed correctly — `_pgsql.SQL(", ").join(_pgsql.Placeholder() for _ in columns)` is exactly the right idiom
- ✅ **Stray instructional comment** is gone
- ✅ Tests updated to use `str(sql)` for assertions on Composed objects

So the SQL composition itself is in good shape. The remaining CI failure is a test-environment thing, and it's my fault for not being precise enough in my earlier review when I said "deferred import." Sorry for the confusion. Let me lay out the exact remaining steps.

## What's still red — and why

1. **`ModuleNotFoundError: No module named 'psycopg2'` in `test_postgres_destination.py`**

   The deferred import fix you did (`from psycopg2 import sql as _pgsql` inside methods) covers the case where someone imports `drt.destinations.postgres` without psycopg2 installed — the module loads fine. But the tests still call `PostgresDestination().load(...)`, which calls `_build_insert_sql()`, which executes the deferred import at runtime and raises `ModuleNotFoundError` because the dev/CI environment doesn't have psycopg2 (it's in the `[postgres]` extra, not in core dev deps).

   **The fix is on the test side.** Add this at the top of `tests/unit/test_postgres_destination.py` (right after the docstring, before other imports):

   ```python
   import pytest
   pytest.importorskip("psycopg2.sql")
   ```

   This skips the entire test module cleanly when psycopg2 isn't installed. It's the standard idiom for tests that depend on optional extras.

2. **`make check-i18n` failed**

   This is just a rebase issue — `main` has moved a lot since this PR was opened (8+ merges including a Snowflake fix, my retry-tests PR, and the README/ROADMAP/i18n-sync churn from the v0.7.0 release). Once you rebase, the i18n-sync hash drift goes away.

## What you'll see after rebase

When you rebase on `main`, you'll pick up `_load_replace_swap` (the zero-downtime swap path from #435 / #448) on Postgres and MySQL — added during the v0.7.0 cycle. **Those new methods also need the same `psycopg2.sql` treatment** for consistency. The diff is small:

```python
# In _load_replace_swap (postgres.py)
cur.execute(_pgsql.SQL("DROP TABLE IF EXISTS {}").format(_pgsql.Identifier(shadow)))
cur.execute(_pgsql.SQL("CREATE TABLE {} (LIKE {} INCLUDING ALL)").format(
    _pgsql.Identifier(shadow), _pgsql.Identifier(table)
))
```

If you'd rather keep this PR's scope to just `_load_replace` / `_load_upsert` / build_sql functions and address `_load_replace_swap` in a follow-up, that's totally fine — let me know in the PR description when you push the rebase.

## Summary of next push

```bash
# 1) rebase
git fetch origin main
git rebase origin/main
# resolve any conflicts (likely just CHANGELOG.md and minor)

# 2) add the importorskip line at top of test_postgres_destination.py
# (right after the """docstring""", before the `from __future__` import)

# 3) verify locally
make lint
make test  # the postgres tests will now skip when psycopg2 isn't installed

# 4) push
git push --force-with-lease
```

Once CI is green I'll merge right away — the substance of the PR is approved. 🙏

Three iterations is honestly impressive patience for a defensive-coding refactor. Almost there.

masukai:
Merged 🙏

Thanks for sticking with this through three iterations @Khush-domadia — the `_pgsql.SQL` migration on `_load_replace`, `_load_upsert`, `_build_insert_sql`, and `_build_upsert_sql` is exactly the right hardening, and `pytest.importorskip("psycopg2.sql")` cleanly handles the optional-extra test environment.

I'm filing a small followup chore (#483) to extend the same `psycopg2.sql` treatment to `_load_replace_swap` / `finalize_sync` (the swap path added in #435 / #448), per the scope-split we discussed earlier — keeping that as a separate landing rather than retrofitting your PR. It's tiny once the pattern is in.

This SQL hardening lands in v0.7.2 alongside the telemetry feature.

PR #460, opened 2026-05-06
fix(snowflake): address bookkeeping and edge cases in destination

### LINKED ISSUE #443: Snowflake destination: align with sibling-destination conventions (resolve_env, MERGE bookkeeping, key-only edge case)
## Summary

Three stylistic / bookkeeping consistency items deferred from the #353 take-over (PFCAaron12's original Snowflake destination). None are blockers; the destination works correctly for the common path. Filing as a follow-up cleanup so #353 stays a clean \"land the contributor's connector\" PR per the same split pattern as #442.

## 1. Use `resolve_env` instead of `os.environ.get` for credentials

[`drt/destinations/snowflake.py`](drt/drt/destinations/snowflake.py) currently:

```python
account = os.environ.get(config.account_env)
user = os.environ.get(config.user_env)
password = os.environ.get(config.password_env)
```

All other destinations (`postgres`, `mysql`, `clickhouse`, `notion`, ...) use `from drt.config.credentials import resolve_env` and call `resolve_env(value, env_var_name)`. The helper is the canonical entry point for the `<value> | env-var` pattern and produces consistent error messages on missing values.

**Fix**: replace the three `os.environ.get` calls with `resolve_env(None, config.<X>_env)`.

## 2. MERGE-mode bookkeeping when staging INSERTs partially fail

```python
elif config.mode == \"merge\":
    ...
    cur.execute(f\"CREATE TEMP TABLE {staging_table} LIKE {table_fq}\")

    for row in records:                       # ← no try/except, no on_error handling
        cur.execute(f\"INSERT INTO {staging_table} ...\", list(row.values()))

    cur.execute(merge_sql)
    result.success += len(records)            # ← always reports all rows succeeded
```

Two issues:

- A row that fails to INSERT into the staging table will raise out of the loop, leaving `result.success == 0` even though earlier rows reached the staging table — but the MERGE never runs, so the live table is unchanged. The user sees `success=0, failed=0` which is misleading.
- If the project ever wants `on_error=skip` for MERGE mode, this loop has nowhere to record per-row errors.

**Fix**: mirror the INSERT-mode loop — wrap each staging INSERT in `try/except`, append to `result.row_errors`, respect `sync_options.on_error`, and at the end set `result.success += len(records) - result.failed`.

## 3. MERGE with all columns part of the key

When `upsert_key == columns` (every column is part of the merge key), the current code generates:

```sql
MERGE INTO ... target USING ... source
  ON ...
  WHEN MATCHED THEN UPDATE SET     -- empty update_clause → syntax error
  WHEN NOT MATCHED THEN INSERT (...) VALUES (...)
```

Snowflake will reject the empty `UPDATE SET`. The Postgres `_build_upsert_sql` and MySQL `_build_upsert_sql` both handle this case (`DO NOTHING` / `INSERT IGNORE`).

**Fix**: when `update_cols` is empty, drop the `WHEN MATCHED` clause entirely so the MERGE only inserts on no-match.

## Test additions

Add tests covering each fix:

- `resolve_env` parity → uses the `secrets.toml` fallback path (other destination tests already do this; copy the pattern from `test_postgres_destination.py`).
- MERGE staging INSERT partial-fail with `on_error=skip` → asserts `result.failed == 1`, `result.success == n - 1`, MERGE still runs.
- MERGE all-columns-are-key → asserts the SQL contains \"WHEN NOT MATCHED\" but not \"WHEN MATCHED THEN UPDATE\".

## Acceptance criteria

- [ ] All three fixes applied to [`drt/destinations/snowflake.py`](drt/drt/destinations/snowflake.py).
- [ ] New tests in `tests/unit/test_snowflake_destination.py` covering each.
- [ ] `make lint` + `make test` clean.

GokulKashyap:
## What does this PR do?

This PR addresses the three cleanup items for the Snowflake destination connector to bring it to full parity with the other destinations:
1. **Credential Resolution**: Swapped out `os.environ.get` for the standard `resolve_env` helper so it properly falls back to `.drt/secrets.toml`.
2. **MERGE Partial Failures**: Wrapped the row-by-row staging inserts in a `try/except` block to correctly log a `RowError` and continue executing the final MERGE if `on_error` is set to `skip`.
3. **Empty Update Clauses**: Updated the SQL generator to drop the `WHEN MATCHED THEN UPDATE SET...` clause if all columns are used in the `upsert_key`, preventing Snowflake syntax errors.

## Related Issue

Closes #443 

## Checklist

- [x] Updated `CHANGELOG.md` (if user-facing change)


GokulKashyap:

yodakanohoshi (dismissed):
Thanks for picking this up, @GokulKashyap — and especially for mirroring the existing INSERT-mode loop so cleanly. It makes the MERGE path easy to read alongside its sibling.

Verified locally:
- `resolve_env` swap is consistent with the other destinations
- MERGE staging `try/except` correctly tracks `RowError` / `on_error` / `len(records) - result.failed`
- `WHEN MATCHED` is dropped when `update_cols == []`, and the captured SQL is valid Snowflake (at least one of MATCHED / NOT MATCHED clauses is required, which is satisfied)
- `pytest tests/unit/test_snowflake_destination.py` → 13 passed

All three items from #443 are fully addressed. Approving — will squash-merge shortly. Thanks again!

yodakanohoshi:
@GokulKashyap apologies — I approved this and said I'd squash-merge shortly without first confirming the merge state. There's actually a CHANGELOG.md conflict to resolve before we can land it. Two entries merged into main between when this branch was opened and now: #386 (BasicAuth `secrets.toml` fix) and #448 (`replace_strategy=swap` × `json_columns`), both in the same `[Unreleased] / Fixed` section as your Snowflake entry.

The fix is just keeping all three entries side-by-side in CHANGELOG — your code changes are unaffected.

If you'd like to handle it, the steps are:

```
git fetch upstream
git rebase upstream/main
# resolve CHANGELOG.md to keep all three Fixed entries (BasicAuth, replace_strategy, your Snowflake one)
git add CHANGELOG.md
git rebase --continue
git push --force-with-lease
```

Or if you'd prefer I push the rebase to your branch directly, happy to — you've enabled `Allow edits by maintainers`, so it's a couple of commands on my end. Whichever is easier for you.

This is queued for v0.7.1 so there's no rush. Sorry again for the premature merge promise, and thanks for your patience 🙏

GokulKashyap:
@yodakanohoshi  No worries at all! I actually just resolved the CHANGELOG.md conflict via the GitHub web UI a few minutes ago. All three entries are sitting side-by-side perfectly now. It should be green and ready for you to squash-merge!

yodakanohoshi (approved):
@GokulKashyap thank you so much for the lightning-fast turnaround! Our comments crossed paths in the timestamps — you'd already pushed the merge commit before my earlier comment landed, so please disregard that one.

Confirmed the resolution is exactly right:

- All three Fixed entries preserved side-by-side in CHANGELOG (BasicAuth, replace_strategy, Snowflake)
- Snowflake code intact, no churn from the merge
- `pytest tests/unit/test_snowflake_destination.py` → 13 passed locally on the merged branch

Merge commit vs rebase: either works perfectly for us — we squash-merge so the final history collapses to a single commit either way.

Re-approving now and will squash-merge shortly. Genuinely impressed by the speed — thanks for both the original cleanup and the quick conflict resolution 🙏

GokulKashyap:
@all-contributors please add @GokulKashyap for code, test

PR #478, opened 2026-05-07
feat(validate): add deprecation warning detection for deprecated sync keys (#467)

Muawiya-contact:
## What does this PR do?

Implements **VERSIONING.md Step 2 ("Add Tooling Support")** — `drt validate` now warns you when your YAML config uses deprecated keys, tells you what to use instead, and links you to the migration guide.

Before this, `drt validate` would silently accept deprecated config keys. Now it doesn't.

## Changes

- **`drt/deprecations.py`** (new) — centralized registry of deprecated features. Add new deprecations here going forward.
- **`drt/config/parser.py`** — `load_syncs_safe()` now checks configs against the registry and collects warnings alongside errors
- **`drt/cli/main.py`** — `drt validate` shows ⚠️ warnings in text output and includes a `deprecations` field in `--output json`. Exit code stays `0` — warnings are non-blocking.
- **`docs/migration/v0.6-to-v0.7.md`** (new) — real migration guide for `batch_size` → `batch_config.size`
- **`docs/migration/v0.4-to-v0.5.md`** (new) — placeholder
- **`tests/unit/test_deprecations.py`** (new) — 12 tests, all passing

## Example output

```
⚠️  sync.batch_size is deprecated (removed in v0.7.0)
     Use sync.batch_config.size instead.
     See docs/migration/v0.6-to-v0.7.md
```

## Checklist

- [x] Tests pass (`make test`)
- [x] Linter passes (`make lint`)
- [x] Updated `CHANGELOG.md`

## Related

Closes #467
Parent: #459

PR #474, opened 2026-05-07
Feat/rest api source

GokulKashyap:
## What does this PR do?

Implements a new source connector to pull data directly from external HTTP 
endpoints  
- Reuses existing `auth.py` models (Bearer, API key, Basic, OAuth2)
- Reuses existing pagination models (Offset, cursor, link-header)
- Adds a `result_path` config to extract records using dot-notation
- Registers `rest_api` in the connectors registry
- Updates CHANGELOG.md

## Related Issue

Closes #422 

## Checklist

- [x] Tests pass (`make test`)
- [x] Linter passes (`make lint`)
- [x] Updated `CHANGELOG.md` (if user-facing change)

Issue #422, opened 2026-04-29
feat: add REST API source — pull from HTTP endpoints


yodakanohoshi:
## Motivation

All current sources are SQL databases. A REST API source would allow syncing data from external HTTP APIs directly — enabling API→destination pipelines without requiring a DWH intermediary. This mirrors the existing `rest_api` destination and uses similar auth/pagination config.

## Proposed config

```yaml
source:
  type: rest_api
  url: https://api.example.com/v1/contacts
  auth:
    type: bearer
    token: ${API_TOKEN}
  pagination:
    type: cursor
    cursor_param: page_token
    cursor_path: next_page_token
  result_path: data           # JSONPath to the records array
```

## Scope

- Bearer / API key / Basic auth (reuse `auth.py` models)
- Offset, cursor, and link-header pagination (reuse existing pagination models)
- `result_path` JSONPath to extract the records array
- Returns `Iterator[dict]` — compatible with existing `Source` protocol
- `drt sources` registry entry

## Out of scope

- GraphQL source
- WebSocket / streaming sources
- Response caching

PR #357, opened 2026-04-18
feat/drt test validators 233

Muawiya-contact:
## What does this PR do?

Adds three new test types to `drt test` for checking data quality after syncs:

- **freshness**: Checks that data is recent (e.g., "no data older than 7 days")
- **unique**: Checks for duplicate values in columns (e.g., "id column has no duplicates")
- **accepted_values**: Checks that values are in an allowed list (e.g., "status is only active/inactive/pending")

Supports human-readable time formats like "7 days", "24 hours", "30 minutes", etc.

## Related Issue

Closes #233

## Checklist

- [x] Tests pass (`make test`)
- [x] Linter passes (`make lint`)
- [x] Updated `CHANGELOG.md` (if user-facing change)

PR #484, opened 2026-05-11
feat: add --check-connection flag to drt validate

### LINKED ISSUE #367: good first issue: add connection test to drt validate for SQL destinations
## Summary

Enhance `drt validate` to optionally test the destination connection (e.g., `SELECT 1` for SQL destinations) to catch config errors before `drt run`.

## Scope

- Add `--check-connection` flag to `drt validate`
- For SQL destinations (Postgres, MySQL, ClickHouse): attempt a lightweight query
- For non-SQL: skip gracefully
- Report pass/fail per sync

GokulKashyap:
## What does this PR do?

This PR implements a new --check-connection flag for the drt validate command.
Added test_connection() method to the Destination protocol.
Implemented lightweight connectivity checks (SELECT 1) for PostgreSQL, MySQL, ClickHouse, and Snowflake.
Added graceful skipping for non-SQL destinations (e.g., Slack, REST API).
Enhanced CLI output to provide clear pass/fail/skip feedback per sync.
Included comprehensive unit tests with mocks for all scenarios.

## Related Issue

Closes #367 

## Checklist

- [x] Updated `CHANGELOG.md` (if user-facing change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- **Connection test in `drt validate`** (#367): Added `--check-connection` flag to test connectivity to SQL destinations (PostgreSQL, MySQL, ClickHouse, Snowflake) before running syncs. Reports pass/fail per sync; non-SQL destinations are skipped gracefully.
+- **Support for Generic REST API source** (#357): Added `RestApiSource` for streaming data from paginated REST APIs. Supports `link-header` and `query-param` pagination, bearer token auth, and dynamic JSON path extraction.
 - **Deprecation warnings in `drt validate`** (#478, closes #467): The `validate` command now reads `drt/deprecations.py` and surfaces ⚠️ warnings for any deprecated config keys it finds in your sync YAMLs — both in text output and as a per-sync `deprecations` array under `--output json`. Exit code stays `0` (warnings are non-blocking). The registry is currently empty (no active deprecations); add entries to `DEPRECATED_SYNC_KEYS` when announcing a new deprecation per VERSIONING.md Step 1. Migration guides live under `docs/migration/`. Closes the Step 2 ("Add Tooling Support") TODO from #457.
 
 ## [0.7.0] - 2026-05-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,8 +37,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+
 - **Connection test in `drt validate`** (#367): Added `--check-connection` flag to test connectivity to SQL destinations (PostgreSQL, MySQL, ClickHouse, Snowflake) before running syncs. Reports pass/fail per sync; non-SQL destinations are skipped gracefully.
 - **Deprecation warnings in `drt validate`** (#478, closes #467): The `validate` command now reads `drt/deprecations.py` and surfaces ⚠️ warnings for any deprecated config keys it finds in your sync YAMLs — both in text output and as a per-sync `deprecations` array under `--output json`. Exit code stays `0` (warnings are non-blocking). The registry is currently empty (no active deprecations); add entries to `DEPRECATED_SYNC_KEYS` when announcing a new deprecation per VERSIONING.md Step 1. Migration guides live under `docs/migration/`. Closes the Step 2 ("Add Tooling Support") TODO from #457.
+
+## [0.7.2] - 2026-05-11
+
+**Theme: Production Ready follow-up #2.** Opt-in anonymous telemetry, deprecation warnings in `drt validate`, Postgres `psycopg2.sql` hardening — closing out the v0.7 cycle items that didn't make v0.7.1.
+
+### Breaking Changes
+
+None. Drop-in upgrade from v0.7.1.
+
+### Added
+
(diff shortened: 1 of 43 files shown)


GokulKashyap:
I have implemented the --check-connection feature using a dynamic runtime check (getattr) in the CLI instead of adding a mandatory test_connection method to the Destination protocol in base.py.

Rationale:

Stability: Modifying the base Protocol would have forced us to update over 35 existing mock objects in the unit test suite, creating significant 'noise' and risk of regression.
Architectural Flexibility: Not all destinations (e.g., Slack, Local File, Discord) require a connectivity test. This 'opt-in' approach allows SQL destinations to provide validation logic without burdening non-SQL connectors with empty boilerplate methods.
Cleanliness: This keeps the core interface lean and avoids forcing 'no-op' implementations in every future connector."

masukai:
Thanks for the contribution @GokulKashyap 🙌 — solid first-mover work on #367. Core implementation looks good (`SELECT 1` per SQL destination, graceful skip for non-SQL, output formatter wired up). Before merge I'd like two blockers addressed and a couple of nice-to-haves are optional.

## 🚫 Blockers

### 1. CHANGELOG has a phantom entry that doesn't belong to this PR

```diff
+ - **Connection test in `drt validate`** (#367): Added `--check-connection`...
+ - **Support for Generic REST API source** (#357): Added `RestApiSource`...
```

The second entry mentions a feature (REST API source) whose code is not in this PR — the diff only touches CLI / Postgres / MySQL / ClickHouse / Snowflake / Twilio destinations. Also, `#357` is the wrong issue/PR reference (#357 is the unrelated merged "test validators" PR; the REST API source lives in #422 / your PR #474).

Please remove the REST API source line from this PR's CHANGELOG. The REST API source CHANGELOG entry should land via #474 when that PR is finalized, with the correct issue reference.

### 2. `--check-connection` is missing from JSON output mode

[`drt/cli/main.py`](drt/cli/main.py) flow:

```python
if output == "json":
    # ... emits results with `valid` + `deprecations` per sync ...
    return       # ← early-return here

# ... text-mode handling ...
if check_connection:
    # ← only runs in text mode
```

The deprecation warnings (#478) added `deprecations: [...]` to the JSON-mode `results` array — `--check-connection` should follow the same pattern with a `connection_test: {success, error, skipped}` (or similar) field per sync. Otherwise `drt validate --output json --check-connection` is silent on the connection test, which breaks CI/automation use cases. Test coverage in `test_cli_validate_connection.py` is text-only and doesn't catch this — please add a JSON-mode test case while you're in there.

## 🔧 Nice-to-haves (not blocking; can fold into the same fix-up commit if convenient)

### 3. Snowflake `test_connection` duplicates `load()` connection logic

The credential resolution + `snowflake.connector.connect(...)` boilerplate is repeated. Postgres extracts a `_connect()` helper — same pattern in Snowflake would tidy this up. Acceptable to leave as-is and address in a followup if you'd rather keep scope tight.

### 4. `drt/destinations/snowflake.py` lost its trailing newline

`\ No newline at end of file` shows up in the diff at the bottom of snowflake.py. Likely an editor save quirk. Just re-save with a final `\n`.

### 5. Non-SQL skip case uses `success=False` semantics

```python
print_connection_test_result(sync.name, success=False, skip=True)
```

The text output probably renders this fine, but a clearer model is `success=None` (or a dedicated `skipped` status) so JSON consumers can distinguish "tried and failed" from "skipped because non-SQL". Tied to the JSON mode gap above — consider while threading the JSON shape.

## Scope note for future PRs

The Twilio change in this PR is just a trailing-newline fix — fine to keep here, but next time consider splitting unrelated cleanups into their own PR so this PR's diff stays scoped to the issue it closes.

## Next step

Push the two blocker fixes plus whichever nice-to-haves you want; once CI re-greens I'll do another pass and merge. CHANGELOG fix is mechanical, JSON mode shape mirrors the deprecation pattern — should be quick.

Apologies for the wait — the #487 collision review with @Photon101 (now redirected) and the v0.7.2 release crunch slowed turnaround. Your PR was the first-mover here and the implementation deserves the merge once these two get cleared.

GokulKashyap:
Good catch @masukai.
I have made the following changes : 
- Implement `--check-connection` for `drt validate --output json`
- Refactor Snowflake connection logic to remove duplication via `_connect()`
- Differentiate between connection failure and "skipped" (non-SQL) status
- Fix missing trailing newline in Snowflake destination
- Clean up CHANGELOG.md by removing unrelated entries
- Add unit tests for connection validation in JSON mode

masukai (approved):
Both blockers cleared cleanly, and the JSON path now carries `connection_test` per sync — exactly the shape I wanted. CI green across 3.10–3.13 + codecov. Approving.

A few notes (none blocking):

1. **`getattr(dest, "test_connection", None)` dynamic dispatch** — your rationale on #484 was solid: keeping `test_connection` off the `Destination` Protocol avoids touching 35+ destinations and their mock tests. Good trade-off. I'll open a follow-up `design:` issue to revisit promoting it to the Protocol once we have a story for non-SQL destinations (REST API webhooks could realistically support `test_connection` too).

2. **CHANGELOG.md formatting** — couple of stray blank lines around `## [Unreleased]` and before `## [0.7.0]`. I'll tidy that up in a follow-up commit on main.

3. **Snowflake `_connect()` extraction** — clean refactor; the credential-resolution path is preserved bit-for-bit. Nothing to change.

Thanks @GokulKashyap 🙌 — this is the first contributor PR landing the connection-check primitive for the v0.7 reliability track. Quality of the test suite (6 CLI cases + per-destination connection tests) was the deciding factor here.

---

If drt has been useful to you, a ⭐ on https://github.com/drt-hub/drt would mean a lot. Feel free to share feedback in https://github.com/drt-hub/drt/discussions.

PR #485, opened 2026-05-11
chore(postgres): extend psycopg2.sql migration to swap path (fixes #483)

### LINKED ISSUE #483: chore(postgres): extend psycopg2.sql migration to swap path (followup from #452)
## Background

PR #452 (merged in v0.7.2) migrated the Postgres destination's `_load_replace`, `_load_upsert`, `_build_insert_sql`, and `_build_upsert_sql` from raw f-string SQL to `psycopg2.sql.SQL` / `psycopg2.sql.Identifier` for safe identifier quoting. The scope was deliberately tight — `_load_replace_swap` and `finalize_sync` (added later via #435 / #448 for the zero-downtime swap path) were excluded as a follow-up to keep the PR small.

This issue tracks completing the migration on the swap path so the entire Postgres destination uses one composition style.

## Goal

Apply the same `psycopg2.sql` treatment to the remaining swap-path SQL in `drt/destinations/postgres.py`:

- `_load_replace_swap`: the `DROP TABLE IF EXISTS {shadow}` and `CREATE TABLE {shadow} (LIKE {table} INCLUDING ALL)` pre-amble currently uses f-string interpolation
- `finalize_sync`: the `ALTER TABLE {shadow} RENAME TO {table}` (transactional swap) currently uses f-string interpolation
- Any other raw-SQL helpers introduced by #435 / #448 that thread table/identifier names through f-strings

## Implementation sketch

Pattern lifted directly from #452:

```python
from psycopg2 import sql as _pgsql  # deferred import inside method

cur.execute(_pgsql.SQL("DROP TABLE IF EXISTS {}").format(_pgsql.Identifier(shadow)))
cur.execute(_pgsql.SQL("CREATE TABLE {} (LIKE {} INCLUDING ALL)").format(
    _pgsql.Identifier(shadow), _pgsql.Identifier(table)
))
cur.execute(_pgsql.SQL("ALTER TABLE {} RENAME TO {}").format(
    _pgsql.Identifier(shadow), _pgsql.Identifier(table)
))
```

Tests in `tests/unit/test_postgres_destination.py` should assert against `str(query)` shape using the existing `pytest.importorskip("psycopg2.sql")` guard — same approach as the tests landed in #452.

## Acceptance

- [ ] All swap-path SQL in `drt/destinations/postgres.py` uses `psycopg2.sql.SQL` / `Identifier` (no remaining f-string interpolation of table names)
- [ ] Unit tests cover the swap-path SQL composition (matching the test pattern from #452)
- [ ] `make lint` clean (ruff + mypy)
- [ ] `make test` passes locally (885+ tests)
- [ ] Optional: a tiny CHANGELOG note under Unreleased that the swap path is hardened consistent with #452

## Why this wasn't done in #452

The contributor (@Khush-domadia) opened #452 against an earlier `main` that didn't yet have the swap path; rather than expand scope mid-review, we agreed to land the SQL composition style across the original four entry points and follow up on the swap path separately. This issue is the follow-up.

## Related

- PR #452 (psycopg2.sql migration — merged in v0.7.2)
- PR #435 / #448 (zero-downtime swap path that introduced the f-string interpolation now pending hardening)
- Issue #442 (original request to switch Postgres destination to psycopg2.sql)

Photon101:
## Summary

Completes the psycopg2.sql migration started in #452 by extending safe SQL composition to the Postgres destination's replace-with-swap path.

## Changes

- `drt/destinations/postgres.py`: All swap-path SQL (DROP TABLE IF EXISTS, CREATE TABLE LIKE, INSERT, ALTER TABLE RENAME, cleanup DROP) now uses `psycopg2.sql.SQL` + `Identifier` instead of f-strings.  
- `tests/unit/test_postgres_destination.py`: Updated test assertions to work with composable SQL objects — uses `str(query)` comparison and verifies composable objects are being sent (not raw strings).  

## Verification

```bash
# Targeted Postgres unit tests
uv run pytest tests/unit/test_postgres_destination.py -v

# Lint
uv run ruff check drt/ tests/

# Type checking
uv run mypy drt/
```

All pass locally.

Closes #483

### DIFF
--- a/drt/destinations/postgres.py
+++ b/drt/destinations/postgres.py
@@ -233,12 +233,20 @@ class PostgresDestination:
         config: PostgresDestinationConfig,
     ) -> SyncResult:
         """Build a shadow table per sync; atomic rename happens in finalize_sync."""
+        from psycopg2 import sql as _pgsql
         result = SyncResult()
         shadow = f"{table}__drt_swap"
 
         if not self._swap_shadow_created:
-            cur.execute(f"DROP TABLE IF EXISTS {shadow}")
-            cur.execute(f"CREATE TABLE {shadow} (LIKE {table} INCLUDING ALL)")
+            cur.execute(
+                _pgsql.SQL("DROP TABLE IF EXISTS {}").format(_pgsql.Identifier(shadow))
+            )
+            cur.execute(
+                _pgsql.SQL("CREATE TABLE {} (LIKE {} INCLUDING ALL)").format(
+                    _pgsql.Identifier(shadow),
+                    _pgsql.Identifier(table),
+                )
+            )
             self._swap_shadow_created = True
             self._swap_table = table
 
@@ -263,7 +271,11 @@ class PostgresDestination:
                     conn.rollback()
                     # Cleanup shadow on hard fail
                     cur = conn.cursor()
-                    cur.execute(f"DROP TABLE IF EXISTS {shadow}")
+                    cur.execute(
+                        _pgsql.SQL("DROP TABLE IF EXISTS {}").format(
+                            _pgsql.Identifier(shadow)
+                        )
+                    )
                     conn.commit()
                     self._swap_shadow_created = False
                     self._swap_table = None
@@ -279,6 +291,8 @@ class PostgresDestination:
         sync_options: SyncOptions,
     ) -> SyncResult | None:
         """Atomic rename: original->old, shadow->original, drop old."""
+        from psycopg2 import sql as _pgsql
+
         if not self._swap_shadow_created or self._swap_table is None:
             return None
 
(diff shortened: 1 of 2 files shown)


Photon101:

Photon101:
recheck

masukai:
Merged 🙏

Thanks @Photon101 — closes #483 cleanly. Scope matched the issue exactly: DROP IF EXISTS / CREATE LIKE / ALTER RENAME / DROP all migrated to `psycopg2.sql.SQL` + `Identifier`, plus the cleanup-after-failure path. Tests use the `_query_text()` helper consistently and assert that no raw f-string fragments slip through.

One small note for future test maintenance (not blocking — keeping the merge moving): the `execute_side_effect` matcher `_query_text(sql).startswith("Composed([SQL('INSERT INTO")` relies on the internal `repr` of `psycopg2.sql.Composed`, which could shift between psycopg2 versions. If this ever breaks on a psycopg2 bump, the simplest fix is matching on `"INSERT INTO"` substring instead. Filing a mental TODO rather than asking for a change here.

This lands as part of the v0.7.x hardening series. Postgres destination's SQL composition is now consistent across replace / upsert / swap paths.

PR #492, opened 2026-05-12
 feat: orphan shadow table cleanup for swap strategy

### LINKED ISSUE #447: design: list_orphan_swap_tables() Protocol signature
### What problem does this solve?

#433 requires a new Protocol method so destinations can expose 
orphan shadow tables for cleanup. Before implementing, we need 
consensus on the signature — it affects every destination that 
supports replace_strategy: swap.

### Proposed solution

Two candidate signatures:

# Option A — simple, keeps Protocol minimal
`def list_orphan_swap_tables(self, config: DestinationConfig) -> list[str]: ...`

# Option B — richer records for CLI formatting
`def list_orphan_swap_tables(self, config: DestinationConfig) -> list[OrphanShadow]: ...`

Leaning Option A — --older-than filtering can happen inside 
the destination (age detection is DB-specific anyway), and CLI 
can fan out a per-table size query if needed.

### Category

New source connector

### Alternatives considered

Option B — but adds Protocol complexity early. Can always 
upgrade later if CLI needs richer output.

### Example YAML (if applicable)

```yaml

```

Muawiya-contact:
### What does this PR do?

Adds a safe cleanup mechanism for orphan `__drt_swap` tables created when swap-based replacements are interrupted.

This ensures leftover shadow tables do not accumulate in production environments.

### Changes included

- Adds `list_orphan_swap_tables()` to destination protocol
- Implements Postgres support for detecting orphan shadow tables
- Introduces `drt clean --orphans` CLI command
- Dry-run is default (safe mode)
- `--execute` explicitly drops orphan tables
- Strict rule: only tables ending in `__drt_swap` are affected
- Fixes transaction safety for per-table drop operations
- Minor CLI improvements (comment + caching)

### Safety guarantees

- No deletion happens by default (dry-run only)
- No impact on non-shadow tables
- Each table drop is handled independently
- Failures on one table do not affect others

### Related Issue

Closes #447  
Depends on #433

### Checklist

- [x] Dry-run safe by default
- [x] Only `__drt_swap` tables targeted
- [x] Postgres implementation added
- [x] Protocol signature unchanged (approved design)
- [x] Tests passing
- [x] No breaking changes


Muawiya-contact:
Hi @masukai 👋
This is just a draft PR any feedback on it? Just leave the comment whenever you're free.
Thanks.

Muawiya-contact:
I will fix the CL tests' time to sleep. Just check the core logic and give feedback.

masukai (commented):
Solid design direction @Muawiya-contact 🙏 — splitting `OrphanCleanup` out as a separate `Protocol` instead of bolting it onto `Destination` is exactly the right call (keeps the existing 35+ destinations valid as-is). The per-table commit isolation and the `LIKE ... ESCAPE '\\'` fix for the underscore wildcard are clean responses to the Copilot review.

That said, here's what I need before this is merge-ready.

## 🚫 Blockers

### 1. CHANGELOG.md is overwriting release history

The diff deletes the **entire `## [0.7.2]` and `## [0.7.1]` sections** (~40 lines of release notes), leaving `[Unreleased]` followed directly by `## [0.7.0]`. This must not land — it would erase the public changelog for two already-shipped releases.

Looks like the branch was cut from a pre-0.7.1 `main` and hasn't been rebased. The new orphan-cleanup entry should be added under `[Unreleased]` only, leaving every existing release section untouched.

```bash
git fetch origin
git rebase origin/main
# resolve CHANGELOG.md by keeping every release section and adding your new entry under [Unreleased]
```

### 2. Cleanup scope is global, not per-sync

`list_orphan_swap_tables` in `drt/destinations/postgres.py:274` queries `information_schema.tables` for everything matching `%\_\_drt_swap` across **all non-system schemas in the DB**, not scoped to the current sync's base table. Every sync that points at the same Postgres DB will report every other sync's swap shadows as orphans, and `drt clean --orphans --execute` will happily drop them all.

Concrete failure: imagine `sync_users` and `sync_orders` both target the same DB; `sync_users` is mid-swap with a live `users__drt_swap` shadow. A `drt clean --orphans --execute` invoked from `sync_orders` will list `users__drt_swap` and drop it, breaking the in-flight `sync_users` write.

Two ways out, your call:

- **(a) Scope by the sync's base table** — orphan candidates restricted to `{sync.destination.table}__drt_swap`. Smallest blast radius. My recommendation.
- **(b) Require `--older-than HOURS`** at the CLI level as a floor — operator-explicit safety. But this currently no-ops on Postgres (creation timestamp not in catalog), so it'd need to log a clear warning rather than silently passing through.

I'd go with **(a)** and keep (b) as a follow-up. Either way, the multi-sync scenario above must not be possible by default.

### 3. CI is red

`test (3.10)` is failing; the rest of the matrix was cancelled. Please get CI green before re-review. You mentioned the sleep timing in CL tests — happy to dig in if the actual failure is something different.

Also: I don't see `_get_destination` being added in this PR, but `drt/cli/main.py:108` calls it. If it's an existing helper, fine — if it was assumed to exist, that may be the CI failure.

## 🟡 Suggestions (non-blocking)

- `test_clean_command.py` uses `mock.patch("drt.cli.main.isinstance", return_value=True)` to bypass the `OrphanCleanup` Protocol check. Mocking `isinstance` makes the test pass regardless of the real Protocol behavior — it'd be stronger to construct a fake destination that actually implements `list_orphan_swap_tables` / `drop_orphan_swap_tables`, so the Protocol check exercises its real path.

- The `execute` phase in `drt/cli/main.py:147-169` looks up `dest_to_use` by iterating `syncs` and picking the first `OrphanCleanup`-typed destination. If scope-fix (a) lands, this loop simplifies considerably (one sync = one orphan set).

- Codecov flags 10.56% patch coverage (110 lines uncovered). Most of that is the new `clean` body and the Postgres orphan methods. Once `isinstance` mocking is dropped (above), coverage should rise organically.

## Out of scope (FYI)

The `google_sheets.py` `# type: ignore[no-untyped-call]` change is unrelated to this PR. Not a blocker, but worth flagging — scoped PRs are easier to revert if needed.

---

To summarize: design is right, implementation details mostly improved since Copilot's pass, the **scope-blast-radius problem (#2)** is the one architecturally significant thing left to settle. Once that's nailed + CHANGELOG restored + CI green, I can re-review fast.

Thanks for sticking with this 🙏

Muawiya-contact:
# 🙏 Help Needed — CI Still Failing After Latest Fixes

Hi @masukai 👋

Thank you for the detailed review on my last push. I've been working through your blockers, but **CI is still red** across all Python versions (3.10–3.13) with the same 5 test failures. I've been going in circles and would really appreciate your guidance on what I'm missing.

---

## What's Still Failing

### 1. `test_clean_help_shows_options` — `tests/unit/test_clean_command.py`

The test asserts `--orphans` appears in `drt clean --help` output, but the option isn't showing up:

```
AssertionError: assert '--orphans' in <help output>
```

The `clean` command exists, but the `--orphans` flag I added doesn't appear to be registering in the CLI help text. I'm not sure if it's a Typer wiring issue or if there's a specific pattern used elsewhere in the codebase I should be following.

---

### 2. `TestDropOrphanSwapTables` — `tests/unit/test_orphan_cleanup.py` (4 tests)

All four tests call `dest.drop_orphan_swap_tables(config, [...])` and expect a `(dropped, failed)` tuple back, but the method is returning `([], [])` in all cases:

- `test_drop_orphan_tables_drops_valid_entries`
- `test_drop_orphan_tables_commits_each_table_independently`
- `test_drop_orphan_tables_rejects_malformed_entries`
- `test_drop_orphan_tables_rejects_non_swap_tables`

The tests mock `dest._connect` and `conn.cursor`, so the connection isn't the issue. The logic for validating table names (`schema.table` format + `__drt_swap` suffix) and per-table independent commits doesn't seem to be executing properly.

---

## My Questions

1. **Is there a pattern in the existing codebase** for how Typer options are registered on subcommands that I might be doing differently?

2. **For `drop_orphan_swap_tables`** — is there something about how `_connect` is expected to be called (e.g., it should be used as a context manager, or returns a specific type) that the mock setup in the tests assumes? I want to make sure my implementation matches the expected call signature.

3. **Should I open a separate issue** for tracking these test failures specifically, or should I keep iterating on this PR? I don't want to pollute the PR thread further if a focused issue would be more helpful.

---

## What I've Already Tried

- Added `--orphans` and `--execute` as `typer.Option` parameters to the `clean` command
- Implemented `drop_orphan_swap_tables` on `PostgresDestination` with per-table commit isolation and the `schema.table` + `__drt_swap` validation guards
- Rebased onto `main` (CHANGELOG conflict resolved, release history preserved)
- Scoped orphan detection to the current sync's base table (per your blocker #2)

---

Any pointers — even just pointing me to a similar method or command in the codebase I can model after — would be a huge help. Happy to jump on a quick call too if that's easier.

Thanks again for your patience 🙏

Muawiya-contact:
# Hey @masukai 👋

Hope you're doing well! Just a small update on **PR #492**.

I've been grinding through the issues you flagged and wanted to
give you a heads-up on where things stand:

- ✅ CHANGELOG restored — 0.7.1 and 0.7.2 sections are intact
- ✅ Orphan scope narrowed to the current sync's base table
- ⚠️ CI still red — I'm stuck on the two test failures and
  would really appreciate a small nudge in the right direction

Specifically:
- The `--orphans` flag isn't registering in Typer help output
- `drop_orphan_swap_tables` keeps returning `([], [])` in tests

No pressure at all — whenever you have a free moment,
even just pointing me to a similar pattern in the codebase
would go a long way.

Really appreciate your patience with me on this one 🙏

— Muawiya

masukai:
@Muawiya-contact Sorry for the slow turnaround on this — thanks for your patience 🙏 I pulled the branch locally and dug into both failures; here's what I found.

## 1. `test_clean_help_shows_options` — actually passing now ✅

I ran the latest `464ca6c` locally and `--orphans` does show up in `drt clean --help`. `test_clean_command.py` is 5/5 green for me. The CI checks shown on this PR are still from 5/14 (commit `51a5141`) — your 3 commits since the merge from `main` haven't kicked a fresh CI run. A re-push (or empty commit) should re-trigger CI and you'll likely see this one green.

## 2. `TestDropOrphanSwapTables` (4 failures) — root cause found 🔴

The real cause is in `drt/destinations/postgres.py:497-499`:

```python
_sql = sql
if _sql is None:
    from psycopg2 import sql as _sql
```

CI installs only `.[dev]` (see `.github/workflows/ci.yml`), so `psycopg2` is **not available** in the test environment. The lazy import raises `ImportError`, the outer `except Exception` catches it, and every table gets appended to `failed` — which is exactly why you see `dropped == []`.

The repo already has a convention for this. Look at the top of `tests/unit/test_postgres_destination.py:10`:

```python
import pytest

pytest.importorskip("psycopg2.sql")
```

Add the same two lines to the top of `tests/unit/test_orphan_cleanup.py`. The 4 tests will skip in CI (matching `test_postgres_destination.py`'s behavior) and run locally when psycopg2 is installed.

If you'd prefer they run in CI too, that's a separate ci.yml change to install `.[postgres]` for the test step — but for this PR, matching the existing pattern keeps the scope tight.

---

Re-push with the `importorskip` line added and I'll re-review quickly. Almost there 🙏

Muawiya-contact:
Hey @masukai 👋
CI is green across all Python versions (3.10–3.13) ✅
Root cause of `--orphans` failure: Rich was injecting ANSI color codes into the help output, breaking the plain string search. Fixed by stripping escape codes before asserting in the test.
All blockers resolved:

- ✅ CHANGELOG restored (`0.7.1` and `0.7.2` intact)
- ✅ Orphan scope narrowed to current sync's base table
- ✅ psycopg2 tests skip correctly in CI
- ✅ Temp scripts removed

Ready for re-review whenever you're free 🙏
— @Muawiya-contact

masukai:
Thanks for the persistence on this @Muawiya-contact 🙏 The orphan-cleanup feature itself is in great shape — the `OrphanCleanup` protocol split, the parameterized catalog query, and the per-table commit isolation in `drop_orphan_swap_tables` all hold up well, with thorough test coverage.

The only thing between this and merge was branch hygiene — none of it your feature code. A merge-conflict resolution on the branch had clobbered a few unrelated files, and since you'd already done three rounds on the same CHANGELOG conflict, I pushed a cleanup commit (`3c73df2`) so you don't have to do a fourth. Exactly what it does:

- **CHANGELOG.md** — restored to `main` (the `[0.7.3]` section + ~11 `[Unreleased]` entries had been dropped) and re-added a single entry for #492.
- **`drt/cli/main.py`** — the `validate` command had reverted to a pre-#503/#497 shape; restored it to `main`, kept only your new `clean` command.
- **`drt/destinations/google_sheets.py` / `drt/sources/sqlite.py`** — reverted unrelated changes back to `main`. (The SQLite one materialized all rows into a list before yielding, which drops lazy streaming — if that was intentional, happy to look at it as its own PR.)
- **`coverage.xml`** — removed (generated test artifact) and added to `.gitignore`.
- Merged latest `main` so the branch is current.

Your feature code (`OrphanCleanup` protocol, the postgres methods, the `clean` command, both test files) is untouched. The PR diff is now exactly the feature — 7 files, +579/-1.

One tiny non-blocking note for a possible follow-up: `test_clean_deduplicates_orphan_tables_across_syncs` is named for dedup, but `clean` processes each sync independently rather than deduplicating — the test passes either way, so it's just the name that's slightly aspirational.

CI is green across Python 3.10–3.13 and the diff is clean — good to merge. Thanks again — solid feature 🙏

masukai (dismissed):
Approving 🎉 CI is green across Python 3.10–3.13, the diff is clean (7 files, feature-only), and the `OrphanCleanup` protocol design has held up well through review. Genuinely solid end-to-end work @Muawiya-contact 🙏 — branch-cleanup details are in the comment above, and your feature code is untouched.

Thanks for sticking with this through the long arc; it's a real reliability win for swap-strategy users. If you've got a moment, a ⭐ on the repo and sharing what you build with drt always helps it reach more data folks 🙏

masukai:
@yodakanohoshi — requesting your formal review here when you have a moment 🙏

Quick context so this is a fast one:
- The orphan-cleanup feature itself was already in good shape — the only thing blocking #492 was branch hygiene (a merge-conflict resolution had clobbered `CHANGELOG.md` and the `validate` command, plus a stray committed `coverage.xml`).
- I pushed a cleanup commit (`3c73df2`) to fix that — full details in the comment above. The diff is now exactly the feature: 7 files, +579/-1.
- CI is green across Python 3.10–3.13.

Because I pushed commits to the branch, my own approval no longer satisfies the code-owner review requirement — so this genuinely needs a second maintainer's eyes. Everything is green and the diff is clean, so it should be a quick look. Thanks 🙏

Muawiya-contact:
Thank you so much @masukai 🙏 Really appreciate you pushing the cleanup commit and walking me through every blocker with such patience. Learned a lot from this one. Looking forward to the next contribution!

About repo, I have already followed the `drt-hub` and started the repo, I often share my achievements on LinkedIn representing **drt-hub**.

masukai:
@Muawiya-contact 🙏 Genuinely glad this one landed well — it was a long arc and you stuck with every round. Love that you're sharing your drt work on LinkedIn too; that kind of visibility really helps the project. Looking forward to the next contribution 🚀

@yodakanohoshi — gentle nudge on the formal review here whenever you get a moment 🙏 CI is green across Python 3.10–3.13 and the diff is feature-only (7 files, +579/-1), so it should be a quick look. No rush — just flagging it's ready and unblocked on everything else.

masukai:
Merged 🎉 Thank you @Muawiya-contact for the persistence through every round — orphan-cleanup landed cleanly via \`2c98f8c\`. Four review rounds + a long arc, and the end result is a well-designed `OrphanCleanup` Protocol that splits the concern out of `Destination` (so the 35+ existing destinations don't need to know about swap shadow tables) and a usable `drt clean` surface for operators.

This unblocks more confident adoption of `replace_strategy: swap` going forward — a real reliability win, especially as more drt deployments hit production environments where SIGTERM-interrupted syncs can leave shadows behind. Genuinely solid contribution.

Phase 2 of OTel (#531) is also yours when you have bandwidth — and you saw #527 just merged too. Three substantive PRs landed in two weeks 🚀

If you have a moment to drop the v0.7.4 release / orphan cleanup arc into a LinkedIn post tagging drt-hub, that helps the project a lot — the storytelling around real OSS contribution (vs marketing-flavored AI posts) is what cuts through right now.

Looking forward to the next one 🙏

PR #498, opened 2026-05-13
fix: quote qualified Postgres table identifiers

### LINKED ISSUE #442: Postgres destination: switch raw SQL f-strings to psycopg2.sql.Identifier for safe quoting
## Summary

The PostgreSQL destination uses raw f-string interpolation for table names in several SQL paths, including `_load_replace`, `_load_replace_swap`, `finalize_sync`, and `_load_upsert` (via `_build_insert_sql`). This breaks for identifiers containing `.`, mixed case, reserved words, or spaces.

Existing precedent for the right pattern is at [`drt/destinations/postgres.py`](drt/destinations/postgres.py) `get_row_count()`, which uses `psycopg2.sql.Identifier` — but even that call passes `config.table` as a single string, so a schema-qualified name like `public.scores` would be quoted as a single identifier (`\"public.scores\"`) instead of schema-qualified (`\"public\".\"scores\"`).

## Background

Surfaced during review of #435 (zero-downtime replace via swap) by @yodakanohoshi:

> The swap path could follow the same approach \\[as `get_row_count`\\]. The existing `_load_replace` also uses raw f-string interpolation, so this PR doesn't add a regression — but it does extend the inconsistency by another method.

Deferred from #435 to keep that PR's scope on the data-loss bug fix and its regression tests. This issue captures the broader cleanup as a follow-up.

## Proposed Approach

1. Add a `_split_qualified(table: str) -> tuple[str | None, str]` helper that splits `\"schema.table\"` into `(schema, relation)` and returns `(None, table)` if unqualified.
2. Add a `_qualified_ident(schema, relation) -> sql.Identifier` helper that produces `Identifier(schema, relation)` (or `Identifier(relation)` if no schema) — both forms quote each part correctly.
3. Replace raw f-string SQL in:
   - `_load_replace` (TRUNCATE / INSERT)
   - `_load_replace_swap` (DROP / CREATE / INSERT)
   - `finalize_sync` swap path (RENAME / DROP)
   - `_load_upsert` and `_build_insert_sql` (INSERT, ON CONFLICT)
   - `get_row_count` (currently passes single-arg `Identifier(config.table)`, breaks for `schema.table`)
4. Audit `drt/destinations/mysql.py` and `drt/destinations/clickhouse.py` for the same issue — each has its own DB-API and quoting style, so the fix may differ per destination.

## Out of Scope

- Adding configuration to support truly arbitrary identifiers (quoted with embedded `.` or `\"`) — those are vanishingly rare in real warehouse table names and not worth supporting until requested.

## Acceptance Criteria

- [ ] All Postgres SQL paths use `psycopg2.sql` composed queries (no f-string interpolation of identifiers).
- [ ] Tests cover schema-qualified table names (e.g., `marketing.email_events`) end-to-end for at least replace, swap, and upsert paths.
- [ ] `get_row_count` correctly handles schema-qualified table names.

Photon101:
Closes #442.

## Summary
- add helpers that split optional `schema.table` names and compose `psycopg2.sql.Identifier` with separate schema/relation parts
- use the helpers across Postgres row count, replace, swap, finalize, insert, and upsert SQL paths
- keep swap shadow/old suffixes on the relation name so `marketing.email_events` becomes `marketing.email_events__drt_swap`, not a single quoted identifier
- add regression coverage for schema-qualified upsert, replace, swap, row-count, and replace error paths

## Validation
- `uv run --extra dev --extra postgres python -m pytest tests/unit/test_postgres_destination.py tests/unit/test_dry_run_row_count.py tests/unit/test_dry_run_summary.py -v`
- `uv run --extra dev --extra postgres python -m pytest --cov=drt.destinations.postgres --cov-report=term-missing tests/unit/test_postgres_destination.py`
- `uv run --extra dev --extra postgres ruff check drt/destinations/postgres.py tests/unit/test_postgres_destination.py`
- `uv run --extra dev --extra postgres mypy drt/destinations/postgres.py`
- `git diff --check`

### DIFF
--- a/drt/destinations/postgres.py
+++ b/drt/destinations/postgres.py
@@ -87,6 +87,39 @@ def _serialize_value(
     return value
 
 
+def _split_qualified(table: str) -> tuple[str | None, str]:
+    """Split an optional ``schema.table`` name into schema and relation parts."""
+    if "." not in table:
+        return None, table
+    schema, relation = table.split(".", 1)
+    return schema or None, relation
+
+
+def _join_qualified(schema: str | None, relation: str) -> str:
+    if schema is None:
+        return relation
+    return f"{schema}.{relation}"
+
+
+def _qualified_ident(table: str) -> Any:
+    """Return a psycopg2 Identifier that quotes each qualified-name part."""
+    from psycopg2 import sql as _pgsql
+
+    schema, relation = _split_qualified(table)
+    if schema is None:
+        return _pgsql.Identifier(relation)
+    return _pgsql.Identifier(schema, relation)
+
+
+def _relation_name(table: str) -> str:
+    return _split_qualified(table)[1]
+
+
+def _with_relation_suffix(table: str, suffix: str) -> str:
+    schema, relation = _split_qualified(table)
+    return _join_qualified(schema, f"{relation}{suffix}")
+
+
 class PostgresDestination:
     """Upsert or replace records into a PostgreSQL table."""
 
@@ -166,7 +199,7 @@ class PostgresDestination:
         try:
             cur = conn.cursor()
             query = sql.SQL("SELECT COUNT(*) FROM {}").format(
-                sql.Identifier(config.table)
+                _qualified_ident(config.table)
             )
             cur.execute(query)
             row = cur.fetchone()
@@ -199,7 +232,7 @@ class PostgresDestination:
         result = SyncResult()
 
         if not self._replace_truncated:
-            cur.execute(_pgsql.SQL("TRUNCATE TABLE {}").format(_pgsql.Identifier(table)))
+            cur.execute(_pgsql.SQL("TRUNCATE TABLE {}").format(_qualified_ident(table)))
             self._replace_truncated = True
 
         query = self._build_insert_sql(table, columns)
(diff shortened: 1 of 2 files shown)


Photon101:
Updated in `c83da81` after the initial Codecov report. I added coverage for the schema-qualified helper branches and replace-mode error paths; the latest `codecov/patch` status check is now green.

yodakanohoshi (approved):
Thanks for the follow-up on this one, @Photon101! 🙌

This is the right finishing touch for #442 — the helper trio
(`_split_qualified` / `_qualified_ident` / `_with_relation_suffix`)
cleans up every call site, and you caught the subtle bit that
`ALTER TABLE ... RENAME TO` only accepts a bare relation on the RHS.
Test coverage across upsert / replace / swap / row_count is exactly
what was asked for.

I'll add a CHANGELOG entry post-merge — no action needed from your side.

The MySQL / ClickHouse audit listed in #442's "Proposed Approach" is
out of scope for this PR and works well as a separate follow-up issue.
If you'd be interested in tackling it, I'd be glad to open one with
the scope written up — entirely no pressure, this PR stands on its own.

Thanks again for the contribution!

PR #502, opened 2026-05-13
feat: add Mermaid docs graph command

Photon101:
## Summary

- Add a `drt.docs` manifest/builder/renderer path for static project graphs.
- Register `drt docs` with `generate` and `serve` subcommands; Mermaid writes to stdout, while `html`/`json` and `serve` return clear placeholders.
- Document the Mermaid round-trip in README with a GitHub-flavored Mermaid example.

Closes #501

## Validation

- `uv run pytest tests/unit/test_docs_mermaid.py -v`
- `uv run pytest tests/unit/test_cli_cloud_stub.py tests/unit/test_docs_mermaid.py -v`
- `uv run ruff check drt/docs drt/cli/main.py tests/unit/test_docs_mermaid.py`
- `uv run mypy drt/docs`
- `uv run mypy drt/cli/main.py`
- `git diff --check`
- `uv run drt docs --help`
- `uv run drt docs generate`
- sample project command: `uv run --project /home/semy/github/drt drt docs generate --format mermaid > /tmp/drt-docs-dag.md`
- Visual Mermaid smoke check: rendered `/tmp/drt-docs-dag.md` to SVG in headless Chrome via Playwright; screenshot captured at `/tmp/drt-docs-mermaid.png`.

PR #497, opened 2026-05-13
feat: add ConnectionTestable destination protocol

Photon101:
Closes #495.

## Summary
- add a runtime-checkable `ConnectionTestable` destination protocol for optional connection probes
- switch `drt validate --check-connection` from dynamic `getattr` dispatch to `isinstance(dest, ConnectionTestable)`
- document SQL destinations that expose `test_connection` and add contract coverage for SQL vs non-SQL destinations
- replace protocol-positive test mocks with a concrete fake so Python 3.12 runtime protocol checks exercise the intended path

## Validation
- `uv run pytest tests/unit/test_cli_validate_connection.py tests/unit/test_destination_contract.py -v`
- `uv run --python 3.12 --extra dev python -m pytest tests/unit/test_cli_validate_connection.py tests/unit/test_destination_contract.py -v`
- `uv run ruff check drt/cli/main.py drt/destinations/base.py drt/destinations/clickhouse.py drt/destinations/mysql.py drt/destinations/postgres.py tests/unit/test_cli_validate_connection.py tests/unit/test_destination_contract.py`
- `uv run --python 3.12 --extra dev ruff check drt/cli/main.py drt/destinations/base.py drt/destinations/clickhouse.py drt/destinations/mysql.py drt/destinations/postgres.py tests/unit/test_cli_validate_connection.py tests/unit/test_destination_contract.py`
- `git diff --check`

PR #496, opened 2026-05-13
polish: restore Rich formatting for cloud stubs

Photon101:
Closes #494.

## Summary
- render both `drt cloud push` and `drt cloud status` through Rich `console.print`
- restore the bold blue `drt Cloud` heading and dim coming-soon footer
- keep the `drt-hub/drt` update link in the stub text

## Validation
- `uv run pytest tests/unit/test_cli_cloud_stub.py -v`
- `uv run ruff check drt/cli/main.py tests/unit/test_cli_cloud_stub.py`
- `uv run drt cloud push`
- `uv run drt cloud status`

PR #503, opened 2026-05-13
feat: warn on hardcoded sync secrets

### LINKED ISSUE #424: feat: detect hardcoded secrets in sync YAML during drt validate
## Motivation

Users occasionally paste API keys or passwords directly into sync YAML. `drt validate` should warn when it detects likely hardcoded secrets so they can be moved to `${ENV_VAR}` references before the file is committed.

## Proposed behaviour

`drt validate` emits a WARNING (not error) for fields that:
- Match a secret field name pattern (`*_key`, `*_token`, `*_password`, `*_secret`, `api_*`) AND
- Contain a value that is not a `${...}` reference AND
- The value matches a high-entropy string heuristic (Shannon entropy > threshold) OR a known secret pattern (e.g. `sk-`, `xoxb-`, `AIza`)

Example output:
```
WARNING  destination.api_key: looks like a hardcoded secret — use ${API_KEY} instead
```

## Scope

- Integrated into `drt validate` (no new command)
- Pattern list is maintained in a small allowlist/denylist in `config/`
- Entropy check uses standard Shannon entropy (no external deps)
- Exit code remains 0 (warning only); `--strict` promotes to error

## Out of scope

- Git pre-commit hook (separate issue)
- Scanning existing committed files

Photon101:
## Summary

- Add a raw-YAML hardcoded secret scanner for `syncs/*.yml` with known token patterns and a Shannon entropy fallback.
- Wire findings into `drt validate` as warnings while keeping the default exit code at 0.
- Add `drt validate --strict` to promote warnings to validation errors, and include warning details in JSON output.

Closes #424

## Validation

- `uv run pytest tests/unit/test_cli_validate.py -v`
- `uv run pytest tests/unit/test_deprecations.py tests/unit/test_cli_validate.py -v`
- `uv run ruff check drt/config/secrets.py drt/cli/main.py tests/unit/test_cli_validate.py`
- `uv run mypy drt/config/secrets.py drt/cli/main.py`
- `git diff --check`
- `uv run drt validate --help`
- Manual sample: hardcoded `destination.auth.token` warns with exit 0, and `--strict` reports the same finding as an error with exit 1.

### DIFF
--- a/drt/cli/main.py
+++ b/drt/cli/main.py
@@ -19,6 +19,7 @@ import typer
 if TYPE_CHECKING:
     from drt.config.credentials import ProfileConfig
     from drt.config.models import SyncConfig
+    from drt.config.secrets import SecretFinding
     from drt.destinations.base import Destination
     from drt.sources.base import Source
     from drt.state.history import HistoryManager
@@ -736,21 +737,27 @@ def validate(
         False, "--check-connection", help="Test connectivity to SQL destinations."
     ),
     output: str = typer.Option("text", "--output", "-o", help="Output format: text or json."),
+    strict: bool = typer.Option(False, "--strict", help="Treat warnings as validation errors."),
 ) -> None:
     """Validate sync definitions against the JSON Schema."""
 
     from drt.config.parser import load_syncs_safe
     from drt.config.schema import write_schemas
+    from drt.config.secrets import find_hardcoded_secrets
 
     result = load_syncs_safe(Path("."))
+    secret_findings = find_hardcoded_secrets(Path("."))
 
     if select:
         result.syncs = [s for s in result.syncs if s.name == select]
         result.errors = {k: v for k, v in result.errors.items() if k == select}
         result.deprecations = {k: v for k, v in result.deprecations.items() if k == select}
+        secret_findings = [finding for finding in secret_findings if finding.sync_name == select]
         if not result.syncs and not result.errors:
             print_error(f"No sync named '{select}' found.")
             raise typer.Exit(1)
+
+    secret_warnings_by_sync = _group_secret_findings(secret_findings)
 
     if output == "json":
         # Collect all deprecations into a flat list for JSON output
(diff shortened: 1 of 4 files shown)


Photon101:
Pushed `7651384` to address the Codecov patch warning on the secret scanner.

Added focused `tests/unit/test_config_secrets.py` coverage for malformed YAML, non-mapping YAML, list recursion/path reporting, `_env`/`_path` suffix exclusions, blank/short/whitespace values, high-entropy detection, and entropy helper boundaries.

Local validation:

- `pytest tests/unit/test_config_secrets.py tests/unit/test_cli_validate.py -v` -> 23 passed
- `pytest --cov=drt.config.secrets --cov-report=term-missing tests/unit/test_config_secrets.py tests/unit/test_cli_validate.py -q` -> `drt/config/secrets.py` 100% covered
- `ruff check drt/config/secrets.py tests/unit/test_config_secrets.py tests/unit/test_cli_validate.py`
- `git diff --check`

yodakanohoshi:
🎉 Thanks for the clean implementation of #424!

A few things I particularly liked:

- **Raw-YAML scan before env expansion** — Reading the raw YAML rather than the post-Pydantic-expansion view is exactly the right call. `${ENV_VAR}` references stay correctly treated as safe, even when the actual env var holds a real secret in the user's environment. No false positives there.
- **Known patterns + entropy fallback, two-tier design** — High-confidence patterns like `sk-` / `xoxb-` / `AKIA` give strong signal, and the Shannon entropy ≥ 3.5 fallback catches the long tail. The balance between false positives and missed detections is well thought out.
- **`_NON_SECRET_SUFFIXES` (`_env`, `_path`)** — Excluding fields like `token_env` or `password_path` (which are secret *references*, not secrets themselves) is exactly the kind of practical consideration that matters in real-world usage.
- **`--strict` behavior** — Keeping the default exit code at 0 with warnings only, while letting users opt into strict mode for CI, matches the intent of #424 exactly.
- **8 tests** — Edge cases are carefully covered (malformed YAML, nested lists, filename fallback, entropy boundary conditions).

I'll add the `[Unreleased]` CHANGELOG entry after merge — no action needed on your side.

Also, looking across #496–#504: polish work (#496), a Protocol extension (#497), a Postgres bug fix (#498), docs visualization (#502), this PR's security feature, and a new destination (#504) — that's a remarkable breadth of contributions across different layers of the codebase in a short span. The fact that each PR is scoped at an independently reviewable granularity is genuinely helpful on the maintainer side too.

Thanks again, and looking forward to more 🙏

yodakanohoshi (approved):

Issue #515, opened 2026-05-16
feat(cli): enrich `drt --version` with Python / install path / OS info


masukai:
## Motivation

`drt --version` currently prints just `drt version X.Y.Z`. For bug triage and \`drt doctor\`-adjacent debugging, knowing the Python version, install path, and OS at a glance saves a back-and-forth.

## Current

\`\`\`
\$ drt --version
drt version 0.7.2
\`\`\`

## Proposed

\`\`\`
\$ drt --version
drt 0.7.2
Python 3.12.4 (CPython)
Install: /Users/foo/.venv/lib/python3.12/site-packages/drt
Platform: Darwin 24.5.0 (arm64)
\`\`\`

## Scope

- Edit \`version_callback\` in \`drt/cli/main.py:276\`
- Use \`sys.version_info\`, \`platform.python_implementation()\`, \`drt.__file__\`, \`platform.system()\` / \`platform.release()\` / \`platform.machine()\`
- Keep the first line backward-compatible-ish (\`drt X.Y.Z\` — any scripts grepping \`drt version\` should still work, so prefer \`drt version 0.7.2\` on line 1)
- One unit test in \`tests/unit/\` verifying the multi-line output contains \"Python\" and \"Install\"

## Out of scope

- Anything that requires a config file or network call — keep it instant
- JSON output (could be a follow-up)

## Estimated effort

30–60 minutes. Good first issue 👋

PR #526, opened 2026-05-19
Feat/cli enriched version

Pawansingh3889:
## Checklist

- [ ] Tests pass (`make test`)
- [ ] Linter passes (`make lint`)
- [ ] Updated `CHANGELOG.md` (if user-facing change)


masukai (approved):
Clean, well-scoped addition @Pawansingh3889 🙏 — the BC contract is exactly right: line 1 stays byte-identical to `drt version X.Y.Z`, and `test_first_line_is_backwards_compatible` + `test_output_has_four_lines` lock both the format and the contract. Lazy imports in the callback, 7 focused tests, 100% patch coverage, excellent CHANGELOG entry.

One nit, non-blocking — `test_output_has_four_lines` is slightly fragile: the four lines go through Rich `console.print`, which wraps at ~80 cols when stdout isn't a TTY. A deep install path (`Install: /Users/.../site-packages/drt`) can wrap to a 5th line and fail the count. None of the four lines use Rich markup, so plain `print()` would be both safer and simpler here. Happy to take it as a fast follow-up if you'd rather not re-push.

(Minor: the PR description is still the blank template — worth a one-liner linking #515 for the record.)

Approving — thanks for the steady triage + contribution work 🙌 If drt's been useful, a ⭐ on the repo always helps it reach more data folks.

PR #527, opened 2026-05-20
 feat: OpenTelemetry integration — traces and metrics for sync execution

### LINKED ISSUE #429: feat: OpenTelemetry integration — traces and metrics for sync execution
## Motivation

`--log-format json` gives structured logs, but there is no way to emit traces or metrics to an observability backend (Grafana Tempo, Jaeger, Datadog, Honeycomb). For teams running drt in production, OTLP export is the standard way to integrate with their existing observability stack.

## Proposed config

```yaml
# profiles.yaml or env vars
observability:
  otel:
    endpoint: http://localhost:4317   # OTLP gRPC
    service_name: drt
    headers:
      Authorization: Bearer ${OTEL_TOKEN}
```

Or via standard OTEL env vars (`OTEL_EXPORTER_OTLP_ENDPOINT`, etc.).

## Scope

- One span per sync run: `drt.sync` with attributes `sync.name`, `sync.mode`, `sync.destination_type`, `rows.processed`, `rows.errors`
- Child spans: `drt.extract`, `drt.load` (per batch)
- Counter metric: `drt.rows.synced`, `drt.sync.errors`
- Uses `opentelemetry-sdk` + `opentelemetry-exporter-otlp` as optional extra (`drt-core[otel]`)
- No-op when not configured (zero overhead)

## Out of scope

- Prometheus scrape endpoint (separate issue)
- Custom span attributes from sync YAML

Muawiya-contact:
Closes #429

---

## What's in this PR so far

This is a **draft PR** for the full OpenTelemetry integration. Posting early for feedback as work progresses.

### Done — Phase 1 (Config & Dependencies)
- Added `otel` optional extra to `pyproject.toml`
- Added `OtelConfig` and `ObservabilityConfig` Pydantic models
- Wired `observability` field into all profile types
- Added `_parse_observability()` to load from `profiles.yaml`
- Unit tests for config parsing and defaults

### Coming next
- Phase 2 — NoOp tracer provider (zero overhead when unconfigured)
- Phase 3 — Sync lifecycle spans (`drt.sync.run`, `drt.sync.extract`, `drt.sync.load`)
- Phase 4 — Metrics + error recording

---

## Checklist
- [x] Phase 1 tests pass
- [x] Ruff formatting applied
- [ ] Phase 2–4 in progress

---

> @masukai — happy to get early feedback on the config shape before I wire up the tracer in Phase 2 🙏

### DIFF
--- a/drt/config/credentials.py
+++ b/drt/config/credentials.py
@@ -27,13 +27,14 @@ Example ~/.drt/profiles.yml:
 from __future__ import annotations
 
 import os
-import re
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Literal
 
 import yaml
 from pydantic import BaseModel, Field
+
+from drt.config.parser import expand_env_vars
 
 
 class OtelConfig(BaseModel):
@@ -44,9 +45,6 @@ class OtelConfig(BaseModel):
 
 class ObservabilityConfig(BaseModel):
     otel: OtelConfig = Field(default_factory=OtelConfig)
-
-
-_ENV_VAR_PATTERN = re.compile(r"\$\{([^}]+)\}")
 
 
 # ---------------------------------------------------------------------------
@@ -62,7 +60,6 @@ class BigQueryProfile:
     method: Literal["application_default", "keyfile"] = "application_default"
     keyfile: str | None = None
     location: str = "US"  # e.g. "US", "EU", "asia-northeast1"
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.project}.{self.dataset})"
@@ -72,7 +69,6 @@ class BigQueryProfile:
 class DuckDBProfile:
     type: Literal["duckdb"]
     database: str = ":memory:"  # path or :memory:
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.database})"
@@ -82,7 +78,6 @@ class DuckDBProfile:
 class SQLiteProfile:
     type: Literal["sqlite"]
     database: str = ":memory:"  # path or :memory:
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.database})"
@@ -97,7 +92,6 @@ class PostgresProfile:
     user: str = ""
     password_env: str | None = None  # env var name
     password: str | None = None  # explicit (non-recommended)
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.host}:{self.port}/{self.dbname})"
@@ -126,7 +120,6 @@ class RedshiftProfile:
     password_env: str | None = None  # env var name
     password: str | None = None  # explicit (non-recommended)
     schema: str = "public"  # Redshift schema
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.host}:{self.port}/{self.dbname})"
@@ -143,7 +136,6 @@ class ClickHouseProfile:
     user: str = "default"
     password_env: str | None = None  # env var name
     password: str | None = None  # explicit (non-recommended)
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.host}:{self.port}/{self.database})"
@@ -170,7 +162,6 @@ class MySQLProfile:
     user: str = ""
     password_env: str | None = None
     password: str | None = None
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.host}:{self.port}/{self.dbname})"
@@ -189,7 +180,6 @@ class SnowflakeProfile:
     schema: str = "PUBLIC"
     warehouse: str = ""
     role: str | None = None
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.account}/{self.database}.{self.schema})"
@@ -207,7 +197,6 @@ class SQLServerProfile:
     password_env: str | None = None
     password: str | None = None
     schema: str = "dbo"
-    observability: ObservabilityConfig = field(default_factory=ObservabilityConfig)
 
     def describe(self) -> str:
         return f"{self.type} ({self.host}/{self.database}.{self.schema})"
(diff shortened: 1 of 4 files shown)


Muawiya-contact:
This PR needs some tests to pass.

Muawiya-contact:
@masukai, any feedback on phase 1?

masukai (commented):
Phase 1 looks good and lines up with the 4-phase plan we agreed on #429 — `drt-core[otel]` as an optional extra, `ObservabilityConfig` / `OtelConfig` parsing the `otel` block, NoOpTracer deferred to Phase 2. Clean starting point 🙏 A few execution-level notes before Phase 2 builds on this — direction's unchanged, this is all about the Phase 1 shape.

## 🔁 Config placement — our suggestion: one top-level block

The plan said "parse the `otel` block from `profiles.yaml`" but didn't pin down *where* in the file it lives. #527 puts `observability` inside each profile (added to all 10 profile dataclasses); #429's example shows it as a **top-level block**, a sibling of the profiles:

```yaml
observability:        # top level — one block
  otel:
    endpoint: http://localhost:4317
```

Our suggestion is the top-level shape, for two reasons:

1. **No source/destination ambiguity.** A sync run reads from a source profile and writes to a destination profile. With per-profile config, Phase 3 has to answer "whose `observability` wins?" on every run — and #429 scopes this as *one span per sync run*, a run-level concern, not a per-connection one. A single block has one unambiguous answer.
2. **No duplication.** Per-profile means re-declaring the same OTLP endpoint + token in every profile a user defines; a top-level block is declared once.

It also collapses most of this PR's diff — 10 dataclass fields back to one parse path.

Treat this as a proposal, not a hard call — if you see per-profile as the better fit (per-connection telemetry routing, say), make the case and we'll settle it together. Either way worth locking in now, since changing config shape post-release is breaking.

## 🔧 One fix — `${VAR}` headers get persisted as plaintext

This is the live half of Copilot's note #2. `load_profile` expands `${OTEL_TOKEN}` at load time, so `OtelConfig.headers` holds the resolved secret in memory. `save_profile` then dumps it straight back — a load→save round-trip rewrites `Authorization: Bearer ${OTEL_TOKEN}` as the **plaintext token** in `profiles.yml`. The new round-trip test locks that in:

```python
assert saved["roundtrip"]["observability"]["otel"]["headers"] == {
    "Authorization": "Bearer secret-token"
}
```

The rest of `credentials.py` deliberately never does this — profiles store `password_env` (the var *name*) and resolve lazily, so secrets never hit disk. The `otel` block should follow the same rule: keep `headers` in raw `${...}` form in the persisted model, expand only where Phase 2 inits the exporter. You handled the `exclude_defaults` half of Copilot's note — this is the other half.

## 🧹 Minor

- **Env expander is a duplicate** — `_expand_env_vars` / `_expand_env_vars_in_str` in `credentials.py` are identical to the public `expand_env_vars` in `drt/config/parser.py:35`; `resolver.py:161` carries a third copy of the regex. Please `from drt.config.parser import expand_env_vars` rather than adding a fourth — or lift it into a shared `drt/config/_env.py`.
- **Coverage** — the 4 uncovered lines are the unset-`${VAR}` `ValueError` path; one small test covers it (and reusing `parser.expand_env_vars` likely makes it moot).

---

Net: direction is exactly as agreed — this is all Phase 1 execution. Sort the secret round-trip and let's pin down the block placement, and Phase 2 has a clean base. Thanks for driving this @Muawiya-contact 🙏

Muawiya-contact:
Hey @masukai 👋
All three points addressed:

- ✅ `observability` moved to a single top-level block — removed from all 10 profile dataclasses
- ✅ `${OTEL_TOKEN}` now stays unexpanded on disk — secrets only resolve at exporter init time (Phase 2)
- ✅ Duplicate env expander removed — using `expand_env_vars` from `parser.py` now
- 

Tests updated to match the new shape. Ready for another look whenever you're free 🙏

masukai (commented):
Re-review 🙏 — all three points from last round landed cleanly:

- ✅ **Top-level `observability` block** — out of all 10 profile dataclasses, now one `load_observability_config` reading a single top-level block. `_profiles_mapping` still reads old flat-format `profiles.yml`, so existing users aren't broken.
- ✅ **Secrets stay raw on disk** — `load_observability_config` validates without expanding, `save_profile` passes the `observability` subtree through verbatim, and `test_observability_profile_round_trip_keeps_raw_secret_headers` sets `OTEL_TOKEN` then asserts the saved file *still* has `${OTEL_TOKEN}` — exactly last round's failing assertion, inverted.
- ✅ **Shared expander** — `from drt.config.parser import expand_env_vars`, no fourth copy.

CI green 3.10–3.13, patch coverage 97%. Nice turnaround.

## 🔎 One leftover — `expand_env_vars` in `load_profile`

`load_profile` still does `raw = expand_env_vars(profiles[profile_name])`. That call existed to expand the OTEL headers — but headers now load separately via `load_observability_config` (unexpanded, by design). So post-redesign this call only ever expands `${VAR}` inside **regular profile credential fields**, which:

- isn't needed for the OTEL feature anymore;
- adds an inline-`${VAR}`-in-profiles behavior that's undocumented and untested (profiles already resolve secrets lazily via `*_env`); and
- is a small regression risk — `expand_env_vars` raises `ValueError` on an unset `${VAR}`, so an existing profile with a literal `${...}` would now hard-fail on load where it didn't before.

Suggest just dropping the call (and the now-unused import) — keeps Phase 1 minimal and one secret-handling pattern for profiles. If inline profile templating is something we want, let's spin it into its own issue with docs + tests. Dropping it breaks no test in the PR (I checked — `load_syncs` expansion is separate).

## 🧹 Minor — non-blocking

- **CHANGELOG** — Phase 1 is config-only with no working exporter yet, so an entry can wait for the phase that emits traces. One thing worth a line whenever it lands: an existing flat `profiles.yml` gets rewritten into the nested `{observability, profiles}` shape on the next `save_profile`.
- `_profiles_mapping` and `save_profile` carry the same `{k:v ... if k != "observability"}` fallback — `save_profile` could just call `_profiles_mapping`.
- `tests/unit/test_config.py` picked up some unrelated reformatting — just diff noise.

Net: shape is right, Phase 2 has a clean base. Drop the `expand_env_vars` call and this is good to go. Thanks @Muawiya-contact 🙏

Muawiya-contact:
Hey @masukai 👋

Dropped the `expand_env_vars` call from `load_profile` and removed the now-unused import — just

 `raw = profiles[profile_name]` now. Ruff passes, nothing else touched.

I think that clears everything from round 2 — ready for a final look whenever you're free 🙏

Also saw you filed #531 for Phase 2 — I'll pick it up once this merges and keep the chain going.

masukai (approved):
Re-review 🎉 — round-2 nit cleanly addressed:

- ✅ `expand_env_vars` call dropped from `load_profile` — `raw = profiles[profile_name]` now, exactly as suggested.
- ✅ Now-unused `expand_env_vars` import also removed (verified — zero remaining references in `drt/config/credentials.py`).
- ✅ Ruff + mypy clean, CI green across 3.10–3.13.

Phase 1 scope is exactly right: config schema for `observability.otel` + top-level block + `[otel]` extras + secrets stay raw on disk. No exporter wiring yet, no engine instrumentation yet — that's Phase 2 (#531) and Phase 3 (will file after Phase 2 lands).

Approving and merging. Phase 2 is yours whenever you have bandwidth — and as you said, the chain continues 🙏

Thanks for the patience through three review rounds. The shape of the config block we landed on (top-level, secrets raw on disk, shared `expand_env_vars` from `parser.py`) is a clean base for the exporter init in Phase 2.

If drt is useful to you, a ⭐ helps a lot — and your LinkedIn drt-hub mentions are appreciated too 🙏

PR #522, opened 2026-05-17
docs: add CLI help examples

nvphungdev:
## What does this PR do?

Adds `Examples:` sections to `drt run`, `drt list`, and `drt validate` help output, using command patterns already documented in the README and quickstart examples. Adds focused help-output tests so the examples stay visible.

## Related Issue

Closes #516

## Checklist

- [x] Tests pass (`.venv/bin/python -m pytest tests/unit/test_cli_help_examples.py`)
- [x] Linter passes (`.venv/bin/python -m ruff check drt/cli/main.py tests/unit/test_cli_help_examples.py`)
- [x] Updated `CHANGELOG.md` (if user-facing change)

Issue #516, opened 2026-05-16
feat(cli): add "Examples:" section to `drt run`, `drt validate`, `drt list` help


masukai:
## Motivation

CLI help currently lists options but no usage examples. For new users, seeing \`drt run --select tag:crm --threads 4\` once is faster than reading the option table and guessing.

## Current (\`drt run --help\`)

```
 Usage: drt run [OPTIONS]

 Run syncs (all by default).

╭─ Options ─...
```

## Proposed

```
 Usage: drt run [OPTIONS]

 Run syncs (all by default).

 Examples:
   drt run                                Run every sync
   drt run --select my_sync               Run a specific sync by name
   drt run --select tag:crm --threads 4   Run all CRM-tagged syncs in parallel
   drt run --dry-run --diff               Preview changes without writing

╭─ Options ─...
```

## Scope

Add an \"Examples:\" section to the docstring (Typer renders it in help output) for these three commands in \`drt/cli/main.py\`:

- \`run\` (line 497)
- \`validate\` (search for \`def validate\`)
- \`list\` (search for \`def list\`)

3–4 examples per command. Pick examples from the existing README or quickstart docs to stay consistent.

## Out of scope

- Other commands (could be follow-up GFIs once the pattern is set)
- Doc generation changes

## Estimated effort

1 hour. Good first issue 👋

PR #529, opened 2026-05-22
feat(cli): drt init - print 'Next steps' block

### LINKED ISSUE #517: feat(cli): `drt init` — print "Next steps:" block after scaffold
## Motivation

After \`drt init\`, a first-time user is dropped into a directory with files they don't yet know. A short \"Next steps:\" block printed to stdout removes the \"now what?\" friction.

## Current

\`drt init\` scaffolds files and exits with a success message (\`print_init_success\` in \`drt/cli/output.py\`).

## Proposed (append after the existing success message)

```
Next steps:
  1. Edit drt_project.yml — set your profile credentials
  2. Edit syncs/example.yml — define your first sync
  3. Run: drt validate              # check config
  4. Run: drt run --dry-run         # preview without writing
  5. Run: drt run                   # execute

Docs: https://github.com/drt-hub/drt#quickstart
```

## Scope

- Extend \`print_init_success\` in \`drt/cli/output.py\` (or add a sibling \`print_init_next_steps\`)
- Use Rich \`console.print\` to keep consistent styling with other CLI output
- Update the one existing \`drt init\` test in \`tests/unit/\` to assert \"Next steps:\" appears

## Out of scope

- Changing what files are scaffolded
- Interactive prompts

## Estimated effort

1 hour. Good first issue 👋

cian-ps:
## What does this PR do?

Expands the `print_init_success` function in `drt/cli/output.py` to print out more detailed next steps instructions and a documentation link.

Adds `tests/unit/test_cli_init_output.py`, as I couldn't find a dedicated test for `drt init` yet.

## Related Issue

Closes: #517 

## Checklist

- [ ] Updated `CHANGELOG.md` (if user-facing change)

### DIFF
--- a/drt/cli/output.py
+++ b/drt/cli/output.py
@@ -35,9 +35,13 @@ def print_init_success(paths: list[str]) -> None:
         console.print(f"  [dim]Created[/dim] {p}")
     console.print()
     console.print("Next steps:")
-    console.print("  1. Edit [bold]drt_project.yml[/bold] if needed")
-    console.print("  2. Add sync definitions to [bold]syncs/[/bold]")
-    console.print("  3. Run [bold]drt run --dry-run[/bold] to preview")
+    console.print("  1. Edit [bold]drt_project.yml[/bold] — set your profile credentials")
+    console.print("  2. Edit [bold]syncs/example_sync.yml[/bold] — define your first sync")
+    console.print("  3. Run: [bold]drt validate[/bold]\t\t[dim](check config)[/dim]")
+    console.print("  4. Run: [bold]drt run --dry-run[/bold]\t[dim](preview without writing)[/dim]")
+    console.print("  5. Run: [bold]drt run[/bold]\t\t[dim](execute)[/dim]")
+    console.print()
+    console.print("Docs: https://github.com/drt-hub/drt#quickstart")
     console.print()
 
 
@@ -116,10 +120,7 @@ def _print_row_count_diff(sync: SyncConfig, destination: object, new_rows: int) 
             )
     except Exception as e:
         # Silently skip row count if unable to connect (not a blocking error)
-        console.print(
-            f"  [dim](Could not retrieve current row count: {type(e).__name__})[/dim]"
-        )
-
+        console.print(f"  [dim](Could not retrieve current row count: {type(e).__name__})[/dim]")
 
 
 def print_sync_result(sync_name: str, result: SyncResult, elapsed: float) -> None:
@@ -196,8 +197,8 @@ def print_validation_error(sync_name: str, errors: list[str]) -> None:
 
 
 def print_connection_test_result(
-    sync_name: str, 
-    success: bool | None, 
+    sync_name: str,
+    success: bool | None,
     error: str | None = None,
 ) -> None:
     if success is None:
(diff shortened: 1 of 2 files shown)


masukai:
Thanks for the careful read on #517 @cian-ps 🙌 — and you're absolutely right that no existing `drt init` test was present. That was my mistake in the issue description; the new `tests/unit/test_cli_init_output.py` you added is in fact the first coverage for `print_init_success`. Good call adding it as a fresh file rather than forcing a fit into something unrelated.

## ✅ What's great

- **Enriched "Next steps:" block** — the expansion from 3 imperative lines to a numbered 5-step recipe with parenthetical hints (`(check config)` / `(preview without writing)` / `(execute)`) and a `Docs:` link is exactly the friction-reducer #517 was aiming at. The leading `Edit drt_project.yml — set your profile credentials` and `Edit syncs/example_sync.yml — define your first sync` give the user the two concrete files to touch before any command runs, which is the part new users were getting stuck on.
- **Test design** — small, focused `capsys` capture asserting the user-visible contract (paths, key commands, "Docs"). Doesn't over-assert on Rich markup so the styling stays free to evolve.
- **Format reflows** in `print_connection_test_result`, `_print_row_count_diff`, and `print_diff_table` look like `make fmt` running over lines that were on the wrong side of ruff's wrap threshold — happy to absorb those, no harm done.

## 🔧 One change to revert before merge

The `drt/state/watermark.py` edit (`from google.cloud import storage` → `import google.cloud.storage as storage`) is outside #517's scope and isn't required for lint. Both `make lint` and the full CI matrix (ruff + mypy on Python 3.10/3.11/3.12/3.13) have been green with the original form across recent merges — I verified locally just before this review:

```
ruff check drt tests
All checks passed!
mypy drt
Success: no issues found in 90 source files
```

The most likely cause on your end is the GCS extras not being installed in your local venv (`pip install drt-core[gcs]` resolves `google.cloud.storage` for mypy), which can make `from google.cloud import storage` surface differently from the aliased form even with the `# type: ignore[import-untyped]` in place. If that turns out to be a reproducible CI-vs-local divergence we'd want to fix at the config level, please file a separate issue with the `make lint` output you saw and I'll dig in — but for this PR, could you revert that one line?

```bash
git checkout origin/main -- drt/state/watermark.py
git commit --amend --no-edit
git push --force-with-lease
```

Codecov is currently flagging that line as the only uncovered patch line, so the revert also drops patch coverage to 100% as a side effect.

## After that

Everything else looks good to go. Once the watermark revert is in, this is ready to merge.

If drt has been useful while you were poking around the codebase, a ⭐ on the repo helps a lot: https://github.com/drt-hub/drt 🌱

Looking forward to the next one!

masukai:
Merged 🚀 Thank you @cian-ps — clean revert on the watermark.py side, exactly the diff we wanted. Your `test_cli_init_output.py` is also the first dedicated coverage for `print_init_success` on the repo, which is a nice side effect.

If you'd like to keep iterating, there are a few more `good first issue` candidates that touch this same area:

- **#516** — adding "Examples:" sections to `drt run`, `drt validate`, `drt list` help output (related CLI UX polish, currently in flight in #522 but the existing PR has a Rich rendering issue worth a fresh take if it doesn't land soon).
- **#515** is already done by another contributor, but **#530** has 5 small REST API source cleanups looking for a home if you're interested in the source side.

Also, if drt has been useful to poke at, a ⭐ on the repo really helps: https://github.com/drt-hub/drt 🌱

Looking forward to the next one!

Issue #530, opened 2026-05-23
refactor(rest-api): polish follow-ups from PR #474


masukai:
Five small follow-ups raised in the PR #474 approval review for the new `RestApiSource`. None are blocking — the source landed and works — but each tightens the implementation that future API sources will copy from. Pick one or all 🙏

## Tasks

- [ ] **Lift `_extract_next_link` into a shared HTTP util.** `drt/sources/rest_api.py:111` reaches into `RestApiDestination._extract_next_link` for Link-header parsing, coupling source ↔ destination through a destination's static method. Move it to a new `drt/_http_utils.py` (or `drt/destinations/_http.py`) and have both the source and the REST API destination import it.
- [ ] **Document the `query` parameter on `RestApiSource.extract`.** The `Source` Protocol passes `query: str` (used by SQL sources for the statement). For REST, the URL is in config so `query` is silently ignored. A one-line docstring noting this would save the next reader a head-scratch.
- [ ] **Replace `# type: ignore[no-any-return]` in `_extract_records` with `cast`.** Per [CLAUDE.md](CLAUDE.md), `type: ignore` is reserved for external-library issues; here the ignores are on pure-Python narrowing paths in `drt/sources/rest_api.py`. After the `isinstance` checks, `cast(list[dict[str, Any]], ...)` satisfies the rule cleanly. There are three such ignores in `_extract_records`.
- [ ] **Remove the unused module-level logger.** `drt/sources/rest_api.py:24` defines `logger = logging.getLogger(\"drt\")` but it is never referenced. Either delete it, or wire it into the pagination paths for trace-level visibility (consider doing this alongside the next bullet).
- [ ] **Add a debug log in `_extract_records` single-dict fallback.** `drt/sources/rest_api.py:136` returns `[data] if isinstance(data, dict) else []`. A debug log noting the fallback shape was used (e.g. `\"REST source: no records array found, wrapping single dict as one record\"`) would prevent silent surprises when an endpoint returns `{\"error\": \"...\"}` and a single error object gets persisted as a row.

## Context

These came from the maintainer review on PR #474. The PR shipped (`d014f36`) and is in `[Unreleased]`. None of these change user-visible behavior — they're internal cleanup so the file stays the canonical template for future non-database sources (the v0.8 source-diversification slice will lean on this pattern).

Each box can land as its own PR if you'd rather scope tight, or as one bundled PR — your call. `good first issue` because each individual item is a self-contained edit in one file.

## Pointers

[...]

PR #556, opened 2026-05-24
refactor(rest-api): lift extract_next_link into shared drt._http_utils

Pawansingh3889:
## What does this PR do?

Moves the RFC 5988 Link-header parser out of `RestApiDestination` and into a new top-level `drt/_http_utils.py` module as a plain function `extract_next_link()`.

`drt/sources/rest_api.py` was importing `RestApiDestination` solely to call its `_extract_next_link` static method — a cross-package coupling that had no business existing. Both the source and destination now call the shared util directly.

The five unit tests in `test_pagination_rest_api.py::TestExtractNextLink` are retargeted to the new symbol and lose their `rest_api_destination` fixture dependency.

No behaviour change.

## Related Issue

Addresses item 1 of #530. Items 2–5 already landed via #534. Origin of the helper: #474.

## Checklist

- [x] Tests pass (`make test`) — 51/51 REST/HTTP unit tests green; full unit run 1058 pass on Linux (6 unrelated Windows-only SQLite teardown errors locally)
- [x] Linter passes (`make lint`) — `ruff check` and `mypy drt` both clean
- [x] Updated `CHANGELOG.md`


masukai (approved):
Clean, surgical refactor @Pawansingh3889 🙏 — pure move + rename, regex byte-identical, all 3 callsites swapped (source, destination, tests), and the docstring even picks up the quote/whitespace/None tolerance that was implicit before. The fixture cleanup in `TestExtractNextLink` (dropping `rest_api_destination` only where it's not needed, keeping it for the `fetch_paginated` tests) is exactly the right scope.

Picking `drt/_http_utils.py` over `drt/destinations/_http.py` is also the right call — source and destination both depend on it, so a destination-owned module would have just moved the cross-domain coupling, not removed it.

CI green across 3.10–3.13, CHANGELOG entry hits the right notes (item 1 + #534 cross-reference). This closes #530 entirely 🎉

Approving — thanks for picking up the last item 🙌

PR #610, opened 2026-06-03
fix(clickhouse): apply _quote_ident to all SQL command paths (#512)

### LINKED ISSUE #512: refactor(clickhouse): harden identifier quoting and audit SQL command paths
## Summary

`ClickHouseDestination` could use a quoting pass on its SQL command paths:

1. `get_row_count` builds a backtick-quoted identifier in a way that doesn't render correctly for schema-qualified table names — the join string is `` .` `` where it should be `` `.` ``.
2. Most `client.command(...)` paths interpolate `config.table` without any quoting. This works fine for plain lowercase names but is fragile for reserved words, mixed case, or database-qualified names.

Surfaced during review of #442 / PR #498. The Postgres equivalent was completed in #498; the ClickHouse audit was listed in #442's "Proposed Approach" but out of scope there.

## get_row_count: quoting doesn't render as intended

[`drt/destinations/clickhouse.py` L205-L210](drt/destinations/clickhouse.py#L205-L210):

```python
escaped_table = (
    ".`".join(config.table.split("."))
    if "." in config.table
    else config.table
)
result = client.query(f"SELECT COUNT(*) FROM `{escaped_table}`")
```

For `config.table = "db.scores"`:

```python
>>> ".`".join("db.scores".split("."))
'db.`scores'
>>> f"`{_}`"
'`db.`scores`'
```

→ The intended output is `` `db`.`scores` `` (4 backticks, two quoted parts). The actual output has 3 backticks and isn't a valid identifier reference.

There is currently no test coverage for `get_row_count` with a schema-qualified table, which is why this hasn't surfaced in practice.

## Audit scope (all `command(...)` call sites)

These interpolate `config.table` / `shadow` / etc. with no quoting:

| Method | Line | SQL |
|---|---|---|
| `load` (non-swap replace) | L79 | `TRUNCATE TABLE {config.table}` |
| `_load_replace_swap` | L124 | `DROP TABLE IF EXISTS {shadow}` |
| `_load_replace_swap` | L125 | `CREATE TABLE {shadow} AS {table}` |
| `_load_replace_swap` | L150 | `DROP TABLE IF EXISTS {shadow}` (cleanup) |
| `finalize_sync` | L179 | `EXCHANGE TABLES {table} AND {shadow}` |
| `finalize_sync` | L181 | `DROP TABLE {shadow}` |
| `get_row_count` | L210 | (see above) |

`client.insert(table, ...)` (L86, L132) goes through the driver's typed path and is probably safer — worth confirming whether `clickhouse-connect` quotes table names internally.

## Design questions to settle before coding

ClickHouse identifier handling differs enough from Postgres / MySQL that this benefits from some design discussion first:

1. **Quoting policy**: ClickHouse accepts both `` `name` `` and `"name"` and unquoted lowercase names. Always-quote is safer and matches what we do elsewhere — confirm this is the direction we want.
2. **Helper location**: extract a `_quote_ident(table)` similar to MySQL's, or inline?
3. **`client.insert` quoting**: does `clickhouse-connect` quote the `table` parameter? If so, `client.insert(shadow, ...)` is already safe and can be left alone.
4. **Server-side parameter binding**: `clickhouse-connect` supports `parameters=` for values but (as far as we know) not for identifiers. Worth confirming.

These open questions are why this isn't flagged `good first issue` — it benefits from someone with ClickHouse familiarity to make the call.

## Acceptance Criteria

- [ ] `get_row_count` produces the intended SQL for schema-qualified table names; regression test covers `db.scores`
- [ ] All `client.command(...)` interpolations of identifiers go through a consistent quoting path
- [ ] Decision on `client.insert` documented (either confirmed safe, or quoted)
- [ ] Unit tests cover schema-qualified table names for replace, swap (including `EXCHANGE TABLES`), and `get_row_count`
- [ ] No new `type: ignore`; `ruff check` and `mypy drt` pass

## Out of Scope

- Adding support for arbitrary quoted identifiers (embedded `.`, `` ` ``, etc.)
- MySQL identifier audit — tracked in #511

## Reference

- PR #498 (Postgres equivalent) — pattern to follow for tests and call-site replacement, though the underlying quoting mechanism differs.

yodakanohoshi:
## Summary

ClickHouse counterpart to the Postgres ([#498](https://github.com/drt-hub/drt/pull/498)) / MySQL ([#514](https://github.com/drt-hub/drt/pull/514)) qualified identifier fixes. The `_quote_ident` helper already existed (added for `sync.mode: mirror` in [#598](https://github.com/drt-hub/drt/pull/598)) but was only used on the mirror DELETE path. This PR routes every remaining SQL command path through it.

### The bug

`ClickHouseDestination.get_row_count` built its quoted identifier inline:

```python
escaped_table = ".`".join(config.table.split(".")) if "." in config.table else config.table
result = client.query(f"SELECT COUNT(*) FROM `{escaped_table}`")
```

For `db.scores` this rendered ``SELECT COUNT(*) FROM `db.`scores` `` — three backticks, a syntax error. Confirmed `Code: 62` against ClickHouse 24.8 for any schema-qualified destination table. There was no `get_row_count` unit test with a qualified table, which is why it never surfaced. Surfaced in practice by `drt run --dry-run` against a destination configured as `table: e2edb.users`.

### The fix

- `get_row_count`: uses `_quote_ident` (which renders ``` `db`.`scores` ``` correctly).
- `TRUNCATE TABLE` (replace mode), `DROP/CREATE TABLE ... AS` (swap shadow setup + on-error cleanup), `EXCHANGE TABLES` / `DROP TABLE` (`finalize_sync` swap): switched from raw interpolation to `_quote_ident` so reserved words / mixed case / `db.table` names are all safe.
- **`client.insert(table, ...)` paths (non-swap + swap shadow): also pre-quoted.** I initially left these alone assuming `clickhouse-connect` quoted internally, but the driver source (`clickhouse_connect/driver/insert.py:132` + `httpclient.py:406` + `asyncclient.py:1800`) shows raw `f"INSERT INTO {table} ... FORMAT Native"` interpolation. Plain `db.scores` works only because both parts are unquoted-safe; reserved words / mixed case would fail at the insert path even after the rest of this PR. Quoting here matches the established Postgres / MySQL precedent of also quoting the insert/upsert paths.

### Tests

- `TestClickHouseIdentifierQuoting` (5 tests): `_quote_ident` semantics, TRUNCATE-quotes-qualified-table on replace mode, CREATE/EXCHANGE both reference fully-quoted qualified identifiers (with a guard that no path leaks an unquoted `analytics.scores`), and `client.insert` quotes the qualified table on both non-swap and swap shadow paths.
- `test_clickhouse_destination_get_row_count_with_qualified_table` (regression for the core bug): asserts the emitted SQL is exactly ``SELECT COUNT(*) FROM `analytics`.`scores` ``.
- Updated existing `TRUNCATE TABLE analytics_scores` + raw-`analytics_scores` insert + swap `endswith("__drt_swap")` assertions to expect the now-quoted form.

### Verification

- `make lint` — ruff + mypy green
- Targeted: `pytest tests/unit/test_clickhouse_destination.py tests/unit/test_clickhouse_mirror_mode.py tests/unit/test_dry_run_row_count.py` → **64 passed**
- Full suite: `pytest` → **1323 passed, 9 skipped**
- End-to-end (separately verified against a real ClickHouse 24.8 with database-qualified tables): the previous SQL is rejected with `Code: 62`; the fixed paths run cleanly across replace (TRUNCATE+INSERT), swap (EXCHANGE), mirror (ALTER DELETE), and `get_row_count`.

Closes #512.

## Acceptance criteria mapping (issue #512)

- [x] `get_row_count` produces the intended SQL for schema-qualified table names; regression test covers `db.scores`
- [x] All `client.command(...)` interpolations of identifiers go through a consistent quoting path (`_quote_ident`)
- [x] Decision on `client.insert` documented: driver does **not** quote internally → destination pre-quotes (matches #498 / #514)
- [x] Unit tests cover schema-qualified table names for replace, swap (including `EXCHANGE TABLES`), `get_row_count`, and `client.insert`
- [x] No new `type: ignore`; `ruff check` and `mypy drt` pass

## Test plan

- [x] Unit tests cover the row-count regression + identifier quoting on TRUNCATE / CREATE / EXCHANGE / `client.insert`
- [x] `make lint` green
- [x] Full `pytest` green
- [x] Manually verified end-to-end on ClickHouse 24.8 with schema-qualified destination table


masukai (approved):
Great catch, @yodakanohoshi — and the level of investigation (including verifying `Code: 62` against ClickHouse 24.8 for the malformed `` `db.`scores` `` case) makes this very easy to accept.

Reviewed the diff:

- ✅ `get_row_count`: the inline `` ".`".join(...) `` was indeed producing 3 backticks; routing through `_quote_ident` (added in #598 for mirror) gives the right `` `db`.`scores` ``.
- ✅ `TRUNCATE TABLE` / `DROP TABLE IF EXISTS` / `CREATE TABLE ... AS` / `EXCHANGE TABLES` / `DROP TABLE` (swap finalize) — all routed through `_quote_ident`.
- ✅ `client.insert(table_q, row, ...)` on both the non-swap and swap paths — the inline comment citing `clickhouse_connect/driver/insert.py` is the right pointer for the next reader who wonders why we pre-quote here.
- ✅ New regression test `test_clickhouse_destination_get_row_count_with_qualified_table` pins the `` `analytics`.`scores` `` rendering, and the docstring documents the `Code: 62` failure mode.
- ✅ CHANGELOG entry under `Fixed` is exhaustive and references the Postgres (#498) / MySQL (#514) precedent — the family of three is now consistent.
- ✅ CI / codecov green.

Approving — closes the ClickHouse leg of the qualified-identifier story to match #498 / #514. Thanks for picking this up post-#598; appreciate the end-to-end verification.

Issue #170, opened 2026-04-02
feat: add Azure Blob Storage destination (CSV/Parquet export)


masukai:
## Summary

Add Azure Blob Storage as a destination connector for exporting data as CSV or Parquet files.

Completes the 3 major cloud storage destinations alongside S3 (#168) and GCS (#169).

Depends on #67 (CSV/JSON file destination) for the file serialization layer.

## Scope

- `drt/destinations/azure_blob.py` — Upload via `azure-storage-blob`
- `drt/config/models.py` — AzureBlobDestinationConfig
- Supports CSV and Parquet output formats
- Auth via connection string env or managed identity
- Configurable container, prefix, file naming
- Unit tests

## Example config

\`\`\`yaml
destination:
  type: azure_blob
  connection_string_env: AZURE_STORAGE_CONNECTION_STRING
  container: data-exports
  prefix: "drt/users/"
  format: csv                    # "csv" | "parquet"
\`\`\`

## Dependencies

- #67 (CSV/JSON file destination) should land first
- `azure-storage-blob` as optional dependency (`drt-core[azure]`)

## Difficulty

★★☆ (Intermediate)

Issue #169, opened 2026-04-02
feat: add GCS destination (CSV/Parquet export)


masukai:
## Summary

Add Google Cloud Storage as a destination connector for exporting data as CSV or Parquet files.

GCS is the natural pairing for BigQuery users. Both Census and Hightouch support GCS as a destination.

Depends on #67 (CSV/JSON file destination) for the file serialization layer.

## Scope

- `drt/destinations/gcs.py` — GCS upload via `google-cloud-storage`
- `drt/config/models.py` — GCSDestinationConfig
- Supports CSV and Parquet output formats
- Auth via Application Default Credentials or service account keyfile
- Configurable bucket, prefix, file naming
- Unit tests

## Example config

\`\`\`yaml
destination:
  type: gcs
  bucket: my-data-exports
  prefix: "drt/users/"
  format: csv                    # "csv" | "parquet"
  credentials_path: /path/to/sa-key.json   # or ADC
\`\`\`

## Dependencies

- #67 (CSV/JSON file destination) should land first
- `google-cloud-storage` as optional dependency (`drt-core[gcs]`)

## Difficulty

★★☆ (Intermediate)

Issue #172, opened 2026-04-02
feat: add Delta Lake source (S3/GCS/local)


masukai:
## Summary

Add Delta Lake as a source connector for reading Delta tables from S3, GCS, or local filesystem.

The Lakehouse architecture (Delta Lake, Apache Iceberg) is growing rapidly. Many organizations store analytical data as Delta tables on object storage rather than in traditional DWH. Supporting Delta as a source enables drt to activate data directly from the lakehouse.

## Scope

- `drt/sources/delta.py` — Read Delta tables via `deltalake` (delta-rs Python bindings)
- `drt/config/credentials.py` — DeltaProfile dataclass
- Supports local paths, S3 URIs (`s3://`), GCS URIs (`gs://`)
- SQL filtering via DuckDB integration (optional)
- Unit tests

## Example config

\`\`\`yaml
delta_lake:
  type: delta
  uri: s3://my-bucket/delta-tables/users    # or local path: ./data/delta/users
  storage_options:                           # optional: cloud auth
    AWS_ACCESS_KEY_ID_ENV: AWS_KEY
    AWS_SECRET_ACCESS_KEY_ENV: AWS_SECRET
\`\`\`

## Reference

- [delta-rs Python docs](https://delta-io.github.io/delta-rs/)
- Distinct from Databricks source (#88) which uses SQL Connector

## Difficulty

★★★ (Advanced — Delta Lake + cloud storage integration)

masukai:
Picking this up myself to keep v0.8 moving, @yodakanohoshi — shout if you'd rather keep it (or the sibling #173). Thanks! 🙌

PR #613, opened 2026-06-05
feat(destinations): add Amazon S3 destination (csv/json/jsonl/parquet + gzip) (closes #168)

### LINKED ISSUE #168: feat: add S3 destination (CSV/Parquet export)
## Summary

Add Amazon S3 as a destination connector for exporting data as CSV or Parquet files.

S3 is one of the most common data exchange points — many downstream systems poll S3 for new files. Both Census and Hightouch support S3 as a destination.

Depends on #67 (CSV/JSON file destination) for the file serialization layer.

## Scope

- `drt/destinations/s3.py` — S3 upload via `boto3`
- `drt/config/models.py` — S3DestinationConfig
- Supports CSV and Parquet output formats
- Auth via AWS credentials (env vars or IAM role)
- Configurable bucket, prefix, file naming
- Unit tests

## Example config

\`\`\`yaml
destination:
  type: s3
  bucket: my-data-exports
  prefix: "drt/users/"
  format: csv                    # "csv" | "parquet"
  region: us-east-1
  # Auth via AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY env vars or IAM role
\`\`\`

## Dependencies

- #67 (CSV/JSON file destination) should land first
- `boto3` as optional dependency (`drt-core[s3]`)

## Difficulty

★★☆ (Intermediate)

masukai:
## Summary

Adds the **Amazon S3 destination** — closes [#168](https://github.com/drt-hub/drt/issues/168). New file-style destination that uploads each sync batch as a single file to an S3 bucket. Builds on the existing FileDestination / ParquetDestination shape; sibling to the planned [#169](https://github.com/drt-hub/drt/issues/169) (GCS) and [#170](https://github.com/drt-hub/drt/issues/170) (Azure Blob) destinations, which will reuse the serialiser / key-naming / credential-threading layer here.

| Aspect | Value |
|---|---|
| Formats | csv / json / jsonl / parquet |
| Compression | gzip for text formats; native column compression for parquet |
| Auth | boto3 standard credential chain by default; explicit overrides via profile, env vars, or endpoint_url |
| Default key | `<prefix><UTC ISO8601 basic>.<ext>` (Census / Hightouch convention) |
| Override | `key_template` with `{timestamp}` placeholder |
| Extras | `pip install drt-core[s3]` (boto3>=1.34); add `[parquet]` for parquet format |

### Failure semantics (worth reviewing)

`load()` distinguishes three classes of failure:

| Class | Disposition | Why |
|---|---|---|
| Missing extras (`ImportError` on `import boto3`) | **bubbles up** | Deployment mistake. Engine should surface it once at the top, not re-fail every batch with the same generic error. |
| Serialisation error (bad rows / pandas can't build DataFrame) | row failure, early return | Per-batch, deterministic |
| AWS-side error (network, auth, permissions) | `result.failed = len(records)` recorded, sync continues | Recoverable, other batches keep going |

This split is intentional — see the comments at the top of the `load()` body.

### Empty-batch contract

The implicit "no driver was imported" pattern from the SQL contract module ([#595](https://github.com/drt-hub/drt/pull/595)) — `load([])` short-circuits before `_client()` is called, so boto3 is never imported. CI's minimal install does NOT include `[s3]` extras, so a regression that lost the short-circuit would crash the test with `ModuleNotFoundError`. The empty-batch test passing proves the short-circuit holds.

I did NOT add S3 to a contract framework module in this PR — there's no good fit (existing modules are HTTP-server / file-system / SQL-lazy-import / API-httpx-tripwire / SMTP / staged) and a single-entry "cloud storage" module is awkward. Once GCS (#169) and Azure Blob (#170) land, a `test_destination_cloud_storage_empty_batch.py` module covering all three will be a clean follow-up.

### Key naming

Default: `<prefix><UTC ISO8601 basic>.<ext>` — e.g. `drt/users/20260605T123000Z.csv`. Timestamping the key (rather than overwriting a fixed name) matches Census / Hightouch and lets downstream consumers reliably detect "new files since last check". `key_template` supports the `{timestamp}` placeholder; if the template includes its own extension, it's preserved (e.g. `key_template: "latest.csv"` for snapshot semantics).

Per-sync routing pattern: use a sync-specific `prefix`. I considered threading `sync_name` through `SyncOptions` but that's a broader engine change; prefix-per-sync is the simpler shape for this PR.

## Test plan

- [x] 21 unit tests in `tests/unit/test_s3_destination.py` covering config validation × 5, empty-batch no-import × 1, CSV upload × 4, JSON / JSONL × 2, gzip × 1, credential threading × 5, failure paths × 2, parquet × 1
- [x] `make test` — 1450 pass on the branch
- [x] `make lint` — ruff + mypy clean
- [x] `make check-i18n` — clean (README.ja roadmap row added alongside README.md)
- [x] `pip install drt-core[s3]` resolves boto3>=1.34 cleanly
- [ ] CI green

## Out of scope / follow-ups

- **GCS destination (#169)** — same shape, swap boto3 for google-cloud-storage
- **Azure Blob destination (#170)** — same shape, swap for azure-storage-blob
- **Empty-batch contract module for cloud storage** — once 2+ cloud storage destinations are landed, refactor into a dedicated contract module
- **Multipart upload for very large batches** — current implementation buffers the full payload in memory; multipart upload would be relevant for syncs producing files larger than a few hundred MB
- **GitHub Topics** — currently 20/20; `s3` would be a natural topic to swap in if `redshift` (source-only) or `data-pipeline` (overlaps with `data-engineering`) is dropped


yodakanohoshi:
🎉 Clean, well-tested addition — the S3 destination is solidly built, and the docs match the implementation.

Particularly good:

- **Auth** — env-var-name-only config (no raw AWS keys in the model) follows the OAuth2 secrets pattern rather than the SQL `password`-in-YAML one, so secrets never land in S3DestinationConfig. Cross-checked docs/connectors/s3.md against `_client()`: credential chain → named profile → explicit env vars → endpoint_url is described exactly as implemented, and `resolve_env` threading matches the other destinations.
- **Failure semantics** — three-way split (missing-extras bubbles up, serialisation + AWS errors stay per-batch) is the right call and clearly documented.
- **Empty-batch contract** — short-circuits before any boto3 import, with a test that deliberately omits the mock so a lost short-circuit fails on CIs minimal install.

CI green across 3.10–3.13.

A couple of non-blocking notes:

- **Possible future direction** — `_serialise_csv` / `jsonl` here overlap with what `FileDestination` already does (`file.py`), and the JSON branch differs slightly (`file.py` uses `indent=2`, `s3.py` doesnt). One option down the line, especially as GCS (#169) / Azure (#170) come in, would be a shared file-serialise helper across `file.py` + `s3.py` + the siblings — your "reusable serialiser layer" note points the same way. Not needed for this PR.
- **Scope** — for the record, this is plain object-PUT file export, not a table-format writer; Iceberg/Delta/S3 Tables stay separate (source-side #172/#173). Correct scope for this issue — just noting it so we dont conflate later.

One thing before merge: there's a trivial CHANGELOG.md conflict — `main` picked
up a `### Documentation` entry under `## [Unreleased]` from #614, and this PR adds
`### Added`. They don't actually clash; both just need to sit under `## [Unreleased]`
(Added first, then Documentation). Could you rebase / merge `main` and keep both?
Everything else looks good to go — thanks!

masukai:
Thanks for the careful review @yodakanohoshi — really appreciate the cross-check of docs against `_client()` and calling out the three-way failure split.

> there's a trivial CHANGELOG.md conflict ... Could you rebase / merge `main` and keep both?

Done — just rebased onto current `main` and resolved the `[Unreleased]` block exactly as you suggested: `### Added` (S3) first, `### Documentation` (#614) second, both intact. Re-pushed as `1034c16`. CI is re-running now.

On your two non-blocking notes:

> `_serialise_csv` / `jsonl` here overlap with what `FileDestination` already does ... shared file-serialise helper across `file.py` + `s3.py` + the siblings

Strongly agree — this is exactly the shape I want once GCS (#169) lands. The pattern I have in mind is a `drt/destinations/_file_serialiser.py` module exporting `serialise_csv` / `serialise_json` / `serialise_jsonl` / `serialise_parquet` (the latter still gated by `[parquet]` extras), called from `FileDestination` / `S3Destination` / `GCSDestination` / `AzureBlobDestination`. I'll keep it as scope-creep-out-of-this-PR but track it as a refactor follow-up once GCS lands so we have at least two callers driving the abstraction. The JSON `indent=2` divergence you spotted is real — I'd reconcile that in the same refactor (probably with a `pretty: bool` config field, defaulting to current per-destination behaviour for back-compat).

> for the record, this is plain object-PUT file export, not a table-format writer; Iceberg/Delta/S3 Tables stay separate

Confirmed. The Iceberg / Delta sources (#172/#173) are the read side; for table-format **destinations** we'd want a separate Iceberg writer destination once the source side ships, and that's a much bigger lift (catalog interaction, schema evolution, etc.). Good to have this clarified explicitly in the thread for future readers.

Thanks again!

Issue #303, opened 2026-04-13
feat: config encryption for secrets at rest


masukai:
## Summary

Add optional encryption for secrets stored in secrets.toml and state files.

## Why

Production deployments may store secrets.toml in CI or shared environments. Encryption at rest is a security best practice and an Enterprise expectation.

## Scope

- `drt encrypt` / `drt decrypt` commands for secrets.toml
- Support: age encryption (simple, no key server needed)
- Optional: environment variable for decryption key
- Encrypted files have `.enc` extension
- `drt run` auto-decrypts if key is available

masukai:
Update found during v0.10 planning (2026-08-14): this issue's premise is partly overtaken by work that shipped since it was filed (2026-04).

[#782](https://github.com/drt-hub/drt/issues/782) (secret provider URIs, shipped v0.9.0) lets any `*_env` field resolve from AWS/GCP Secret Manager or Vault instead of `secrets.toml` — `[#967](https://github.com/drt-hub/drt/pull/967)` just closed the gap where 10 connectors silently ignored a provider URI and fell through to a plain env lookup, so it now works uniformly across every connector. For anyone using a provider URI, there's no plaintext secret sitting in `secrets.toml` to encrypt in the first place — this issue's stated motivation ("production deployments may store secrets.toml in CI or shared environments") is largely addressed by pointing them at #782 instead.

The remaining real case is someone who deliberately doesn't want an external secret manager dependency and still wants `secrets.toml` itself encrypted at rest. That's a legitimate but narrower use case than originally framed — worth re-scoping the "Why" section to name that explicitly (and cross-link #782) before this gets picked up, rather than implementing against the original framing as written.

masukai:
Closed via #1001 (merged). `drt encrypt`/`drt decrypt` for `.drt/secrets.toml`, age X25519 via the optional `pyrage` extra, transparent in-memory auto-decrypt on every existing load path. Codex review caught and fixed 2 real issues (silent secrets loss on Python 3.10 without the extra, and a TOCTOU race in the no-overwrite guard).

Issue #402, opened 2026-04-22
refactor: registry-based load_profile() / save_profile()


masukai:
## Summary

`load_profile()` (credentials.py:320-438) and `save_profile()` (456-552) use long if-else / isinstance chains for 10 source types. Each new source requires editing both functions.

## Proposed changes

Use a profile registry pattern (similar to `connectors/registry.py`):
- Each profile dataclass registers its own YAML ↔ dataclass conversion
- `load_profile()` and `save_profile()` become ~10 lines each

## Priority

Low — the current code works and new sources are added infrequently. Target v0.7+.

PR #635, opened 2026-06-11
feat(cli): add drt profile commands to manage credential profiles (closes #423)

### LINKED ISSUE #423: feat: drt profile — CLI commands to manage credential profiles
## Motivation

Currently profiles (source credentials) are managed by hand-editing YAML. There is no CLI to list, add, test, or remove profiles — unlike most DB tools (psql, bq, snowsql) that ship profile management commands.

## Proposed commands

```
drt profile list                    # show all profiles in profiles.yaml
drt profile add <name>              # interactive prompt → write to profiles.yaml
drt profile remove <name>           # remove profile entry
drt profile test <name>             # run a test query against the source
drt profile show <name>             # print profile (mask secrets)
```

## Scope

- Sub-command group `drt profile` (Typer sub-app)
- `profile test` runs `SELECT 1` (or equivalent) to verify connectivity
- `profile add` supports BigQuery, PostgreSQL, DuckDB at minimum (others prompt for required fields via registry)
- Secrets are never printed in plain text (`profile show` masks values)

## Out of scope

- Encrypted storage (→ v0.9 #303)
- Profile sharing / export

masukai:
## Summary

Profiles in `~/.drt/profiles.yml` were hand-edited YAML; this adds a `drt profile` Typer sub-app matching the profile commands `psql` / `bq` / `snowsql` ship. Closes #423.

```bash
drt profile list            # all profiles + source type
drt profile show <name>     # one profile, secrets masked
drt profile test <name>     # connectivity check (source.test_connection)
drt profile add <name>      # interactive prompts → write to profiles.yml
drt profile remove <name>   # delete an entry (confirm unless --yes)
```

## Subcommands

| Command | Behaviour |
|---------|-----------|
| `list` | All profiles + their source type (rich table) |
| `show <name>` | One profile with inline secret **values** masked |
| `test <name>` | Load profile → build source via the connector registry → run its `test_connection` |
| `add <name>` | Interactive prompts; supports bigquery / duckdb / sqlite / postgres / redshift / mysql / clickhouse / snowflake (port coerced to int) |
| `remove <name>` | Delete an entry; confirmation prompt unless `--yes` |

## Secret handling

`show` never prints inline secrets in plain text — it masks values for keys matching `password` / `token` / `secret` / `api_key` / `account_sid`, while leaving **env-var-name references (`*_env`)** and **file paths (`keyfile`)** visible. This fits drt's design: real secrets live in env vars / `secrets.toml`, so a profile usually only references env-var *names* (which are safe to show). `add` / `remove` preserve the top-level `observability` block and other profiles when rewriting the file.

## Implementation

Three raw-profile helpers in `drt/config/credentials.py` — `load_raw_profiles` / `write_raw_profile` / `remove_profile` — operate at the dict level so the CLI can display/edit profiles of any type without the typed `ProfileConfig` round-trip (`load_profile` / `save_profile` stay the engine's typed path).

## Tests

18 unit tests in `tests/unit/test_cli_profile.py` — **100% coverage of `drt/cli/commands/profile.py`** — via Typer's `CliRunner` with a monkeypatched tmp `~/.drt`:

- **list** — populated, empty
- **show** — long + short secret masking, env-var-name visibility, unknown profile
- **test** — OK, raises, returns-false, unknown
- **add** — duckdb, postgres (int-port coercion), unsupported type, overwrite-declined, observability-block preservation
- **remove** — `--yes`, declined, unknown, no-file

The new credentials helpers are fully covered by these tests too.

## Skill integration

The `/drt-troubleshoot` skill's **profile** (step 2) and **connectivity** (step 3) steps now point at `drt profile show` / `drt profile test` — exactly the right tools for those layers.

## Test plan

- [x] `pytest tests/unit/test_cli_profile.py` — 18 passed, 100% file coverage
- [x] `pytest .../test_cli_commands_smoke.py .../test_cli_help_examples.py` — pass (command-enumeration tests unaffected)
- [x] `ruff check` + `mypy` clean
- [x] `make check-drift` exits 0, `make check-skills` + `make check-i18n` clean
- [x] `drt profile --help` smoke

- Encrypted storage (→ v0.9 #303)
- Profile sharing / export

Interactive `add` covers the common source types; others can be hand-added to `profiles.yml`.


yodakanohoshi (dismissed):
Thanks for the contribution! Reviewed this one. The `drt profile` sub-app (`list` / `show` / `test` / `add` / `remove`) and the dict-level helpers in `credentials.py` (`load_raw_profiles` / `write_raw_profile` / `remove_profile`) are nicely put together.

I checked the security side (credential storage at rest) and it's fine — `add` never solicits a plaintext secret; it only prompts for env-var *names* (`*_env`), so this PR doesn't write any plaintext to disk. The new command module is at 100% test coverage, the CLI registration follows the registry pattern (side-effect import in `commands/__init__.py`, no `main.py` edit), and no `type: ignore` was added. Approving 🚀

A few non-blocking follow-up notes:

- **`show` masking**: `_mask` exposes the first 2 chars for values >4 chars (`profile.py:198-202`). Values are expected to be env-var names so the risk is low, but full masking (`***`) would be safer.
- **`profiles.yml` permissions**: `write_raw_profile` / `_rewrite_profiles` write with the default umask, so the file is world-readable on POSIX. That said, the pre-existing `save_profile` does the same, so this is **not a regression in this PR**. A `chmod 0o600` would be better handled as a separate hardening ticket (out of scope here).
- **Legacy flat layout migration**: `add` / `remove` silently migrate a flat `profiles.yml` to the nested `profiles:` form. Harmless (still valid and readable), but this rewrite behavior isn't tested — a small test would be reassuring.
- **Relation to #402**: this PR deliberately bypasses the typed isinstance chains with raw dict helpers, so `_ADD_FIELD_SPECS` (`profile.py:138-185`) becomes a 3rd per-type table. Once #402 (registry-based) lands, it'd be nice to source this from that registry to reduce the number of per-source-type edit sites (you own #402 too, so you're likely already aware).

None of these are merge blockers.

masukai:
Thanks for the thorough review — and for checking the credential-storage-at-rest angle specifically. Rebased onto current `main` (resolves the CHANGELOG conflict) and folded the two cheap nits into a follow-up commit (`f2e95e3`):

- **`show` masking** → `_mask` now returns `***` unconditionally, dropping the `s[:2] + "***"` head-reveal. Env-var names are rare to be secrets, but on the off chance an inline value lands there we no longer leak a recognisable prefix. Strengthened `test_show_masks_inline_secrets` to assert no `su***`-style prefix appears.
- **Flat-layout migration** → added `test_add_migrates_flat_layout_to_nested` + `test_remove_migrates_flat_layout_to_nested`: both write a legacy flat `profiles.yml` (no `profiles:` wrapper), run `add`/`remove`, and assert the file is rewritten under `profiles:` with the other entries preserved.

Deferring the other two, per your framing:
- **`profiles.yml` perms (`0o600`)** → agreed it's better as a separate hardening ticket. As you noted it's not a regression here (`save_profile` already writes with the default umask), so I'll file it to cover the whole profile-write surface at once rather than half-fix it in this PR.
- **Sourcing `_ADD_FIELD_SPECS` from the #402 registry** → yep, I own #402; I'll collapse this third per-type table into the registry when that lands so there's one edit site.

Coverage stays 100% on `profile.py`. Re-requesting review 🙏

masukai:
Filed the file-permissions hardening as #650 (covers the whole `credentials.py` write surface — `save_profile` + `write_raw_profile`/`_rewrite_profiles` — with `0o600` applied atomically, Windows no-op, and a mode-assertion test). Labelled it `good first issue` since it's well-scoped.

Issue #469, opened 2026-05-06
refactor: introduce Destination.fetch_existing() Protocol method (replace hardcoded _QUERYABLE_TYPES)


masukai:
## Background

\`drt/destinations/query.py\` currently uses hardcoded \`isinstance\` checks against a tuple of destination config classes (\`_QUERYABLE_TYPES\`) to decide whether a destination supports queries (for \`lookups\`, \`drt test\` validation, and now \`--diff\` from #413).

This works but doesn't scale:

- Adding a new queryable destination requires touching the central \`query.py\` file
- The check happens at runtime via \`isinstance\`, not at the type level
- Doesn't fit the Protocol-based design that drives the rest of the project

## Proposed change

Add an optional Protocol method to the \`Destination\` interface:

\`\`\`python
@runtime_checkable
class QueryableDestination(Protocol):
    \"\"\"Optional protocol for destinations that can be queried.\"\"\"

    def fetch_existing(
        self,
        config: DestinationConfig,
        match_columns: list[str],
        select_columns: list[str],
    ) -> list[dict[str, Any]]:
        ...
\`\`\`

Each SQL destination (Postgres/MySQL/ClickHouse, plus Snowflake once #413-followup-1 lands) implements this. The query.py dispatcher becomes a duck-typed call: \`if hasattr(dest, "fetch_existing"): ...\`.

## When

**v0.9 milestone** — this fits the Protocol freeze prep (#300) work. We want to lock in the Protocol surface before v1.0, and adding this method is part of that.

## Migration plan

1. Add the Protocol method as **optional** (duck-typed, doesn't break existing implementations)
2. Implement on Postgres / MySQL / ClickHouse / Snowflake
3. Migrate \`drt/destinations/query.py\`, \`drt/engine/diff.py\` (after #413), and \`drt/destinations/lookup.py\` to use the new method
4. Deprecate \`_QUERYABLE_TYPES\` with a removal target

## Out of scope

- Adding query support to non-DB destinations
- BigQuery / Databricks etc. (separate destination issues)

## References

- #300 Protocol freeze prep (parent v0.9 work)
- #413 \`drt run --dry-run --diff\` (v0.7.1) — current consumer of \`_QUERYABLE_TYPES\`
- \`drt/destinations/query.py\` — current location to refactor

masukai:
Two updates found during v0.10 planning (2026-08-14):

[...]

PR #636, opened 2026-06-11
feat(snowflake): make destination queryable — enables --diff, lookups, drt test (closes #468)

### LINKED ISSUE #468: feat(snowflake): add lookups + query support (enable --diff for Snowflake destination)
## Background

Snowflake was added as a destination in v0.7 (#353), but it's not yet integrated with the lookup / query infrastructure used by Postgres/MySQL/ClickHouse. This means:

- \`lookups\` config doesn't work on Snowflake destinations
- The new \`--diff\` flag (#413, v0.7.1) falls back to sample mode for Snowflake instead of true diff

## Scope

1. Add \`lookups: dict[str, LookupConfig] | None\` to \`SnowflakeDestinationConfig\`
2. Add \`_query_snowflake(...)\` to \`drt/destinations/query.py\`
3. Add \`SnowflakeDestinationConfig\` to \`_QUERYABLE_TYPES\`
4. Wire up \`get_table_name(...)\` for Snowflake
5. Tests: lookup integration tests + query smoke tests

## Out of scope

- BigQuery / Databricks / S3 / GCS query support — separate v0.8 issues
- Protocol method abstraction — covered by the \`Destination.fetch_existing()\` Protocol redesign issue (Protocol freeze prep, v0.9)

## References

- v0.7 ROADMAP: Snowflake shipped early
- #316 / #354 — established the lookup pattern on the existing 3 destinations

masukai:
## Summary

Snowflake shipped as a destination in v0.7 (#353) but was never wired into the query infrastructure that Postgres / MySQL / ClickHouse use, so **three features silently degraded** on it:

- `drt run --dry-run --diff` fell back to **sample mode** instead of a true record-level diff
- the `lookups` FK-resolution config **did nothing**
- `drt test` validators **couldn't run** against the target table

Closes #468.

## What changed

`drt/destinations/query.py` — Snowflake added to the type-dispatched query layer:
- `_QUERYABLE_TYPES` += `SnowflakeDestinationConfig`
- `get_table_name` branch → fully-qualified `<database>.<schema>.<table>`
- `execute_test_query` → new `_query_snowflake`
- `fetch_rows` → new `_fetch_rows_snowflake`

`drt/config/models.py` — `lookups: dict[str, LookupConfig] | None` field on `SnowflakeDestinationConfig` (a forward reference to `LookupConfig`, defined later in the file; Pydantic v2 resolves it at validation time).

The new query helpers instantiate `SnowflakeDestination()` to reach its **instance-method** `_connect` (Snowflake's `_connect` isn't a `@staticmethod` like Postgres's) and run through a context-managed cursor.

With this, `is_queryable(snowflake_config)` is True and the `--diff` / `lookups` / `drt test` paths **light up automatically** — they were already type-dispatched, just missing the Snowflake case.

## Tests

5 new tests in `tests/unit/test_destination_query.py`:
- `is_queryable(snowflake)` → True
- `get_table_name` → fully-qualified `ANALYTICS.PUBLIC.USER_SCORES`
- `lookups` field parses
- `execute_test_query` → int (via a mocked connection)
- `fetch_rows` → list-of-dicts

The existing 34 query / snowflake-destination / registry tests still pass. My new lines are fully covered (the file's overall coverage number reflects the pre-existing untested postgres/mysql/clickhouse helpers, which are exercised elsewhere).

## Docs

`docs/connectors/snowflake.md`:
- new `lookups` config row
- a Notes entry explaining the v0.7.9 queryability and what it unlocks. (The doc's prior "drt test works with Snowflake" line was **aspirational** before this PR — Snowflake wasn't queryable, so `drt test` skipped it. It's now accurate.)

## Test plan

- [x] `pytest tests/unit/test_destination_query.py` — 8 passed
- [x] `pytest .../test_snowflake_destination.py .../test_connector_registry.py` + diff/lookup/test_runner tests — all pass
- [x] `ruff check` + `mypy` clean
- [x] `make check-drift` exits 0, `make check-skills` + `make check-i18n` clean
- [x] CI green

- BigQuery / Databricks / S3 / GCS query support — separate v0.8 issues
- `Destination.fetch_existing()` Protocol abstraction — deferred to v0.9 Protocol-freeze prep


yodakanohoshi (approved):
Thanks for the contribution! Reviewed this one.

Wiring `SnowflakeDestinationConfig` into the type-dispatched query layer in `query.py` (adding it to `_QUERYABLE_TYPES`, the FQN `get_table_name` branch, `_query_snowflake` / `_fetch_rows_snowflake`, and the `lookups` field) is clean and consistent with the existing Postgres / MySQL / ClickHouse siblings. This lights up true record-level `--diff`, `lookups`, and `drt test` for the Snowflake destination. Approving 🚀

Notes (all non-blocking):

- **Identifier quoting**: `get_table_name` returns an unquoted dotted FQN (`database.schema.table`), which matches the existing Postgres/MySQL/ClickHouse branches and isn't a new defect here — the `_quote_ident` fix family lives in the write paths, and this query layer was already unquoted. Snowflake parses the 3-part identifier correctly, so no practical impact, but case-sensitive / reserved-word identifiers are something to keep in mind for a future query-layer hardening pass.
- **`_QUERYABLE_TYPES` hardcoding vs #469**: this extends the existing type-tuple + isinstance dispatch rather than the `Destination.fetch_existing()` Protocol from #469, deferring the abstraction to the v0.9 Protocol-freeze prep. That's the pragmatic call — a one-off Protocol migration for a single connector would be the worse outcome. Just flagging that it adds a 4th hardcoded type, slightly raising the migration cost when #469 lands. If we'd rather do the Protocol first, let's discuss.
- **`lookups` field placement**: correct — the Snowflake config is a standalone `BaseModel` (doesn't inherit `BaseSqlDestinationConfig`, where the other three get `lookups`), so a dedicated field is the right move.

**Merge order vs #642**: no code conflict, but `docs/connectors/snowflake.md` and `CHANGELOG.md` will conflict textually. No semantic dependency either way — let's merge this one (#636) first and have #642 do a trivial rebase of those two files.

PR #637, opened 2026-06-11
feat(engine): dead letter queue + drt retry — persist & replay failed records (closes #278)

### LINKED ISSUE #278: feat: dead letter queue — persist failed records for retry
## Summary

When a sync partially fails (some records succeed, some fail), persist the failed records locally so they can be retried later.

## Motivation

In production, partial failures are common (API rate limits, validation errors on specific records). Currently failed records are logged but lost. A dead letter queue enables \`drt retry [sync_name]\` to re-process only the failed records.

## Scope

- Store failed records in \`~/.drt/dlq/[sync_name]/\` as JSON
- Include: original record, error message, timestamp, attempt count
- \`drt retry [sync_name]\` command to re-process DLQ
- \`drt status\` shows DLQ depth per sync
- Configurable max DLQ size

masukai:
Closes #278.

## What

A **Dead Letter Queue** for per-record load failures, plus a `drt retry` command to replay them. Partial failures (a per-record `429`, a few rows failing downstream validation while the batch otherwise succeeds) are routine in production — until now those records were counted in `result.failed` and previewed via `RowError`, but the **records themselves were dropped**. Recovering them meant re-running the whole sync.

## How it works

- **Opt-in per sync** — `sync.dlq.enabled: true` (default `false`). Unlike `RowError` (which keeps only a 200-char preview to avoid logging PII), the DLQ writes the **full** record to `.drt/dlq/<sync_name>.jsonl` — a privacy decision the operator makes explicitly. `sync.dlq.max_records` (default 10000, `0` = unbounded) caps growth FIFO, keeping the newest failures.
- **Stays within the engine purity boundary** — a new `SyncObserver.on_records_failed(sync_name, dead_letters)` hook carries failures out of `engine/sync.py`. The engine pairs each `RowError.batch_index` back to the **full post-mapping record it sent** (a pure correlation — no I/O); a new `DlqObserver` does the persistence. `tests/unit/test_engine_observer.py`'s boundary checks still pass.
- **`drt retry <sync>`** — replays in `sync.batch_size` chunks, drops now-succeeding records, writes the rest back with bumped `attempts`. Flags: `--limit N`, `--dry-run`, `--clear`. Needs only the destination (records replay verbatim). Attribution is conservative: keeps a whole chunk queued when `row_errors` don't fully account for `result.failed`.
- **`drt status`** — `⚠ <sync>: N record(s) in dead letter queue` line (text) + `dlq_depth` per sync (json), including syncs with no state row yet.

## Scope

Works across the ~24 destinations that emit `row_errors` (REST API, Slack/Discord/Teams, HubSpot/Notion/Intercom/Linear/Jira, the SQL destinations, …). Staged destinations and lookup-filtered rows are out of scope for v1 — their original records aren't available at the failure point.

## Tests

35 new tests; **full unit suite 1502 green**, `ruff` + `mypy` (126 files) clean, `make check-drift` / `check-skills` clean.

- `tests/unit/test_dlq_store.py` (13) — store mechanics, FIFO cap, corrupt-line skip
- `tests/unit/test_engine_observer.py` — `DlqObserver` persistence / fire-and-forget swallow / no-op-on-other-events + engine-emits-full-records correlation + no-emit-on-clean-batch + composite forwards 6 methods
- `tests/unit/test_cli_retry.py` (7) — empty queue, unknown sync, full drain, partial-keeps-and-bumps-attempts, dry-run, clear, limit
- `tests/unit/test_cli_status_coverage.py` (2) — text DLQ line + json `dlq_depth`
- `tests/unit/test_config.py` (4) — opt-in default, parse, max_records default, negative rejected

## Docs

`docs/llm/API_REFERENCE.md` sync block, new `docs/guides/dead-letter-queue.md`, and the `/drt-troubleshoot` skill's partial-failure step. CHANGELOG `[Unreleased]` updated.


yodakanohoshi (dismissed):
Thanks for the contribution! Big PR, but it's well put together.

I verified the most important part — the **engine I/O boundary is respected**. `engine/sync.py` only does a pure correlation (pairing `result.row_errors[].batch_index` back to `record_batch[idx]` into `DeadLetter` objects and calling `observer.on_records_failed(...)`); there are no direct `logger.*` / `.save_sync(` / `watermark_storage.save(` calls, and all persistence is concentrated in `DlqObserver`. The boundary tests (`tests/unit/test_engine_observer.py`) are intact, the new hook is propagated across `SyncObserver` (Protocol) / `NullObserver` / `LoggingObserver` (silent) / `CompositeObserver` / `StatePersistingObserver`, and the CLI command follows the registry pattern. Test coverage is solid too. Approving 🚀

A few non-blocking nits — feel free to fold into a quick push or a follow-up issue:

- **Negative `--limit`**: `entries[:max(0, limit)]` clamps a negative value to 0, so it silently becomes a no-op. Consider erroring out, or a doc note.
- **Re-queuing on un-attributable batches**: when `row_errors` don't fully account for `result.failed` (`pinpointed=False`), the whole chunk is re-queued, so rows that actually succeeded get re-sent and may be counted as `failed_again` next time. This is a reasonable conservative tradeoff ("prefer re-send over silent drop"), but a one-line comment noting "succeeders may be re-queued" would make the intent clearer.
- **Lock scope**: `DlqStore._lock` is a `threading.Lock`, i.e. process-local. It protects `--threads N` workers within a single process, but concurrent `drt run` and `drt retry` in separate processes doing read-modify-write on the same file is last-writer-wins. Worth aligning the docstring with that.

None of these are merge blockers.

masukai:
Thanks for the careful read — especially confirming the engine I/O boundary holds, since that's the load-bearing constraint for this one. Folded all three nits into a single follow-up commit (`9604395`), rebased onto current `main` (picks up #642's Snowflake `replace`):

- **Negative `--limit`** → now errors instead of silently no-op'ing. Added an explicit guard before the store is touched (`--limit must be >= 0.` → exit 1) and dropped the `max(0, limit)` clamp so the slicing reads straight (`entries[:limit]` / `entries[limit:]`). New test `test_retry_negative_limit_errors` covers it.
- **Re-queuing on un-attributable batches** → extended the inline comment in `retry.py` to spell out the trade-off: on a batch we can't attribute per-record, rows that actually succeeded get re-queued and may be re-sent next retry — we prefer a re-send (idempotent for upsert destinations) over a silent drop.
- **Lock scope** → `DlqStore` docstring now states `_lock` is a **process-local** `threading.Lock`: it serialises `--threads N` workers within one process but does **not** coordinate across processes, so a concurrent `drt run` + `drt retry` on the same file is last-writer-wins.

No logic change beyond the negative-limit guard. Re-requesting review when you have a moment 🙏

masukai:
Also pushed a docs-only follow-up (`027c1ea`): a **"Make retries idempotent"** section in `docs/guides/dead-letter-queue.md`. `drt retry` re-sends payloads verbatim, so a record that reached the destination but was reported failed (timeout-after-write, ambiguous 5xx) would be written twice on append-only sinks — the section documents the per-destination dedupe knobs (`upsert_key`, Elasticsearch `id_field`, REST idempotency-key endpoints) so users configure one before relying on the queue. No code change.

PR #642, opened 2026-06-13
feat(snowflake): mode: replace — truncate default + replace_strategy: swap (closes #434)

### LINKED ISSUE #434: feat: Snowflake support for replace_strategy: swap

masukai:
Closes #434.

## What

Adds `sync.mode: replace` to the **Snowflake** destination, including the zero-downtime `replace_strategy: swap`.

**Surprise found while implementing:** Snowflake had **no `mode: replace` handling at all** — `load()` only routed `config.mode` insert/merge + `sync.mode: mirror`, so `mode: replace` silently fell through to insert/merge. (Databricks has the same gap — see follow-up below.) So this adds the **full replace branch**, not just swap, at parity with Postgres/MySQL/ClickHouse:

- **`replace_strategy: truncate`** (default) — `TRUNCATE TABLE` once, then INSERT each batch.
- **`replace_strategy: swap`** — shadow `<table>__drt_swap` via `CREATE OR REPLACE TABLE … LIKE` (carries clustering keys) → write batches → `finalize_sync` does an atomic `ALTER TABLE … SWAP WITH` then drops the old shadow. `SWAP WITH` preserves grants on the original name. First run (target absent) writes directly and skips the swap.
- **Orphan cleanup** — implements the `OrphanSwapCleanup` protocol (`list_orphan_swap_tables` / `drop_orphan_swap_tables`, `__drt_swap`-suffix validated) so `drt clean --orphans` works on Snowflake too.

## Scope / safety

- **Isolated to `drt/destinations/snowflake.py`** — no `drt/config/models.py` change (`replace_strategy` already lives on `SyncOptions`), so this is conflict-free with the in-flight Snowflake-queryable PR.
- 11 new tests in `tests/unit/test_snowflake_destination.py` (existing tests untouched): truncate once-across-batches, swap shadow-create + finalize SWAP/DROP, swap first-run direct-write, `on_error=fail` mid-swap shadow cleanup, no-op finalize for insert mode, orphan list/drop suffix safety.
- `ruff check` + `mypy drt` clean; full unit suite green.

## Follow-up (not in this PR)

- **Databricks** has the same missing-`replace` gap — should get the same treatment in a separate issue/PR.


Muawiya-contact:
Nice work, @masukai — the implementation looks solid.

The silent fallthrough on `mode: replace` is a real catch. Good that it got folded in here rather than left for a separate cleanup ticket.

A few things I wanted to confirm before approving:

- **Grants on swap** — you mentioned `SWAP WITH` preserves grants on the original table name, which lines up with [Snowflake's docs on object rename/swap](https://docs.snowflake.com/en/sql-reference/sql/alter-table). No concerns there.
- **First-run path** — skipping swap and falling through to a plain `CREATE` on the first run makes sense. Worth double-checking that the code has a short comment explaining the branch so the next person doesn't wonder why.
- **Orphan cleanup** — didn't expect this in scope but it's a good bonus. The `__drt_swap` suffix validation is important — glad it's in the tests.
- **Test coverage** — 11 tests covering truncate, swap, first-run, mid-swap error cleanup, and orphan list/drop is thorough. Codecov coming back clean is reassuring.

One minor thought: since `CREATE TABLE … LIKE` copies clustering keys but not policies, it might be worth a one-line note in the docstring or inline comment — not a blocker, just good for whoever hits a policy use case later.

Happy to approve from my side. Since `yodakanohoshi` is already on review, no need to wait on me — but if you want a second set of eyes, LGTM. 🟢

masukai:
Thanks @Muawiya-contact 🙏 — appreciate the careful read.

- **First-run**: there's already an inline comment on that branch (`# First run: nothing to swap against — write straight to target.`) so the next person isn't left guessing.
- **Policies**: good catch. `CREATE TABLE … LIKE` carries clustering keys but **not** masking / row-access policies or tags — and since `SWAP WITH` swaps the underlying objects, a policy-bearing target would lose them. Added a note to the docstring + connector doc (and tightened the doc, which previously over-claimed that policies survive — they don't; only role grants do). Pushed in 2af611d.
- **Orphan cleanup**: figured if we're creating shadow tables we owe users a cleanup path — `drt clean --orphans` already existed for the Postgres / MySQL / ClickHouse legs, so it was mostly wiring Snowflake into the same `OrphanSwapCleanup` protocol.

Thanks for the 🟢 — no pressure at all given Phase 3 + finals are on your plate. yodakanohoshi can take the formal pass.

(FWIW the same pattern just landed for Databricks in #644 — Delta has no `SWAP WITH`, so that one uses an atomic `INSERT OVERWRITE` instead.)

yodakanohoshi (dismissed):
Thanks for the contribution! And thanks for fixing the silent fallthrough where `sync.mode: replace` was falling through to insert/merge.

The two strategies — `truncate` (default) and `swap` (zero-downtime via `ALTER TABLE … SWAP WITH`) — plus the `OrphanSwapCleanup` implementation faithfully mirror the existing Postgres implementation. Type narrowing via `assert isinstance`, no `type: ignore`, and the suffix-validated orphan drop are all convention-compliant, and the 11 new tests cover the paths well. Approving 🚀

One thing I'd love to see aligned before merge (non-blocking):

- **State reset on finalize failure**: the `finally` in `finalize_sync` unconditionally resets the in-memory state (`_swap_shadow_created` / `_swap_table`) even when `SWAP WITH` / `DROP` raises. That means an in-process retry is no longer possible and the shadow is left behind (recoverable via `drt clean --orphans`). Postgres splits the DROP into a separate transaction specifically to avoid this (around `postgres.py:421`) — aligning with that would match the sibling implementation.

Other notes (no action needed):

- **Empty source in `replace` is a no-op**: the `if not records: return SyncResult()` early return means an empty source doesn't truncate/replace, so existing rows survive. This matches Postgres behavior and the mirror "transient empty source" safety stance, so it's fine — just confirm #434 doesn't expect "empty source = empties the table" semantics.
- **f-string SQL identifier interpolation**: `TRUNCATE TABLE {table_fq}` etc. are unquoted, but this matches the existing convention in this file (the mirror DELETE and insert/merge paths do the same), and the values come from config, not user rows, so it's not a new injection surface. The `_quote_ident` hardening that Postgres/MySQL/ClickHouse got isn't applied here, but that's out of scope for this PR.

**Merge order**: this conflicts textually with #636 (Snowflake queryable) only in `docs/connectors/snowflake.md` and `CHANGELOG.md`. We'll merge #636 first, so a trivial rebase of those two files here afterwards is all that's needed.

masukai:
Thanks for the thorough review @yodakanohoshi 🙏 Addressed:

- **State reset on finalize failure** (your before-merge ask): moved the in-memory state reset (`_swap_shadow_created` / `_swap_table`) out of the `finally` so it runs **only after `SWAP WITH` succeeds** — a failed swap now leaves state intact, so the shadow stays recoverable via `drt clean --orphans` and a retry is possible. Mirrors postgres.py's separate-transaction split (Snowflake autocommits, so the SWAP commits before the DROP). Added `test_finalize_swap_failure_preserves_state`. (e35c876)
- **Rebased on main** after #636 landed — the only conflict was the textual one you predicted in `docs/connectors/snowflake.md` + `CHANGELOG.md`, resolved trivially.
- **Empty source = no-op**: confirmed — #434 doesn't expect 'empty source empties the table'; matching the Postgres / mirror 'transient empty source' safety stance is intended.
- **f-string SQL identifiers**: agreed it's the existing in-file convention (config-sourced, not row data) — left for the future `_quote_ident` hardening pass, out of scope here.

Green + mergeable. 🚀

yodakanohoshi:
Thanks for addressing the feedback! `e35c876` adds the finalize-failure state preservation (keeping the shadow around so a retry is possible, matching Postgres), which resolves the one non-blocking point I'd raised. The rebase also cleared the CHANGELOG conflict.

Took another full pass — both `truncate` / `swap` strategies, the `OrphanSwapCleanup` implementation, the suffix validation, and the tests all look good. Merging 🚀 Thanks for the work!

yodakanohoshi (approved):

PR #651, opened 2026-06-17
fix(cli): drt destinations was missing 9 registered connectors + close drift gap

masukai:
## Problem

`drt destinations` (and `--detailed` / `--format json`) was listing only **23 of 32** registered destinations. The nine missing — `bigquery`, `databricks`, `snowflake`, `s3`, `gcs`, `azure_blob`, `elasticsearch`, `mixpanel`, `salesforce_bulk` — were properly registered in `drt/connectors/registry.py` and fully functional in syncs, but invisible to the CLI. **All six cloud destinations shipped in v0.7.9 were affected**, so a user running `drt destinations` right after upgrading would not see BigQuery, Databricks, S3, GCS, Azure Blob, or Elasticsearch.

### Root cause
`drt destinations` reads a **hand-maintained** `DESTINATIONS` list in `drt/config/connectors.py` (and a parallel `DESTINATION_CONFIG_CLASSES` map in `drt/cli/_connector_detail.py` for `--detailed`), not the registry. Both drifted from the registry over time.

### Why it slipped through
The existing parity tests checked `DESTINATIONS ⊆ DESTINATION_CONFIG_CLASSES` (every listed destination has detail metadata) — but **nothing checked the other direction** (`registry ⊆ DESTINATIONS`). A connector could be registered and never listed without failing any test. The post-merge `check_drift.sh` audit covers docs / skills / README / the MCP inventory, but **not** this CLI list.

## Fix
- Synced `DESTINATIONS` (`drt/config/connectors.py`) and `DESTINATION_CONFIG_CLASSES` (`drt/cli/_connector_detail.py`) to all 32 registered destinations.
- **Closed the drift blind spot**: added `test_DESTINATIONS_matches_registry` + `test_SOURCES_matches_registry` asserting the CLI lists equal the registry **exactly**. These fail at **PR time** (stronger than the post-merge drift workflow), so a future connector that's registered but not listed breaks the build immediately.

`SOURCES` was already in sync — verified and now guarded too.

## Verification
- `drt destinations` now lists all 32 (the 9 previously-missing confirmed present).
- New parity tests pass; existing `test_cli_list_connectors.py` suite green (14 passed).
- Full unit suite: 1466 passed, 11 skipped. `ruff` + `mypy` clean.

## Note / possible follow-up
These lists are redundant with the registry (which already stores each type's config class). A future refactor could derive both from the registry and delete the hand-maintained lists entirely — out of scope here to keep the fix small and reviewable; the new parity tests make the current duplication safe in the meantime.


yodakanohoshi (approved):
Thanks for this! I checked whether the drift gap I flagged in the #646 review is actually closed — all three points are resolved:

- **`drt destinations` now enumerates the full registry**, so the 9 destinations that were previously invisible from the CLI (the six v0.7.9 cloud/DWH connectors + `mixpanel` / `salesforce_bulk` / `snowflake`) now show up. I verified on the branch that the `DESTINATIONS` set exactly equals `_destination_registry` (32 entries).
- **`salesforce_bulk` is now registered** in both `drt/config/connectors.py` and `drt/cli/_connector_detail.py`.
- Most importantly, the new **bidirectional parity test** (`DESTINATIONS == registry`) enforces the registry as the single source of truth, so the same drift will fail CI if it ever recurs. Closing the hand-maintained-list blind spot this way is the right approach, and using a PR-time test to cover what `check_drift.sh` (post-merge) doesn't reach for this CLI list is a good call.

Scope is tightly focused on the fix. Approving 🚀

Optional nit (no action needed): `DESTINATION_CONFIG_CLASSES` is currently guarded transitively (`DESTINATIONS ⊆ detail_map` + `DESTINATIONS == registry`); adding a direct `detail_map == registry` assertion would also catch a stale *extra* entry in the detail map. The PR's own follow-up note (derive the lists from the registry) is the proper long-term fix, with this as a stopgap.

Issue #317, opened 2026-04-14
epic(destinations): schema-aware serialization via INFORMATION_SCHEMA introspection


masukai:
## Vision

Eliminate the need for users to declare \`json_columns\` (#316) or hit per-type adapter crashes (#311, #315) by **introspecting the destination schema at sync start** and routing each value through the correct serializer based on the actual column type.

This is the \"fundamental\" layer of the JSON serialization design space. Layer 1 (per-destination patches: #311, #315) is shipping now; Layer 2 (config-driven \`json_columns\`: #316) is for v0.6; this issue is **Layer 3 for v0.7+**.

## Approach

1. At sync start, query \`INFORMATION_SCHEMA.COLUMNS\` once for the destination table:
   \`\`\`sql
   SELECT column_name, data_type
   FROM information_schema.columns
   WHERE table_schema = ? AND table_name = ?
   \`\`\`
2. Build a \`column_name -> type_category\` map, where \`type_category\` is one of: \`json\`, \`array\`, \`scalar\`, \`unknown\`.
3. For each record value, dispatch serialization based on the destination column's category:
   - \`json\` (JSONB / JSON): use the destination's native JSON wrapper
   - \`array\` (Postgres ARRAY, MySQL — N/A): pass list through to driver adapter
   - \`scalar\`: passthrough
   - \`unknown\`: fall back to current behavior
4. Cache the map for the duration of the sync (don't re-query per row).

## Why this is the right destination

- Zero config burden on users
- Handles edge cases that \`json_columns\` cannot (e.g. \`list → JSONB\` ambiguity)
- Matches behavior of mature reverse-ETL tools (dlt, Airbyte) and ORMs (SQLAlchemy)
- Foundation for **DWH-as-destination** work (#milestone v0.7) — Snowflake/BigQuery/Databricks/ClickHouse all expose information_schema and benefit from the same machinery

## Trade-offs

- One extra round-trip per sync (negligible vs typical sync cost)
- Requires read access to \`information_schema\` — most users have this, but locked-down environments may not. Fallback: if introspection fails, revert to Layer 2 \`json_columns\` config (graceful degradation).
- Increased test surface: each destination needs introspection coverage

## Scope (for v0.7 epic)

- [ ] Define a \`DestinationSchema\` protocol (or extend the existing destination interface) with \`describe_columns(table) -> dict[str, ColumnType]\`
- [ ] Implement for: Postgres, MySQL, Snowflake, BigQuery (as destination), ClickHouse
[...]

PR #653, opened 2026-06-19
feat(snowflake): schema-aware VARIANT serialization via PARSE_JSON (#317 Layer 3, Phase 2a)

masukai:
**Part of #317 (Layer 3 — schema-aware serialization). Phase 2a: Snowflake.** Stacked on #652 (Phase 1 — Postgres/MySQL); base will retarget to `main` once #652 lands.

## Problem
Snowflake's semi-structured columns (`VARIANT` / `OBJECT` / `ARRAY`) can't take a `dict`/`list` as a plain bound parameter — the connector stores a stringified `repr`, not parsed JSON. You need `PARSE_JSON`. Today drt binds raw values, so a dict→VARIANT lands wrong.

## Fix
`describe_columns` (from #652) extended to Snowflake: reads `INFORMATION_SCHEMA.COLUMNS` (per-database, `UPPER(...)` so quoted/unquoted identifiers both resolve), maps `VARIANT`/`OBJECT`/`ARRAY` → `json`.

Since **Snowflake disallows functions in a `VALUES` clause**, when any column needs wrapping the destination switches the INSERT to the `SELECT` form and wraps those columns with `PARSE_JSON`:
```sql
INSERT INTO db.schema.t (id, payload) SELECT %s, PARSE_JSON(%s)   -- payload is VARIANT
```
binding `json.dumps(value)`. With no JSON column the clause is the unchanged `VALUES (%s, …)` — byte-identical to before. Applied to **all** write paths: insert / merge-staging / replace-truncate / replace-swap (via a shared `_value_clause` + `_bind_row`).

`introspect_schema: true` (new field, default on; opt out for roles without `information_schema` read) + per-sync cache mirror the Phase-1 design. SQL-gen tests pin `introspect_schema=False` to isolate; 9 new tests cover the introspection mapping/query and `load()` integration (VARIANT → `SELECT`/`PARSE_JSON`/`json.dumps`, all-scalar → `VALUES`, cache, opt-out). Full suite 1559 passed; `ruff`/`mypy` clean.

## ⚠️ Please validate against a real warehouse before merge
`INSERT … SELECT … PARSE_JSON(%s)` is the canonical Snowflake VARIANT-load form, but **this suite is mock-based** (as is the entire Snowflake destination — no live account in CI). A quick sanity insert of a dict into a VARIANT column on a real Snowflake would de-risk the SQL shape.

## Scope
**ClickHouse / BigQuery: verified no serialization gap** — `clickhouse-connect` is schema-aware and the BigQuery client is JSON-native, so no speculative introspection wiring (would be dead code). **Databricks**: tracked as a later #317 phase.


masukai:
Tracked the real-warehouse validation as #654 (a single tracker for all the mock-only DWH paths — this PR's PARSE_JSON form is the top item there). So the "sanity-check on a live Snowflake" note won't get lost.

PR #608, opened 2026-06-01
feat(destinations): add Mixpanel destination (profiles + events)

Pawansingh3889:
## What does this PR do?

Adds a **Mixpanel** reverse-ETL destination (closes #417) with the two endpoints from the issue scope, selected by `endpoint`:

- **`people_set`** (default) → `/engage#profile-set`. Sets user-profile properties; auth is the project token, carried inside each record (no header).
- **`import_events`** → `/import`. Ingests events; auth is a service account (HTTP Basic) + the numeric `project_id` query param.

Both batch up to **2000 records/request** (Mixpanel's limit) and support **EU residency** (`api-eu.mixpanel.com`). Modelled on the existing Amplitude destination and reuses the shared `RateLimiter` / `with_retry` / `SyncResult` / `RowError` plumbing — no new dependencies beyond httpx.

Event imports get a **deterministic `$insert_id`** (derived from the row when `insert_id_field` is unset) so re-running a sync doesn't double-count events.

## Config

```yaml
# profile set
destination:
  type: mixpanel
  endpoint: people_set
  project_token_env: MIXPANEL_TOKEN
  distinct_id_field: user_id

# event import
destination:
  type: mixpanel
  endpoint: import_events
  project_id: "1234567"
  service_account_username_env: MIXPANEL_SA_USERNAME
  service_account_secret_env: MIXPANEL_SA_SECRET
  event_name: signup_completed
  time_field: event_time
```

## Related Issue

Closes #417.

## How was this tested

- [x] 13 unit tests in `tests/unit/test_mixpanel_destination.py` — both endpoints, payload shapes (`$token`/`$distinct_id`/`$set` and `event`/`properties`), EU routing, service-account auth + `project_id` param, deterministic `$insert_id`, batch splitting, HTTP error → row errors, config validation. Run alongside the Amplitude suite: 31 passed.
- [x] `ruff check` + `ruff format` clean on all touched files.
- [x] Registry round-trip verified: `type: mixpanel` resolves to both config and destination class for each endpoint.

Wire format checked against Mixpanel's API reference ([profile-set](https://developer.mixpanel.com/reference/profile-set), [import-events](https://developer.mixpanel.com/reference/import-events)).

## Checklist

- [x] Tests pass (`make test`)
- [x] Linter passes (`make lint`)
- [x] Registry entries added (config union + `connectors/registry.py`)
- [x] Docs added (`docs/connectors/mixpanel.md` + README destination table)

[...]

PR #655, opened 2026-06-19
feat(config): write ~/.drt credentials 0o600 / dir 0o700 on POSIX

### LINKED ISSUE #650: Harden ~/.drt/profiles.yml file permissions (0o600) at all credentials.py write sites
### Summary

`drt` writes credential-adjacent files (`~/.drt/profiles.yml`, and `~/.drt/secrets.toml` if present) with the process's default umask, so on POSIX they typically land `0644` (world-readable). drt's design keeps *real* secrets in environment variables — `profiles.yml` usually only references env-var *names* (`*_env`) — but the file can still contain inline values (hosts, users, and occasionally an inline password for quick local setups), and a world-readable credential file is a poor default regardless.

### Surface (write sites to harden)

All in `drt/config/credentials.py`:

- `save_profile` — typed write path used by the engine
- `write_raw_profile` / `_rewrite_profiles` — the `drt profile add` / `remove` path (added in #423 / #635)

These should write with `0o600` (and create the parent `~/.drt` as `0o700`), applied **atomically** with the write (`os.open(..., O_CREAT|O_WRONLY|O_TRUNC, 0o600)`, or a `chmod` on the fd immediately after create and before content) so there's no world-readable window between create and write.

### Why a separate ticket

Surfaced in review of #635 (`drt profile` CLI). It's **not a regression there** — the pre-existing `save_profile` already writes with the default umask — so fixing it piecemeal inside #635 would only cover half the write surface. This ticket covers the whole `credentials.py` write path in one place.

### Scope

- POSIX `0o600` on `profiles.yml` (and `~/.drt` dir `0o700`) at every write site.
- Windows: no-op — NTFS ACLs differ; guard the `chmod` behind a `os.name == "posix"` check so it doesn't error on Windows.
- A test asserting the written file's mode is `0o600` on POSIX (`stat.S_IMODE(os.stat(path).st_mode) == 0o600`).

### Out of scope

- Encrypted-at-rest profile storage — tracked separately for v0.9 (#303).

Pawansingh3889:
Closes #650.

## What

`~/.drt/profiles.yml` was written with the process umask, so on POSIX it usually lands `0o644` (world-readable) even though it can hold inline credentials (hosts, users, the occasional inline password for a quick local setup).

This hardens every write site in `drt/config/credentials.py`:

- `save_profile` (the typed engine path)
- `_rewrite_profiles`, used by both `write_raw_profile` and `remove_profile` (the `drt profile add` / `remove` path added in #635)

Two small helpers do the work:

- `_open_private` opens the file with `os.open(..., O_CREAT | O_WRONLY | O_TRUNC, 0o600)`, so a newly created file is never world-readable, and `chmod`s the descriptor so a pre-existing `0o644` file gets tightened too.
- `_ensure_private_dir` creates `~/.drt` as `0o700`.

Both are guarded behind `os.name == "posix"` and are a no-op on Windows (NTFS ACLs differ), as the issue suggested.

## Notes

- `os.fchmod` isn't in the Windows typeshed stub and the repo doesn't allow `type: ignore`, so I used `os.chmod(fd, ...)` instead — descriptor-based, same effect, mypy-clean cross-platform.
- Out of scope, per the issue: encrypted-at-rest profile storage (tracked in #303).

## Tests

New `tests/unit/test_credentials_permissions.py`: asserts `save_profile` / `write_raw_profile` write `0o600`, the `~/.drt` directory is `0o700`, and a pre-existing `0o644` file is tightened. The mode assertions are POSIX-gated (they run on Linux in CI); a platform-agnostic round-trip test guards content parity. ruff and mypy pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **`~/.drt/profiles.yml` is now written `0o600` (owner-only) on POSIX** (#650) — credential-adjacent files were written with the process umask (typically world-readable `0o644`). All `credentials.py` write sites (`save_profile` and the `drt profile add` / `remove` path via `_rewrite_profiles`) now create the file atomically with mode `0o600` via `os.open(..., O_CREAT, 0o600)` (no world-readable window) and `chmod` the descriptor to tighten a pre-existing `0o644` file; the `~/.drt` directory is created `0o700`. No-op on Windows (NTFS ACLs differ), guarded behind `os.name == "posix"`.
+
 ### Fixed
 
 - **`drt destinations` was missing 9 registered connectors** — `bigquery`, `databricks`, `snowflake`, `s3`, `gcs`, `azure_blob`, `elasticsearch`, `mixpanel`, and `salesforce_bulk` were registered in the connector registry but absent from the hand-maintained `DESTINATIONS` list in `drt/config/connectors.py`, so `drt destinations` / `drt destinations --detailed` / `--format json` silently listed only 23 of 32 (every cloud/DWH destination, including all six added in v0.7.9, was invisible to the CLI even though it worked in syncs). Synced both `DESTINATIONS` and the `DESTINATION_CONFIG_CLASSES` map in `drt/cli/_connector_detail.py` to the registry. **Closed the drift blind spot**: the existing parity tests only checked `DESTINATIONS ⊆ _connector_detail`, never `registry ⊆ DESTINATIONS` — new `test_DESTINATIONS_matches_registry` / `test_SOURCES_matches_registry` assert the CLI lists equal the registry exactly, failing the build at PR time (the post-merge `check_drift.sh` audit never covered this surface). `SOURCES` was already in sync.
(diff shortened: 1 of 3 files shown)


masukai (approved):
🎉 This is exactly right — thank you @Pawansingh3889! Clean, complete hardening of the whole `credentials.py` write surface:

- **Atomic** `os.open(..., O_CREAT | O_WRONLY | O_TRUNC, 0o600)` so there's never a world-readable window, **plus** the descriptor `os.chmod(fd, 0o600)` to tighten a pre-existing `0o644` file — both directions covered.
- `~/.drt` created `0o700`, and every write site (`save_profile` + the `drt profile add`/`remove` path via `_rewrite_profiles`) goes through it.
- POSIX-guarded (`os.name`), clean no-op on Windows.
- Real `stat.S_IMODE` assertions (fresh-write `0o600` + tightens-preexisting) — exactly the test I'd hoped for.

CI all green, 100% patch coverage, CHANGELOG entry present. Approving + merging 🚀

If drt's been useful, a ⭐ on the repo is always appreciated — and this is your second solid contribution (after Mixpanel #608 🙌), so if you'd like more in this vein, the [good first issues](https://github.com/drt-hub/drt/issues?q=is%3Aopen+label%3A%22good+first+issue%22) are open. Thanks again!

Issue #619, opened 2026-06-07
feat(otel): Phase 3 — engine instrumentation spans (sub-issue of #429)


masukai:
Phase 3 of the OpenTelemetry epic [#429](https://github.com/drt-hub/drt/issues/429). Phase 1 ([PR #527](https://github.com/drt-hub/drt/pull/527)) shipped the config schema for `observability.otel` + the `drt-core[otel]` extras. Phase 2 ([PR #615](https://github.com/drt-hub/drt/pull/615) by @Muawiya-contact) just landed and gives us a lazy, idempotent `get_tracer()` / `get_meter()` that returns OTLP-backed providers when configured and zero-overhead no-ops in every other case. This issue is the next step toward actually emitting traces from a running sync.

## Scope

Wrap the engine's sync execution with structured spans so a sync run can be observed end-to-end via any OTLP-compatible collector (Jaeger, Tempo, Grafana, Honeycomb, etc.).

### Acceptance criteria

- [ ] `engine/sync.py` `run_sync` (or its equivalent entry point) wraps the full sync in a top-level span: `drt.sync.run`.
  - Span attributes: `sync.name`, `source.type`, `destination.type`, `sync.mode`, `batch_size` (taken from `SyncOptions`).
  - Status set to `OK` on clean completion, `ERROR` with `record_exception` on uncaught failures.
- [ ] Within `drt.sync.run`, the extract phase is wrapped as `drt.sync.extract` (child span).
  - Span attributes: `extract.rows_extracted` (set after extraction completes).
- [ ] Within `drt.sync.run`, each batch sent to `destination.load()` is wrapped as `drt.sync.load` (child span).
  - Span attributes: `batch_index`, `batch_size`, `load.success`, `load.failed`, `load.skipped` (taken from the returned `SyncResult`).
- [ ] **Zero-cost when OTel is off.** `get_tracer()` returns `_FallbackNoOpTracer` when `[otel]` isn't installed or no endpoint is configured (Phase 2 contract); the engine should call `with tracer.start_as_current_span(...)` unconditionally — no `if otel_enabled:` branches — and rely on the no-op fallback for the zero-cost path. Confirm with a benchmark or comment that the no-op spans add negligible overhead.
- [ ] Documented in `docs/connectors/` or a new `docs/observability.md` — at minimum the YAML config to enable OTLP export, a sample collector setup (e.g. Jaeger via docker-compose), and what the span tree looks like for a typical sync.

### Verification plan

[...]

PR #664, opened 2026-06-23
feat(otel): BatchSpanProcessor + flush-on-exit for trace export (#658)

### LINKED ISSUE #658: feat(otel): use BatchSpanProcessor + flush-on-exit for trace export (#619 follow-up)
### What problem does this solve?

The OpenTelemetry tracer set up in `drt/observability/otel.py` (`_initialize_if_needed`) exports spans with a **synchronous** processor:

```python
trace_provider.add_span_processor(SimpleSpanProcessor(tracer_exporter))
```

`SimpleSpanProcessor` exports every span by making a **blocking gRPC call on `span.end()`**, on the calling thread.

This was harmless while OTel was just scaffolding (Phase 1 #527 / Phase 2 #531 emitted no spans). But Phase 3 (#619) now emits a span **per batch** (`drt.sync.load`) plus `drt.sync.extract` and `drt.sync.run` — so a sync with N batches blocks the sync thread on **N+2 synchronous exports on the critical path**. A slow or unreachable collector now directly slows down (or stalls) the actual data sync, which is the opposite of what observability should cost.

There's also an inconsistency: **metrics** in the same function already use the non-blocking batched reader (`PeriodicExportingMetricReader`), while **traces** use the synchronous `SimpleSpanProcessor`. The asymmetry suggests the simple processor wasn't a deliberate production choice.

Secondary gap: there is no `provider.shutdown()` / `force_flush()` anywhere. That's fine today (SimpleSpanProcessor flushes on `end()`), but it becomes a correctness requirement the moment export is batched — a CLI process that exits would drop buffered spans.

> Note: this predates Phase 3 — it's part of the Phase 2 exporter setup. `SimpleSpanProcessor` is a perfectly reasonable default for local debugging; the request is to make batched export the default for real (multi-batch, possibly `--threads`) runs.

### Proposed solution

1. Use `BatchSpanProcessor` instead of `SimpleSpanProcessor` in `_initialize_if_needed`, so spans export off the sync's critical path (batched + on a background thread), mirroring the existing `PeriodicExportingMetricReader` for metrics.

2. Add a flush/shutdown hook so buffered spans are not lost when the CLI exits. The graceful-shutdown path from #279 (SIGTERM/SIGINT) is the natural place to call `force_flush()` / `shutdown()`, alongside a normal end-of-run flush.

3. (Optional) Make the processor selectable via config for users who want synchronous export when debugging locally, defaulting to `batch`:
   `observability.otel.span_processor: batch | simple`.

**Acceptance criteria**
- Trace export no longer blocks the sync thread per span.
- Spans are flushed on normal completion and on graceful shutdown (#279).
- The `[otel]`-absent / no-endpoint path is unchanged (still a zero-cost no-op).
- Existing `tests/unit/test_otel.py` still pass; add a test asserting a `BatchSpanProcessor` is registered when an endpoint is configured.

### Category

Performance

### Alternatives considered

- **Keep `SimpleSpanProcessor`** — simplest, no flush-on-exit needed, fine for low-volume/debug runs, but blocking per-span export doesn't scale with per-batch spans and couples sync latency to collector health.
- **Cap span volume** (e.g. only emit `run`/`extract`, drop per-batch `load` spans) — loses the per-batch visibility that #619 deliberately added.
- **Async/OTLP-HTTP exporter tuning** — orthogonal; the processor, not the exporter protocol, is what blocks the calling thread here.

### Example YAML (if applicable)

```yaml

```

Muawiya-contact:
Switches OTel trace export off the sync's critical path.

Before, spans were sent one at a time with a blocking network call right when each span ended — on the same thread doing the actual sync. Once Phase 3 (#619) started emitting a span per batch, a big sync would stop and wait on hundreds of those sends. If the collector was slow or down, your real sync slowed down or stalled. (Metrics already used the non-blocking batched sender — only traces were stuck on the blocking one.)

**What changed:**
- Traces now use `BatchSpanProcessor` — spans are queued and shipped in the background, so the sync thread never waits on export.
- Added `shutdown_telemetry()`, registered via `atexit`, so buffered spans/metrics get flushed before a short-lived `drt run` exits. This covers normal completion and the #279 graceful-shutdown path (both exit cleanly). The only case it skips is the 30s watchdog hard-exit, where dropping a few spans is acceptable.
- New optional `observability.otel.span_processor: batch | simple` (default `batch`) — `simple` restores synchronous export for local debugging.
- The no-OTel / no-endpoint path is unchanged and still a zero-cost no-op.

**Design note:** flush-on-exit is via `atexit` rather than explicit calls in `run.py`, so it's handled in one place and can't be forgotten — and it still covers the #279 graceful path, which exits normally.

**Tests:** existing `test_otel.py` tests updated for the new default; added tests for Batch-by-default, the `simple` override, and the flush / idempotency / no-op behavior. All green; mypy + ruff clean.

Closes #658


Muawiya-contact:
@masukai, Now it's your turn to review that.

Muawiya-contact:
Heads-up @masukai — CI is red, but the 4 failures are all in `test_credentials_permissions.py` (the `0o600` / `0o700` mode checks from the perms work), failing with `assert 420 == 384` (i.e. the runner's tmpfs isn't honoring the chmod). This PR only adds a `span_processor` field to `OtelConfig` and swaps the span processor — it doesn't touch `credentials.py`'s write path or permissions, and all the OTel tests pass. Looks like a runner/filesystem issue rather than anything in this change. Same failures would hit any PR right now. Lmk if you want me to do anything.

masukai:
Thanks for digging in @Muawiya-contact — but it's not a runner/tmpfs flake, it's real. `assert 420 == 384` is `0o644 != 0o600`: the file is being written world-readable instead of owner-only.

The cause is that **this branch is missing #655's permission hardening entirely** — on this PR's head, `drt/config/credentials.py` has no `_ensure_private_dir` and none of the `0o600` / `os.open(..., O_CREAT, 0o600)` write sites, while `main` has the function plus its two call sites. Quick check:

```bash
git show origin/main:drt/config/credentials.py | grep -c "_ensure_private_dir\|0o600"   # → several
git show HEAD:drt/config/credentials.py        | grep -c "_ensure_private_dir\|0o600"   # → 0
```

So the perms test is doing exactly its job — it's catching that the `0o600` logic is gone. What most likely happened: this branch was cut before #655 landed (2026-06-21), and pulling `main` in resolved the `credentials.py` conflict in favor of this branch's side — keeping your `OtelConfig.span_processor` addition but wiping #655's changes.

The OTel change itself is good and I want it merged — it just needs a clean rebase that **keeps #655's code**:

```bash
git fetch origin && git rebase origin/main
# in the credentials.py conflict, keep BOTH:
#   - #655's _ensure_private_dir + the 0o600 / O_CREAT write sites
#   - your new OtelConfig.span_processor field
```

After that the perms tests pass and CI goes green. Ping me if the conflict's fiddly and I'll point at the exact hunks to keep 🙏

Muawiya-contact:
You're right @masukai — my mistake, not a flake. The branch was built in a way that dropped #655's permission hardening from credentials.py while keeping my OtelConfig.span_processor change. Rebased onto origin/main keeping both #655's _ensure_private_dir + the 0o600 write sites and my span_processor field. Perms + OTel tests green locally, force-pushed. Thanks for catching it and pointing at the exact cause 🙏

Muawiya-contact:
Fixed @masukai — restored #655's _ensure_private_dir + 0o600 write sites in credentials.py alongside my span_processor field, as you said. Perms + OTel tests pass locally, force-pushed. Thanks for the precise diagnosis 🙏

masukai (approved):
Verified and approving 🎉 — `credentials.py` now matches `main` exactly (`_ensure_private_dir` + all four `0o600` / `O_CREAT` write sites restored) with just your `OtelConfig.span_processor` field added on top; the true diff is OTel-only and CI is green.

The export change reads well: `BatchSpanProcessor` by default so the per-batch `#619` spans export off the run thread, `simple` kept as the local-debug escape hatch, and `shutdown_telemetry()` registered via `atexit` to flush buffered spans before a short-lived `drt run` exits — with the idempotent clear-then-shutdown and the note that the 30s watchdog `os._exit` intentionally skips it. Nicely thought through.

Merging now. Thanks for the quick turnaround on the rebase 🙏

Issue #660, opened 2026-06-22
feat: PII masking — add truncate strategy (follow-up to #427)


masukai:
## Summary

Follow-up to #427 (PII field masking — `hash` / `redact`). v1 ships with a **flat** config form only:

```yaml
sync:
  mask:
    email: hash      # SHA-256 hex
    phone: redact    # -> "[REDACTED]"
```

This issue tracks adding a `truncate` strategy, which needs a parameter (`length`), so it introduces an **object form** for masks that require options — added on top of v1 as a strictly backwards-compatible change.

## Proposed config

The flat form keeps working; an object form is accepted only when a strategy needs a param:

```yaml
sync:
  mask:
    email: hash                              # flat form (unchanged)
    name: {strategy: truncate, length: 2}    # object form for param-bearing strategies
```

## Scope

- Add `truncate` strategy: keep the first `length` chars, drop the rest (e.g. `length: 2` → initials).
- Introduce the object form `{strategy: ..., length: ...}` alongside the existing flat `field: strategy` form (discriminated on whether a value is a string vs. mapping).
- Same seam as v1: pure transform alongside `field_mappings` (outside the `engine/sync.py` SyncObserver boundary), applied to the post-`field_mappings` field name.
- Null passes through untouched; non-strings stringified first (consistent with v1).
- Unit tests for `truncate` + the flat/object parsing.

## Out of scope

- Encryption / FPE (#303)
- Per-destination policy

## Notes

- Must remain backwards compatible: any existing v1 `mask:` block (flat form) keeps working unchanged.
- Decide default behavior when `length` >= the string length (return as-is) and for `length: 0`.

cc @Pawansingh3889 — splitting this out of #427 per our discussion so v1 stays focused on `hash | redact`.

PR #668, opened 2026-06-26
feat(engine): add sync.mask for PII masking before load

### LINKED ISSUE #427: feat: PII field masking — hash or redact fields before load
## Motivation

Sending PII (email, phone, name) to third-party destinations (Amplitude, HubSpot, Slack) creates compliance risk. Users need a way to hash or redact specific fields before they leave drt — without modifying the source SQL.

## Proposed config

```yaml
sync:
  pii_fields:
    email:
      strategy: hash       # SHA-256 hex digest
    phone:
      strategy: redact     # replace with "[REDACTED]"
    name:
      strategy: truncate
      length: 2            # keep first N chars (e.g. initials)
```

## Scope

- `pii_fields` optional key in `SyncConfig`
- Applied in engine after extraction, before destination.load() — same layer as `field_mappings`
- Strategies: `hash` (SHA-256), `redact` (replace with literal), `truncate`
- `drt validate` warns if a field listed in `pii_fields` does not appear in the query schema (best-effort)
- Works with all destinations

## Out of scope

- Encryption (→ v0.9 config encryption #303 is a separate concern)
- Format-preserving encryption
- Per-destination PII policy

Pawansingh3889:
Closes #427.

Adds `sync.mask` — hash or redact named fields after extraction and just before the destination, without changing the source SQL. Flat v1 shape per the discussion above (option A):

```yaml
sync:
  mask:
    email: hash      # SHA-256 hex
    phone: redact    # -> "[REDACTED]"
```

## What

- `mask: {field: "hash" | "redact"}` on `SyncOptions`. Pydantic validates the values, so an unknown strategy (or the object form) is rejected at config load.
- New `drt/engine/masking.py` with a pure `apply_mask(records, mask)` — same shape as `field_mappings.py`, so it stays outside the `engine/sync.py` SyncObserver boundary and is unit-testable on its own.
- Applied right after `apply_field_mappings`, so `mask` keys reference the destination-facing (post-rename) field name. Works across every destination.
- Null passes through untouched; non-strings are stringified before hashing; a configured field that's absent from a row is skipped; the input batch is never mutated.

## Out of scope (v1)

Param-bearing strategies — `truncate` etc. — are a backwards-compatible object-form follow-up (#660), so they can land later without breaking existing `mask:` blocks.

## Tests

`tests/unit/test_masking.py` — 15 tests: config validation (rejects unknown strategy + object form), hash is a deterministic SHA-256, redact, null passthrough, non-string stringify, absent-field skip, unmasked fields untouched, no input mutation, multi-field/multi-row. ruff + mypy clean; the `test_engine_observer.py` boundary checks still pass.


masukai (approved):
Verified @Pawansingh3889 🙌 — textbook pure transform: `engine/masking.py` mirrors `apply_field_mappings` and slots in at the same pipeline seam, with **no engine-purity violation** (the `test_engine_observer` boundary stays green) and **no PII leak path** — the masked batch is what reaches `destination.load()`, the diff/dry-run accumulator, and the DLQ / `row_errors`. Cursor tracking runs *before* the mask, so watermarks use the true source value — nicely ordered. Closes #427.

One non-blocking follow-up: worth a one-line doc note that `hash` is unsalted SHA-256 = pseudonymization (not anonymization); the param-bearing strategies are tracked in #660. Merging — thanks!

PR #659, opened 2026-06-21
feat: VS Code extension for drt YAML (validation + autocomplete) (#293)

Muawiya-contact:
Adds a small VS Code extension that gives you live validation, autocomplete, and hover for drt sync files — right as you type, before you ever run `drt validate`.

It's deliberately thin: no custom language server, no runtime code. It just hands drt's own JSON Schema (generated from `drt/config/schema.py`, the same source as `drt validate --emit-schema`) to the `redhat.vscode-yaml` extension, which does the real work. So nothing is hand-written and there's very little to maintain.

Lives at `integrations/vscode-drt/`, next to `dagster-drt/` — monorepo, as we agreed in #293.

Quick demo — invalid value → instant red line, and `Ctrl+Space` → full connector list: https://streamable.com/z4ttpb

A few follow-ups, not in this PR:
- Marketplace publish, once a `drt-hub` publisher is set up (happy to maintain it).
- A small PR adding field descriptions to the config models, so hover docs get richer.
- A CI step to regenerate the schema on each drt release.

Closes #293

PR #669, opened 2026-06-26
ci: guard bundled VS Code extension schemas against drt-core drift (#662)

### LINKED ISSUE #662: CI: guard the bundled VS Code extension schemas against drt-core drift
## Problem

The VS Code extension (`integrations/vscode-drt/`, added in #659) bundles two JSON Schemas:

- `integrations/vscode-drt/schemas/drt_project.schema.json`
- `integrations/vscode-drt/schemas/sync.schema.json`

These are **generated** from drt-core via `scripts/regenerate-schemas.sh` → `drt.config.schema.write_schemas`, and committed because the extension ships standalone (no Python at extension runtime). They match the current `drt-core` byte-for-byte today (verified at merge time of #659).

The risk: as drt-core's config models evolve, the bundled copies can **silently drift** out of sync. The extension would then validate against a stale schema and there's no signal until a user notices.

## Proposal

Add a CI job that regenerates the schemas and fails on any diff:

```bash
cd integrations/vscode-drt
./scripts/regenerate-schemas.sh
git diff --exit-code -- schemas/
```

So a drt-core change that alters the schema fails the build until the bundled schemas are regenerated (and the extension version + CHANGELOG bumped, as the regen script already reminds).

## Notes

- Cheap to run (pure Python, no extra deps beyond `drt-core`).
- Pairs with the existing drift-audit philosophy (`scripts/check_drift.sh`).
- Follow-up from #659 / #293.

Muawiya-contact:
## What does this PR do?
Adds a CI workflow (`vscode-schema-drift`) that regenerates the bundled VS Code extension schemas from drt-core and fails the build on any diff — so the committed `integrations/vscode-drt/schemas/*.json` can't silently drift from the config models.

Runs the recipe from the issue:

```bash
cd integrations/vscode-drt

./scripts/regenerate-schemas.sh

git diff --exit-code -- schemas/
```
A `paths:` filter scopes it to `drt/config/**`, the bundled schemas, and the generator script, so it stays cheap and skips unrelated PRs. Unlike `drift-check.yml` (maintainer-facing, post-merge, files a tracking issue), this is a hard PR gate by design — stale bundled schemas should block merge.

## Related Issue
Closes #662

## Checklist
- [x] Tests pass — N/A, CI-only change (no `drt/` code touched)
- [x] Linter passes — N/A, CI-only change
- [ ] CHANGELOG — N/A `[skip changelog]` (CI/tooling only, not user-facing)


Muawiya-contact:
@masukai The guard caught real drift on its first run — the bundled `sync.schema.json` was missing the **airtable** and **klaviyo** destinations (and `introspect_schema` on the SQL destinations) added since #659. Regenerated the schemas to match drt-core, so this PR both adds the drift guard and fixes the existing drift it found. CI is green.

masukai (approved):
Verified end-to-end @Muawiya-contact 👏 — the job installs drt-core, regenerates via `write_schemas` (reusing `regenerate-schemas.sh`, no reimplementation), and fails on `git diff` of `schemas/`. I confirmed it both **catches real drift** (a drt-core model change → exit 1) and **passes on a synced tree**, on the right `pull_request` + `push` paths. Closes #662 exactly.

The 194-line `sync.schema.json` regen in this PR is itself proof the guard was overdue 👍. (Tiny, non-blocking: the workflow YAML has CRLF line endings + no final newline — harmless at runtime, normalize to LF when convenient.) Merging.

Issue #673, opened 2026-06-28
DWH smoke validation — BigQuery (real account)


masukai:
Part of #654.

Verify on a live BigQuery account:
- streaming insert path
- temp-table MERGE + cleanup (#645)

Runs through the harness (#674). Sign-off by @masukai.

masukai:
**Verified ✓ on a real BigQuery project.**

Provisioned via `tests/integration/dwh/provisioning/bigquery.sh` (org-owned billing-enabled project `drt-smoke`, throwaway `smoke` dataset, least-privilege SA — `bigquery.jobUser` at project + dataset-scoped WRITER — plus a monthly budget alert), registered the `SMOKE_BIGQUERY_*` secrets, and ran `dwh-smoke`. All four legs green on the live project ([run 29063707019](https://github.com/drt-hub/drt/actions/runs/29063707019)):

```
test_bigquery_insert_roundtrip       PASSED   # streaming insert
test_bigquery_merge_roundtrip        PASSED   # temp-table MERGE + cleanup (#645)
test_bigquery_connection             PASSED
test_bigquery_complex_type_roundtrip PASSED
4 passed in 23.32s
```

One real-world gotcha for the runbook: a project created *after* billing signup isn't auto-linked to the billing account — the first run failed with `billingNotEnabled` / "Streaming insert is not allowed in the free tier" until `gcloud billing projects link` was run. Worth a line in the provisioning docs.

Smoke-side cost is inside the free tier (~$0). Nightly covers all three warehouses from tonight. Closing.

Issue #672, opened 2026-06-28
DWH smoke validation — Databricks (real account)


masukai:
Part of #654.

Verify on a live Databricks account:
- `INSERT OVERWRITE` atomicity for `replace_strategy: swap` (#644)
- complex-type / VARIANT-equivalent serialization

Runs through the harness (#674). Sign-off by @masukai.

masukai:
**Verified ✓ on a real Databricks workspace.**

Provisioned a **Databricks Free Edition** workspace (org email, no card — the new permanent free tier supports exactly what the smoke needs: a Serverless Starter Warehouse, PATs, and Unity Catalog's default `workspace` catalog + a `drt_smoke` schema), added the `SMOKE_DATABRICKS_*` secrets, and ran `dwh-smoke`. All four legs green on the live workspace:

```
test_databricks_insert_roundtrip           PASSED
test_databricks_replace_swap_roundtrip     PASSED   # INSERT OVERWRITE swap
test_databricks_complex_type_serialization PASSED   # from_json / parse_json (#317 / #680)
test_databricks_connection                 PASSED
4 passed in 51.86s
```

This validates the #706 inline-params fix **and** the #317 Databricks leg (from_json DDL reconstruction) on a real workspace — closing the loop @yodakanohoshi opened when his real-machine run caught the connector-3.0 paramstyle bug. Runs nightly from here; Free Edition means this leg is **permanently $0**. Next: running the same smoke against the #732 branch to live-validate the native-`?` migration before it merges. Closing.

Issue #671, opened 2026-06-28
DWH smoke validation — Snowflake (real account)


masukai:
Part of #654. **Priority 1.**

Verify on a live Snowflake account (mock CI can't exercise these):
- VARIANT/OBJECT/ARRAY load via `PARSE_JSON` (#317 Layer 3 / #653)
- `replace_strategy: swap` (`ALTER TABLE … SWAP WITH`) + orphan cleanup (#434)

Runs through the harness (#674). Sign-off by @masukai.

masukai:
**Verified ✓ on a real Snowflake account.**

Provisioned a throwaway trial account (org email, XS warehouse w/ auto-suspend + a resource-monitor credit cap, least-privilege role/user — reproducible via `tests/integration/dwh/provisioning/snowflake.sql`), added the `SMOKE_SNOWFLAKE_*` repo secrets, and ran `dwh-smoke` against it. All four legs green on the live warehouse:

```
test_snowflake_insert_roundtrip           PASSED
test_snowflake_replace_swap_roundtrip     PASSED   # ALTER TABLE ... SWAP WITH (#434)
test_snowflake_complex_type_serialization PASSED   # VARIANT/OBJECT/ARRAY via PARSE_JSON (#317 / #653)
test_snowflake_connection                 PASSED
4 passed in 37.74s
```

This is drt's first real-cloud validation — the #317 schema-aware serialization and #434 swap replace are now confirmed on live Snowflake, not just mocks. It'll also run nightly from now on. Cost: a few seconds of XS-warehouse runtime = a tiny fraction of the trial credits (~$0). Closing the Snowflake leg of #654. 🎯

PR #676, opened 2026-06-28
ci(test): DWH smoke harness — gated/nightly, skip when no secrets (#674)

### LINKED ISSUE #674: DWH smoke validation — CI harness (gated/nightly, skip when no secrets)
Part of #654.

Build the credential-agnostic harness so the per-warehouse smoke checks have somewhere to run:
- Manually-dispatched + nightly workflow running a **minimal end-to-end sync per DWH destination**, mirroring the v0.7.5 DuckDB E2E harness.
- **Skips gracefully when the cloud secrets aren't present** (no-op for forks / contributors; runs for real only when creds exist).
- Start with the **Snowflake leg as the reference shape**, then replicate for Databricks + BigQuery.

Cloud secrets + the actual "verified on a real service ✓" sign-off are owned by @masukai (his accounts / repo secrets).

Muawiya-contact:
## What does this PR do?
Adds the credential-agnostic DWH smoke harness from #674. A manually-dispatched + nightly workflow runs a minimal seeded-DuckDB → engine → real-warehouse sync per cloud destination and reads the rows back, mirroring the v0.7.5 DuckDB E2E harness.

Snowflake is the reference leg (`tests/integration/dwh/test_snowflake_smoke.py`); Databricks and BigQuery mirror it — only the destination config + read-back/cleanup driver differ.

**Safe no-op until secrets exist** — gated three ways:
- `pytest.importorskip(...)` — skips unless the warehouse extra is installed
- `require_env(...)` — skips unless the `DRT_SMOKE_*` secrets are set
- workflow jobs are `if: github.repository == 'drt-hub/drt'` + a per-job credential check that no-ops green when the secret is empty

So forks, local runs, and the upstream repo *before secrets are added* are all green without touching a warehouse. Required secret names are documented in `tests/integration/dwh/README.md`.

Cloud secrets + the "verified on a real service ✓" sign-off are yours (#671 / #672 / #673). I can't run these against live warehouses, so the read-back SQL / connector calls follow the documented APIs but haven't hit a real service yet — the first nightly run is the real check.

Verified locally: contributor case → 3 skipped cleanly; `ruff` clean; `mypy drt` clean; full suite collects and existing integration tests pass (24 passed, 3 skipped).

## Related Issue
Closes #674

## Checklist
- [x] Tests pass (`make test`) — added; skip safely without secrets
- [x] Linter passes (`make lint`) — ruff + mypy clean locally
- [ ] Updated `CHANGELOG.md` — N/A `[skip changelog]` (CI/test harness, not user-facing)

### DIFF
--- a/.github/workflows/dwh-smoke.yml
+++ b/.github/workflows/dwh-smoke.yml
@@ -1,0 +1,143 @@
+# DWH smoke validation — real-warehouse end-to-end checks (#674, part of #654).
+#
+# The cloud-warehouse destinations (Snowflake / Databricks / BigQuery) are
+# CI-tested only with sys.modules-injected mocks — that covers drt's own logic
+# but can't catch dialect-specific behaviour only a live service exercises.
+# This workflow runs a minimal seeded-DuckDB -> engine -> real-warehouse sync
+# and reads the rows back, mirroring the v0.7.5 DuckDB E2E harness.
+#
+# Credential-agnostic by design:
+#   * Only runs on the upstream repo (forks have no secrets).
+#   * Each job checks whether its DRT_SMOKE_* secrets are present and, if not,
+#     no-ops green — so this is safe to merge before any secrets exist and
+#     "activates" per-warehouse as masukai adds them.
+# Cloud secrets + the "verified on a real service" sign-off are owned by the
+# maintainer (his accounts / repo secrets); see #671 / #672 / #673.
+#
+# Required repo secrets (per warehouse) are listed in
+# tests/integration/dwh/README.md.
+
+name: dwh-smoke
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Nightly 03:00 UTC — offset from CI (00:00) and the weekly audits.
+    - cron: '0 3 * * *'
+
+permissions:
+  contents: read
+
+jobs:
+  snowflake:
+    name: Snowflake smoke
+    runs-on: ubuntu-latest
+    if: github.repository == 'drt-hub/drt'
+    env:
+      DRT_SMOKE_SNOWFLAKE_ACCOUNT: ${{ secrets.SMOKE_SNOWFLAKE_ACCOUNT }}
+      DRT_SMOKE_SNOWFLAKE_USER: ${{ secrets.SMOKE_SNOWFLAKE_USER }}
+      DRT_SMOKE_SNOWFLAKE_PASSWORD: ${{ secrets.SMOKE_SNOWFLAKE_PASSWORD }}
+      DRT_SMOKE_SNOWFLAKE_DATABASE: ${{ secrets.SMOKE_SNOWFLAKE_DATABASE }}
+      DRT_SMOKE_SNOWFLAKE_SCHEMA: ${{ secrets.SMOKE_SNOWFLAKE_SCHEMA }}
+      DRT_SMOKE_SNOWFLAKE_WAREHOUSE: ${{ secrets.SMOKE_SNOWFLAKE_WAREHOUSE }}
+    steps:
+      - uses: actions/checkout@v7
+      - name: Check credentials
+        id: creds
+        run: |
+          if [ -z "$DRT_SMOKE_SNOWFLAKE_ACCOUNT" ]; then
+            echo "No Snowflake smoke secrets configured — skipping (no-op)."
+            echo "has_creds=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_creds=true" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Set up Python
+        if: steps.creds.outputs.has_creds == 'true'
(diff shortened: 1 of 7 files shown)


Muawiya-contact:
Hi @masukai, what's up!
I want to work with you on a specific task to get a co-author award with you on DrT. So, can you give me a draft PR with some commits from you? Then I will pick that one and finish them, please.
Thanks 😅.

Pawansingh3889 (commented):
The gating reads as genuinely safe to merge before secrets exist: repo check plus per-job creds plus `require_env`/`importorskip` is a clean triple no-op.

One thing on the read-back, since you flagged the warehouse legs as unrun. The Snowflake destination writes with unquoted identifiers (`snowflake.py` notes "UPPERCASE for unquoted DDL"), so Snowflake folds the table and column names to uppercase. The read-back quotes them, `SELECT "name" FROM "{table}"`, which is case-sensitive, so it may not match the `NAME` and uppercased table the sync actually created. Would an unquoted read-back (`SELECT name FROM {table}`) match the write path better?

Same pattern in the Databricks and BigQuery legs.

masukai:
The secret gating here is genuinely solid — `workflow_dispatch` + `schedule` only (no `pull_request` / `pull_request_target`), the `if: github.repository == 'drt-hub/drt'` guard, secrets referenced only as env keys and `-z`-checked (never echoed), the BigQuery SA written to `$RUNNER_TEMP`. The skip path is right too, so it's a clean no-op for forks.

The thing to fix before it can pass a real run: the smoke tests sync into a freshly-generated unique table that's never created, but drt's `insert` mode only `INSERT INTO`s an existing table — it doesn't create it (`docs/connectors/bigquery.md` notes this). So every leg would fail at `run_sync` on the first nightly. Each test needs to pre-create the target table with the seed schema before `run_sync`, and drop it in the existing `finally`.

One more, on the Snowflake leg: the INSERT column list is unquoted (`id, name, email`), so Snowflake stores `NAME`, but the read-back queries `SELECT "name"` (quoted lowercase) — that won't match. Reading back unquoted (`SELECT name`) matches how the write path stored it.

These are basically what the first real nightly would have surfaced — once the table lifecycle is in, a single manual `workflow_dispatch` on one leg confirms the round-trip.

masukai:
@Pawansingh3889 we posted at the same minute — the Snowflake casing is exactly right, and that's the same fix I landed on (unquoted read-back to match the unquoted INSERT). Thanks for reviewing this one. I also flagged a second blocker above: the target table never gets created, so the `insert` legs need a pre-create (with the seed schema) before `run_sync`.

Pawansingh3889:
Glad it matched. And agreed, the missing table-create is the real blocker; the casing would only have surfaced after that. Pre-creating with the seed schema before run_sync, dropped in the existing finally, makes sense.

masukai:
Happy to, @Muawiya-contact — opened #680 as a draft. It's the introspection foundation for the Databricks schema-aware serialization (#317): `describe_columns()` now reads Databricks' `information_schema` and categorises the complex types, with tests. The write-path wiring + tests + docs are left to finish, and there's a checklist in the PR body.

Build on top of `feat/schema-aware-databricks` and your commits sit alongside mine, so the squash-merge carries us both as `Co-authored-by`. Ping me if any of the completion steps need clarifying.

Muawiya-contact:
Thanks both — really useful, these are exactly the things the first nightly would have surfaced.

- **Table lifecycle (@masukai)** — fixed. All three legs now pre-create the target table with the seed schema before `run_sync` (Snowflake `CREATE TABLE`, Databricks `... USING DELTA`, BigQuery `CREATE TABLE`), dropped in the existing `finally`.
- **Snowflake casing (@masukai, @Pawansingh3889)** — switched the read-back and DROP to unquoted (`SELECT name FROM {table}`) so they match the destination's unquoted INSERT that Snowflake folds to UPPERCASE. Pre-create is unquoted too, so the whole leg is consistent.
- **One more I folded in** — BigQuery's insert path is `insert_rows_json` (streaming buffer), so rows aren't immediately visible to SQL `SELECT`. Added a short poll on the read-back so it isn't flaky once the table exists.

No-creds path still skips clean (3 skipped), ruff + mypy green. Still haven't run the live legs from my side — happy for a `workflow_dispatch` on the Snowflake leg whenever to confirm the round-trip.

masukai (approved):
Verified — re-checked all three legs: each now pre-creates the target table with the seed schema before `run_sync` and drops it in `finally`, and the Snowflake read-back is unquoted (`SELECT name`) to match how the unquoted INSERT stores the column. Gating is intact too — still skips cleanly without creds (`require_env` runs before any connect/create), and the triggers stay `workflow_dispatch` + `schedule` only.

Good to merge. Since it adds a workflow file it'll go in via the web UI on our side. Thanks for the quick turnaround, @Muawiya-contact.

Issue #500, opened 2026-05-13
ADR(docs): manifest schema v1 + Jinja template structure + theme tokens


masukai:
## Context

Sub-issue of #499. Settles the architectural decisions that downstream phases (P1 Mermaid, P2 manifest.json, P3 static HTML, P4 serve) all depend on.

This issue produces three deliverables in tree:

1. `docs/adr/0001-drt-docs-manifest.md` — manifest schema specification
2. `docs/adr/0002-drt-docs-theme.md` — visual design tokens (palette, spacing, type scale)
3. `drt/docs/manifest.py` — dataclasses matching the schema (no rendering logic, used by P1–P4)

## Manifest schema v1 — proposed fields

```json
{
  "schema_version": 1,
  "drt_version": "0.8.0",
  "generated_at": "2026-05-13T01:00:00Z",
  "project": {
    "name": "my-drt-project",
    "root": "/abs/path/to/project"
  },
  "syncs": [
    {
      "name": "users_to_hubspot",
      "source": "bigquery_prod",
      "destination": "hubspot_main",
      "mode": "upsert",
      "upsert_key": ["email"],
      "schedule": "0 */4 * * *",
      "lookups": [
        { "ref_sync": "accounts_to_hubspot", "join_on": "account_id" }
      ],
      "query_template_path": "queries/users.sql",
      "tags": ["crm", "production"],
      "state": {
        "last_sync_at": "2026-05-13T00:00:00Z",
        "last_cursor_value": "2026-05-12T23:55:00Z",
        "rows_synced": 1248,
        "last_status": "success",
        "last_error": null
      }
    }
  ],
  "sources": [
    {
      "name": "bigquery_prod",
      "type": "bigquery",
      "consumed_by": ["users_to_hubspot", "accounts_to_hubspot"]
    }
  ],
  "destinations": [
    {
      "name": "hubspot_main",
      "type": "hubspot",
      "produced_by": ["users_to_hubspot"]
    }
  ],
  "edges": [
    { "kind": "source_to_sync", "from": "bigquery_prod", "to": "users_to_hubspot" },
    { "kind": "sync_to_destination", "from": "users_to_hubspot", "to": "hubspot_main" },
    { "kind": "lookup", "from": "accounts_to_hubspot", "to": "users_to_hubspot" }
  ]
}
```

### Schema decisions to confirm in ADR

- `state` field omitted entirely (not `null`-valued) when `--no-state` is passed
[...]

Issue #499, opened 2026-05-13
epic(docs): `drt docs` — sync catalog & lineage UI


masukai:
# epic(docs): `drt docs` — sync catalog & lineage UI

## Motivation

drt is a YAML-first reverse-ETL tool. As the number of sync definitions grows (most production deployments will have 10–50+), operators need a single surface to answer:

- **What syncs exist, and what do they actually do?** (source → destination, mode, schedule, lookups)
- **How do they relate to each other?** (cross-sync `lookups` dependencies from #354, shared sources, shared destinations)
- **What's the current state?** (last run status, last cursor value, last row counts) — without grep'ing through `.drt/state/`
- **What's the upstream / downstream blast radius?** (if I change `dim_users`, which syncs care?)

dbt solved the same UX problem for transforms via `dbt docs generate` / `dbt docs serve`. drt should ship the same primitive for reverse ETL.

## Confirmed decisions

The four axes that were settled before opening this epic:

| Decision | Value | Rationale |
|---|---|---|
| **CLI naming** | `drt docs generate` / `drt docs serve` | Mirrors dbt — zero-surprise for the user population most likely to adopt. `drt docs` (no subcommand) prints help. |
| **Output dir** | `target/docs/` | dbt convention, treated as build artifact (gitignored). `.drt/` stays reserved for state / secrets. |
| **Lookups in Mermaid (Phase 1)** | Included | Cross-sync `lookups` edges (#354) render in the DAG from day 1 — lineage value is highest, cost is one extra edge color. |
| **Manifest schema** | Public surface (semver-relevant) | `schema_version: 1`. Enables Backstage / OpenMetadata / Hightouch ingestion. Breaking changes require major bump and migration notes. |
| **State in docs** | Included by default, `--no-state` to opt out | `last_sync_time`, `last_cursor`, `rows_synced`, `last_status` rendered in sync detail view. Operators who commit docs to a public repo can strip with `--no-state`. |
| **Phase 4 frontend ceiling** | HTMX + Alpine + Tailwind via CDN | No bundler, no React/Vue. Install must stay trivial. |
| **Phase split** | P1–P3 (generate) → v0.8 · P4 (serve) → v0.8.x | Static-first ships value with minimal blast radius; live UI follows once IA is validated. |

### Functional

- `drt docs generate [--output target/docs] [--format html|mermaid|json] [--no-state]`
- Default `--format html`, default `--output target/docs`
- Output artifacts when `--format html`:
[...]

PR #677, opened 2026-06-28
feat(docs): drt docs generate --format html — multi-file static site (P3 of #499)

### LINKED ISSUE #510: feat(docs): `drt docs generate --format html` — multi-file static site (P3 of #499)
## Context

Sub-issue of #499. Third implementation slice: render the manifest into a navigable HTML catalog under `target/docs/`.

## Depends on

- P1 #501 / #505 (foundation: `Manifest` dataclasses + `build_manifest`)
- P2 #507 / #508 (manifest schema v1 + state injection, `--format json` already wired)
- ADR #500 (visual mockup at [`docs/design/drt-docs-prototype.html`](https://github.com/drt-hub/drt/blob/main/docs/design/drt-docs-prototype.html))

## Design decisions (ADR follow-ups settled before this issue)

Studied dbt docs / elementary `edr report` / Great Expectations data docs and converged on the **Great Expectations + elementary hybrid**: multi-file static site, **no runtime fetch**, data inlined per-page.

| Axis | Decision | Rationale |
|---|---|---|
| **Output structure** | multi-file static site (one HTML per view) | Splits load, deep-linkable, but no SPA hydration cost |
| **Data delivery** | per-page `<script type="application/json" id="drt-data">` block inlines the relevant subset of the manifest | No `fetch()`, no CORS issues on `file://`, works in every browser |
| **`manifest.json`** | still emitted (P2), but as a public artifact for external tools (Backstage / OpenMetadata) — HTML render does not consume it at runtime | Decouples downstream consumers from our internal HTML pipeline |
| **CSS** | hand-written `style.css` (~500 lines, no Tailwind CDN at runtime) vendored to `assets/` | Portable, ~10KB gzipped, no runtime JS to enforce Tailwind classes |
| **JS** | small `app.js` (~3–5KB) for: sidebar `<details>` localStorage persistence, tab toggle, search filter against inline JSON | Light enough that runtime cost stays trivial |
| **YAML highlighting** | `pygments` at **build time**, output embedded as styled HTML directly into the sync detail page; `pygments-default.css` vendored to `assets/` | Zero runtime dependency; build-time `pygments` becomes a `[dev]` extra |
| **Mermaid** | CDN `mermaid.min.js` for now (P3) — keeps file size down; vendoring is a P3 follow-up if offline use becomes blocking | Pragmatic compromise; vendored mermaid is ~1MB |
| **Theme** | light + dark via `prefers-color-scheme`; theme tokens in `:root` / `@media` blocks | Mockup-consistent, no JS required for theme |
| **Self-hosting** | `target/docs/` copies cleanly to nginx / GitHub Pages / Netlify / S3; `file://` works for local preview | The whole point of static-first |

### Functional

`drt docs generate --format html [--output target/docs] [--no-state]`

Produces:

```
target/docs/
├── index.html                      # Overview
├── dag.html                        # DAG (Mermaid)
├── sync/<name>.html                # per sync
├── source/<name>.html              # per source
├── destination/<name>.html         # per destination
├── manifest.json                   # P2 artifact (still emitted)
└── assets/
    ├── style.css
    ├── app.js
    └── pygments-default.css
```

### Acceptance

- `drt docs generate --format html` produces a `target/docs/` tree matching the layout above.
- Opening `target/docs/index.html` directly via `file://` works fully — no CORS errors, no missing assets, no JS errors in console.
- The five view types are navigable via the sidebar.
- Mermaid DAG renders on `dag.html` (light/dark theme aware).
- YAML on sync detail pages is syntax-highlighted via pygments.
- `assets/` is self-contained; output directory is portable to any static host.

### Out of scope

- `drt docs serve` (→ P4, v0.8.x)
- Vendoring Mermaid (P3 follow-up if needed)
- Run history timeline (`HistoryManager`) — current `state` block is "latest only"

Refs: #499, #500 (ADR + mockups), #501, #505, #507, #508

Muawiya-contact:
## What does this PR do?
Implements `drt docs generate --format html` — P3 of #499. Renders the project manifest into a navigable multi-file static site under `target/docs/`, following the ADR (#500) decisions.

```
target/docs/

├── index.html            # overview + counts + sync table

├── dag.html              # Mermaid DAG (light/dark aware)

├── sync/<name>.html      # per sync — Pygments-highlighted YAML + last-run state

├── source/<name>.html    # per source — "used by"

├── destination/<name>.html  # per destination — "fed by"

├── manifest.json         # still emitted (P2 artifact)

└── assets/{style.css, app.js, pygments-default.css}
```
**Per the ADR:**
- **Multi-file static site, no runtime fetch** — each page inlines its data subset in a `<script type="application/json" id="drt-data">` block, so opening `index.html` directly via `file://` works with no CORS errors.
- **Hand-written vendored CSS** (mockup's brand-violet tokens, light/dark via `prefers-color-scheme`) — no Tailwind runtime.
- **Mermaid** via CDN for the DAG; **Pygments** at build time for YAML highlighting (new `[docs]` extra).
- `--output` / `--no-state` honored; `--format mermaid` / `--format json` paths unchanged.

## Implementation
- New `drt/docs/html.py` (`render_html`) + `drt/docs/_html_assets.py` (vendored CSS/JS).
- Wired the `html` branch in `drt/cli/commands/docs.py`; exported `render_html` from `drt/docs/__init__.py`.
- `pyproject.toml`: `[docs]` extra (`pygments`), pygments added to `[dev]`, per-file `E501` ignore for the two template/asset modules.

**Note on the sync YAML:** the per-sync page renders the catalog entry from the **manifest** (name/source/destination/mode/tags/description). Manifest schema v1 intentionally doesn't carry the original `model` SQL, so that's not shown — consistent with the P2 schema. Happy to revisit if you'd prefer reading raw sync YAML off disk.

## Testing
- New `tests/unit/test_docs_html.py` (8): output-tree layout, no-`fetch()` file:// safety, per-page JSON parses, manifest.json emitted, highlighted YAML + state block, relative asset paths, empty-project render.
- Updated the two obsolete "html not implemented" placeholders in `test_docs_mermaid.py` / `test_docs_manifest_json.py` to assert the site now generates.
- Local: `ruff` clean, `mypy drt` clean (135 files), 31 docs tests pass. CLI verified end-to-end incl. `--no-state` and `--output`.

`drt docs serve` (P4), vendoring Mermaid, run-history timeline.

## Related Issue
Closes #510

## Checklist
- [x] Linter passes (`make lint`) — ruff + mypy clean
- [x] Updated `CHANGELOG.md`


Muawiya-contact:
@masukai Quick  notes:

- **ADR fidelity** — followed #500 as written: multi-file static site, per-page inlined `<script id="drt-data">` (no `fetch()`, so `file://` works), vendored CSS, Mermaid via CDN, build-time Pygments. `manifest.json` is still emitted as the P2 artifact.
- **One decision worth your eye** — the per-sync page renders YAML from the **manifest** fields, not the original sync file. Schema v1 doesn't carry `model` SQL, so it isn't shown. That keeps the render manifest-only (consistent with P2), but if you'd rather the detail page show the *raw* sync YAML off disk, that's a small follow-up — happy to do it either way.
- **Obsolete tests** — the two `--format html → NotImplementedError` placeholders in `test_docs_mermaid.py` / `test_docs_manifest_json.py` are now updated to assert the site generates.
- **Scope** — stopped exactly at the issue boundary: no `drt docs serve` (P4), no Mermaid vendoring, no run-history timeline.

Tested the `file://` path manually too — opens clean, no console errors, light/dark both render.

masukai:
Nice work on the multi-file structure. Two XSS issues in the generated HTML I'd want to close before this goes in:

1. autoescape isn't actually engaging. `select_autoescape(["html"])` keys off the template filename suffix, but the `DictLoader` templates are named `"sync"`, `"index"`, etc. (no `.html`), so it never turns on — every manifest string (sync / source / destination names, descriptions, SQL, `last_error`) renders raw. A sync named `evil<script>alert(1)</script>` executes in the page. Switching the `Environment` to `autoescape=True` (the boolean form, since the templates have no extension) fixes it; keep the deliberately-raw bits (`yaml_html`, the mermaid source) on `|safe`.

2. The inlined `<script type="application/json" id="drt-data">` block can break out — `json.dumps` doesn't escape `</script>`, so a `</script>` in any embedded name escapes into live markup (autoescape doesn't cover this one). The usual fix is escaping `<` / `>` / `&` to their `\u00xx` forms after `json.dumps` — stays valid for `JSON.parse`.

One more worth doing: clear `target/docs/` before writing so a removed sync doesn't leave an orphan page.

The reason the tests didn't catch these is the fixtures are all alphanumeric — a case feeding `<script>` / `</script>` / `&` into a sync name + description + `last_error` would lock both down. The rest (structure, `file://` handling, `manifest.json`, the empty-project path) looked solid.

Muawiya-contact:
@masukai, Good catches — both fixed:

1. **autoescape** — you're right, `select_autoescape(["html"])` never engaged (no `.html` suffix on the DictLoader names). Switched to `autoescape=True`. `yaml_html` stays `|safe` (Pygments output is already escaped). I left the **mermaid** source *not* `|safe`: autoescaping it neutralizes injected markup, and Mermaid still gets the intended `<br/>`/`<i>` because the browser decodes entities in `textContent` — so it's safe and renders correctly. Shout if you'd rather I escape inside `render_mermaid` and `|safe` it instead.
2. **`</script>` breakout** — `dumps()` now escapes `<` `>` `&` → `\u003c` `\u003e` `\u0026` after `json.dumps` (valid for `JSON.parse`); block marked `|safe` so the quotes survive.
3. **orphan pages** — output dir is cleared before writing.

Added two tests that feed `</script><script>` + `&` into the sync name, description, tags, and `last_error` (asserting no raw script and no JSON-island breakout) and one for the orphan-removal case — that's the fixture gap you flagged. Resolved the CHANGELOG conflict with main too.


Fell free to ask if any change is needed.

masukai:
@Muawiya-contact #677 is really strong — the escaping, the multi-file structure, and the Pygments highlighting all hold up, and you nailed the ADR #500 visual language (the brand-violet tokens, light/dark via `prefers-color-scheme`).

First, the important bit: **please keep going on #677 and your other work — none of this is a blocker or a request to pause.** I just want to open up the docs UI's design direction in parallel, since it's a surface I care a lot about.

The visual target is the **ADR #500 prototype already in the repo** (`docs/design/drt-docs-prototype.html` — the mock you built #677 against). #677 is close; a few things from the prototype aren't in yet: the Overview **type bars** + **Recent Runs** panel, the sync-page **YAML / Lineage / State tabs**, and **search**. I'd like to bring those in — and to act as **design lead** for the docs UI (own the visual direction + design system and the final aesthetic call), since the UX is product-critical for me. Your design sense is clearly strong though, so I want your input throughout, not just on implementation — the thing I'd most love to **pair** on is the **Lineage view**: turning it into a real visual graph (dbt docs' lineage / dagster's asset graph, but static) is the unsolved problem.

North star for the look: **dbt docs, dagster's asset UI, elementary** — lightweight but rich, fully static.

I'm going to do a focused design pass on top of #677 (probably tomorrow) and share a spike here with concrete before/afters. Treat all of it as provisional — I expect to iterate hard on the actual look. Sound good? Anything you'd push back on or want to own?

Muawiya-contact:
@masukai 
Scoped the Lineage view so I can come into your design pass with something concrete.

The key thing about drt's graph specifically: it's not an arbitrary DAG, it's near-perfectly layered — source → sync → destination — so the expensive part of general graph layout (rank assignment) is free; the rank just *is* the node type. The only wrinkle is lookup edges, which run backward. Here's the shape:

```mermaid
graph LR
  subgraph s0["sources · rank 0"]
    duckdb[("duckdb")]
  end
  subgraph s1["syncs · rank 1"]
    customers(["customers"])
    users(["users"])
    events(["events"])
  end
  subgraph s2["destinations · rank 2"]
    discord["discord"]
    postgres["postgres"]
    s3["s3"]
  end
  duckdb --> customers & users & events
  customers --> discord
  users --> postgres
  events --> s3
  postgres -. lookup .-> customers
```

That layering turns "needs dagre/graphviz" into a few hundred lines of deterministic, dependency-free layout we fully control:
- ranks from node type; lookup edges treated as back-edges
- within-rank ordering via barycenter sweeps (fixed iterations → byte-identical output, clean git diffs)
- forward edges as béziers port-to-port; lookup edges dashed + bowed over the top (as above)
- emitted as inline themeable SVG (not ``) so it inherits the page CSS vars → light/dark for free, and every node wraps an `<a>` into its detail page — clickable lineage, fully static

Build-vs-buy: I'd lean hand-rolled (zero deps, total control, fits the static-first `[docs]` spirit) over `grandalf` (pure-Python Sugiyama — good fallback) or graphviz (best layouts, but a system binary that fails "lightweight/static"). Your call though.

The parts I think are the actual *unsolved* bit and would love to pair on: crossing-minimization on real fan-in/fan-out shapes, back-edge routing when several lookups hit one sync, and how far we push interactivity (hover-highlight / pan-zoom as a thin non-layout JS layer) vs. keeping it dead-static. Happy to bring a layout spike if useful — but it's your surface, so I'll follow your visual lead on the look.

(Diagram's just Mermaid for illustration — the actual feature replaces runtime Mermaid with the static pre-laid-out SVG above.)

masukai:
@Muawiya-contact This is really lovely thinking — and I think your framing is spot on: drt's graph isn't an arbitrary DAG, it's pre-ranked by node type, so the expensive part of layered layout (rank assignment) comes for free. That's actually the whole reason I'd rather not buy it.

**Build-vs-buy — I'm with you, let's go hand-rolled.** grandalf/graphviz earn their keep on exactly the rank-assignment + network-simplex machinery we don't need here — pulling them in would pay for layout intelligence we already get for free off the `source → sync → destination` structure, while saddling us with either a runtime dep (grandalf) or a system binary that breaks the "lightweight, `pip install drt[docs]`, works on `file://`" promise (graphviz). Inline themeable SVG with `<a>`-wrapped nodes feels like the right emission target. And honestly the static pre-laid angle is a *differentiator*, not a compromise — dbt docs and dagster both lay out at runtime in JS, so a diffable, zero-JS-for-layout SVG is something neither of them does.

**If you're up for building the layout spike, I'd genuinely love that** — it's exactly the kind of thing I'd want to react to. And no need to wait on me for it: the layout engine is design-independent (it just emits node positions + edge paths; the *look* sits on top), so it's a great thing to keep moving on whenever you have the energy.

A couple of honest notes on timing, so I'm not leaving you guessing:

- I'm a bit swamped right now, so realistically it'll be around **7/2 (JST)** before I can bring concrete visual direction + before/afters. If that ends up slipping, I'll give you a heads-up rather than leave you hanging. The aesthetic call is the part I'd like to own as design lead, so please treat *styling* as TBD — but none of that blocks the engine.
- So the split is roughly: **layout engine = green light, go whenever you like**, and **final look + how much interactivity = my pass around 7/2.**

On the three you flagged as unsolved — just my early lean, so we've got something to pair against (very much open to being talked out of any of it):

1. **Crossing-min** — at our scale (3 ranks, usually <20 nodes) this is nearly free; barycenter with fixed iterations + a stable name tiebreak should give byte-identical output and near-optimal crossings. Might be worth A/B-ing barycenter vs the **median heuristic** in the spike — median often edges out crossing count and is just as deterministic.
2. **Back-edges (several lookups → one sync)** — agreed this is the genuinely hard part, since it's where the "perfect layering" assumption leaks. My lean: lane-assignment above the graph (subway-style), each back-edge its own vertical lane + a distinct top-edge port on the target sync, so N lookups never stack into one curve. This is the bit I'd most love to pair on with you.
3. **Interactivity** — maybe architect it as a *separable* progressive-enhancement JS layer (hover-highlight upstream/downstream, optional pan-zoom) that never touches build-time layout and degrades cleanly to dead-static on `file://`? How far we push it is a nice design-pass conversation — keeping it separable means we don't have to decide anything now.

Thank you again for getting this far ahead of me — it makes the design pass so much easier to have your scoping to build on. 🚀

Muawiya-contact:
@masukai — happy to keep #677 moving whichever way suits your design pass. Two options as I see it:
- **merge now** as the clean foundation, and we iterate the Overview bars / Recent Runs / tabs / search / lineage on top in follow-up PRs, or
- **hold** it and you fold your 7/2 direction in first.

No preference on my end — just let me know which keeps your design pass simplest, and I'll line the follow-ups up accordingly.

masukai:
@Muawiya-contact quick scheduling note — the design direction I said I'd bring on 7/2 is going to be 7/3-ish instead. Today went to shipping **v0.7.10** (tracked/scoped mirror, #686/#687 — filed and released same-day 😅). Not answering the merge-now-vs-hold question yet; I'll come back with the design pass and we'll settle it then. Thanks for the patience 🙏

Muawiya-contact:
@masukai Take your time 😉.

masukai:
@Muawiya-contact Let's go with **merge now** — thanks for offering both paths. Quick reasoning:

- It's green across the board (CI / CodeQL / codecov), and the XSS hardening + tests made this a genuinely solid foundation — no reason to sit on it.
- The design work was always going to land as follow-up PRs anyway (tokens, sync-page tabs, lineage). Iterating on merged main is simpler than stacking more onto a live PR — and it unblocks anything else that wants to build on the docs engine.
- And you shouldn't have to babysit a mergeable PR while I finish the visual direction. 😄

One thing I'd like to do for the follow-ups: **I'll request your review on the design PRs as they come** — I want your eyes on them (especially anything lineage-shaped), not just your patience. The design direction itself is close and will arrive as the first of those follow-ups.

Thanks again for building such a clean base to iterate on. 🚀

masukai:
@Muawiya-contact Design pass, as promised (a day late — thanks for the patience 😄). Everything below lives on [`design/docs-spike`](https://github.com/drt-hub/drt/tree/design/docs-spike) — the mock itself is [`docs/design/drt-docs-lineage-mock.html`](https://github.com/drt-hub/drt/blob/design/docs-spike/docs/design/drt-docs-lineage-mock.html) (self-contained, open it raw in a browser; light/dark follow `prefers-color-scheme`). Treat all of it as a target to react to, not a spec set in stone.

## The one design idea everything hangs off: ownership at a glance

Looking at the Mermaid DAG, the thing a viewer can't answer is *"which part of this does drt actually manage?"* — every node looks the same. But drt's graph has three semantically different kinds of thing in it, and they happen to be exactly your three ranks:

| zone | what it is | who owns it |
|---|---|---|
| **sources** | external data assets | not drt's — it reads them |
| **syncs** | **version-controlled YAML — code** | **drt's domain** |
| **destinations** | external tables / channels / SaaS | not drt's — it writes them |

So the design leans into it: the middle rank gets a violet **"managed by drt"** band, the outer zones stay neutral with `external · read` / `external · write` labels. Node anatomy carries the same taxonomy — syncs are *code cards* (doc icon, mono name, mode incl. the new `mirror`, status dot + word), data stores get a cylinder, external apps a window glyph. Lookups are dashed violet, entering the consumer sync via **distinct top-edge ports** — your lane concept, drawn.

**Before (runtime Mermaid):**

before — mermaid DAG

**After (the visual target — static SVG, brand tokens, light & dark):**

after — lineage light

after — lineage dark

## Drill-down: three grains, all static

- **L0 — project DAG** (above): connection / sync grain. Width is fixed by the three ranks, so growth is vertical only — scroll holds up to ~30 syncs, and past that we switch to a compact density + group-collapse view rather than shrinking the graph.
- **L1 — ego-graph**: every node wraps an `<a>` (as agreed); clicking a sync lands on its detail page's Lineage tab, which shows just that sync's neighborhood, pre-rendered. Zero JS, works on `file://`.
- **L2 — column grain**: and here's the fun drt-specific bit — `field_mappings` and `mask` are *declared* in the sync YAML, so column-level lineage on the write side needs **no SQL parsing at all**. dbt has to parse SQL to get this; we read it off the manifest. (Source-side table/column extraction *does* need model-SQL parsing — flagged as a schema-v2 follow-up, drawn dashed in the mock.)

drill-down — column grain

Expand-in-place (dagster-style) stays exactly where we put it earlier: a separable progressive-enhancement layer on top, degrading to plain links. The static baseline never depends on it.

## What this means for your layout engine — almost nothing 🎉

The emission contract just gains: zone bands per rank (a `<rect>` + label each), fixed-size node boxes per kind, and named top-edge ports for back-edges (which your lane assignment already produces). Algorithm untouched — barycenter/median A/B and the back-edge lanes are still the interesting part, and still the part I'd love to pair on.

## Overview polish (already on the spike branch)

Type-distribution bars + Recent Runs, per the ADR mock:

overview before

overview after

## Coverage check + what's next

Everything drt does today maps onto this language: 5 modes (incl. `mirror`), run state, lookups, `mask` (the L2 badge), tags → chips/filter. Two deliberately not drawn yet: a **DLQ badge** on failed syncs (`n rows → DLQ`) and a **drift-audit badge** on destinations — both feel like they'll push the DAG from "catalog" toward "operational dashboard", coming in the next iteration.

Sequencing: once #677 is merged I'll line the design up as follow-up PRs (tokens file → sync-page tabs (YAML / Lineage / State) → this lineage look on your engine) — **and I'll request your review on each**. Tear into anything above; the back-edge port/lane geometry especially is as much yours as mine. 🚀

Muawiya-contact:
@masukai Good automated catches — triaged all three:

1. **`rmtree` safety (High)** — agree, this is the real one. `--output` can point anywhere, so an unconditional wipe is a footgun. Will add the guard: refuse unless the dir is empty or looks like a prior drt-docs build (`index.html` + `assets/`), and error if it's a file.
2. **ImportError hint (Medium)** — agree, should match the `drt mcp run` pattern and print `pip install drt-core[docs]` instead of a stack trace. Easy consistency fix.
3. **Slug collision (Medium)** — valid; punctuation-only-distinct names (`a_b` vs `a__b`) could collide. Will fail-fast with a clear error on duplicate slugs.

None of these block the current merge (the rmtree one only triggers on a mis-pointed `--output`), so I'd rather not churn the branch now that your design work's on it — I'll roll all three into a small hardening follow-up right after this lands, alongside the `from_json` DDL-quote fix from #680. @masukai good to merge from your side?

masukai:
Housekeeping before this merges: everything agreed in this thread that isn't on `main` yet is now tracked — **#702** (post-#677 design & lineage follow-up tracker) and **#701** (the static DAG layout engine — @Muawiya-contact, that's the natural home for your layout spike). Design artifacts are committed under `docs/design/` on `design/docs-spike`, so nothing here gets lost when this PR closes. 🚀

masukai (approved):
Yes — good to merge from my side 🚀 Reviewed the generator end to end:

- **Sound + safe rendering**: Jinja2 with `autoescape=select_autoescape(['html'])`, and the manifest-string / inline-JSON XSS hardening from `a3c463e` is in. Slugs computed once and reused for filenames + cross-links; `write()` mkdir-parents cleanly; `target/docs` default.
- **The Copilot trio is genuinely deferrable, and now tracked** in #703 so it doesn't live only in this thread: (1) the `rmtree` is **default-safe** — `--output` defaults to `target/docs` (a regenerable build dir), so the footgun only fires on an explicitly mis-pointed `--output`; (2) the `[docs]` ImportError hint is cosmetic; (3) slug-collision is a punctuation-only edge case. All three fine as your post-merge hardening PR.
- Tests (211 lines) + CI green; multi-file structure, cross-links, and the assets pipeline all exercised.

Merging (squash). I know the squash diverges `design/docs-spike` from main — all yours to rebase/reconstruct as you plan the follow-up PRs. Thanks for driving this one, and for the clean triage 🙏

masukai:
Merged 🎉 P3 of #499 is in — `drt docs generate --format html` ships a real static site. Genuinely nice work carrying this the whole way, @Muawiya-contact. Over to the design follow-ups (#701/#702) + the hardening pass (#703) whenever you're ready — ping me for review on each. 🙏

PR #675, opened 2026-06-28
feat(sources): add Delta Lake + Apache Iceberg lakehouse sources (#172, #173)

masukai:
Adds the two lakehouse **source** connectors, positioning drt for the open-table-format ecosystem.

## What
- **`type: deltalake`** — read [Delta Lake](https://delta.io/) tables via `deltalake` (delta-rs)
- **`type: iceberg`** — read [Apache Iceberg](https://iceberg.apache.org/) tables via `pyiceberg` (REST / SQL / Hive / local catalog)

Both read from **local / S3 / GCS**. Each loads the table into Arrow, registers it in an in-memory **DuckDB**, and runs your model SQL against it — so column selection and incremental cursor filters work like any other source.

## Design
- Lazy-imported **optional extras** `drt-core[deltalake]` / `drt-core[iceberg]` (no heavy deps in core; DuckDB is already bundled).
- `DeltaLakeProfile` / `IcebergProfile` in `credentials.py`; cloud auth via `storage_options` (Delta) / catalog `properties` (Iceberg) with a new `resolve_env_dict` helper — any key ending in `_ENV` resolves from that env var.
- Registered in the connector registry; **all 9 drift surfaces** updated (SOURCES, `_connector_detail`, README + README.ja, drt-init skill, MCP inventory, `docs/connectors/{deltalake,iceberg}.md`). `scripts/check_drift.sh` is clean (13 sources).

## Tests
15 unit tests across `tests/unit/test_{deltalake,iceberg}_source.py` — protocol compliance, `describe()`, the `_table_name` / `_catalog_properties` helpers, `ImportError` when the extra is absent, a **fully-mocked extract** (Arrow → DuckDB register → query → dict rows, asserting storage-option `_ENV` resolution), and profile save/load round-trip. `ruff` + `mypy` clean.

## Notes
- **Implementation only.** Live-warehouse verification (the mock suite can't exercise real S3/GCS/catalog behaviour) is tracked under the **#654** epic + its harness (#674).
- i18n marker bump is a warn-only follow-up (README.md + README.ja.md were updated identically).

Closes #172, closes #173.

PR #678, opened 2026-06-28
feat(engine): add truncate mask strategy and object config form (#660)

### LINKED ISSUE #660: feat: PII masking — add truncate strategy (follow-up to #427)

Pawansingh3889:
Closes #660. Follow-up to #427 / #668 (the v1 `hash | redact` masking).

## What

Adds a `truncate` strategy and the backwards-compatible **object form** for param-bearing strategies, as scoped in the issue:

```yaml
sync:
  mask:
    email: hash                              # flat form, unchanged
    name: {strategy: truncate, length: 2}    # object form for param-bearing strategies
```

- `truncate` keeps the first `length` characters (e.g. `length: 2` -> initials).
- The flat form (`field: hash | redact`) keeps working unchanged.
- New `MaskRule` model + `MaskSpec` union in `config/models.py`. Validation: `truncate` requires a non-negative `length`; the parameter-less strategies reject `length`; unknown strategies are rejected.
- The engine handles both forms at the same pure transform seam (outside the `engine/sync.py` SyncObserver boundary), preserving null passthrough and non-string stringify.

## Edge cases (per the issue)

- `length: 0` -> empty string; `length` >= the value length -> returned as-is (string slicing covers both).
- Null passes through untouched; non-strings stringified first, consistent with v1.

## Schema

Regenerated the bundled VS Code `sync.schema.json` (the schema-drift gate) so the object form is exposed, and bumped the extension version 0.1.0 -> 0.1.1 as the regenerate script reminds.

## Tests

`tests/unit/test_masking.py` extended: object-form acceptance and validation (truncate-needs-length, `length` on a parameter-less strategy rejected, unknown strategy rejected) plus `truncate` behaviour (first-N, `length: 0`, length >= value length, non-string, null, and mixed flat + object). Full suite green; ruff and mypy clean.

cc @masukai - split out of #427 per your note on the issue.


masukai:
Reviewed — this looks good to merge: both config forms work side by side, the engine seam is unchanged, and the regenerated schema matches. One thing in the way: #675 (the lakehouse sources) just landed and collides with this on the CHANGELOG `[Unreleased]` block, so GitHub now flags it as conflicting. A quick rebase onto `main` (just the CHANGELOG part) and I'll merge. Sorry for the churn — that one's on my merge timing, @Pawansingh3889.

masukai:
Merged — thanks for the quick rebase, @Pawansingh3889. `truncate` + the backwards-compatible object form are in `main` now. Closes #660.

Issue #654, opened 2026-06-19
Real-warehouse smoke validation for DWH destinations (mock-only paths)


masukai:
## Why

The DWH destinations are CI-tested **only with mocks** — there are no live cloud credentials in CI, so the connector libraries (snowflake-connector-python, google-cloud-bigquery, databricks-sql-connector) are `sys.modules`-injected `MagicMock`s. That fully covers drt's own logic (SQL shape, branching, error mapping), but it can't catch a **dialect-specific behaviour** that only a real warehouse exercises. Several recently-shipped / in-flight features hinge on exactly such specifics and deserve a one-time real-service sanity check.

This is a single tracker for those "mock-only, validate-once-on-real-service" items, ordered by urgency.

## Checklist

- [x] **Snowflake — VARIANT/OBJECT/ARRAY serialization** (PR #653, #317 Phase 2a) ⚠️ **highest priority** (newest, uses a specific SQL form). Verify a `dict`/`list` lands as proper semi-structured data via the `INSERT INTO t (a, v) SELECT %s, PARSE_JSON(%s)` form (drt switches to `SELECT` because Snowflake disallows functions in `VALUES`). Confirm: parsed JSON (not a stringified `repr`), and that the same path works through `merge` (staging) + `replace` (truncate/swap).
- [x] **Snowflake — `sync.mode: replace`** truncate + `replace_strategy: swap` (`ALTER TABLE … SWAP WITH`), v0.7.9 (#434). Confirm the atomic swap + grant preservation on a real table.
- [x] **Databricks — `sync.mode: replace`** via `INSERT OVERWRITE` (snapshot isolation), v0.7.9 (#643). Confirm atomicity + table-object/grant preservation.
- [x] **Databricks — complex-type serialization** (#317 Phase 2b, not yet built). First confirm whether dict/list → STRUCT/VARIANT-equivalent needs handling at all on a real workspace before wiring.
- [x] **BigQuery — insert (streaming) + MERGE-via-temp-table**, v0.7.9 (#165). Confirm streaming-insert buffer behaviour + the temp-table MERGE/cleanup on a real project.

## How (options)

- **Cheapest**: a manual run-through against personal dev accounts, checking each box once, before the next release that ships these.
- **Durable**: a separate **gated / nightly** CI job that reads real credentials from secrets and runs a minimal end-to-end sync per destination (kept off the per-PR path so contributors aren't blocked). Mirrors the DuckDB E2E harness (v0.7.5) but for cloud services.

[...]

PR #680, opened 2026-06-28
feat(databricks): schema-aware serialization via information_schema (#317)

masukai:
## Schema-aware serialization for Databricks (#317) — foundation

This is the **introspection foundation** for the Databricks leg of #317 (Layer 3). I'm opening it as a draft for **@Muawiya-contact to finish** — a co-authored pair on a real piece of work.

### Done (this draft)
- `drt/destinations/schema.py`: `describe_columns()` now routes a `DatabricksDestinationConfig` to `_describe_databricks()`, which reads `<catalog>.information_schema.columns` and maps each column to a type category (case-insensitive schema/table match).
- `_categorize_databricks()`: STRUCT / MAP / ARRAY / VARIANT → `json` (Databricks loads these via `from_json` / `parse_json`; no driver-side typed adapter), everything else → `scalar`. Mirrors the Snowflake treatment.
- `tests/unit/test_schema.py`: introspection mapping, query shape, empty→`None`. Green + ruff/mypy clean.

### To finish (the completion — @Muawiya-contact)
- [ ] Wire `describe_columns(config)` into `drt/destinations/databricks.py` — cache the category map per sync (see `snowflake.py`'s `_schema_cache`), and build the value clause so `json`-category columns bind via `from_json(?, <ddl>)` / `parse_json(?)` while scalars pass through.
- [ ] Cover the **insert**, **merge/staging**, and **replace** write paths (all build `INSERT INTO ... VALUES`).
- [ ] `json.dumps` the `json`-category values before the bind (so the SQL function receives a JSON string), mirroring `snowflake.py`'s `_order_row_values`.
- [ ] Unit tests for the write-path encoding + an end-to-end round-trip case.
- [ ] `docs/connectors/databricks.md` note + a `CHANGELOG.md` `[Unreleased]` entry.

### Co-authorship
Build on these two commits (branch from `feat/schema-aware-databricks`) and your commits ride alongside mine — the squash-merge carries both as `Co-authored-by`, so we both land the credit. 🙂

Part of #317 (Databricks phase; the epic stays open for ClickHouse / BigQuery).


Muawiya-contact:
@masukai picking this up — foundation reads cleanly, thanks. Two quick things before I wire the write path, since `from_json` needs the column's type DDL that the category map doesn't carry:

1. **Where should the `from_json` DDL come from?** Your `_describe_databricks` returns just the category (json/scalar), so the write path doesn't have the column shape (e.g. `STRUCT<...>`) that `from_json(?, '<ddl>')` needs. Two ways:
   - **A:** the write path fetches `full_data_type` itself for the json columns (your introspection stays as-is), or
   - **B:** you extend `_describe_databricks` to also surface `full_data_type`, and I just consume it.
   
   Either's fine by me — your call on where the introspection lives.

2. **Binding split** — planning `from_json(?, '<full_data_type>')` for STRUCT/ARRAY/MAP and `parse_json(?)` for VARIANT (which has no DDL form), scalars pass through unchanged. Value clause switches to the `SELECT` form when wrapping (mirroring the Snowflake leg), stays `VALUES` otherwise so the common path is byte-identical. Sound right?

Once you point me at A vs B I'll build the value clause + thread it through insert / merge-staging / replace, with shape-asserting tests + a round-trip + docs/CHANGELOG.

masukai:
Good question — that's a real gap. The DDL comes from `full_data_type` in `information_schema.columns` (the parameterised form: `ARRAY<STRING>`, `STRUCT<a: INT, b: STRING>`) — the category query only reads `data_type`, which drops it.

Pushed `describe_databricks_ddls(config)` (6d81055) for exactly this: it returns `{column: full_ddl}` for **STRUCT / ARRAY / MAP only**. VARIANT is deliberately omitted — it loads via `parse_json(?)` with no DDL — so the write-path routing per column is:

- `category[col] == "json"` **and** `col in ddls` → `from_json(?, '<ddl>')`
- `category[col] == "json"` **and** `col not in ddls` → `parse_json(?)` (VARIANT)
- else → `?` (scalar passthrough)

A couple of notes for the wiring:
- Cache both maps once per sync (like `snowflake.py`'s `_schema_cache`). It's a second light introspection query — fine to keep separate, or fold both columns into one `SELECT` later if it ever matters.
- `json.dumps` the value before the bind so the SQL function receives a JSON string.
- The DDL is interpolated as a literal into `from_json(?, '<ddl>')` (it comes straight from `information_schema`, so no injection surface) — the value itself stays a `?` bind.

Ping me if `from_json` chokes on any nested type and we'll adjust.

Muawiya-contact:
@masukai! 
B it is — thanks for pushing `describe_databricks_ddls`, that's the clean split. Wiring now: cache both maps per sync, route per-column (from_json with the literal DDL for STRUCT/ARRAY/MAP, parse_json for VARIANT, `?` for scalars), json.dumps the json values at bind, SELECT-form only when wrapping so the no-json path stays byte-identical. Shape-asserting tests + round-trip + docs/CHANGELOG to follow. Will ping if any nested type trips from_json.

masukai:
Write path landed via #685 (reviewed + squashed into this branch — thanks @Muawiya-contact 🚀). This PR is now the complete Databricks leg of #317: introspection foundation (`describe_columns` + `describe_databricks_ddls`) + per-column `from_json`/`parse_json` routing across all four write paths, with `introspect_schema` opt-out. Undrafting — ready for review. Live-warehouse validation stays tracked under #654/#672 via the #674 harness before release.

Muawiya-contact:
Thanks @masukai — glad the routing matched. On the DDL-quote nit: happy to add the `ddl.replace("'", "''")` belt-and-braces as a tiny follow-up PR once #680 lands in main, so we don't churn this branch pre-merge. Exotic as it is, it's a free hardening and keeps the from_json literal bulletproof. Will open it unless you'd rather just leave it.

masukai:
Follow-up PR after this lands is exactly right, @Muawiya-contact — please do open it. Free hardening, zero churn here, and it'll be a 2-minute review. 🙌

yodakanohoshi (dismissed):
🎉 Really clean work — this completes the Databricks leg of #317 and stays faithfully symmetric with the Snowflake implementation. The per-column routing (`from_json` with the literal DDL for STRUCT/ARRAY/MAP, `parse_json` for VARIANT, `%s` for scalars) is exactly right, and switching to the `SELECT` form only when wrapping — keeping the no-json path byte-identical to `VALUES (...)` — means zero risk to existing Databricks syncs.

Test coverage is thorough: unit-level `_value_clause` / `_bind_row` routing, all four write paths (insert / merge-staging / replace-truncate / replace-swap), case-insensitive column matching, the `introspect_schema: false` opt-out, and the JSON encoding round-trip. Docs, CHANGELOG, and the VS Code schema are all updated too. CI is green across the board.

One tiny non-blocking observation: `_insert_rows` keeps `columns` / `json_cols` optional with a `list(row.values())` fallback, but both call sites always pass them, so the `else` branch is currently unreachable (it's what the codecov patch-miss points at). Nice thing about the `_bind_row` path being the live one is that it's also order-safe — e.g. a REST API source can legitimately yield rows with different key order/set per object (`{id, name}` then `{name, id}`), which `list(row.values())` would misalign but `_bind_row`'s `row.get(c)` handles correctly. Mirrors `snowflake.py` exactly, so totally fine to leave as parity; if we ever tighten it, making `columns` required drops the dead branch in both. Nothing needed here.

LGTM — nice addition. 🚀

---
_Generated by [Claude Code](https://claude.ai/code)_

yodakanohoshi:
Approved & CI green — just needs a rebase onto `main` to clear the conflict. 🙂

---
_Generated by [Claude Code](https://claude.ai/code)_

Muawiya-contact:
Thanks @yodakanohoshi — and good eye on the `_insert_rows` fallback. You're right that it's the codecov patch-miss, and the order-safety point is exactly why I kept `_bind_row` as the live path: `row.get(c)` per column stays correct even when a source yields rows with varying key order/set (REST especially), where `list(row.values())` would silently misalign. Left the optional fallback for Snowflake parity as you noted — happy to make `columns` required in both destinations as a follow-up if we ever want to drop the dead branch symmetrically. Appreciate the thorough review 🙏

masukai:
Rebased onto `main` to clear the conflict — GitHub auto-dismissed the approval on the new commits, but the change vs your approved version is **mechanical only**, no logic touched:

1. **vscode schema** — regenerated `sync.schema.json` so it carries **both** main's `MirrorConfig` (#688) **and** this branch's `introspect_schema`; extension version `0.1.3 → 0.1.4` (schema-drift gate is green).
2. **CHANGELOG** — the Databricks #317 entry had landed inside the already-shipped `[0.7.10]` section after the rebase; moved it up to `[Unreleased]` where it belongs (v0.7.10 shipped without this leg).

No changes to `databricks.py` / `models.py` / `schema.py` or any test beyond the rebase replay. Full suite 1745 green, ruff+mypy clean, schema-drift + check-changelog + check-i18n all pass locally. Re-requesting your 👍 when you have a sec 🙏 — this closes the #317 epic.

Muawiya-contact (approved):

PR #685, opened 2026-07-01
feat(databricks): schema-aware write path — from_json/parse_json for complex columns (#317)

Muawiya-contact:
Completes the Databricks leg of #317 on top of your foundation (`6d81055`). Wires `describe_columns` + `describe_databricks_ddls` into the write path.

## What it does
Per-column routing for json-category columns, exactly as you spec'd:
- `STRUCT` / `ARRAY` / `MAP` → `from_json(%s, '<ddl>')` (DDL from `describe_databricks_ddls`)
- `VARIANT` → `parse_json(%s)` (no DDL form)
- scalars → `%s` (unchanged)

`json.dumps` on the json values at bind; the DDL is interpolated as a literal (verbatim from `information_schema`, no injection surface), the value stays a bind. Because Databricks won't accept these functions in a `VALUES` clause, wrapping switches the INSERT to the `SELECT` form; **with no json columns the clause is the unchanged `VALUES (%s, …)` — byte-identical to before.** Both maps cached once per sync (`_resolve_schema` / `_resolve_ddls`), mirroring `snowflake.py`.

Threaded through all four write paths: **insert, merge-staging, replace-truncate, replace-swap.**

## Two decisions worth your eye
1. **Added `introspect_schema: bool = True` to `DatabricksDestinationConfig`.** It only existed on Snowflake; I added it for parity + a user opt-out, and `_resolve_*` gate on it. Config-model change, so flagging it.
2. **Existing databricks tests now set `introspect_schema=False` in their `_config` fixture** — with Layer 3 on by default, the mock-based tests were seeing the introspection queries (extra `execute`/`close` calls). This is the same fix you used for the Snowflake destination tests.

## Tests
13 new in `tests/unit/test_databricks_schema_aware.py`: value-clause routing (incl. case-insensitive column match), `json.dumps` encoding round-trip, all four write paths, `introspect_schema: false` opt-out, no-json byte-identical `VALUES`. Local: ruff clean, `mypy drt` clean (133 files), 77 databricks/schema tests pass, your `test_schema.py` still green.

**Live validation:** the `from_json` / `parse_json` insert form is the canonical Databricks complex-load pattern, but the suite is mock-based like the rest of the destination — worth a `workflow_dispatch` sanity check via the #674 harness before release. Ping me if `from_json` chokes on any nested type.

Part of #317 (epic stays open for ClickHouse / BigQuery).


Muawiya-contact:
@masukai this is the #680 completion — opened as a PR into `feat/schema-aware-databricks` (not main) so my commit rides alongside yours and the #680 squash carries us both.

Built straight on your `6d81055`: `describe_columns` + `describe_databricks_ddls` wired into the write path with your exact per-column routing — `from_json(%s, '<ddl>')` for STRUCT/ARRAY/MAP, `parse_json(%s)` for VARIANT, scalars through as `%s`. `json.dumps` at bind, DDL interpolated as a literal, SELECT-form only when wrapping so the no-json path stays byte-identical. Threaded through all four write paths (insert / merge-staging / replace-truncate / replace-swap), both maps cached per sync.

Two calls I'd flag for you (both in the PR body): I added `introspect_schema` to the Databricks config for Snowflake parity + opt-out, and I set `introspect_schema=False` in the existing databricks test fixture — same fix you used on the Snowflake dest tests, since Layer 3 being on-by-default was perturbing the mock call counts.

ruff + mypy clean, 77 databricks/schema tests green, your `test_schema.py` untouched. The `from_json`/`parse_json` round-trip is mock-verified only (no live Databricks my side) — happy to do a `workflow_dispatch` on the #674 harness whenever you want the live sanity check. Ping me if `from_json` trips on any nested type and I'll adjust.

masukai (approved):
This is exactly the completion I hoped for, @Muawiya-contact — the per-column routing matches the spec precisely, and switching to the `SELECT` form only when a json column is present (byte-identical `VALUES` otherwise) is the right conservatism. 🙌

On your two flagged decisions:

1. **`introspect_schema` on `DatabricksDestinationConfig` — yes, right call.** Parity with Snowflake is exactly what we want here, and the opt-out matters for workspaces where the sync role can't read `information_schema`. Field name, default, and docstring all match the Snowflake precedent, so the config surface stays predictable.
2. **Existing tests pinning `introspect_schema=False` — also correct**, same fix I applied to the Snowflake destination tests for the same reason. The new suite covers the on-path, the old suite keeps covering the write logic in isolation. 👍

Verified locally on your branch: the 13 new tests + existing databricks/schema suites (72) pass, full unit suite 1657 green, ruff + mypy clean.

One optional nit, fine as a follow-up (or ignore): the DDL is interpolated into a single-quoted literal — `from_json(%s, '<ddl>')`. It comes verbatim from `information_schema` so there's no injection surface, but a pathological backtick-quoted field name containing `'` would break the literal. A cheap belt-and-braces is `ddl.replace("'", "''")` at the interpolation site. Genuinely exotic, so not blocking.

Merging into `feat/schema-aware-databricks` and undrafting #680 — the squash to main will carry us both, as you set it up. Live validation stays tracked under #654/#672 via the #674 harness before release. Thanks for the fast, clean handoff on this one 🚀

Issue #340, opened 2026-04-16
feat: sync.mode: mirror — differential delete for stale rows


masukai:
## Summary

Add `sync.mode: mirror` that detects and deletes destination rows no longer present in the source, without truncating the entire table.

## Motivation

`replace` mode (v0.5.2) solves stale-row removal by truncating and re-inserting. This works well for small-to-medium tables, but for large tables (millions of rows) where only a few rows are deleted, TRUNCATE + full re-insert is wasteful. `mirror` would only delete the diff.

## Proposed Behavior

```yaml
sync:
  mode: mirror
  upsert_key: [user_id, session_id]
```

1. UPSERT all source rows (same as `full` mode)
2. DELETE destination rows whose `upsert_key` is not in the current source result set

## Implementation Considerations

- Requires `upsert_key` (needed to identify which rows to delete)
- Deletion strategy options:
  - **NOT IN subquery**: `DELETE FROM dest WHERE key NOT IN (SELECT key FROM source)` — simple but requires materializing source keys
  - **Temp table approach**: Load source keys into temp table, then `DELETE FROM dest WHERE NOT EXISTS (SELECT 1 FROM temp WHERE ...)`
  - **Application-side diff**: Collect all source keys in memory, then issue DELETE for missing keys
- Memory/performance tradeoffs depend on table size — research needed
- This is a CDC-like pattern but without requiring a change stream

## Prior Art

- Fivetran: "mirror" mode (full replace with delete detection)
- Census: "mirror" sync behavior
- Hightouch: "mirror" mode with soft/hard delete options

## Depends on

- #337 (sync.mode: replace — foundation for table refresh patterns)

PR #688, opened 2026-07-02
feat(destinations): tracked & scoped mirror — co-writer-safe differential deletes (closes #686, closes #687)

### LINKED ISSUE #686: feat(engine): tracked mirror — state-based differential deletes (only delete rows drt itself synced)
## Motivation

`sync.mode: mirror` (#340) computes deletions by diffing against the **destination table**: after upserting all source rows, it deletes destination rows whose `upsert_key` was not observed in the run. This is correct only when drt exclusively owns the table.

The canonical Reverse ETL destination is a **product's operational database, where the application itself also writes**. In that setting, destination-diff mirror deletes application-written rows — so users with co-writing destinations are stuck on `mode: full` and accumulate stale rows forever (we hit exactly this: a parent/child writeback where 54 of 55 previously-synced link rows went stale, but we cannot enable mirror because the app may also insert into the table).

Industry precedent — this is how the SaaS tools define mirror:

- **Census**: "mirror syncs identify deletions by comparing against the data they have already sent — not the data that might or might not already exist in the destination. […] It doesn't delete records that existed in the destination before the sync was set up or that were created by other means." (https://docs.getcensus.com/syncs/overview)
- **Hightouch**: delete behavior is driven by its diff engine tracking previously-synced records, separate from sync mode. (https://hightouch.com/docs/syncs/types-and-modes)

drt's destination-diff mirror is the simpler variant; a state-based ("tracked") mirror would close the biggest semantic gap vs. the SaaS tools and make mirror safe on shared tables — arguably the most common real-world destination.

## Proposed semantics

Persist, per sync, the set of `upsert_key` tuples drt has successfully written. On each run:

1. Upsert all source rows (unchanged).
2. `to_delete = previously_synced_keys − current_source_keys` — delete **only those rows**.
3. Persist `current_source_keys` as the new state.

Rules:

- **First run**: baseline only — record keys, delete nothing (same as Census: "the first sync will be an upsert for all records, and the second and following syncs will account for deletions").
- **State missing/lost**: skip deletes + WARN, re-baseline from this run (safety-first, mirroring the existing empty-source guard in #340).
- Only keys **actually written** count (rows dropped by `on_error: skip` / lookup `on_miss: skip` are not recorded).

## Config sketch (naming up for discussion)

```yaml
sync:
  mode: mirror
  mirror:
    strategy: tracked   # default "destination" = current #340 behavior, fully backward compatible
```

## Implementation notes

- State plumbing exists: per-sync `StateManager` (`.drt/state.json`) and destination-side watermark storage. However key sets can be large and runners are often **ephemeral containers**, so local JSON state is not durable enough.
- Suggest a drt-managed side table in the destination (e.g. `_drt_synced_keys` with `sync_name` + key columns/hash): keeps state adjacent to the data, survives runner redeploys, and lets the diff run as a set-based SQL JOIN instead of loading keys into memory.
- Composes with a future scope option (see companion issue on scoped mirror) to cheapen the delete query.

## Edge cases to cover

- Partial batch failure → record only written keys.
- Lookup `on_miss: skip`: a previously-synced row whose FK target disappeared is skipped this run → its key leaves the tracked set → it gets deleted. Document this as intended (the FK target is gone).
- Concurrent runs of the same sync (state write races) — likely out of scope, document.
- `drt run --dry-run --diff` should preview tracked deletions.

### LINKED ISSUE #687: feat(engine): mirror scope — restrict differential deletes to parent keys observed in the run
## Motivation

A common Reverse ETL shape is **1:N regeneration**: a parent entity is periodically regenerated together with its child link rows (`parent_id`, `child_id`). When the parent regenerates, its child set is replaced wholesale — stale links under that parent must be deleted, but rows under parents **not present in this run** (written by other pipelines or by the application) must not be touched.

Current options don't fit:

- `mode: full` — upsert-only, stale child rows accumulate (we measured 54 of 55 previously-synced link rows stale after a parent regeneration).
- `mode: mirror` (#340) — diffs the **whole table**, so it deletes rows under parents the sync never saw.
- `mode: replace` — TRUNCATE, worst of both.

## Proposal

```yaml
sync:
  mode: mirror
  mirror:
    scope: [parent_id]
```

Semantics: after upserting, DELETE only rows where

- the row's `scope` column values **appeared in this run's source**, AND
- the row's `upsert_key` was **not observed** in this run.

Rows under unobserved parents are untouched. Stateless, idempotent, self-healing — every run recomputes the diff within the observed scope, so there is no state to lose (contrast with tracked mirror, #686).

## Positioning vs. tracked mirror (#686)

Complementary, not competing:

- **scope** = stateless and robust, but still assumes drt owns all rows *under the observed parents*. Right tool when co-writers only touch other parents (or only read).
- **tracked** (#686) = fully co-writer safe even within the same parent, but state-dependent.
- They compose: `strategy: tracked` + `scope: [...]` lets the delete query prune to observed parents cheaply.

## Implementation sketch

The MySQL/Postgres/ClickHouse/Snowflake mirror path already accumulates observed `upsert_key` tuples across batches (`_mirror_keys`) and deletes in `finalize_sync`. Extend it to also accumulate distinct scope values, then delete per scope group:

```sql
DELETE FROM t
WHERE parent_id IN (<observed scopes>)          -- batched
  AND (parent_id, child_id) NOT IN (<observed keys>)
```

Existing safety carries over: no batches with records → skip deletes entirely.

## Validation

- `scope` columns must be a subset of the model's output columns (config-time error otherwise).
- Works on all four SQL destinations that already support mirror; non-SQL destinations reject the option.

masukai:
Closes #686.

## What

`sync.mirror.strategy: tracked` — a state-based mirror that only deletes rows **drt itself synced**, making `mode: mirror` safe on destination tables the application also writes to (the canonical Reverse ETL destination). Adopts the Census/Hightouch semantics: deletions are computed against the keys drt previously sent, not against whatever happens to be in the table.

```yaml
sync:
  mode: mirror
  mirror:
    strategy: tracked   # default "destination" = #340 behaviour, fully backward compatible
```

## Semantics

| | `destination` (default, #340) | `tracked` (this PR) |
|---|---|---|
| Delete set | whole destination table − observed keys | previously-synced keys − observed keys |
| Co-writer rows | **deleted** | never candidates |
| State | none | `_drt_synced_keys` in the destination |
| First run | deletes immediately | baselines, no deletes |
| Lost state | n/a | WARN + re-baseline (safety-first) |

- State table (`sync_name`, `key_hash`, `key_json`) is created lazily in the target table's schema/database — adjacent to the data, durable across ephemeral runners (issue's implementation note).
- Target DELETE + state rewrite run in **one transaction**.
- #340 guards carry over: transient empty source → no deletes **and** baseline untouched; only successfully-written keys enter the state (`row_errors` filtered).
- Sync name reaches the destination via a `SyncConfig`-injected `PrivateAttr` on `SyncOptions` — no YAML surface change, no Destination-protocol widening.
- int/str keys round-trip exactly; non-JSON-native key types are stringified in state (documented limitation).

## Scope

**Postgres + MySQL** in this phase (same phasing as #317). ClickHouse / Snowflake / Databricks **fail fast** with a clear `ValueError` instead of silently falling back to the co-writer-unsafe destination diff — their legs, `--dry-run --diff` preview of tracked deletions, tracked+scope composition (#687), and a SQL-JOIN state diff for very large key sets are follow-ups (issues to be filed on merge).

## Tests / verification

- 20 new tests: config surface (`test_config.py`), key canonicalisation (`test_mirror_state.py`), Postgres + MySQL tracked flows (baseline / stale-key delete / no-op parity / composite keys / state-table qualification / WARN logging), and the three fail-fast guards.
- Full unit suite: **1714 passed**, ruff + mypy (136 files) clean, `make check-changelog` OK, `make check-i18n` OK (docs/ja updated + marker bumped in the same PR).
- VS Code bundled schema regenerated for `MirrorConfig` + extension version bumped (schema-drift convention). Note: #680 also bumps the extension version on its branch — whichever lands second rebases + re-regenerates.

## Docs

`docs/connectors/postgres.md` + `mysql.md` tracked-mirror sections (when to use which strategy, state-table semantics), `docs/llm/API_REFERENCE.md` sync block, `docs/ja/configuration.md` (strategy table + marker bump), CHANGELOG.


masukai:
Consolidated #689 (mirror.scope, #687) into this branch via squash — this PR now delivers **both** mirror extensions behind the one shared `MirrorConfig` surface: `strategy: tracked` (#686, state-based co-writer-safe deletes) and `scope` (#687, stateless observed-parents restriction). Closes #686 and #687 together. CI re-running on the combined diff.

PR #645, opened 2026-06-15
feat(bigquery): add BigQuery destination — insert + merge (closes #165)

masukai:
Closes #165 (the oldest open connector request). **Builds on @PFCAaron12's contribution in #584** — see below.

## What

The BigQuery destination — the fourth major DWH destination (Snowflake / Databricks / BigQuery now cover the big three).

- **INSERT** (append) via the streaming insert API (`insert_rows_json`), per-row errors → `result.row_errors` (`on_error` skip/fail).
- **MERGE** (upsert, `config.mode: merge`, requires `upsert_key`) — load the batch into `<table>_drt_tmp` (`load_table_from_json`), run a single `MERGE … ON <upsert_key>` (UPDATE matched / INSERT not-matched), drop the temp table. Composite keys AND-join; all-columns-are-key skips UPDATE. (BigQuery load/MERGE are job-level → batch-level error handling, documented.)
- **Auth** mirrors the BigQuery source: ADC default, or service-account `keyfile`.

## Relationship to #584 (@PFCAaron12 / Gloria Aaron)

@PFCAaron12 opened #584 with a working BigQuery destination. After 14 days without activity (and an unanswered takeover offer from 2026-06-01), I'm landing it as a fresh, Protocol-compliant PR so it can ship in v0.8 — **with @PFCAaron12 credited as co-author**.

| Kept from #584 (theirs) | Reshaped (to the `Destination` protocol) |
|---|---|
| MERGE-via-temp-table approach | `write(rows, mode)` → `load(records, config, sync_options) -> SyncResult` |
| ADC / keyfile auth | `@dataclass` config → Pydantic `BigQueryDestinationConfig` in the `DestinationConfig` union |
| `insert_rows_json` for append | `logging`+`raise` → `SyncResult` / per-row `row_errors` / `on_error` |
| | positional `mode` arg → `config.mode` dispatch |

(Their #584 branch is ~60 files diverged from current main, so a rebase wasn't viable — hence the fresh PR.)

## Scope / safety

- New config + one union line in `models.py`; isolated otherwise.
- **18 tests, 100% coverage** on `bigquery.py` (mocked `google.cloud.bigquery`); `ruff check` + `mypy drt` clean; full suite green; **0 new drift** (doc / skill / README / MCP inventory all updated).
- `sync.mode: mirror` / `replace` for BigQuery are **follow-ups**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #700, opened 2026-07-04
test(bigquery): add MERGE-path smoke coverage + provisioning docs (#673, #645)

yodakanohoshi:
## What

Adds the missing BigQuery leg coverage to the DWH smoke harness, per the plan agreed on #654.
The BigQuery smoke previously covered only the streaming `insert` path; this adds the two tests
@masukai green-lit:

- **`test_bigquery_merge_roundtrip`** — drives `mode: merge` end-to-end against a real project:
  `load_table_from_json` into `<table>_drt_tmp` → single `MERGE` (UPDATE matched `id=1` / INSERT
  not-matched `id=2,3`) → temp-table drop. Pre-seeds a stale `id=1` so **both** MERGE branches run,
  and asserts the `_drt_tmp` staging table is cleaned up afterwards (#645).
- **`test_bigquery_connection`** — fast credential check via `test_connection` (`SELECT 1`),
  mirroring `test_snowflake_connection`.

Also adds BigQuery provisioning notes to `tests/integration/dwh/README.md` (billing-enabled
project, keyfile auth, least-privilege roles, `disableServiceAccountKeyCreation` caveat).

## Why

Part of #673 (BigQuery DWH smoke validation), under epic #654. With the existing
`test_bigquery_insert_roundtrip` covering the streaming leg, merge + connection close out the
BigQuery verification set. Composite-key MERGE is intentionally left as an optional follow-up to
keep this PR's scope tight (single-`id` `upsert_key`, matching the reference shape).

## Gating (safe no-op)

Both tests use the same double gate as the rest of the harness — `pytest.importorskip` (driver) +
`require_env` (credentials) — so a plain `pytest` run, every fork, and CI without
`DRT_SMOKE_BIGQUERY_*` secrets all skip cleanly without opening a connection.

## Verification

Ran green locally against a real BigQuery project (`drt_smoke` dataset) — `insert` + `merge` +
`connection` all pass; created tables (`drt_smoke_*` and `_drt_tmp`) are dropped in `finally`, so
the dataset returns to its prior state. `ruff` + CI `mypy drt` clean.

The cloud secrets (`SMOKE_BIGQUERY_*`) and the official "verified on a real service ✓" sign-off
remain @masukai's per the #654 split — happy to hand off once this is in.

Part of #673 · Epic #654 · Cleanup #645

### DIFF
--- a/tests/integration/dwh/README.md
+++ b/tests/integration/dwh/README.md
@@ -61,6 +61,21 @@ the cloud accounts and the "verified ✓" sign-off are the maintainer's
 | `SMOKE_BIGQUERY_DATASET` | a throwaway dataset |
 | `SMOKE_BIGQUERY_KEYFILE_JSON` | full service-account JSON; the workflow writes it to a temp file |
 
+BigQuery prerequisites (#673):
+
+- A **billing-enabled** project (BigQuery jobs require an active billing account,
+  even for tiny throwaway tables).
+- **Service-account keyfile** auth. The BigQuery leg drives two write paths — the
+  streaming `insert` path and the temp-table `MERGE` path (`<table>_drt_tmp` →
+  `MERGE` → drop, #645) — both against a throwaway dataset.
+- Least-privilege roles for the service account: `roles/bigquery.jobUser` on the
+  project (to run load/query/MERGE jobs) **plus** dataset-scoped
+  `roles/bigquery.dataEditor` on the smoke dataset (to create/insert/drop the
+  target + `_drt_tmp` tables). Avoid granting project-wide `dataEditor`.
+- Caveat: if the org enforces the `iam.disableServiceAccountKeyCreation`
+  constraint, a keyfile can't be minted for the SA — provision the key in a
+  project/folder where that policy is not enforced, or use an exempted SA.
+
 ## Adding a warehouse leg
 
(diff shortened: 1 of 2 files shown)


masukai (approved):
LGTM — exactly the shape we agreed on #654, and the BigQuery-specific details are right.

Verified against the destination:
- **temp-table assertion matches the impl** — `bigquery.py` builds `f"{table_id}_drt_tmp"`, and the test's `{project}.{dataset}.{table}_drt_tmp` + `NotFound` check pins exactly that cleanup (#645). 👍
- **stale `id=1` seeded via DML INSERT, not streaming** — sidesteps the streaming-buffer visibility lag, so the MERGE reliably sees a matched row (UPDATE) alongside the unmatched INSERTs. Good call.
- Gating is the same double gate as the rest of the harness (`importorskip` + `require_env` → graceful skip), so it's a clean no-op on forks / local.
- `test_connection` mirrors `test_snowflake_connection`; single-`id` scope with composite-key MERGE left as an optional follow-up is the right call to keep this tight.

Merging. I'll add the `DRT_SMOKE_BIGQUERY_*` repo secrets + run the real-project sign-off (the #654 split) and check the box there. Thanks for the clean handoff 🚀

PR #644, opened 2026-06-14
feat(databricks): mode: replace — truncate default + replace_strategy: swap (closes #643)

masukai:
Closes #643. Follow-up to #434 / #642 (Snowflake) — same gap on Databricks.

## What

Databricks had **no `sync.mode: replace` handling** — `load()` only routed insert/merge/mirror, so `mode: replace` silently fell through to insert/merge. Adds the **full replace branch** at parity with Postgres/MySQL/ClickHouse/Snowflake:

- **`replace_strategy: truncate`** (default) — `TRUNCATE TABLE` once, then INSERT each batch.
- **`replace_strategy: swap`** — zero-downtime. **Delta Lake has no `ALTER TABLE … SWAP WITH`** (Snowflake's primitive), so drt stages batches into a shadow `<table>__drt_swap` (`CREATE OR REPLACE TABLE … AS SELECT * … WHERE 1=0`), then `finalize_sync` runs an atomic **`INSERT OVERWRITE <table> SELECT * FROM <shadow>`** and drops the shadow. `INSERT OVERWRITE` commits a new Delta version atomically (snapshot isolation) — readers see full-old or full-new, never partial — and **preserves the target table object** (grants / properties / clustering survive; the table is never recreated, which is actually a cleaner grant story than Snowflake's object-swap). First run (target absent) writes directly + skips swap.
- **Orphan cleanup** — implements the `OrphanSwapCleanup` protocol so `drt clean --orphans` works.

## Scope / safety

- **Isolated to `drt/destinations/databricks.py`** — no `drt/config/models.py` change (`replace_strategy` already on `SyncOptions`).
- 12 new tests (existing tests untouched), **100% coverage** on databricks.py; `ruff check` + `mypy drt` clean.

## Design note

| | Snowflake (#642) | Databricks (this PR) |
|---|---|---|
| atomic swap | `ALTER TABLE … SWAP WITH` | `INSERT OVERWRITE … SELECT * FROM shadow` (Delta has no SWAP WITH) |
| grants | preserved (objects swapped) | preserved (table object untouched, data-only overwrite) |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #705, opened 2026-07-05
test(databricks): DWH smoke — insert + swap-overwrite + complex-type/VARIANT + connection (#672)

yodakanohoshi:
## What

Adds the Databricks leg coverage that #672 calls for to the DWH smoke harness (#674), mirroring the BigQuery shape from #700. On top of the existing `insert` roundtrip, the leg now covers:

- **`test_databricks_replace_swap_roundtrip`** — `INSERT OVERWRITE` atomicity for `replace_strategy: swap` (#644). Pre-seeds a stale row the source never emits, then checks (a) the finalize-time `INSERT OVERWRITE` replaced the table's data — only the 3 source rows remain — and (b) the `<table>__drt_swap` shadow was dropped in `finalize_sync`.
- **`test_databricks_complex_type_serialization`** — complex-type / VARIANT serialization (#317 Databricks leg). Seeds a DuckDB source whose row carries a Python `list` + `dict`, syncs into `ARRAY` / `STRUCT` / `VARIANT` columns, and reads typed sub-fields back (`element_at(tags, 1)`, `attrs.theme`, `meta:source`) to confirm the values reconstruct via `from_json` / `parse_json` as real complex types rather than opaque JSON strings.
- **`test_databricks_connection`** — fast credential check via `test_connection` (`SELECT 1`), mirroring the Snowflake / BigQuery legs.

README gains Databricks provisioning notes (Unity Catalog / Delta, SQL warehouse HTTP path incl. the VARIANT channel requirement, least-privilege `CREATE`/`MODIFY` grants scoped to a throwaway schema).

All four tests have been run against a live Databricks workspace (combined locally with #706) and pass — see Verification below.

## Gating (safe no-op)

Same double gate as the rest of the harness — `pytest.importorskip("databricks.sql")` + `require_env` — so a plain `pytest` run, forks, and CI without `DRT_SMOKE_DATABRICKS_*` secrets all skip cleanly (verified: `4 skipped`).

## Depends on #706 — merge that first

Running this suite against a real workspace surfaced a pre-existing issue: the destination's pyformat `%s` binds aren't compatible with databricks-sql-connector >=3.0's native-paramstyle default, so parameterised writes fail server-side with `PARSE_SYNTAX_ERROR` (details in #706). The fix is destination code, so it ships separately; this PR stays test-only. Until #706 merges, the three roundtrip tests fail against a live account (the connection test passes) — CI is unaffected either way since the secrets aren't set yet.

## Verification

Against a live workspace (catalog `workspace`, schema `drt_smoke`):

- **This branch + #706 merged locally: all 4 tests pass** (~90s), on connector 3.7.5 and 4.3.0.
- Without #706: the three roundtrips reproduce the `PARSE_SYNTAX_ERROR`; `test_databricks_connection` passes.
- The smoke schema is empty afterwards — every leg drops what it creates in `finally`.

Cloud secrets (`SMOKE_DATABRICKS_*`) and the official "verified on a real service ✓" sign-off remain @masukai's per the #654 split.

Part of #672 · Epic #654 · Cleanup #644 · Serialization #317 · Depends on #706

### DIFF
--- a/tests/integration/dwh/README.md
+++ b/tests/integration/dwh/README.md
@@ -61,12 +61,16 @@ Databricks prerequisites (#672):
   created **Delta** (`USING DELTA`): the `replace_strategy: swap` leg relies on
   Delta `INSERT OVERWRITE` snapshot-isolation atomicity.
 - A running **SQL warehouse**; its HTTP path is `SMOKE_DATABRICKS_HTTP_PATH`.
+  The complex-type leg uses a `VARIANT` column, so the warehouse must be on a
+  channel that supports VARIANT (current serverless/pro warehouses do).
 - Least-privilege grants for the token principal: `USE CATALOG` + `USE SCHEMA`,
   plus `CREATE TABLE` / `MODIFY` on the smoke schema — the swap leg builds and
-  drops a `<table>__drt_swap` shadow. Scope the grants to the throwaway schema,
-  not the whole catalog.
-- The `insert` + `replace_strategy: swap` + `test_connection` legs run against
-  the same throwaway schema and drop everything they create in `finally`.
+  drops a `<table>__drt_swap` shadow, and the complex-type leg creates ARRAY /
+  STRUCT / VARIANT tables. Scope the grants to the throwaway schema, not the
+  whole catalog.
+- The `insert` + `replace_strategy: swap` + complex-type (`from_json` /
+  `parse_json`) + `test_connection` legs all run against the same throwaway
+  schema and drop everything they create in `finally`.
 
 **BigQuery** (#673)
 | Secret | Notes |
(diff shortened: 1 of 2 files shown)


yodakanohoshi:
While validating this against a real Databricks environment, I found cases where it does not work as expected. Since we cannot merge this safely as-is, I’m closing this PR for now.

We can revisit this once the behavior is fixed and verified in an actual environment.

yodakanohoshi:
Reopening — the "does not work as expected" cases from the closing comment are now root-caused and fixed.

What we found: the failures were not in these tests but in the destination itself. The Databricks destination binds values with pyformat `%s` markers, while databricks-sql-connector >=3.0 defaults to native paramstyle and forwards `%s` to the server unexpanded, so every parameterised write fails with `PARSE_SYNTAX_ERROR` on a real workspace.

The fix ships separately in #706 (one-line `use_inline_params` opt-in plus a regression guard). With this branch and #706 combined locally, all four tests pass against a live workspace (~90s, verified on connector 3.7.5 and 4.3.0), and the smoke schema is clean afterwards.

The PR body has been updated to reflect the current four-test scope and the dependency: please merge #706 first, then this.

PR #706, opened 2026-07-05
fix(databricks): use inline params for connector >=3.0 paramstyle compatibility

yodakanohoshi:
## What

A one-line fix (plus a regression guard) for parameterised writes failing on databricks-sql-connector >=3.0:

- `_connect` now passes `use_inline_params="silent"`, restoring the client-side inline rendering this destination's pyformat `%s` binds were written against.
- The connect-kwargs contract test (`test_connect_uses_databricks_sql_kwargs`) now also checks the flag, so a future refactor that drops it shows up in `make test` rather than only in the credential-gated smoke run.
- CHANGELOG `[Unreleased]` → Fixed entry.

## Why

The destination binds values with pyformat `%s` markers, while connector >=3.0 defaults to native paramstyle (`?` / `:name`, server-side binding) and passes `%s` through to the server as-is. Parameterised statements — plain INSERT, merge-staging INSERT, both replace strategies — then fail with a server-side parse error:

```
[PARSE_SYNTAX_ERROR] Syntax error at or near '%'
== SQL ==
INSERT INTO workspace.drt_smoke.drt_smoke_… (id, name, email) VALUES (%s, %s, %s)
```

The parameterless `test_connection` (`SELECT 1`) is unaffected. Some context:

- drt-core declares `databricks-sql-connector>=3.0` without an upper bound, so fresh installs resolve to an affected connector version as well — this isn't limited to 4.x.
- The mocked unit suite has no server parse step, so this wasn't visible there. The `%s` style follows the Snowflake destination shape; snowflake-connector defaults to client-side pyformat, so the same code works fine there — the differing paramstyle default was easy to miss.
- It surfaced on the first run of the #674 DWH smoke harness against a real workspace (#672) — the kind of dialect-specific behaviour the harness is meant to catch.

Schema introspection (#317) picks up the fix through the shared `_connect`. The Databricks source is unaffected (it executes model SQL verbatim, no binds).

## Verification

Against a live workspace (catalog `workspace`, schema `drt_smoke`):

| connector | without fix | with fix |
| --- | --- | --- |
| 3.7.5 | insert / swap / complex-type roundtrips fail with `PARSE_SYNTAX_ERROR` | all smoke legs pass |
| 4.3.0 | same | all smoke legs pass |

Unit: 74 passed (`test_databricks_destination.py` / `test_databricks_schema_aware.py` / `test_schema.py`); ruff clean. Once this merges, the #672 smoke PR (#705) should pass against a real account.

## Follow-up (separate issue)

The connector deprecates inline params, so the longer-term direction is migrating this destination to native `?` markers. That's kept out of this PR on purpose: it touches every bind site (`_value_clause` scalar / `from_json` / `parse_json`, the mirror `NOT IN` assembly) and ~20 SQL-string assertions across 3 unit test files, and the mirror-mode `NOT IN` delete builds one parameter per key, which would need chunking under native mode's per-statement parameter limit. It also warrants its own live re-verification (including mirror).

Found-by: #672 · Related: #644 #317 · Unblocks: #705


masukai (approved):
Excellent root-cause fix 🎯 This is exactly the real-machine problem — and it explains why the mock suites were green all along: mocks never exercise the connector's param binding, so the pyformat `%s` → native-paramstyle mismatch only surfaced server-side on a live warehouse. Every parameterised write (insert / merge-staging / both replace strategies + the #317 introspection path through the shared `_connect`) was broken on connector >=3.0 since it shipped; `test_connection` surviving because it's the one parameterless call is the perfect tell.

`use_inline_params="silent"` is the right restore (matches the snowflake-connector default this file's SQL was modelled on; connector does its own value escaping so no new injection surface), and folding the assertion into the connect-kwargs contract test means a future refactor that drops it fails `make test` instead of only the gated smoke run. Verified green on 3.7.5 + 4.3.0, CHANGELOG entry in, native-`?` migration + the mirror `NOT IN` per-key chunking caveat noted as follow-up. Clean. Merging.

masukai:
Merged all three (#706 fix → #705 Databricks smoke → #708 Snowflake smoke) 🎉 Exemplary loop, @yodakanohoshi — you ran the harness against a real workspace, root-caused a bug that had been silently breaking **every** Databricks write since it shipped (mocks couldn't catch the paramstyle mismatch), fixed it cleanly, *and* backfilled full swap + complex-type smoke coverage for both Databricks and Snowflake in the same pass. The complex-type legs proving `from_json`/`PARSE_JSON` actually reconstruct on live warehouses is exactly the validation #317 needed. DWH smoke test coverage is now complete across all three — thank you 🙏🚀

(The Databricks CHANGELOG note about the connector-3.0 write regression means it's worth flagging in the next release notes, since it affects anyone on databricks-sql-connector >=3.0 with v0.7.9/v0.7.10.)

PR #704, opened 2026-07-04
design(docs): docs UI follow-ups — tokens, sync tabs + ego lineage, connector badges, tag pages (post-#677)

masukai:
Design follow-ups from [design pass 1](https://github.com/drt-hub/drt/pull/677#issuecomment-4876016764), landing the `design/docs-spike` work on main now that #677 is in. One PR carrying the three announced themes — the commits are split by theme if you prefer reviewing them one at a time. Tracker: #702.

## What's in here (by commit)

1. **Overview: type-distribution bars + Recent Runs panel** — syncs counted by source/destination type, colored run statuses (dot + word, never color alone).
2. **Lineage visual language artifacts** — `docs/design/drt-docs-lineage-mock.html` (ownership zones, node taxonomy, lookup ports, DLQ/drift badge design) + design-pass before/after screenshots; the visual contract #701 emits into.
3. **Sync-page tabs (YAML / Lineage / State) + per-sync ego lineage + connector badges** — tabs are progressive enhancement (no JS → stacked sections); the Lineage tab renders a static, deterministic ego-graph SVG (reads left, writes right, dashed violet lookups into distinct ports, every node an `<a>`); brand-initial connector badges (`_BADGES`, decision recorded in `docs/design/drt-docs-icon-strategy.html`) with a neutral fallback for unknown plugins. All SVG-interpolated strings are `html.escape`d.
4. **Icon-strategy decision record** (A/B/C comparison, B adopted).
5. **Design tokens + tag pages + source/destination polish + deterministic footers** — all tokens in one `:root` block (+ dark override); sidebar Tags group + `tag/<slug>.html` pages; source/destination pages get breadcrumbs, connector badges, and a Last-status column; page footers drop `generated_at` (the page-side of #697 — regenerating an unchanged project is now byte-identical outside `manifest.json`).

## Screenshots

| | |
|---|---|
| Overview | overview |
| Sync page — Lineage tab (ego graph) | see `sync/orders-to-pg.html` in a local render |
| Lineage visual target (light) | lineage light |
| Lineage visual target (dark) | lineage dark |

## Tests

33 docs unit tests green, including the XSS fixtures from #677. New surfaces (tag pages, ego SVG, badges) piggyback on the existing escape/autoescape guarantees; happy to add dedicated fixtures if you want them in this PR rather than a follow-up.

[...]

PR #711, opened 2026-07-05
hardening(docs): rmtree guard + [docs] ImportError hint + slug-collision fail-fast; databricks from_json DDL-quote escape (#703)

### LINKED ISSUE #703: hardening(docs): drt docs generate — rmtree guard, [docs] ImportError hint, slug-collision fail-fast
Follow-up to #677 (merged). Three items flagged in review, triaged as non-blocking by @Muawiya-contact + @masukai and deferred so the design spike branch wasn't churned. Tracking them here so the deferral isn't buried in the PR thread.

1. **`rmtree` guard (High)** — `render_html` does an unconditional `shutil.rmtree(output_dir)` before writing. **Default-safe** (`--output` defaults to `target/docs`, a regenerable build dir), but a user pointing `--output` at a populated directory would have it wiped. Guard: refuse unless the dir is empty or looks like a prior drt-docs build (`index.html` + `assets/`), and error if it's a file.
2. **`[docs]` ImportError hint (Medium)** — missing extras should print `pip install drt-core[docs]` (mirroring the `drt mcp run` pattern), not a stack trace.
3. **Slug-collision fail-fast (Medium)** — `_slug()` can map punctuation-only-distinct names (`a_b` vs `a__b`) to the same file, silently overwriting. Fail-fast with a clear error on duplicate slugs.

@Muawiya-contact offered to roll all three into one small hardening PR (alongside the #680 `from_json` DDL-quote fix). No behaviour-critical rush — default usage is safe today.

Muawiya-contact:
Closes #703. Follow-up to #677 (merged) and #680 (merged) — the three Copilot-flagged docs items + the `from_json` DDL-quote nit, bundled as agreed.

## Fixes

1. **`rmtree` guard (High)** — `render_html` no longer unconditionally deletes `--output`. It refuses a populated directory that isn't a prior drt-docs build (missing `index.html` + `assets/`), and errors if the path is a file. Default `--output target/docs` (a regenerable build dir) is unaffected; only a mis-pointed `--output` is now protected.
2. **`[docs]` ImportError hint (Medium)** — `--format html` without the extra prints `pip install drt-core[docs]` instead of a stack trace, mirroring the `drt mcp run` pattern.
3. **Slug-collision fail-fast (Medium)** — `_slug_map` raises a clear rename error when two names slugify to the same page (`a_b` vs `a__b`), instead of silently overwriting.
4. **Databricks `from_json` DDL-quote escape (#680 nit)** — the DDL literal now doubles single quotes (`'` → `''`) as defence-in-depth. The DDL still comes verbatim from `information_schema`; this just closes the pathological-column-name edge.

## Tests
6 new: slug-collision raise; rmtree guard across non-docs / file / empty / prior-build; the `[docs]` ImportError hint; and the DDL single-quote escape. Local: ruff clean, `mypy drt` clean (138 files), 91 docs/databricks tests pass.

## Checklist
- [x] Updated `CHANGELOG.md` — `[Unreleased]` › Fixed


masukai:
Really clean hardening PR — the rmtree guard + slug-collision fail-fast are exactly right, and the 6 tests cover the edges nicely. 🙏

One thing to fix before I merge:

**CHANGELOG** — the `[Unreleased] › Fixed` entry currently *replaces* the existing Databricks inline-params fix (#706) instead of adding to it. On `main` that bullet is the only Fixed entry, so merging as-is would drop #706's release note. Could you keep #706 and add this #703 work as a *new* bullet under Fixed? (You may have branched before #706's changelog line landed.)

Optional (fine as a follow-up): the new `render_html` guards raise `ValueError`, but `docs.py` only catches `ImportError`, so `drt docs generate` shows a traceback for the "refusing to delete / slug collision / file target" cases. Wrapping the `render_html(...)` call to `print_error` + `raise typer.Exit(1)` would make those guards land as clean CLI errors — matches the nice `[docs]` hint you just added.

Heads-up: #704 also touches `drt/docs/html.py`, and I've just landed it on `main`, so a quick rebase here will pull in both the changelog fix and the html.py merge. Thanks again! 🚀

Muawiya-contact:
@masukai Good call — done. Wrapped the `render_html(...)` call so the guards (refuse-to-delete / slug-collision / file-target) come out as clean `print_error` + exit 1, matching the `[docs]` hint. Added a test asserting the refuse-to-delete case exits 1 with the message and no traceback (and leaves the dir intact). Pushed to this branch.

Muawiya-contact:
@masukai I'm a bit confused about `CHANGELOG.md`. If any changes are required, could you please handle that? 😅😁 Using co-author <contactmuawia@gmail.com>

masukai:
Thanks for the fast turnaround — the `render_html` guards now landing as clean `print_error` + exit 1 (with the traceback-free test) is exactly what I wanted. 🙏

On **CHANGELOG** — you're all set, nothing more needed. After the rebase it's already in the right shape: the #706 inline-params line is preserved and your #703 bullet sits right below it under Fixed. That was my only merge blocker, so we're good there.

On Copilot's **databricks.py** note — that one's a misread. This PR only touches the `from_json` DDL escape; `_connect` still passes `use_inline_params="silent"` (line 671), untouched — the asserting test + green CI confirm it. Safe to resolve.

(Optional, non-blocking: Copilot's `html.py` `is_dir()` point for FIFO/device paths is a fair edge — fold in a one-line `is_dir()` guard if you like, or leave as a follow-up.) Ready to merge on my side. 🚀

masukai (approved):
LGTM — the hardening guards are exactly right, the CHANGELOG now preserves #706's inline-params note and adds #703 below it, and the render_html errors land as clean CLI failures. Copilot's databricks flag is a confirmed false positive (`use_inline_params="silent"` is still on line 671, untouched). Thanks @Muawiya-contact 🙏

Issue #701, opened 2026-07-04
feat(docs): static pre-laid project DAG — hand-rolled layout engine replaces runtime Mermaid


masukai:
## Summary

Replace the runtime-Mermaid DAG page with the **hand-rolled static layout engine** scoped in the #677 thread. This is the implementation issue for what was agreed there, so the plan survives #677 merging and closing.

## Agreed contract (from the #677 discussion — see the design pass comment and the layout scoping comment there)

**Layout (deterministic, dependency-free):**
- Ranks are free: rank = node type (`source → sync → destination`); lookup edges treated as back-edges.
- Within-rank ordering via barycenter sweeps with **fixed iterations + stable name tiebreak** → byte-identical output, clean git diffs (ties into #697). Worth A/B-ing barycenter vs the **median heuristic** in the spike.
- Back-edges (N lookups → one sync): **lane assignment** above/around the graph, each back-edge its own lane + a **distinct top-edge port** on the target sync — N lookups never stack into one curve.

**Emission (the visual contract from design pass 1):**
- Inline **themeable SVG** (CSS variables, light/dark for free), every node wrapped in an `<a>` to its detail page. No runtime layout JS, works on `file://`.
- **Ownership zones**: neutral `SOURCES — external · read` band, violet **"managed by drt"** band for syncs, neutral `DESTINATIONS — external · write` band.
- Node taxonomy: sync = code card (doc icon, mono name, mode, status dot + word); data store / external app = connector **brand-initial badges** (`_BADGES` in `drt/docs/html.py`, decision recorded in `docs/design/drt-docs-icon-strategy.html`).
- Visual target: `docs/design/drt-docs-lineage-mock.html` (committed, self-contained). The per-sync **ego-graph** on sync pages already ships this visual language on `design/docs-spike` — the project DAG generalizes it.

**Interactivity** stays a separable progressive-enhancement layer (hover-highlight up/downstream, optional pan-zoom) that never touches build-time layout — out of scope here, tracked on the design tracker.

@Muawiya-contact has been scoping/spiking the layout side (thank you 🙏) — this issue is the natural home for that spike when it lands. The crossing-minimization A/B and the back-edge lane geometry are the parts we planned to pair on.

Related: #499 (epic), #510 (P3), #697 (determinism), ADR #500.

Muawiya-contact:
[...]

Issue #697, opened 2026-07-03
chore(docs): byte-identical output — move generated_at out of every page footer


masukai:
## Summary

Every generated docs page renders `generated_at` in its footer, so regenerating docs for an **unchanged project** produces a diff on every page. That undermines the docs site's diffability story — the same property we're designing the static lineage SVG around ("fixed iterations → byte-identical output, clean git diffs", see #677 discussion) — and makes docs-in-git workflows (review the docs diff alongside the config diff) noisy to useless.

## Proposal

Byte-identical output for identical input:

- `generated_at` lives in **`manifest.json` only** (one predictable diff line) — page footers keep `drt <version> · schema v<n>` and drop the timestamp.
- Audit the builder for any other nondeterminism (iteration order of syncs/sources/destinations/edges, JSON key order) and pin it with explicit sorts. Most paths already iterate in manifest order; this makes it a stated contract.
- Optional escape hatch: `--stamp` to re-enable in-page timestamps for teams that want them.

Acceptance: `drt docs generate && drt docs generate` → second run writes zero changed bytes outside `manifest.json`'s `generated_at`, and unchanged projects produce a one-line git diff.

Related: #510, #499, #677.

masukai:
Verified this is **already done on `main`** — closing with evidence rather than opening a no-op PR.

The acceptance bar was: `drt docs generate && drt docs generate` → zero changed bytes outside `manifest.json`'s `generated_at`. Ran exactly that against `examples/docs-demo` (10 syncs, 30 files):

```
$ diff -rq gen1 gen2
Files gen1/manifest.json and gen2/manifest.json differ   # the only one
$ diff gen1/manifest.json gen2/manifest.json
<   "generated_at": "2026-07-16T04:34:42Z",
>   "generated_at": "2026-07-16T04:34:43Z",             # the only line
```

How it got done piecemeal:
- **Page-side**: #704 dropped the timestamp from every footer (`html.py` now carries a `{# no timestamp here … (#697) #}` comment; `generated_at` is passed into the template context but no template references it).
- **Regression test**: #753 added `test_regeneration_is_byte_identical`, which renders the whole site twice and byte-compares every file.
- **Determinism audit**: sync iteration is sorted-glob order, nav/tags are explicitly sorted, and the 30-file empirical diff above confirms no other nondeterminism leaks.

The proposal's optional `--stamp` escape hatch (re-enable in-page timestamps) was never asked for by anyone — skipping it until someone wants it, per YAGNI.

[...]

PR #713, opened 2026-07-06
feat(docs): static DAG layout engine — deterministic ranks + ordering + back-edge lanes (toward #701)

Muawiya-contact:
Draft — the layout-engine core for #701, the design-independent part (safe ahead of the #704 visual work). Opening as a draft for the pairing on crossing-min + back-edge geometry we planned.

## What's here
`drt/docs/layout.py` — `compute_layout(manifest, strategy=…, config=…)` → node positions + edge paths. Pure geometry: no SVG, styling, or JS (that emission layer sits on top later and consumes this).

- **Ranks are free** — rank = node kind (source → sync → destination).
- **Ordering** — barycenter *and* median heuristics, selectable via `strategy=`, so we can A/B crossing counts. Fixed iterations + name tiebreak → **byte-identical output** (ties into #697).
- **Lookup back-edges** — lane assignment done: each gets its own lane + a distinct top-edge port on the consumer sync, so N lookups never stack into one curve.
- Returns `Layout(nodes, edges, width, height, crossings, strategy)`.

## Open for pairing (the geometry you flagged)
Risers currently go straight up at the producer's x, which can cross same-column nodes when a producer sits directly below its consumer. The lane **assignment** is solved; the riser **routing** around the graph is the open refinement. Also happy to run the barycenter-vs-median A/B on real project shapes so we pick with data.

## Tests
10 in `tests/unit/test_docs_layout.py`: ranks, determinism, crossing-uncross, both strategies, distinct lanes/ports, empty/single-node, config scaling. Local: ruff clean, `mypy drt/docs/layout.py` clean.

## Not in scope
No #704 visual work — no tokens, badges, or ownership zones. This is the layout core only.


Muawiya-contact:
@masukai the layout-engine core from the #677/#701 scoping — opened as a draft since this is the part we planned to pair on.

Design-independent on purpose (safe ahead of your #704 work): `compute_layout(manifest, strategy=…)` → node positions + edge paths, pure geometry, no SVG/tokens/badges.

Working:
- ranks = node kind; barycenter **and** median both in + selectable, so we can A/B crossing counts
- fixed iterations + name tiebreak → byte-identical output (ties into #697)
- lookup back-edges: lane assignment done — each gets its own lane + distinct top-edge port, so N lookups never stack
- 10 tests (ranks, determinism, crossing-uncross, A/B, distinct lanes/ports, empty/single-node)

The bit I'd genuinely like to pair on — exactly the geometry you flagged: risers currently go straight up at the producer's x, so a producer sitting directly below its consumer can cross same-column nodes. Lane *assignment* is solved; riser *routing* around the graph is the open refinement. And happy to run the barycenter-vs-median A/B on a few real project shapes so we pick the heuristic with data rather than vibes.

Left it as the layout core only — the emission layer (your tokens/zones/badges) consumes this once #704 lands. Shout with how you'd like to sequence.

Muawiya-contact:
Both fair — 

1. **`_RANK` unused** — correct, it's dead (ranks are derived in `_ranks()` from the sorted node lists). Removing it. (ruff doesn't flag it since it's module-level, but it's still dead code.)

2. **O(n²) inversion count** — right on the complexity. It runs per rank-pair per sweep, so ~8× per layout. At drt's DAG scale (dozens of nodes, low hundreds at the extreme) it's negligible — but the mergesort version is identical in result and a clean swap, so I'm happy to do it as future-proofing. @masukai — preference? I'd lean "leave it simple until a real project makes it hurt," but no strong view; will swap if you'd rather have the O(n log n) in from the start.

masukai:
Great core — CI green and the determinism story ties into #697 nicely.

1. **`_RANK`** — agreed, remove it. Dead once ranks derive from `_ranks()`.

2. **O(n²) inversion count** — let's keep it simple. The docs DAG is the reverse-ETL last-mile subset (bounded small — dozens, low-hundreds worst case) and this is offline docs-gen, not a sync hot path, so the O(n²) never bites. The real constraint here is byte-identical determinism, and the plain double-loop is trivially correct + trivially deterministic — I'd rather not add mergesort tie-handling surface against that guarantee. One ask: drop a one-line comment marking the O(n²) as intentional (bounded-n + determinism-first, swap only if a real project hurts) so it reads as a decision, not an oversight.

3. **Riser routing** (the bit to pair on) — since lane *assignment* is already solved, complete it by routing the riser through its assigned lane's reserved gutter instead of straight up at the producer's x: jog into the lane immediately, travel vertically in the inter-column gutter (clear of every node column), then into the consumer's top port. That keeps risers out of node columns by construction.

4. **barycenter vs median A/B** — yes, run it on a few real project shapes and let's pick with data. My prior is median tends to win on sparse DAGs like ours, but happy to be shown otherwise. Glad to pair on 3 whenever.

Muawiya-contact:
All three done:

- **`_RANK`** — removed.
- **O(n²)** — kept, with a comment marking it a decision (bounded-n + determinism-first, swap only if a real project hurts). Agree the double loop's trivial determinism is worth more than shaving complexity that never bites here.
- **Riser routing** — done your way: risers now jog into the reserved inter-column gutter and rise vertically at a distinct x per lane, then into the consumer's top port. Clear of node columns by construction. Added a test asserting no riser's vertical segment lands inside any node's x-span.

**barycenter vs median A/B** — ran it, and the data's interesting: on realistic drt shapes they **tie on crossings, every time**. Structural reason — each sync is degree-1 on each side (one source, one dest), so a sync's barycenter == median (single value); they can only diverge on source/dest ordering under heavy fan. Forcing that (50 synthetic heavy-fan graphs) they diverged ~16% of the time, and it was **mixed** — median won slightly more often but barycenter had the biggest single win. So for our shapes it's a wash.

Given that, one option instead of picking a side: since both are cheap and deterministic, default to computing **both and keeping the lower-crossing result** (deterministic tiebreak by strategy name) — best of both, no heuristic religion. Happy to wire that, or just keep it selectable and default to one. Your call — glad to pair on it whenever.

masukai:
All three landed exactly right — and the A/B result is genuinely the useful kind of finding. The degree-1 ⇒ `barycenter == median` observation is spot on (each sync is single-value on both sides, so the two heuristics can only diverge under source/dest fan), and it's worth a one-line comment in the code so nobody re-litigates the choice later.

**On the knob: default to `median`, keep both selectable, and let's hold off on the compute-both-keep-lower-crossing machinery for now.** Same reasoning as the O(n²) call — on real drt shapes they *tie*, so "run both and pick" earns nothing on the graphs we actually render, while adding surface (a second layout pass + a cross-strategy comparison) against the byte-identical-determinism guarantee we're protecting. If a real project ever surfaces the heavy-fan divergence, we revisit with that project as the test case. Please capture the A/B numbers (tie on real shapes; ~16% mixed divergence on synthetic heavy-fan) in the commit/PR so the decision is on record.

Riser routing is the clean version — jogging into the reserved gutter with a distinct x per lane, plus the test asserting no vertical segment lands in any node's x-span, is exactly what I was after.

With `median` as the default + the O(n²) decision comment, I think this is **good to merge**. Flip the default if it isn't already median, make sure the degree-1 finding is captured as a comment, and I'll approve. Really nice work on this — the layout core is in great shape for the emission layer to consume. 🚀

Muawiya-contact:
@masukai All three in:
- **Default → median**, both still selectable.
- **Degree-1 finding** captured as a comment by `_STRATEGIES` — why bary==median on real drt shapes (sync is single-value on both sides), the ~16% mixed heavy-fan divergence, and why we skip both-and-pick.
- **A/B numbers on record** in the commit body (tie on real shapes; ~16% mixed on synthetic heavy-fan).

Agree on holding off the compute-both machinery — same logic as the O(n²) call, and if a real project ever surfaces the divergence it becomes the test case. Thanks for the thorough pairing on this one 🙏

masukai (approved):
All landed exactly as discussed — default `median` with both strategies still selectable, the degree-1 ⇒ bary==median finding captured on `_STRATEGIES` (with the ~16% heavy-fan divergence and why we skip both-and-pick), and the A/B numbers on record in the commit. Determinism-first, no speculative machinery. CI green. This is the clean layout core the emission layer (#704) can consume — great pairing on it. 🚀

PR #731, opened 2026-07-08
refactor(destinations): make columns required in Snowflake/Databricks _insert_rows, drop the values() fallback (#699)

### LINKED ISSUE #699: refactor(destinations): make columns required in Snowflake/Databricks _insert_rows, drop the values() fallback
Follow-up from the #680 review (agreed by @yodakanohoshi + @Muawiya-contact to defer).

Both `snowflake.py` and `databricks.py` `_insert_rows` keep an optional `columns=None` path that falls back to `list(row.values())`. Since Layer 3 (#317), the **live** path is always `_bind_row(row, columns, json_cols)` — which orders values by `columns` via `row.get(c)` per column, so it stays correct even when a source yields rows with varying key order/set (REST especially, where `list(row.values())` would silently misalign). That makes the `columns=None` branch **dead code** — it's the codecov patch-miss yodakanohoshi flagged on #680.

**Proposal:** make `columns` (and `json_cols`) required parameters on `_insert_rows` in both destinations and delete the `list(row.values())` fallback, symmetrically. Pure cleanup — no behaviour change on the live path; removes an untested branch and the ordering foot-gun.

- `drt/destinations/snowflake.py` — `_insert_rows` signature + all call sites
- `drt/destinations/databricks.py` — same
- adjust/trim any test that exercised the fallback

Small, mechanical; good candidate for @Muawiya-contact who offered to take it.

Muawiya-contact:
Closes #699. Follow-up from the #680 review (deferred cleanup, agreed by @yodakanohoshi + @masukai).

`_insert_rows` in both `snowflake.py` and `databricks.py` now takes `columns` + `json_cols` as **required** params; the `list(row.values())` fallback is removed. The live path was already always `_bind_row(row, columns, json_cols)` — which orders by `columns` via `row.get`, so it's correct even when a source yields rows with varying key order/set (REST). The fallback was dead + untested (the codecov patch-miss from #680), and `list(row.values())` was the ordering foot-gun. All four call sites already passed both args, so this is signature + body only — no behaviour change on the live path.

Added `test_bind_row_orders_by_columns_regardless_of_row_key_order` to both destinations: reordered keys and a missing key bind by `columns` (missing → `None`), pinning the order-safe path as the only path.

ruff + mypy clean; 113 destination tests pass.

- [ ] CHANGELOG — N/A `[skip changelog]` (internal cleanup, no user-facing change)

### DIFF
--- a/drt/destinations/databricks.py
+++ b/drt/destinations/databricks.py
@@ -437,23 +437,21 @@ class DatabricksDestination:
         records: list[dict[str, Any]],
         sync_options: SyncOptions,
         result: SyncResult,
-        columns: list[str] | None = None,
-        json_cols: list[str] | None = None,
+        columns: list[str],
+        json_cols: list[str],
     ) -> None:
         """Execute a parameterised INSERT per row, honouring ``on_error``.
 
-        When ``columns`` is given, values are ordered to it and json columns
-        (``json_cols``) are ``json.dumps``'d so the ``from_json`` / ``parse_json``
-        bind receives a JSON string (Layer 3, #317).
+        Values are ordered to ``columns`` and json columns (``json_cols``) are
+        ``json.dumps``'d so the ``from_json`` / ``parse_json`` bind receives a
+        JSON string (Layer 3, #317). Ordering by ``columns`` (via ``row.get`` in
+        ``_bind_row``) keeps binds correct even when a source yields rows with a
+        varying key order/set — so ``columns``/``json_cols`` are required, not an
+        optional ``list(row.values())`` fallback (#699).
         """
         for i, row in enumerate(records):
             try:
-                values = (
-                    _bind_row(row, columns, json_cols or [])
-                    if columns is not None
-                    else list(row.values())
-                )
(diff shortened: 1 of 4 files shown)


Muawiya-contact:
@masukai #699 ready — columns/json_cols required, fallback gone, order-safe `_bind_row` test added to both destinations. Copilot clean, CI green, `[skip changelog]` since it's internal cleanup.

masukai (approved):
LGTM — reviewed the diff and verified the core claim: all four `_insert_rows` call sites (2 each in snowflake.py / databricks.py, the direct and swap paths) already pass `columns` + `json_cols` as real lists, so making them required just drops genuinely dead code — the `list(row.values())` fallback was never reached on the live path (the #680 codecov patch-miss). `json_cols or []` → `json_cols` is safe since the callers always pass a list, so no behaviour change on the live path. The new `_bind_row` ordering tests pin exactly why the fallback was the foot-gun (varying key order / missing key → correct alignment via `row.get`). Clean. Thanks @Muawiya-contact 🙏

PR #732, opened 2026-07-09
feat(destinations): migrate Databricks to native ? paramstyle (#707)

### LINKED ISSUE #707: Databricks destination: migrate from inline params to native paramstyle
## Background

#706 fixed the Databricks destination failing on databricks-sql-connector >=3.0 (pyformat `%s` binds vs. the connector's native-paramstyle default) by opting back into client-side inline rendering with `use_inline_params="silent"`.

That keeps us on a path the connector explicitly deprecates — upstream's direction is native parameters (`?` / named markers, server-side binding), and `use_inline_params` will eventually be removed. We should migrate before that happens. Native binding also removes the escaping-based injection surface that inline rendering carries by design.

## Scope (surveyed as of #706)

**Product code — 2 files:**

- `drt/destinations/databricks.py` — all bind sites go `%s` → `?`:
  - `_value_clause`: scalar `%s`, `from_json(%s, '<ddl>')`, `parse_json(%s)` (+ docstrings)
  - `_finalize_mirror`: the `NOT IN` placeholder assembly (single-key and composite-key branches)
  - drop `use_inline_params="silent"` from `_connect`
- `drt/destinations/schema.py` — the two Databricks introspection queries (`lower(%s)` in `_describe_databricks` / `describe_databricks_ddls`). **Leave the Postgres / MySQL / Snowflake queries untouched** — those drivers are natively pyformat; the file will intentionally carry two paramstyles, worth a comment.
- `drt/sources/databricks.py` — **out of scope**: it executes model SQL verbatim with no `parameters` argument, so neither paramstyle applies.

**Tests — 3 files:**

- `tests/unit/test_databricks_schema_aware.py` — ~15 assertions on `%s`-bearing SQL strings
- `tests/unit/test_schema.py` — a few Databricks introspection query assertions
- `tests/unit/test_databricks_destination.py` — the connect-kwargs contract test pins `use_inline_params` (added in #706); update it to assert the flag is *gone*
- `tests/integration/dwh/test_databricks_smoke.py` — docstring mentions only (cosmetic)

## ⚠️ Not a mechanical rename: mirror-mode parameter limit

`_finalize_mirror` builds `DELETE ... WHERE key NOT IN (?, ?, ...)` with **one parameter per observed key**. Native mode enforces a per-statement parameter limit (inline mode had none), so a mirror sync with more keys than the limit would start failing only after this migration. The DELETE needs chunking (`AND`-composed `NOT IN` batches) or a rework (e.g. anti-join against a staging table — likely nicer anyway given the destination already stages for MERGE). Confirm the exact limit against the connector/server docs during implementation.

## Verification

- Unit suite (assertions updated)
- Live re-run of the #674 Databricks smoke leg (insert / replace-swap / complex-type / connection)
- A live mirror-mode run with enough keys to cross the chunking boundary
- Confirm the per-connection deprecation warning is gone

Refs: #706 (interim fix) · #672 (surfaced the bug) · #317 (`from_json` / `parse_json` bind sites) · #340 (mirror leg)

masukai:
Closes #707 (pending live smoke — see ⚠️ below).

Retires the interim `use_inline_params="silent"` from #706 and moves the Databricks destination onto the connector's default **native `?` paramstyle** (server-side binding). This drops the per-connection deprecation warning and the escaping-based injection surface that client-side inline rendering carries by design.

## What

- **`databricks.py`** — `%s` → `?` in `_value_clause` (scalar / `from_json` / `parse_json`), and `use_inline_params` dropped from `_connect`.
- **`schema.py`** — the two Databricks introspection queries go `?`. **Postgres / MySQL / Snowflake stay pyformat** (native to their drivers) — the file now carries two paramstyles on purpose, called out in a comment.
- **`_finalize_mirror` reworked to a staging anti-join.** The old `DELETE … WHERE key NOT IN (?, ?, …)` bound one parameter per observed key, which native mode's per-statement parameter limit would cap — so a large mirror would start failing *only* after this migration. Instead, observed keys are staged into a scratch Delta table (reusing the MERGE path's staging approach) and removed via `DELETE … WHERE key NOT IN (SELECT key FROM staging)` — **no key parameters bound in the DELETE**, so it scales to any number of keys.

## Why anti-join over chunking

The issue floated AND-composed `NOT IN` chunking *or* an anti-join. Chunking only helps if the limit is per-IN-list; if it's per-statement-total, AND-composing doesn't reduce the parameter count. The staging anti-join is **limit-immune either way** and reuses machinery the destination already has for MERGE.

## Tests

- ~15 `%s`→`?` assertions in `test_databricks_schema_aware`
- the Databricks introspection assertion in `test_schema`
- the connect-kwargs contract test now asserts `use_inline_params` is **gone** (it pinned it *present* in #706)
- the mirror-finalize tests assert the full staging + anti-join sequence (CREATE keys table → per-key INSERT → `NOT IN (SELECT …)` DELETE with **no bound key params** → DROP)

## ⚠️ Needs a live Databricks smoke run before merge

The unit suite is **mock-based**, like the rest of this destination — I can't exercise a real workspace here. Before merging, please run:
- the #674 Databricks smoke leg (insert / replace-swap / complex-type / connection)
- a **live mirror run with enough keys to cross the old inline parameter limit** (validates the anti-join path end to end)
- confirm the per-connection deprecation warning is gone

## Verified (offline)

- Full unit suite: **1838 passed, 5 skipped**
- `ruff` + **mypy across the whole package** clean · `make check-drift` → 0 drift
- `databricks.py` / `schema.py` changed lines covered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Databricks destination migrated to native `?` paramstyle** ([#707](https://github.com/drt-hub/drt/issues/707)): removes the interim `use_inline_params="silent"` opt-in from #706 — all bind sites (`_value_clause` scalar / `from_json` / `parse_json`, and the two schema-introspection queries) now use the connector's default native `?` markers with server-side binding, dropping the per-connection deprecation warning and the escaping-based injection surface inline rendering carried. The `sync.mode: mirror` end-of-sync DELETE is reworked from an inline `NOT IN (?, ?, …)` (one parameter per observed key, which native mode's per-statement parameter limit would cap) to a **staging anti-join**: observed keys are staged into a scratch Delta table and removed via `DELETE … WHERE key NOT IN (SELECT key FROM staging)`, binding no key parameters so it scales to any number of keys. Postgres/MySQL/Snowflake stay pyformat (native to their drivers). **Needs a live Databricks smoke run (incl. a mirror large enough to cross the old parameter limit) before release** — the unit suite is mock-based like the rest of this destination.
+
 - **README modernised; Japanese i18n retired** ([#712](https://github.com/drt-hub/drt/pull/712)): README.ja.md and its translation upkeep are removed; the English README is the single maintained entry point.
 
 ### Fixed
--- a/drt/destinations/databricks.py
+++ b/drt/destinations/databricks.py
@@ -70,19 +70,20 @@ def _value_clause(
     Layer 3 (#317): json-category columns are wrapped so Databricks reconstructs
     the complex value from a JSON string —
 
-    - STRUCT / ARRAY / MAP -> ``from_json(%s, '<ddl>')`` (DDL from ``information_schema``)
-    - VARIANT -> ``parse_json(%s)`` (no DDL form)
+    - STRUCT / ARRAY / MAP -> ``from_json(?, '<ddl>')`` (DDL from ``information_schema``)
+    - VARIANT -> ``parse_json(?)`` (no DDL form)
 
-    Scalars pass straight through (``%s``). Databricks — like Snowflake — won't
+    Scalars pass straight through (``?``). Databricks — like Snowflake — won't
     accept these functions in a ``VALUES`` clause, so any wrapping switches the
-    statement to the ``SELECT`` form (``INSERT INTO t (cols) SELECT %s,
-    from_json(%s, '<ddl>')``). With no json columns the clause is the unchanged
-    ``VALUES (%s, %s, ...)`` — byte-identical to pre-#317 behaviour.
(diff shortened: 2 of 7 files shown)


masukai:
**⚠️ resolved — live smoke run done, green.** Ran `dwh-smoke` against **this branch** (`workflow_dispatch --ref feat/databricks-native-paramstyle-707`) on the real Databricks workspace ([run 29060869631](https://github.com/drt-hub/drt/actions/runs/29060869631)):

```
test_databricks_insert_roundtrip           PASSED
test_databricks_replace_swap_roundtrip     PASSED
test_databricks_complex_type_serialization PASSED   # from_json/parse_json under native ?
test_databricks_connection                 PASSED
4 passed in 45.54s
```

For contrast, the same suite also passed on `main` (inline params) minutes earlier — so this is an apples-to-apples live validation of the native-paramstyle migration, on the same workspace/warehouse. The exact failure class that #706 caught (mock-green / live-red on param binding) is what this run rules out. Merge-ready from the validation side.

Issue #710, opened 2026-07-05
chore(infra): cost monitoring for the DWH smoke accounts — budget caps + spend digest


masukai:
## Why

The DWH smoke accounts (Snowflake / BigQuery / Databricks — epic #654, provisioned via #709) run nightly and incur **real, if tiny, cost**. We want a minimal but reliable way to know we're not silently spending. This is **cost/spend monitoring only** — smoke pass/fail is already covered by the GitHub Actions failure email, so **no smoke-result notification is needed** here.

Goal: keep every environment in the "daily, ultra-light, ~free" range, with a safety net that (a) *prevents* runaway spend even if no one is watching, and (b) *surfaces* month-to-date spend so it's visible at a glance.

### Layer 1 — prevention (hard caps) — the real safety net, minimum required
Stops compute / rejects spend at a threshold regardless of whether anyone reads a notification.

- [ ] **Snowflake** — Resource Monitor `SUSPEND` at 100% of a small credit quota. *(already in `tests/integration/dwh/provisioning/snowflake.sql`)*
- [ ] **BigQuery** — Cloud Billing budget (e.g. $5/mo) **+ a custom query-bytes quota** to physically cap runaway scans.
- [ ] **Databricks** — budget / usage limit + minimum auto-stop on the SQL warehouse (if/when the Databricks leg is provisioned).

### Layer 2 — visibility (notification)
- [ ] **(a) Native budget email alerts** per cloud (GCP budget / Snowflake resource-monitor / Databricks budget) — zero-code backstop.
- [ ] **(b) Unified cost digest to Discord** — a scheduled GitHub Actions cron (daily or weekly) that reads each DWH's own usage views and posts one line, e.g. `MTD: SF $0.4 / BQ $0.0 / DBX $2.1 — under budget ✅`, to a private ops channel (a `DISCORD_COST_WEBHOOK` secret). **Separate from the community-Discord decision (#378)** — this is an internal ops channel.
  - Snowflake: `SNOWFLAKE.ACCOUNT_USAGE.WAREHOUSE_METERING_HISTORY` (credits)
  - BigQuery: `INFORMATION_SCHEMA.JOBS` bytes-billed → est. $, or a billing export for exact $
  - Databricks: `system.billing.usage` (DBU)

## Design notes

- **Cost queries need broader read than the least-privilege smoke role** → provision a **separate read-only "cost-watcher" identity / grant** per DWH; don't widen the smoke user.
[...]

PR #745, opened 2026-07-10
feat(snowflake): key-pair authentication for destination + source (closes #737)

### LINKED ISSUE #737: feat(snowflake): key-pair authentication for destination + source (unblock TYPE=SERVICE users)
## Why

Snowflake is deprecating password-only sign-ins for human users, and the transitional `TYPE = LEGACY_SERVICE` escape hatch is **already unavailable on newly created accounts** (`SQL compilation error: invalid value 'LEGACY_SERVICE' for property 'TYPE'` — hit live on the org smoke account, 2026-07-10). The proper programmatic identity is `TYPE = SERVICE`, but SERVICE users **cannot authenticate with a password** — they need key-pair (or OAuth).

drt's Snowflake destination (and source profile) currently support **password auth only** (`password_env`). That means:
- our own nightly smoke runs on a password-authenticated user that Snowflake will eventually block, and
- any drt user in a security-conscious org (where SERVICE + key-pair is mandated) can't use the Snowflake connector at all.

## Proposal

- Add key-pair auth to `SnowflakeDestinationConfig` (and the Snowflake profile): `private_key_env` (PEM contents) or `private_key_path_env`, optional `private_key_passphrase_env`. `snowflake-connector-python` accepts `private_key=<DER bytes>` — decode PEM → DER at connect time.
- Precedence: key-pair wins over password when both are set; clear validation error when neither.
- Docs: `docs/connectors/snowflake.md` auth section + provisioning runbook (`ALTER USER … SET TYPE = SERVICE; ALTER USER … SET RSA_PUBLIC_KEY = '…';`).
- Migrate the smoke account: generate an org key pair, switch `DRT_SMOKE_USER` to `TYPE = SERVICE` + `RSA_PUBLIC_KEY`, replace `SMOKE_SNOWFLAKE_PASSWORD` with a private-key secret, update `tests/integration/dwh/` + `dwh-smoke.yml` env mapping.

## Notes

- Until this lands, the smoke user stays a password-authenticated default-type user — works today, but it's the thing Snowflake's "upcoming deprecation of password-only sign-ins" banner is warning about.
- BigQuery already uses keyfile auth; Databricks uses PAT — Snowflake is the only password-only leg left.

masukai:
Closes #737. **This is the fix for the red nightly Snowflake smoke leg** — the account now enforces MFA on password sign-ins (`250001: Multi-factor authentication is required`, 2 consecutive nightly failures), and `TYPE = LEGACY_SERVICE` is unavailable on new accounts, so `TYPE = SERVICE` + RSA key-pair is the only durable path.

## What

- **`private_key_env`** (PEM/PKCS#8 *contents*) + **`private_key_passphrase_env`** on `SnowflakeDestinationConfig` **and** the `snowflake` source profile. Key wins over password; `password_env` becomes optional with a one-of-the-two validator — **existing password configs work unchanged**.
- Shared `load_snowflake_private_key` (PEM→DER) in `config/credentials.py` — `cryptography` is already a snowflake-connector dependency, imported lazily.
- **Smoke harness**: dual-auth gate (`DRT_SMOKE_SNOWFLAKE_PRIVATE_KEY` preferred, password fallback), all four tests + the raw verification connects switched; `dwh-smoke.yml` maps the new secret.
- **Docs**: `docs/connectors/snowflake.md` auth section rewritten (retires the long-standing "no key-pair support" caveat with the openssl + `ALTER USER` recipe); provisioning runbook gains the SERVICE-user section.
- vscode schema regen + extension `0.1.5`; CHANGELOG entry.

## Tests

5 new: config validation (password-only / key-only / neither→error) + destination `_connect` kwargs (real RSA key generated in-test → DER bytes passed, password absent; password fallback path). Full suite **1843 passed**, ruff + mypy clean.

## After merge (same flow that fixed the cost watcher today)

1. Generate the smoke key pair; register `SMOKE_SNOWFLAKE_PRIVATE_KEY`
2. `ALTER USER DRT_SMOKE_USER SET TYPE = SERVICE, RSA_PUBLIC_KEY='…'; UNSET PASSWORD`
3. Dispatch `dwh-smoke` → Snowflake leg back to green

The same battle-tested sequence just fixed `DRT_COST_USER` (#710) — verified locally against the live account before secret rotation.


masukai:
**Rotation complete — Snowflake leg recovered.** ✅

Post-merge sequence: generated the smoke key pair → `DRT_SMOKE_USER` converted to `TYPE = SERVICE` + `RSA_PUBLIC_KEY` (password unset) → key verified against the live account (auth + CREATE/DROP privileges) **before** rotating `SMOKE_SNOWFLAKE_PRIVATE_KEY` → old password secret deleted → dispatch: **Snowflake 5/5 passed in 46.76s** (incl. the mirror test after #746's one-liner), Databricks ✓, BigQuery ✓ — [all three warehouses green](https://github.com/drt-hub/drt/actions/runs/29073367727).

With both programmatic users (`DRT_SMOKE_USER`, `DRT_COST_USER`) now `TYPE = SERVICE` + key-pair and the human account MFA-enrolled, the account's "password-only sign-ins" deprecation warning is fully addressed — zero at-risk users.

Issue #702, opened 2026-07-04
docs(ui): post-#677 design & lineage follow-up tracker


masukai:
## Why this exists

Design pass 1 for the docs UI happened in the #677 thread, and #677 is about to merge and close. This tracker carries everything **agreed but not yet on `main`**, so nothing drifts when that thread goes quiet. Design artifacts live in-repo: `docs/design/drt-docs-lineage-mock.html` (lineage visual language incl. DLQ/drift badges), `docs/design/drt-docs-icon-strategy.html` (icon decision B), both on `design/docs-spike` together with the implemented spike.

## Follow-up PRs (implemented on `design/docs-spike`) — all three land via **#704**

- [x] #704 — design tokens consolidated into one `:root` block + deterministic page footers (page-side of #697)
- [x] #704 — sync-page tabs (YAML / Lineage / State), per-sync **ego lineage** (static SVG), connector **brand badges**, KPI cards, breadcrumb
- [x] #704 — Tags sidebar group + `tag/<slug>.html` pages; source/destination page polish (badges, read-by/fed-by, Last-status column)

## Agreed, not yet implemented

- [ ] **Project-DAG static layout engine** (replaces runtime Mermaid) → #701 (layout core landed via #713; emission in progress)
- [ ] **DLQ badge** (`n → DLQ` on failed/partial syncs) — needs run/DLQ data in the manifest (#698)
- [ ] **Drift badge** (`⚠ drift` on destinations) — needs a persisted drift-audit artifact (storage TBD)
- [ ] **Column-level lineage (L2)** — expose `field_mappings` / `mask` in manifest schema v2 (write side is declared, no SQL parsing); source-side table extraction from model SQL (needs SQL parsing — later)
- [ ] **Interactivity layer** (separable JS, degrades to links on `file://`): expand-in-place ego view, hover-highlight up/downstream, optional pan-zoom, manual theme toggle
- [ ] **Empty/error state system** across pages → #751; per-sync raw YAML view (schema v1 carries no model SQL)
- [ ] **Scale mode** for large projects: compact card density, tag/type group collapse; search-results UX
- [ ] Responsive/mobile treatment for the DAG page (general mobile pass landed via #752; DAG page follows #701)
- [ ] Finishing pass: favicon/meta, a11y (skip link, focus rings, tab aria), print styles, search empty state → #753 (incl. a byte-identical regression test enforcing #697)
[...]

Issue #696, opened 2026-07-03
feat(docs): default-safe connection labels — docs output currently publishes endpoints, phone numbers, emails


masukai:
## Summary

`drt docs generate` embeds destination connection details verbatim — in `manifest.json`, in every page's sidebar, and on destination pages. The labels come from each config's `describe()`, and several of them expose more than a shareable catalog should:

| connector | `describe()` output | exposure |
|---|---|---|
| rest_api | `rest_api (https://internal-host/path)` | **full endpoint URL** |
| twilio | `twilio (+81…)` | **phone number** |
| sendgrid | `sendgrid (ops@company.com)` | **email address** |
| s3 / opensearch etc. | `… (endpoint, region)` | infra endpoints |
| github | `github (owner/repo)` | private repo names |
| snowflake / bigquery / databricks | full `db.schema.table` path | usually fine, but full paths |

The moment a team hosts `target/docs/` on static hosting (the intended use), this becomes a published inventory of internal endpoints and contact identities. drt already treats data carefully (`mask`, PII hashing — #427); the docs output should hold the same line. Note the precedent already in the codebase: webhook-style destinations deliberately describe as just `(webhook)`.

## Proposal

Default-safe labels in docs output, opt-in for full detail:

- Docs builder uses a **redacted label** by default: keep the *object* identity (table, channel, sheet, index, object type — that's the useful catalog info), drop network locations and personal identifiers (URL → path-less type label, phone/email → masked, endpoint/region → dropped).
- `drt docs generate --full-labels` (name bikesheddable) restores today's verbatim `describe()` for trusted/internal contexts.
- Applies to both the HTML pages and the emitted `manifest.json` (it ships inside the site output).

Implementation seam is open to discussion — a `describe_public()` alongside `describe()`, a structured `(type, object, endpoint)` triple the builder can filter, or per-type redaction rules in the builder. Slight preference for the structured option so the builder never string-parses.

Related: #510 (P3 site), #499 (epic). Run-history exposure (#698) has the same concern for `errors[]` strings.

PR #751, opened 2026-07-10
design(docs): empty/error state system — zero-sync hero + last_error on the sync KPI card

masukai:
Next design follow-up off the #702 tracker: a small empty/error-state pass.

- **Zero-sync projects get a real empty state** on Overview and DAG — title + the three CLI steps to get going (`drt init` → `drt run` → `drt docs generate`) instead of a wall of zeros. The DAG page keeps the tested `No syncs found` copy.
- **Failed/partial syncs explain themselves at a glance**: the Last-run KPI card on the sync page now shows `last_error` (truncated to one line, full text in the `title` attribute and the State tab).
- Two tiny additions beside the existing `.empty` (`.empty--hero`, `.kpi__error`); tokens untouched.

| | |
|---|---|
| Empty overview | the `/empty` preview in a local render — `render_html(empty_manifest, …)` |
| Error surfacing | `sync/accounts-to-pg.html` in the demo fixture (failed + `psycopg2.OperationalError…`) |

Docs suite green (50 passed / 2 skipped).

@Muawiya-contact same deal as before — no rush, a glance whenever suits. This one deliberately doesn't touch `layout.py` or anything your #701 emission work sits on. ☕

Related: #702 (tracker), #704 (previous batch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #752, opened 2026-07-10
design(docs): YAML tab shows the file as written — code chrome (linenos/copy/collapse) + dark palette + mobile pass

masukai:
Design follow-ups off the #702 tracker, three themes in reviewable commits:

## 1. The YAML tab shows the sync as written on disk

`collect_sync_yaml_texts()` (builder) best-effort maps sync name → (relative path, raw text) from `syncs/*.yml`; the CLI hands it to `render_html`, which prefers the file over the manifest-derived view. **This is how the YAML tab gets `model` SQL without touching manifest schema v1** — display-only, nothing new enters `manifest.json`. Syncs whose file isn't found keep the manifest view plus a note.

## 2. Code chrome

GitHub-style block: header bar (`syncs/orders_to_pg.yml · YAML · 41 lines`), **line numbers** (table mode — selecting code doesn't grab them), a **Copy** button (clipboard, JS-only), and **collapse** for files over 32 lines (fade + "Show all N lines"; no-JS renders everything). Dark mode gets a proper Pygments palette (`native` behind `prefers-color-scheme`) — the block used to stay light on dark pages.

## 3. Mobile pass

The <860px breakpoint used to hide BOTH sidebar and topnav — no navigation at all. Now: nav stays, search goes full-width, cards stack, wide tables scroll in place.

## Tests

53 docs tests green — raw-vs-fallback, hostile-YAML escaping, collector best-effort skips, path + linenos assertions.

@Muawiya-contact no rush as ever ☕ — one thing worth your eye: `collect_sync_yaml_texts` reads `syncs/*.yml` directly; if the hardening instincts from #711 see a sharp edge there (symlinks, weird encodings), I'd rather hear it now.

Related: #702 (tracker), #696 (note: raw YAML shows destination config as written — the redaction policy discussion applies here too).

### DIFF
--- a/drt/cli/commands/docs.py
+++ b/drt/cli/commands/docs.py
@@ -61,9 +61,12 @@ def docs_generate(
             print_error("HTML docs generation requires: pip install drt-core[docs]")
             raise typer.Exit(1)
 
+        from drt.docs.builder import collect_sync_yaml_texts
+
         manifest = build_manifest(Path("."), include_state=include_state)
+        sync_yaml_texts = collect_sync_yaml_texts(Path("."))
         try:
-            written = render_html(manifest, output)
+            written = render_html(manifest, output, sync_yaml_texts=sync_yaml_texts)
         except ValueError as e:
             print_error(str(e))
             raise typer.Exit(1) from e
--- a/drt/docs/_html_assets.py
+++ b/drt/docs/_html_assets.py
@@ -208,10 +208,15 @@ td.right, th.right { text-align:right; }
 .group a .count { float:right; }
 
 @media (max-width:860px) {
-  .cards { grid-template-columns:1fr; }
-  .two-col { grid-template-columns:1fr; }
+  .cards, .two-col { grid-template-columns:1fr; }
   .sidebar { display:none; }
-  .topnav { display:none; }
+  .topbar { flex-wrap:wrap; row-gap:6px; padding:10px 12px; }
+  .project { display:none; }
+  .search { flex-basis:100%; margin-left:0; }
+  .search input { width:100%; }
+  .main { padding:20px 14px; }
+  table { display:block; overflow-x:auto; white-space:nowrap; }
+  .kv { grid-template-columns:120px 1fr; }
 }
 """
 
--- a/drt/docs/_html_assets.py
+++ b/drt/docs/_html_assets.py
@@ -154,6 +154,7 @@ td.right, th.right { text-align:right; }
 .empty--hero pre.code { display:inline-block; text-align:left; margin:14px auto 0; }
 .kpi__error { font-family:var(--mono); font-size:11.5px; color:var(--error); margin-top:4px;
   overflow:hidden; text-overflow:ellipsis; white-space:nowrap; }
+.yaml-note { font-size:12px; color:var(--muted); font-style:italic; margin-top:8px; }
 
 /* Sync detail — KPI cards, tabs (progressive: stacked without JS), ego lineage */
 .kpi { border:1px solid var(--line); border-radius:var(--radius); padding:12px 16px; }
--- a/drt/docs/_html_assets.py
+++ b/drt/docs/_html_assets.py
@@ -156,6 +156,31 @@ td.right, th.right { text-align:right; }
   overflow:hidden; text-overflow:ellipsis; white-space:nowrap; }
 .yaml-note { font-size:12px; color:var(--muted); font-style:italic; margin-top:8px; }
 
+/* code block chrome — header bar + line numbers + collapse */
+.codeblock { border:1px solid var(--line); border-radius:var(--radius); overflow:hidden; position:relative; }
+.codehdr { display:flex; align-items:center; gap:10px; padding:7px 12px;
+  border-bottom:1px solid var(--line); background:var(--chip); font-size:12px; }
+.codehdr__path { font-family:var(--mono); font-weight:600; color:var(--fg); }
+.codehdr__meta { color:var(--muted); }
+.copy-btn, .expand-btn { display:none; }
+.js .copy-btn { display:inline-block; margin-left:auto; border:1px solid var(--line);
+  background:var(--surface); color:var(--fg); border-radius:6px; padding:2px 10px;
+  font:500 11.5px var(--sans); cursor:pointer; }
+.js .copy-btn:hover { border-color:var(--brand-500); color:var(--brand-700); }
+.codebody .highlight { margin:0; }
+.codebody .highlight pre { margin:0; padding:12px 14px; overflow-x:auto; font-size:12.5px; line-height:1.6; }
+.codebody .highlight table { border-collapse:collapse; width:100%; }
+.codebody .highlight table td { padding:0; border:none; }
+.codebody .highlight .linenos { width:1%; user-select:none; -webkit-user-select:none; }
+.codebody .highlight .linenos pre { color:var(--muted); opacity:.55; text-align:right; padding-right:4px; }
+.js .collapsible:not(.open) .codebody { max-height:520px; overflow:hidden; }
+.js .collapsible:not(.open) .codebody::after { content:""; position:absolute; left:0; right:0; bottom:34px;
+  height:56px; background:linear-gradient(transparent, var(--surface)); pointer-events:none; }
+.js .collapsible:not(.open) .expand-btn { display:block; width:100%; border:none;
+  border-top:1px solid var(--line); background:var(--chip); color:var(--brand-700);
+  padding:7px; font:500 12px var(--sans); cursor:pointer; }
+.js .collapsible.open .expand-btn { display:none; }
+
 /* Sync detail — KPI cards, tabs (progressive: stacked without JS), ego lineage */
 .kpi { border:1px solid var(--line); border-radius:var(--radius); padding:12px 16px; }
 .kpi__label { font-size:11px; text-transform:uppercase; letter-spacing:0.05em; color:var(--muted); font-weight:600; }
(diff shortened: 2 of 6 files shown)


Muawiya-contact:
**Verdict: request changes** — two blocking items, both in the areas you flagged. I ran the branch in an isolated worktree with constructed fixtures rather than eyeballing; details below.

## Priority 1 — `collect_sync_yaml_texts()` file safety

**① [high] Uncaught `AttributeError` on non-dict YAML crashes the whole docs build.**
`(yaml.safe_load(raw) or {}).get("name")` assumes a mapping, but `except (OSError, UnicodeDecodeError, yaml.YAMLError)` doesn't catch `AttributeError`. A `syncs/*.yml` that is valid YAML but not a dict blows up:
- `syncs/list.yml` = `- one\n- two` → `AttributeError: 'list' object has no attribute 'get'`. It propagates out of the collector, and `docs.py` calls it unguarded, so `drt docs generate --format html` aborts with a traceback — contradicting the "silently skipped / best-effort" contract. `test_collect_sync_yaml_texts_best_effort` misses this (only a YAMLError + a nameless dict).
- Fix: `isinstance(parsed, dict)` guard, or add `AttributeError` to the `except`.

**② [medium] Symlinks are followed with no guard, and the displayed path masks the real source.**
`glob("*.yml")` returns symlinks and `read_text()` follows them — no `is_symlink()`/`resolve()`-within-project check. `syncs/innocent.yml → ../OUTSIDE_secret.yml` was read and embedded, with the code header showing `syncs/innocent.yml`, hiding that the content came from outside `syncs/`. (A symlink to a non-dict external file is still read into memory, then trips ①.)

**③ [medium] No size cap.** A ~2 MB `huge.yml` embedded all ~2.08M chars into the sync page — no truncation. With ② (symlink to an unbounded/large source) this is a read-amplification + page-bloat vector. Suggest a byte cap + "truncated" note.

**Safe:** path traversal — `*.yml` doesn't cross separators or match `..`, so a crafted *filename* can't escape (only symlinks can, per ②). Non-UTF-8 / binary — `UnicodeDecodeError` is caught → skipped cleanly. Missing `syncs/` or unreadable file → `{}` / manifest fallback + note, as intended.

## Priority 2 — #696 redaction

**④ [high] The raw-YAML tab publishes `destination:` secrets verbatim, with no redaction.**
Rendered a sync inlining connection details; every one appeared verbatim in the sync page (after stripping Pygments spans):

| `destination:` field | in rendered HTML |
|---|---|
| `password: hunter2-SUPERSECRET` | leaked |
| `token: xoxb-…REALSLACKTOKEN` | leaked |
| `host: prod-db.internal.acme.com` | leaked |
| `admin_email: oncall@acme.com` | leaked |
| `phone: +1-555-123-4567` | leaked |

The manifest-derived view never carried the connection block at all (schema v1 = name/source/destination/mode/description/tags/state), so this is a net-new exposure surface for a docs site that may be committed/hosted — no key-masking or `*_env`-only policy on the raw path. Needs a redaction pass (mask known-sensitive keys) or explicit opt-in + a visible warning before shipping.

## Standard correctness — all good
- **Escaping ✓** hostile file content (`</script>`) renders escaped; `test_raw_yaml_is_escaped` puts the payload in the raw YAML text, so it genuinely covers file-content injection.
- **Determinism (#697) ✓** `glob` is `sorted()`, collects are byte-identical, paths use `.as_posix()` (stable on Windows). (Reads live disk, so it's "same working tree → same bytes" — inherent to the feature.)
- **No-JS ✓** `.copy-btn`/`.expand-btn` are `display:none` by default and gated on `.js`; the collapse `max-height` is `.js .collapsible:not(.open)` — full file renders without JS. Line numbers `linenos="table"` + `user-select:none`.
- **ruff ✓** clean · **mypy** only the pre-existing `credentials.py:363` (not this PR) · **docs tests** 53 passed · **full suite** 1816 passed, 6 failed = all the known Windows `test_sqlite.py` `os.unlink` flake (passes on CI Linux), zero docs-area failures.

## Non-blocking notes
- **Merge collision with #701:** this branch predates that merge — both rewrite `render_html`'s signature and the `_SYNC`/`_DAG` templates, so whichever lands first, the other needs a real rebase on `html.py`.
- **`layout.py` +23/−18 is pure formatting churn** (AST-identical) — unrelated to this PR and touches the frozen engine; worth dropping from a raw-YAML PR.

Nice feature overall — the code chrome, dark palette, and mobile pass are clean, and escaping/determinism/no-JS are all handled. Just want ① and ④ sorted before merge.

Muawiya-contact (changes requested):
Good effort @masukai, just do that changes please.

masukai:
Thanks @Muawiya-contact — thorough as always, and the worktree-with-hostile-fixtures approach caught exactly the right things. Pushed fixes for all four:

**① [high] non-dict YAML crash** — `isinstance(parsed, dict)` guard; a top-level list / scalar (`syncs/list.yml` = `- one\n- two`) is now skipped per the best-effort contract instead of AttributeError-ing out of the collector. New test `test_collect_sync_yaml_texts_non_dict_yaml_is_skipped`.

**④ [high] secret leakage (#696)** — inline values are masked before embedding. A line-based redactor (so the "as written" layout survives) replaces the scalar after any sensitive key — `password / passwd / passphrase / secret / token / api_key / access_key / private_key / client_secret / credential / connection_string / dsn / auth / webhook / url / endpoint / host / hostname / email / phone` (case-insensitive substring). `*_env` keys are preserved (they're env-var references, drt's recommended pattern), and a leading `# drt docs: inline secrets / PII masked` note appears when anything was redacted. Verified against your exact table — password/token/host/admin_email/phone all masked, `password_env` and `batch_size` untouched. New test `test_collect_sync_yaml_texts_redacts_secrets`.

**② [med] symlink follow** — `path.is_symlink()` → skip, so a link can't read outside `syncs/` under a masked in-project path. New test `test_collect_sync_yaml_texts_skips_symlinks`.

**③ [med] size cap** — capped at 64 KiB with a `# … truncated by drt docs` note. New test `test_collect_sync_yaml_texts_caps_size`.

**Non-blocking**
- **layout.py churn** — good catch, dropped it back to `main` (it was AST-identical formatting and would've collided with #701). The PR's `layout.py` diff is now empty.
- **#701 collision** — the branch shows `MERGEABLE` against current `main` after these pushes, so no rebase needed right now; flagging in case that changes before merge.

One known limitation worth noting: redaction covers **inline scalar values**. A sensitive value written as a *block scalar* (`private_key: |` + an indented PEM body) isn't masked line-by-line — the recommendation there stays `*_env`. Happy to extend to block scalars if you'd rather close that too.

`ruff` + `mypy` clean, 23 docs tests green. Re-review whenever suits 🙏

masukai:
Rebased onto current `main` — this was **CONFLICTING** after the docs train landed (#796 static-SVG DAG → #751 empty-states → #753 quick-wins all touched `html.py` / `_html_assets.py`), which would have blocked your re-review. Now **MERGEABLE and green** again.

**What the rebase involved** (all mechanical, nothing behavioural):
- `html.py` — dropped this branch's stale inline `_slug` / `_badge` / `_node_card` copies; they live in `drt/docs/_svg.py` since #796 and are imported from there.
- `_html_assets.py` — kept **both** sides: `.empty--hero` / `.kpi__error` (from #751/#753) *and* this branch's `.yaml-note`.
- `test_docs_html.py` — kept **both**: `test_regeneration_is_byte_identical` (#697, from #753) *and* the YAML-tab tests.

**Your four requested changes are all still in and verified after the rebase:**
- ① `isinstance(parsed, dict)` guard (non-mapping YAML skipped, no `AttributeError`)
- ② `path.is_symlink()` skip
- ③ 64 KiB cap + truncation note
- ④ inline secret / endpoint / PII redaction (`*_env` preserved) — #696

Full suite **2028 passed**, `ruff` + `mypy` clean, CI green incl. codecov. Ready for your re-review whenever suits — no rush 🙏

Muawiya-contact (approved):
**Approving — all four review blockers are fixed and verified. Really nicely done, @masukai. 🙏**

I re-ran my original adversarial fixtures against the rebased head (`5eda325`) rather than taking the summary on trust — every point holds up end-to-end:

- **① non-dict YAML** — a top-level list / scalar `.yml` is now skipped (`isinstance(parsed, dict)` guard); no more `AttributeError` aborting `drt docs generate`. ✓
- **② symlinks** — `path.is_symlink()` skip confirmed: a `syncs/innocent.yml → ../OUTSIDE.yml` symlink is no longer read, and the external file's contents never reach the docs. ✓
- **③ size cap** — a >64 KiB file is truncated with the `# … truncated by drt docs` note, output bounded. ✓
- **④ #696 redaction (the security blocker)** — inline `password` / `token` / `host` / `admin_email` / `phone` in a `destination:` block are all masked to `'« redacted »'`, **`host_env` (and any `*_env`) is preserved**, and the leading "inline secrets / PII masked" note fires. Verified none of the secret values survive into the rendered HTML. ✓

Also confirmed: hostile file content is still HTML-escaped after redaction (no live markup), and you added a focused regression test for each case (`_non_dict_yaml_is_skipped`, `_skips_symlinks`, `_caps_size`, `_redacts_secrets`) — so this can't silently regress. `test_docs_html.py` green (24), ruff + mypy clean locally, and the rebase kept both sides on `html.py` / `_html_assets.py` / `test_docs_html.py` (the `#697` byte-identical test *and* the YAML-tab tests coexist).

The line-based redaction that preserves the "as written" layout while masking only the sensitive scalar is a genuinely tasteful touch — it keeps the feature's whole point (show the file as authored) without the leak. Thanks for the thorough follow-through on every note, including the ones that needed real hardening rather than a one-liner.

LGTM. 🚀

PR #753, opened 2026-07-10
design(docs): quick-wins — favicon, a11y, print styles, search empty state + byte-identical regression test

masukai:
Small finishing pass off the #702 tracker — no schema, no layout, just polish:

- **Favicon + meta description** — an inline violet-`d` SVG; every page has been 404ing on `favicon.ico` since P3 landed.
- **A11y**: skip-to-content link, `:focus-visible` rings on all interactives, tabs carry `role=tab` / `aria-selected` (synced by JS).
- **Print stylesheet**: topbar/sidebar/buttons hidden, all tab panels print, code collapse released — static docs get PDF'd.
- **Search empty state**: the sidebar filter says "No matches" instead of silently emptying.
- **Determinism enforced**: new test renders the site twice and asserts byte-identical output — #697's acceptance bar, now a regression gate.

Note: written against `main`; if #752 lands first I'll rebase (both touch the assets file).

@Muawiya-contact tiny one — a skim whenever ☕

Related: #702, #697.


Muawiya-contact:
@masukai, just give me some time, I will review all PRs thoroughly soon.

masukai:
No rush at all, @Muawiya-contact — take whatever time you need 🙏 Nothing in the queue is time-sensitive, so please review only as deeply as feels useful and whenever your schedule allows. Quality over speed, always. Thanks for picking them up 💜

Muawiya-contact:
@masukai 
Approve — nice low-risk polish pass, and the byte-identical regeneration test is a good addition. A few non-blocking notes:

1. **Determinism test could be stronger.** `test_regeneration_is_byte_identical` renders twice in the same process, so it won't catch `PYTHONHASHSEED`-dependent set/dict ordering — which is the exact failure mode #697 guards. I checked out-of-band and the site *is* cross-process stable (identical SHA-256 under different hash seeds), so it passes today — but to make it a real regression gate, rendering one side in a subprocess with a different seed (or comparing to a committed golden) would close that gap.

2. **Print CSS has some forward-referenced classes** — `.copy-btn`, `.expand-btn`, `.collapsible`, `.codebody` only appear inside the `@media print` block; nothing emits them yet. Look like they're anticipating #752's collapsible-code feature? Inert today either way — just flagging in case you'd rather land them with #752.

3. Tiny one: the diff touches `layout.py` (+23/−18) despite the "no layout" note — it's pure `ruff format` reflow (AST-identical, zero behavior change, no conflict with the emission work), so harmless, but might be worth splitting into its own housekeeping commit to keep this diff docs-only.

Verified the docs suite passes (28/28) and determinism holds. 🙂

Muawiya-contact (approved):

Issue #773, opened 2026-07-10
feat(cli): drt run --failed — sync-level re-run of failures from the previous invocation


masukai:
## Problem

When `drt run` executes 20 syncs and 3 fail (one flaky API, one rate-limit blowout, one bad credential), the recovery story is manual: read the output, then `drt run --select a --select b --select c`… which today isn't even expressible in one command (single-valued `--select`). In cron/CI, "re-run only what failed" is the difference between a 2-minute recovery and a full re-run that re-sends everything that already succeeded.

Note `drt retry` exists but is *record-level* (DLQ replay for one sync). This is *sync-level* re-run.

## Prior art

- **dbt retry**: re-runs the previous invocation from its `run_results.json`, only failed/skipped nodes, preserving flags.

## Proposal

```bash
drt run --failed            # re-run syncs whose last recorded status != success
drt run --failed --dry-run  # preview what would be re-run
```

- Source of truth: `.drt/state.json` already stores per-sync `status` + `error` — zero new persistence needed for the local case; with the remote state backend it works across runners
- Semantics: selector-style — `--failed` intersects with `--select`/`--exclude` if given; error if combined state is empty ("nothing failed")
- Watermark interplay: failed incremental syncs didn't advance their watermark (existing behavior), so re-run is naturally correct

## Scope

**In:** `--failed` flag on `run` (and `drt test --failed` for symmetric test re-runs), status-based selection from StateManager, docs.
**Out:** full dbt-style "replay last invocation's exact flag set" (needs run artifacts — separate issue; reference it), record-level replay (#278 DLQ covers).

## Related

- #278 (`drt retry` record-level), run-artifacts issue (invocation-level replay), remote-state issue (cross-runner `--failed`)

## Milestone fit

Candidate: v0.8.x — small, immediately lovable ops win; data already persisted.

Issue #772, opened 2026-07-10
feat(cli): state:modified selector — run only syncs changed vs a baseline manifest (slim CI for reverse ETL)


masukai:
## Problem

drt's whole thesis is Git-native reverse ETL — but CI can't answer "which syncs did this PR change?". Today a PR touching one sync YAML either dry-runs *everything* (slow, noisy, needs all credentials) or nothing. The drt-action README's PR-preview example runs the full project for any change.

## Prior art

- **dbt `state:modified` + `--defer`**: compare the working project against a baseline manifest artifact; run only changed nodes. This single feature is the backbone of dbt CI ("slim CI").

## Proposal

drt already generates a machine-readable project manifest (`drt docs generate --format json`, schema v1, #507). Add:

```bash
# in CI, against the baseline manifest saved from main
drt run --select state:modified --state path/to/baseline/manifest.json --dry-run --diff
```

- Comparison unit = sync definition: hash the *resolved* sync config (post env-substitution, model text, destination config minus secrets) into the manifest; `state:modified` selects syncs whose hash differs or which are new
- `state:new` as the narrower variant
- Composes with the selection-v2 resolver (`--select state:modified --exclude tag:heavy`)
- drt-action follow-up: `select: state:modified` input + a docs recipe (save manifest artifact on main; restore in PR jobs)

## Scope

**In:** stable config-hash in manifest (needs a documented canonicalization + exclusion of volatile fields — #697's generated_at work is the precedent), `--state` flag, `state:modified|new` methods, CI guide update.
**Out:** `--defer` (no cross-sync refs to defer), watermark/state reuse across environments (remote-state issue), auto-detection of changed SQL files referenced via `ref('name')` from `syncs/models/*.sql` — include file-content hash so it *is* covered.

## Related

- #507 (manifest v1 — the baseline artifact), #697 (byte-identical docs output — same determinism requirement), #370 (CI/CD guide), drt-action (#292)

## Milestone fit

Candidate: v0.9 — headline "Git-native" differentiator vs GUI SaaS; needs selection-v2 first.

masukai:
Shipped across the 4-PR stack: #950 (manifest v3 / config_hash), #951 (state:modified/state:new selector methods), #952 (--state CLI flag), #953 (docs + CI recipe). All merged.

Issue #786, opened 2026-07-10
research: streaming / event-triggered syncs — what is the OSS answer to Census Live Syncs & Hightouch Streaming?


masukai:
## Problem (research framing — decide, not build)

drt is batch-only: syncs run when invoked. The SaaS competition has moved to sub-minute activation — **Census Live Syncs** and **Hightouch Streaming/CDC** ship "row lands in warehouse → seconds later it's in the destination". As warehouses grow streaming ingestion (Snowpipe Streaming, BigQuery CDC/Storage Write API, Databricks DLT), "fresh data, stale activation" becomes the visible gap for exactly the operational use cases reverse ETL exists for (abandoned-cart to ads, fraud flags to support tools).

This issue is for **investigating** what an OSS, CLI-first answer even looks like — explicitly *not* a commitment to build a daemon.

## Questions to answer

1. **Trigger taxonomy**: what can we detect cheaply per warehouse? (BQ table change notifications via Pub/Sub? Snowflake STREAMs polled? Databricks table triggers? Postgres LISTEN/NOTIFY/logical decoding?) Which are pull-poll vs push?
2. **Runtime shape**: does this fit `drt serve` grown up (long-lived process consuming triggers → running selected syncs), an orchestrator recipe (Dagster sensors — dagster-drt may already be 80% of the answer), or warehouse-native scheduling invoking drt-action? Bias: recipes before daemons (CLAUDE.md: CLI-first, no heavy infra).
3. **Engine prerequisites**: micro-runs need cheap per-run overhead — which of (diff-based incremental, remote state, rate-limit sharing across frequent runs) are hard prerequisites?
4. **Where's the OSS line**: is "sub-minute, exactly-once-ish activation" an OSS feature, or an Enterprise/Cloud boundary item (OPEN_CORE.md)?

## Deliverable

A short ADR under `docs/adr/` (next number: 0004) with: per-warehouse trigger matrix, recommended v1 posture (likely: Dagster-sensor recipe + `drt serve` webhook hardening + docs), prerequisites list, and an explicit build/no-build recommendation for a native watcher.

## Related

- #428 (scheduler — time-based cousin), #218/`drt serve` (existing trigger endpoint), #426 (depends_on), diff-based incremental + remote state issues (prerequisites), dagster-drt sensors

## Milestone fit

Research spike — v0.9 window (informs the v1.0 protocol freeze and OPEN_CORE boundary).

Muawiya-contact:
[...]

PR #791, opened 2026-07-10
feat(cli): selection v2 — globs, repeated --select, --exclude, destination: selectors (closes #771)

### LINKED ISSUE #771: feat(cli): selection v2 — glob patterns, repeated --select union, --exclude, destination:/source: selectors
## Problem

`--select` accepts exactly three forms: a sync name, `tag:<t>`, or `*`/`all`. As projects grow past a handful of syncs (the drt-data pattern: one repo, many syncs, three warehouses), routine operations become impossible to express:

- "everything except the two big backfills" → no `--exclude`
- "all HubSpot syncs" / "everything reading from bigquery_prod" → no `destination:`/`source:` selectors
- "the `users_*` family" → no glob
- "these three by name" → `--select` is single-valued

dbt users arriving at drt expect this vocabulary — it's the muscle memory of `dbt run -s`.

## Prior art

- **dbt node selection**: multiple `--select` values (union), `--exclude`, glob/fnmatch, method selectors (`tag:`, `source:`, `path:`), intersection (`,`).

## Proposal

Adopt the dbt-familiar subset that maps to drt's flat (non-DAG) model:

```bash
drt run --select 'users_*'                        # fnmatch glob
drt run --select tag:crm --select tag:ads         # union across repeated flags
drt run --select tag:crm --exclude users_backfill # subtraction
drt run --select destination:hubspot              # by destination type
drt run --select source:bigquery_prod             # by source profile
drt test / drt validate                           # same engine everywhere
```

Selector resolution becomes one shared module used by `run`/`test`/`validate` (today `test --select` is name-only, `validate --select` separate — unify). Unknown selector → error listing available methods.

## Scope

**In:** glob, repeated-flag union, `--exclude`, `destination:`/`source:` methods, shared resolver across the three commands, shell-completion for selector values.
**Out:** graph operators (`+model+`) — meaningless until #426 depends_on exists; `state:` method (separate issue); named YAML selectors (future sugar).

## Related

- #426 (depends_on) would later enable graph operators on the same resolver
- #288 (`--all`), #254 (`--select` on validate) — the increments this consolidates

## Milestone fit

Candidate: v0.8.x — pure CLI ergonomics, no engine changes; the resolver is also the prerequisite for `state:modified`.

masukai:
## Summary

Implements [#771](https://github.com/drt-hub/drt/issues/771): the dbt-familiar selection vocabulary, shared across `drt run` / `drt test` / `drt validate` via one resolver (`drt/cli/_selection.py`) — previously three divergent inline filters (`test`/`validate` were name-only).

```bash
drt run --select 'users_*'                        # fnmatch glob
drt run --select tag:crm --select tag:ads         # union via repetition
drt run --select tag:crm --exclude users_backfill # subtraction
drt run --select destination:hubspot              # by destination type
drt test / drt validate                           # same grammar everywhere
```

## Semantics

- A `--select` token matching **nothing is an error** (a typo never silently runs zero syncs) — message shapes preserved from pre-#771 (`No sync named 'x' found.` etc.); `--exclude` tokens may match nothing (dbt behaviour)
- Definition order preserved, results deduplicated
- Unknown method (`source:x`) errors and names the available methods
- `validate`: bare-name/glob tokens still match **unparseable** syncs' error keys, so `drt validate --select broken_sync` surfaces its errors
- Best-effort shell completion for `--select`/`--exclude` values (sync names + `tag:`s + `destination:`s; never crashes the shell)

## Notes

- **`source:` dropped from the issue scope on purpose** — syncs share the project profile (one source per run), so there is nothing per-sync to select on. Noted in the module docstring; revisit if per-sync sources land.
- Groundwork for `state:modified` (#772) and `drt run --failed` (#773) — both plug into this resolver.
- The `run --help` examples changed; the #516 help-examples test was updated accordingly.
- Trivial `CHANGELOG [Unreleased]` conflict with #788/#789/#790 — I'll rebase whichever lands later.

## Tests

20 resolver unit tests (grammar, union/exclude/ordering, error messages, completion) + 6 CLI-level tests (glob, union, exclude, exclude-all, unknown method) + existing run/test/validate suites green. Full suite **1989 passed, 12 skipped**; `ruff` + `mypy` clean.

Closes #771


masukai:
@yodakanohoshi — batched review request for a 5-PR CLI stack (all green, all rebased on latest `main`).

These are the DX pull-ins for **v0.8** from the 2026-07-10 dbt/dlt/competitor gap batch (#755–#786). They're stacked, so please review **in this order** — GitHub auto-retargets each to `main` as the parent merges:

| Order | PR | Closes | One-liner |
|---|---|---|---|
| 1 | #791 | #771 | **selection v2** — one shared resolver (`_selection.py`) for `run`/`test`/`validate`: name globs, `tag:`, `destination:`, repeated `--select` union, `--exclude` |
| 2 | #792 | #773 | `drt run --failed` — re-run only syncs whose last status ≠ success (reads `.drt/state.json`; clean state exits 0) |
| 3 | #793 | #774 | `drt run --limit N` — extract-side sampling (`islice`); watermark frozen; refused for mirror/replace |
| 4 | #794 | #775 | `--fail-fast` on `run`+`test` — stop scheduling after first failure; `skipped` becomes a first-class JSON status |
| 5 | #795 | #777 | `drt build` — run each sync then its tests in one pass (sequential v1; `execute_tests_for_sync` extracted from `drt test`) |

**Review notes:**
- #791 is the load-bearing one — the other four build on its resolver. If you only have time for one deep pass, that's the one.
- Each PR: full suite green (2000+ tests), `ruff` + `mypy` clean, CHANGELOG/docs updated. #790's `tomllib` 3.10 skip is already in `main`.
- Two deliberate scope calls, both noted in the PR bodies: (a) #791 drops the `source:` selector from the issue (syncs share the project profile — nothing per-sync to select on); (b) #795 `drt build` is sequential-only in v1 (true `--threads` interleaving needs the #723 run-loop extraction).

No rush — batch it whenever suits. Happy to split any of these smaller if a single diff is easier to review. 🙏

masukai:
Quick nudge on this 5-PR CLI stack (#791 → #792 → #793 → #794 → #795) — all still green and rebased on latest `main` (unaffected by the #798 config split; none of these touch `models.py`).

@Muawiya-contact — added you as a reviewer since you've kindly been working through the queue; absolutely no rush, and please only take these if/when you have bandwidth alongside everything else you're carrying. #791 (the shared selection resolver) is the load-bearing one — if you only have time for a single deep pass, that's the one; the other four build on it.

@yodakanohoshi — whenever you get a chance to look at these, that'd be great 🙏 No urgency, just flagging they've been sitting green for a bit.

Happy to split any of them smaller if a single diff is easier to review.

Muawiya-contact (approved):
**Approving — the load-bearing one, and it's clean.** Unlike the rest of the stack, this targets `main` and its full suite actually ran on CI (green on 3.10–3.13 + codecov + CodeQL); I also read the resolver closely and re-ran the relevant suites locally.

**Resolver (`_selection.py`) — correct where it counts:**
- `matches` covers `*`/`all`, `tag:<glob>`, `destination:<glob>`, unknown-method error (names the available methods), and bare-name/glob via `fnmatchcase`. ✓
- `select_syncs` resolves via `[s for s in syncs if s.name in matched_names]` — **definition order preserved + deduplicated** in one stroke, which is exactly right. A `--select` token matching nothing raises `SelectionError` (typo never silently runs zero syncs); `--exclude` tokens may match nothing (dbt behaviour). ✓
- `complete_selector` swallows every exception → completion can't crash the shell. ✓

**Integrations:**
- One resolver now backs `run` / `test` / `validate` (previously three divergent name-only filters) — good consolidation.
- `validate` is handled thoughtfully: `_fnmatch_token` restricts unparseable syncs (which exist only as error-dict keys) to bare-name/glob tokens, so `drt validate --select broken_sync` still surfaces its errors while `tag:`/`destination:` correctly can't match an unparsed sync. ✓

**Verification:** 71 local tests pass across `test_cli_selection` + run/test/validate/help-examples; `ruff` clean; `mypy` clean on the changed files (only the pre-existing, transitive `credentials.py:368`). The 26 new tests (20 resolver + 6 CLI) match the described grammar/error/completion coverage.

Dropping `source:` with a documented reason (one source per run — nothing per-sync to select on) is the right call, and this cleanly sets up `--failed` (#773) and `state:modified` (#772). One tiny by-design edge worth noting only for the record: a bare sync name containing a literal `:` would be read as a method token — fine given the grammar, just flagging it exists. LGTM 🚀

Muawiya-contact:
Thanks for the clear map of the stack, @masukai — and no worries at all, I actually went ahead and did a full pass on all five rather than just this one, since they build on each other and it was easier to hold the whole picture in my head at once. Really enjoyable set to read through.

Where things landed:

- **#791** (this one) — approved ✅. The shared resolver is the right call; order-preserving dedup in a single pass, the typo-protection on `--select`, and the `validate` handling for unparseable syncs (bare-name/glob only, since method tokens can't match an unparsed sync) are all clean. This is the load-bearing piece and it carries the weight well.
- **#792** `--failed`, **#793** `--limit`, **#794** `--fail-fast` — all approved ✅. I ran each locally (their CI only had the `cla` check since they target stacked bases, so I verified on the heads directly). The `--limit` watermark-non-advance guard and the `--fail-fast` graceful-drain-vs-cancel split in particular are exactly the careful bits I'd have worried about, and you'd already handled them + locked them with tests.
- **#795** `drt build` — left a request-changes 🔧, one real blocker but a tiny fix: its new `_FakeResult` is missing `watermark_lag`, so all 6 build tests error on the current head (`AttributeError` via `_run_one`) — classic stacked-rebase drift, the base grew the field after the fixture was written. Production `drt build` is fine; it's just the test double. One line (`self.watermark_lag = None`) makes all 6 pass — I verified. There's also a small `--quiet` gap where the test output still prints; details in the comment.

So really it's four green and one one-liner away from green. **Happy to do whatever helps land these** — re-review #795 the moment you push the fixture fix (I'll turn it around fast), take another look if you split anything, or re-run any of them. Whatever's easiest on your end. No rush from my side either; thanks for keeping the whole stack rebased and tidy, it made the review a pleasure. 🙏

PR #792, opened 2026-07-10
feat(cli): drt run --failed — sync-level re-run of the previous invocation's failures (closes #773)

masukai:
## Summary

Implements [#773](https://github.com/drt-hub/drt/issues/773):

```bash
drt run --failed            # re-run syncs whose last status != success
drt run --failed --select tag:crm   # intersects with selection
```

- Reads per-sync status from `.drt/state.json` (zero new persistence) — `failed` **and** `partial` qualify; never-run syncs are excluded
- Clean previous state → note + **exit 0** (recovery loops shouldn't page when there's nothing to recover); JSON mode emits `"note": "nothing_failed"`
- Failed incremental syncs never advanced their watermark, so re-runs are naturally correct
- Complementary to `drt retry` (record-level DLQ replay #278) — this is the dbt-`retry`-shaped sync-level counterpart

**Deviations from the issue, with reasons:** `drt test --failed` deferred — test results aren't persisted anywhere yet (run-artifacts #778 is the enabler). "Error when nothing failed" became exit-0-with-note for cron/recovery ergonomics.

## Stacked PR

⚠️ **Based on #791 (selection v2)** — merge #791 first; GitHub will retarget this to `main` automatically. The diff vs #791 is this feature only.

## Tests

4 new CLI tests (failed-only filter incl. `partial`, never-run exclusion, `--select` intersection, clean-state exit 0). Affected suites green; `ruff` + `mypy` clean.

Closes #773

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #765, opened 2026-07-10
perf(sources): streaming extraction — replace fetchall() with server-side cursors / fetchmany across SQL sources


masukai:
## Problem

Every SQL source except BigQuery extracts with `cur.fetchall()` — the entire result set materializes in memory before the first record reaches the engine (BigQuery streams via its row iterator; Delta/Iceberg full-materialize too, tracked separately in #679). The engine then batches by `sync.batch_size` — but by that point peak memory has already been paid.

For reverse ETL at warehouse scale (the exact BigQuery → Snowflake/Databricks patterns we now smoke-test nightly), a few-GB result set on a GitHub Actions runner (7 GB RAM) is an OOM kill, and even successful runs hold the full set for the entire load duration. This also blocks the windowed-backfill story from being memory-safe.

## Prior art

- **dlt**: extracts in chunks end-to-end; buffered writers keep memory flat regardless of table size.
- **dbt**: N/A (transform-in-place) — but adapter `fetchmany` patterns are standard.
- Drivers all support it natively: psycopg2 named/server-side cursors (`itersize`), MySQL `SSCursor`, snowflake-connector `fetchmany`, databricks-sql `fetchmany_arrow`, ClickHouse blocks, pyodbc `fetchmany`.

## Proposal

Switch `Source.extract()` implementations from `fetchall()` to streaming iteration with a per-source fetch size:

```yaml
# profiles.yml (per source profile, optional)
fetch_size: 10000        # rows per driver round trip; default keeps current behavior until validated
```

- Postgres: server-side (named) cursor; MySQL: `SSCursor`; Snowflake/Databricks/SQLServer/Redshift/ClickHouse/DuckDB: `fetchmany` loops
- `Iterator[dict]` contract is unchanged — this is purely inside each source
- Add a memory note per connector doc + a boundary test asserting no source holds >2×fetch_size rows

## Scope

**In:** the 8 fetchall sources, fetch_size profile knob, contract test.
**Out:** Delta/Iceberg (#679 predicate pushdown covers their path), engine-side backpressure (batching already exists), async extraction.

## Related

- #679 — same theme for lakehouse sources
- #280 (benchmark suite) — should capture before/after peak RSS
- Windowed-backfill issue — streaming extract makes windows memory-safe

## Milestone fit

Candidate: v0.8.2 (warehouse hardening) or v0.9 with the benchmark suite — measurable, contained, high-leverage for the "production ready" claim.

masukai:
[...]

PR #793, opened 2026-07-10
feat(cli): drt run --limit N — extract-side sampling for safe first sends (closes #774)

### LINKED ISSUE #774: feat(cli): drt run --limit N — extract-side sampling for safe first sends (no watermark advance)
## Problem

There is no way to run a sync against a *small sample*. Today's options: `--dry-run` (no load at all) or full send. When developing a new sync — especially against a production CRM/ads API — operators want "actually send 10 rows, look at them in the destination UI, then open the tap". Editing the model SQL to add `LIMIT 10` and remembering to remove it is the current workaround, and it mutates a file that's under review.

## Prior art

- **Census/Hightouch**: "test sync with sample records" before enabling a schedule — a core onboarding affordance.
- **dlt**: `add_limit(n)` on any resource for exactly this.
- **dbt**: `--empty` / `limit` flags on show/build for dev loops.

## Proposal

```bash
drt run --select new_sync --limit 10          # extract-side cap, then real load
drt run --select new_sync --limit 10 --diff   # (with --dry-run) already-capped preview
```

- Implement extract-side: wrap the resolved model as `SELECT * FROM (<model>) AS _drt_base LIMIT {n}` for SQL sources (the subquery-wrap seam exists for cursor injection); REST source caps records/pages; file sources cap rows
- Mark the run `sampled: true` in result/history so a limited run is never mistaken for a full sync (and **skip watermark advancement** — sampling must not eat unsampled rows' cursor progress)
- Refuse `--limit` with `mode: mirror` / `replace` (a sampled mirror would delete real rows) — validate-time error

## Scope

**In:** flag, per-source-family cap strategy, watermark/no-advance + result marking, mode guards, docs.
**Out:** percentage/stratified sampling, destination-side truncation.

## Related

- #413 (`--dry-run --diff`) — preview without side effects; `--limit` is the "small real send" complement
- Windowed-backfill issue — same subquery-wrap machinery

## Milestone fit

Candidate: v0.8.x — small engine+CLI feature with outsized onboarding value (safe first sends).

masukai:
## Summary

Implements [#774](https://github.com/drt-hub/drt/issues/774):

```bash
drt run --select new_sync --limit 10   # actually send 10 rows, then stop
```

## Design

- **Engine-side cap** (`itertools.islice` on the extraction iterator) — dialect-agnostic: SQL, REST, and file sources behave identically, with no per-dialect `LIMIT`/`TOP` SQL rendering. (Post-#765 streaming extraction, the islice also stops driver consumption early.)
- **Watermarks never advance on a sampled run** — cursor tracking is skipped under `--limit`, so rows the sample skipped can't fall permanently behind the cursor (engine test locks this).
- **Sampled runs are marked** — `SyncResult.limit_applied`, a `limit` field in `--output json` entries, and a CLI notice; a limited run can't masquerade as a full sync.
- **Refused for `mode: mirror` / `replace`** — a sampled mirror would DELETE destination rows the sample skipped; a sampled replace would truncate a full table to N rows. Validate-time error naming the offending syncs.
- Composes with `--dry-run --diff` for a capped no-send preview.

## Stacked PR

⚠️ **Based on #792 (`--failed`)** → #791 (selection v2) → main. Merge order: #791 → #792 → this. Diff vs #792 is this feature only.

## Tests

3 engine tests (cap + `rows_extracted`, None passthrough, watermark non-advance via `StatePersistingObserver`) + 3 CLI tests (`extract_limit` kwarg forwarding, `--limit 0` rejection, mirror/replace guard). Affected suites green; `ruff` + `mypy` clean.

Closes #774


Muawiya-contact (approved):
**Approving — clean, correct, well-tested.** Verified by running the code locally on `f695aa4` (the head only carries the `cla` check — the test/mypy workflows haven't run on this stacked head, so I checked directly).

**Engine (the part that matters most):**
- **Cap** — `islice(records_iter, extract_limit)` on the extraction iterator is the right call: dialect-agnostic (REST/file/SQL identical, no per-dialect `LIMIT`/`TOP`), and post-#765 streaming it stops driver consumption early. ✓
- **Watermark non-advance** — the guard `if cursor_field and extract_limit is None:` skips cursor tracking under `--limit`, so `new_cursor_value` stays at `last_cursor_value` and the persisted watermark can't move. This is the critical safety property (a sampled run must not let skipped rows fall behind the cursor), and `test_extract_limit_never_advances_watermark` locks it. ✓
- `SyncResult.limit_applied` + the `limit` field in `--output json` means a sampled run can't masquerade as a full sync. ✓

**CLI guards:** `--limit < 1` rejected; `mode: mirror`/`replace` refused at validate-time naming the offending syncs (a sampled mirror would DELETE skipped rows; a sampled replace would truncate to N) — exactly the right dangerous cases to block. ✓

**Verification:** all 6 new tests pass (3 engine: cap + `rows_extracted`, `None` passthrough, watermark non-advance via `StatePersistingObserver`; 3 CLI: kwarg forwarding, `--limit 0` rejection, mirror/replace guard), broader engine + CLI suites green on the head (78 passed), no #795-style fixture drift. `ruff` clean; `mypy` clean on the changed files (only the pre-existing, transitive `credentials.py:368`).

One tiny non-blocking nit: the `limit: int = typer.Option(None, …)` param is really `int | None` (its default is `None`); `int | None` would match the semantics (mypy stays green either way since `typer.Option` returns `Any`). Cosmetic — not a merge concern.

Nice, well-scoped feature; the engine-side islice over per-dialect SQL is the right design. LGTM 🚀

PR #794, opened 2026-07-10
feat(cli): --fail-fast on drt run / drt test — stop scheduling after the first failure (closes #775)

### LINKED ISSUE #775: feat(cli): --fail-fast — stop scheduling syncs after the first failure (run + test)
## Problem

`drt run` (multi-sync, with or without `--threads`) always runs every selected sync to completion even after failures. In CI that wastes minutes and API quota on runs you already know are red; when the failure is systemic (expired credential, warehouse down), *every* sync fails one by one, producing N alerts (#414) instead of one.

## Prior art

- **dbt `--fail-fast`**: abort the invocation on first node failure — standard CI flag.

## Proposal

```bash
drt run --fail-fast
```

- Sequential mode: stop scheduling after the first `status != success`; remaining syncs reported as `skipped` (new result status, distinct from failure) in output/history
- `--threads` mode: don't schedule new syncs after a failure; let in-flight syncs finish (they already have graceful-shutdown semantics via #279's machinery — reuse the same drain path rather than hard-cancelling mid-load)
- Exit code unchanged (non-zero); `--output json` includes `skipped` entries so tooling can distinguish "didn't run" from "ran and failed"
- Also accept on `drt test`

## Scope

**In:** flag on run/test, skipped status in results/history/JSON output, thread-pool drain behavior, docs.
**Out:** failure-threshold variants (`--fail-fast=3`), per-sync criticality tiers (interesting later with depends_on #426).

## Related

- #279 (graceful shutdown — the drain machinery to reuse), #414 (alerts — one systemic failure should page once)

## Milestone fit

Small CLI/engine-loop change; `good first issue` candidate once the skipped-status shape is agreed. v0.8.x filler.

masukai:
## Summary

Implements [#775](https://github.com/drt-hub/drt/issues/775): `drt run --fail-fast` (and `drt test --fail-fast`) stop scheduling work after the first failure — a red build stops burning minutes and API quota, and a systemic failure (expired credential, warehouse down) pages once instead of N times.

## Design

- **Sequential**: stop scheduling immediately; remaining syncs are reported as `status: "skipped"` / `reason: "fail_fast"`.
- **`--threads`**: `shutdown(wait=False, cancel_futures=True)` on first failure — not-yet-started syncs are cancelled and reported skipped, **in-flight syncs drain normally** (no hard mid-load cancellation; consistent with #279's graceful-shutdown philosophy — the `with`-block's final `shutdown(wait=True)` still waits for them).
- **"Didn't run" ≠ "ran and failed"**: per-sync skipped entries + a top-level `skipped` count join `succeeded`/`failed` in `--output json`.
- `drt test --fail-fast`: stops after the first sync with a failing test; remaining syncs-with-tests reported skipped.
- Exit codes unchanged (non-zero on failure).

## Stacked PR

⚠️ Merge order: **#791 → #792 → #793 → this** (based on `feat/774-run-limit`). Diff vs #793 is this feature only.

## Tests

4 new: sequential skip-list assertions, no-flag regression (all syncs still run), threaded accounting invariant (every sync is exactly one of succeeded/failed/skipped — exact split is timing-dependent by design), and a `drt test --fail-fast` case. Full suite **2003 passed, 12 skipped**; `ruff` + `mypy` clean.

Closes #775


Muawiya-contact (approved):
**Approving — clean, correct, and well-tested.** Verified by running the code, not just reading it (the head only carries the `cla` check — the test/mypy workflows haven't run on it, presumably because it targets `feat/774-run-limit` in the stack — so I checked locally on `494f3a5`).

**Fail-fast logic — both paths correct:**
- **Sequential** — `stop_scheduling` flag; the first failed sync flips it, remaining syncs are reported `status: "skipped"` / `reason: "fail_fast"`. ✓
- **Threaded** — `shutdown(wait=False, cancel_futures=True)` on first failure cancels only *not-yet-started* futures; the `future.cancelled()` branch reports them skipped, in-flight syncs drain via the `with`-block's final `shutdown(wait=True)`. Matches #279's graceful-shutdown philosophy. ✓
- `drt test --fail-fast` breaks after the first sync with a failing test and reports the rest skipped. ✓

**Verification:**
- The 4 new tests pass, including the threaded accounting invariant (`succeeded + failed + skipped == len(syncs)` — the right property test for timing-dependent cancellation, not a vacuous one). Broader CLI batch green too — no stale-fixture drift.
- `ruff` clean; `mypy` clean on the changed files (only the pre-existing, transitive `credentials.py:368` remains).
- "didn't run" vs "ran and failed" is modelled cleanly — the per-sync skipped entries + top-level `skipped` count are a genuinely useful distinction in `--output json`.

Nice, well-scoped change. The graceful drain (rather than a hard mid-load cancel) is the right call. LGTM 🚀

Issue #723, opened 2026-07-07
refactor: mcp/server.py tool split, schema.py table-drive, run.py logging extract, source-connector docs


masukai:
## Summary

Lower-priority structural cleanups + a docs/coverage gap, grouped so they don't clutter the tracker individually.

- [x] **`mcp/server.py` (506 lines, one `create_server`)** — 9 `@mcp.tool()` closures each re-import + re-load project/sync config inline (`load_syncs`/`load_project` repeated in `drt_run_sync`, `drt_run_test`, `drt_list_syncs`, `drt_validate`…). Split into a `mcp/tools/` package (or top-level functions) with one shared `_load_ctx(project_dir)` helper — removes the boilerplate and makes each tool unit-testable without standing up the server.
- [x] **`schema.py` introspection (`47-233`)** — `describe_columns` is a 4-branch `isinstance` chain; each `_describe_{postgres,mysql,snowflake,databricks}` repeats the same skeleton (build `SELECT … FROM information_schema.columns WHERE …`, connect, execute, fetch, categorize). Table-drive it. ⚠️ **Partial collision with #707** (which rewrites the Databricks queries) — do Postgres/MySQL/Snowflake now, fold Databricks in after #707. Also normalizes the leaky private access (`SnowflakeDestination()._connect`) — see below.
- [x] **`_connect` signature normalization** — `postgres/mysql/clickhouse` expose `@staticmethod _connect`, `snowflake/databricks` expose instance-method `_connect`; `schema.py` calls them with two different conventions. Normalize to one `connect(config)` classmethod (folds naturally into the BaseSqlDestination / warehouse-base work).
- [x] **`cli/commands/run.py` (573 lines)** — extract the generic infra: `_JsonFormatter`/`_configure_json_logging` (`81-108`) → `cli/_logging.py`, and the signal→exit-code mapping, out of the run command module. `_run_one` (`139-300`) is a ~160-line orchestration function that could shed the observer-composition into a small factory.
- [x] **Source-only connector docs (chore)** — `duckdb`, `sqlite`, `redshift`, `sqlserver` have no `docs/connectors/*.md`, and `check_drift.sh` Check 1 only maps *destinations* → docs, so a source without a doc is structurally uncheckable. Either add stub source docs or extend Check 1 to sources with an explicit baseline for accepted omissions.

**Effort/risk:** S–M / LOW each.

masukai:
[...]

Issue #759, opened 2026-07-10
feat(engine): watermark.lag — overlap window to catch late-arriving rows in incremental syncs


masukai:
## Problem

Cursor-based incremental uses a strict `WHERE cursor_field > <last_watermark>` predicate. Rows that arrive *late* — upstream batch jobs that backfill yesterday's partition, clock skew between writers, dbt models rebuilt with corrected timestamps — fall permanently behind the watermark and are never synced. For reverse ETL this silently strands CRM/ads updates, and the only remedies today are manual (`--cursor-value` re-run) or nuclear (full re-sync).

## Prior art

- **dlt incremental `lag` (attribution window)**: re-fetches a configurable overlap window behind the cursor on every run, relying on merge/upsert idempotency to make the overlap safe.
- **Fivetran/Airbyte** connectors ship the same concept as "lookback windows".

## Proposal

```yaml
sync:
  mode: incremental
  cursor_field: updated_at
  watermark:
    default_value: "2024-01-01T00:00:00Z"
    lag: 1h            # duration for timestamp cursors, integer for numeric
```

Effective predicate becomes `cursor_field > (last_watermark - lag)`. The stored watermark still advances to the true max — `lag` only widens the read window. Document that `lag` requires an idempotent mode (`upsert`, or `full`-per-window) and warn at validate time if combined with plain append-style `incremental` into REST-ish destinations where duplicates would be user-visible.

## Scope

**In:** duration parsing (reuse `freshness.max_age` units), timestamp + numeric cursors, validate-time warning, docs.
**Out:** per-run adaptive lag; late-arrival *detection* (that's the diff-based strategy's job).

## Related

- #390 (`default_value`), #475 (tz-aware cursor rendering) — same code paths
- Diff-based incremental issue — the general solution for sources with no reliable cursor at all

## Milestone fit

Small, self-contained engine/config change — good `v0.8.x` filler or `good first issue` once the predicate-rendering seam is pointed at.

PR #795, opened 2026-07-10
feat(cli): drt build — run each sync and its tests in one invocation (closes #777)

### LINKED ISSUE #777: feat(cli): drt build — run each sync and its tests in one invocation
## Problem

The canonical loop is `drt run` then `drt test` — two invocations, two selector expressions to keep in sync, and in CI two steps whose interleaving users must script themselves (run all, then test all — even though sync B's tests could run while sync C loads). dbt users type `dbt build` for exactly this.

## Prior art

- **dbt build**: run + test (+snapshot/seed) per node in dependency order — the recommended default command since 1.0.

## Proposal

```bash
drt build [--select ...] [--threads N] [--fail-fast]
```

- Per sync: run → on success, immediately execute that sync's `tests:` (skipping non-queryable destinations exactly as `drt test` does today) → sync's overall status = run ∧ tests
- Failed tests mark the sync failed in state/history (so `--failed` re-runs it) but never roll back loaded data — document this plainly
- Output = unified table (run cols + test cols); `--output json` nests test results under each sync
- Flag parity with `run` (selectors, threads, dry-run = run-dry + test-plan preview via #371's machinery)

## Scope

**In:** command composing existing run/test paths per sync, unified reporting, flag parity, docs + CI guide update.
**Out:** cross-sync ordering (waits for #426), seeds/snapshots analogs (don't exist), retry semantics beyond `--failed` integration.

## Related

- #233 (tests), #371 (test dry-run), #426 (depends_on — would upgrade build to DAG order "for free" later)

## Milestone fit

Candidate: v0.8.x — thin composition over existing machinery; large DX familiarity payoff for the dbt crowd.

masukai:
## Summary

Implements [#777](https://github.com/drt-hub/drt/issues/777) — the `dbt build` muscle-memory command:

```bash
drt build                          # per sync: run → tests
drt build --select tag:crm --fail-fast
drt build --dry-run                # run preview + test plan
```

## Semantics

- **Per sync: run → on success, its `tests:`** — a failed load skips its now-meaningless assertions; non-queryable destinations skip tests exactly like `drt test`
- **A failing test marks the sync failed** — exit non-zero, `status: "tests_failed"`, per-sync `tests` array nested in each `--output json` entry; already-loaded data is **not** rolled back (documented, same as dbt)
- Composes with selection v2, `--fail-fast` (skipped entries), `--dry-run`, `--profile`, `--quiet`
- The per-sync test execution was **extracted from `drt test` into `execute_tests_for_sync()`** — both commands run identical logic (pure refactor; `drt test` behaviour unchanged, suites green)

## Scope note

v1 is **sequential by design** — true `--threads` interleaving of run+test stages needs the run-loop extraction tracked in the #723 refactor family. `drt run --threads` + `drt test` remain the parallel path (noted in module docstring + CHANGELOG).

## Stacked PR

⚠️ Merge order: **#791 → #792 → #793 → #794 → this** (based on `feat/775-fail-fast`). Diff vs #794 is this feature only.

## Tests

6 new in `tests/unit/test_cli_build.py` (composition, `tests_failed` status, failed-run-skips-tests, fail-fast, dry-run plan, selection) + refactored `drt test` suites green. Full suite **2009 passed, 12 skipped**; `ruff` + `mypy` clean.

Closes #777


Muawiya-contact:
**Request changes** — one blocker + one minor bug. The feature design is sound and the `execute_tests_for_sync()` extraction is clean, but the PR doesn't pass its own suite on the current head. All findings are from running the code.

## ① [blocker] All 6 `test_cli_build.py` tests fail on the head
```
6 failed — AttributeError: '_FakeResult' object has no attribute 'watermark_lag'
```
`build.py` reuses `run.py._run_one`, which at `run.py:237` reads `result.watermark_lag` (a real field from #759, already on the base `feat/775-fail-fast` and present on production results + `run`'s own test fakes). The **new** `_FakeResult` in `test_cli_build.py` omits it, so every build test crashes.

- **Production `drt build` is unaffected** — real `SyncResult`s have `watermark_lag`; only the test double is stale. Looks like stacked-base drift (the base gained `watermark_lag` after these tests were written).
- The "2009 passed / suites green" note is stale for this head — its only check-run is `cla`; the test/mypy workflows never ran on `3000569`.
- **One-line fix, verified:** add `self.watermark_lag: str | None = None` to `_FakeResult` (next to `watermark_source`) → all 6 pass.

## ② [medium] `--quiet` doesn't suppress the test output (text mode)
`drt build --quiet` empirically leaks the per-sync test output:
```
QUIET: a_with_tests / ✓ not_null(id): 0     ← run output suppressed, test output still prints
```
`execute_tests_for_sync()` gates prints on `json_mode` only (it predates `--quiet`, since `drt test` has none), and `build.py` passes `quiet` to `_RunContext` but not to the test call — contradicting the `--quiet` help. Untested: all 6 tests use `--output json`, so no text-mode path is covered. Fix: thread `quiet` into `execute_tests_for_sync` (or gate its prints on it too) + add a text-mode `--quiet` assertion.

## Minor / non-blocking
- `execute_tests_for_sync(sync: Any, …)` loosens typing to `Any` — a `SyncConfig` (or `Protocol`) keeps it strict.
- JSON shape is inconsistent: a sync whose **run** failed has no `"tests"` key while run-ok syncs always do — a `"tests": []` default would help consumers.
- The extraction is behavior-preserving for `drt test` (verbatim move, `had_failures` OR-ed by the caller, skip returns `False`) — no concern there.

The run→test composition, fail-fast skip entries, dry-run plan, `tests_failed` status, and no-rollback semantics all read correctly. Once ① and ② are in, this is a solid, well-scoped feature. Nice reuse of `_run_one` + `execute_tests_for_sync` rather than duplicating either loop.

Muawiya-contact (changes requested):
@masukai just do that changes please.

masukai:
All four addressed — thank you, the blocker was a real one and I'd have shipped it broken.

**[blocker] `_FakeResult.watermark_lag`** — fixed. You diagnosed it exactly: `#788` added `watermark_lag` to `SyncResult`, `run.py` reads it, and the fake didn't have it, so all 6 tests died on `AttributeError`. Worth naming *why* it survived to review, because it's a repo-level hole rather than a slip on this PR: **`ci.yml` only triggers on `pull_request: branches: [main]`**, and this PR was stacked on a feature branch — so the full matrix never fired on it, only `cla`. Green checkmarks on a stacked PR here mean almost nothing today. I've now retargeted the whole stack to `main`, and the matrix runs for real. Filing that CI gap separately.

**[medium] `--quiet`** — `execute_tests_for_sync()` takes `quiet` and `drt build --quiet` now silences test output in text mode exactly like `drt run --quiet`. The result dict is untouched, so JSON output and exit codes still carry every failure. Added a text-mode test asserting the loud run prints the test name + build summary and the quiet run prints neither, while both still execute the queries.

**`sync: SyncConfig`** instead of `Any` — done (TYPE_CHECKING import).

**`"tests": []` default** — done, and generalised: *every* entry now carries `tests`, empty when none ran (no tests defined, run failed, or skipped by `--fail-fast`). Two assertions flipped from `"tests" not in entry` to `== []` accordingly.

Local: `1969 passed`, ruff + mypy clean. No rush on the re-look — and if you'd rather leave it, that's fine too, you've already done the expensive part here.

Muawiya-contact (approved):
**Approving — both blockers fixed, verified on the head (`c9a0f9d`).** Thanks for the quick, thorough turnaround, @masukai. Re-ran my original probes rather than taking the fix on trust:

- **① The `watermark_lag` fixture** — all **7** `test_cli_build.py` tests now pass (was 6/6 erroring on `AttributeError` via `_run_one`). The repaired fake result resolves it. ✓
- **② `--quiet` leak** — `execute_tests_for_sync` now takes `quiet` (`show = not json_mode and not quiet`) and `build.py` threads it through, so `drt build --quiet` is fully silent in text mode. Confirmed empirically: LOUD emits the run + test lines + summary, `--quiet` output is now `''`. ✓

Nice extras: you also tightened `sync: Any` → `sync: SyncConfig` (the typing nit I'd flagged) and added a test for the quiet path. CI is green across 3.10–3.13 now that it's rebased onto `main`; `ruff` clean, `mypy` clean on the changed files locally.

One tiny non-blocking note (codecov flagged it, not me): ~16 lines in `build.py` are uncovered — mostly the fail-fast skip-entry branch and the JSON summary path. Totally fine to leave for a follow-up; the core semantics are well covered by the 7 tests.

The run→test composition, graceful no-rollback, `tests_failed` status, and fail-fast skip accounting all read cleanly. LGTM 🚀

masukai:
Thanks for the thorough re-review, @Muawiya-contact — and for re-running your own probes rather than trusting the fix. 🙏

On your one non-blocking note (the ~16 uncovered lines): I decided to just cover them now rather than defer. The new codecov **patch gate** (80%, landed in #803) exists precisely because untested branches hid the `--fail-fast --threads` deadlock — so leaving `build.py`'s own fail-fast/skip branches uncovered would've been a bit hypocritical the day after gating on exactly that.

Added six tests (pushed on top of your approval — test-only, your code is untouched): the `tests_skipped_reason` branch (sync with `tests:` on a non-queryable dest), the text-mode `--fail-fast` skip summary, no-syncs-found, empty-selection-after-exclude, unknown selector, and missing-project. **build.py 81% → 96%**; the last 3 lines are the profile-load failure path, unreachable while the fixture stubs `load_profile`.

CI green across 3.10–3.13. Merging — thanks again for catching the blocker; it was real.

PR #754, opened 2026-07-10
feat(examples): docs-demo — showcase fixture for drt docs generate (10 syncs, lookup lineage, mixed run states)

masukai:
Foundation for the docs-site public demo (plan: showcase the generated site via drt-web, announce after #701's DAG lands).

`examples/docs-demo/` is a fictional but **fully valid** project tuned so `drt docs generate` has something worth showing:

- 10 syncs over postgres / slack / hubspot — modes `full`/`incremental`/`upsert`, `field_mappings`, `mask`
- lookup lineage including the **two-lookups-into-one-sync** case (`orders_to_pg` ← users + products)
- hand-authored `.drt/state.json` (gitignore negation added) → a realistic `success`/`partial`/`failed` mix in Recent Runs; hosts are unresolved env references, nothing connects anywhere
- `test_docs_demo_example` guards it: 10 syncs must validate, lineage edges must exist, site must render — the fixture can't silently rot

Once merged, the drt-web sync workflow can `drt docs generate` here and publish the output under its `static/` (tracked on the drt-web side).

Related: #702 (tracker, showcase line), #510.

@Muawiya-contact FYI more than review — this fixture doubles as a real-shape input for your #701 layout A/B (it's the same shape I offered there, now committed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #796, opened 2026-07-10
feat(docs): static SVG DAG emission from the layout engine (#701)

### LINKED ISSUE #701: feat(docs): static pre-laid project DAG — hand-rolled layout engine replaces runtime Mermaid

Muawiya-contact:
Phase 2 of #701. Consumes the merged layout engine (`drt/docs/layout.py`) and renders its output as one inline, themeable SVG, replacing the runtime-Mermaid DAG page. Draft for pairing on the visual pass (@masukai leads design).

## What it does
- **`drt/docs/dag.py`** — `render_dag_svg(manifest)` calls `compute_layout` (median default) and draws back-to-front:
  - **Ownership zone bands** — sources (external · read) | syncs (violet "managed by drt" band, `--zone-drt-bg`/`--zone-drt-line`) | destinations (external · write).
  - **Edges** from `LayoutEdge.points` — forward beziers styled `--edge`; lookup back-edges as dashed `--edge-lookup` lanes over the top with port dots, both using the marker defs.
  - **Nodes** via `_node_card` — syncs as code cards (doc icon, mono name, mode, status dot+word), sources/destinations as connector cards with `_badge` brand-initial badges. Every node wrapped in `<a>` to its detail page.
- Replaces the Mermaid/CDN DAG page with the static SVG; empty projects fall back to an empty-state card. Works via `file://` — no fetch, no CDN, no runtime layout JS.

## Reusing #704's helpers
Extracted `_node_card`, `_badge`/`_badge_svg`, `_slug`/`_slug_map`, `_clip`, and the marker defs into **`drt/docs/_svg.py`**; `html.py` now imports them from there, so the per-sync ego-graphs and the full DAG share one code path (no duplicated card/badge/escaping logic). `_node_card` gained an optional `status` arg; `_marker_defs(prefix)` takes an id prefix so ego/DAG markers don't collide. `--zone-drt-bg` promoted from the mock into the real token file.

## Determinism (#697)
Same manifest → byte-identical SVG. A test asserts two renders match and that two site builds produce identical `dag.html` bytes. Coordinates come pre-rounded from the engine; iteration follows the layout's stable ordering; no clocks/random/set-ordering.

## Tests
`tests/unit/test_docs_dag.py` (new): byte-identity, zones present, forward+lookup edges, every node `<a>`-wrapped, `file://` safety (no `fetch(`/Mermaid/CDN), empty project, lookup edge uses `--edge-lookup` + marker, escaping. Updated `test_docs_html.py` (Mermaid → static-SVG assertion).

`ruff check` + `mypy drt` clean; full suite green (one pre-existing Windows-only sqlite teardown flake, unrelated).

## For the pairing session
- Dropped the mock's secondary "version-controlled YAML" label under SYNCS — the band can't carry it plus the "managed by drt" pill without overlap. Easy to restore.
- DLQ/drift badges are out of scope (need manifest schema v2, per the mock's footnote).
- Layout engine untouched — consumed, not modified.

Closes #701 (pending design review).

### DIFF
--- a/drt/docs/_svg.py
+++ b/drt/docs/_svg.py
@@ -1,0 +1,175 @@
+"""Shared SVG/emission helpers for the docs site.
+
+Extracted from ``drt.docs.html`` (#704) so the full-DAG emission (#701) and the
+per-sync ego-graphs render nodes, badges, and arrow markers through one code
+path. Everything here is pure string emission — deterministic for a given
+input, all manifest-derived text HTML-escaped at the point of use.
+"""
+
+from __future__ import annotations
+
+import re
+from html import escape
+
+_SLUG_RE = re.compile(r"[^A-Za-z0-9]+")
+
+
+def _slug(value: str) -> str:
+    return _SLUG_RE.sub("-", value).strip("-").lower() or "x"
+
+
+def _slug_map(names: list[str], kind: str) -> dict[str, str]:
+    """Map each name to a stable slug, failing fast on a collision.
+
+    ``_slug`` collapses runs of non-alphanumerics, so punctuation-only-distinct
+    names (``a_b`` vs ``a__b``) can slugify to the same file — which would
+    silently overwrite a page and break cross-links. Detect that and raise.
+    """
+    out: dict[str, str] = {}
+    seen: dict[str, str] = {}
+    for name in names:
+        slug = _slug(name)
+        if slug in seen and seen[slug] != name:
+            raise ValueError(
+                f"Two {kind} names slugify to the same page {slug!r}: "
+                f"{seen[slug]!r} and {name!r}. Rename one so their pages don't collide."
+            )
+        seen[slug] = name
+        out[name] = slug
+    return out
+
+
+# Connector badges — brand-color initials (decision: docs design pass 1).
+# Curated map for known types; anything else falls back to a neutral brand
+# badge with the first two letters, so new plugins render without changes.
+_BADGES: dict[str, tuple[str, str, str]] = {
+    "bigquery": ("BQ", "#4285f4", "#ffffff"),
+    "postgres": ("PG", "#336791", "#ffffff"),
+    "duckdb": ("DK", "#fff100", "#3a3a00"),
+    "snowflake": ("SF", "#29b5e8", "#062d3d"),
+    "databricks": ("DX", "#ff3621", "#ffffff"),
+    "clickhouse": ("CH", "#faff69", "#3a3d00"),
+    "sqlite": ("SQ", "#0f80cc", "#ffffff"),
+    "mysql": ("MY", "#00758f", "#ffffff"),
+    "elasticsearch": ("ES", "#00bfb3", "#00312e"),
+    "s3": ("S3", "#ff9900", "#3e2500"),
+    "gcs": ("GC", "#4285f4", "#ffffff"),
+    "azure_blob": ("AZ", "#0078d4", "#ffffff"),
+    "slack": ("SL", "#4a154b", "#ffffff"),
+    "discord": ("DC", "#5865f2", "#ffffff"),
+    "hubspot": ("HS", "#ff7a59", "#3e1c00"),
+    "salesforce": ("SF", "#00a1e0", "#ffffff"),
+    "airtable": ("AT", "#fcb400", "#3d2c00"),
+    "klaviyo": ("KL", "#232426", "#ffffff"),
+    "rest_api": ("API", "#5a6068", "#ffffff"),
+    "webhook": ("WH", "#5a6068", "#ffffff"),
+}
+
+# Status dot/word pairs (never color-alone) — token names from STYLE_CSS.
+_STATUS_VARS: dict[str, str] = {
+    "success": "--success",
+    "partial": "--warning",
+    "failed": "--error",
+}
+
+
+def _badge(conn_type: str) -> tuple[str, str, str]:
+    known = _BADGES.get(conn_type)
+    if known:
+        return known
+    initials = (conn_type[:2] or "??").upper()
+    return (initials, "#7c3aed", "#ffffff")
+
+
+def _badge_svg(conn_type: str, x: int, y: int) -> str:
+    """A 26x26 brand-initial badge at (x, y). All text escaped."""
+    initials, bg, fg = _badge(conn_type)
+    fs = 9 if len(initials) > 2 else 10
+    return (
+        f'<rect x="{x}" y="{y}" width="26" height="26" rx="7" fill="{bg}"/>'
+        f'<text x="{x + 13}" y="{y + 17}" font-size="{fs}" font-weight="700" '
+        f'class="mono" fill="{fg}" text-anchor="middle">{escape(initials)}</text>'
+    )
+
+
+def _clip(value: str, limit: int = 24) -> str:
+    return value if len(value) <= limit else value[: limit - 1] + "…"
+
+
+def _marker_defs(prefix: str, size: int = 6) -> str:
+    """Arrowhead ``<defs>`` for forward (``{prefix}-arr``) and lookup
+    (``{prefix}-arr-lk``) edges, colored by the ``--edge`` / ``--edge-lookup``
+    tokens. ``prefix`` keeps marker ids unique per SVG."""
+    return (
+        "<defs>"
+        f'<marker id="{prefix}-arr" viewBox="0 0 8 8" refX="7" refY="4" markerWidth="{size}" '
+        f'markerHeight="{size}" orient="auto-start-reverse">'
+        '<path d="M0,0.6 L7.4,4 L0,7.4 Z" fill="var(--edge)"/></marker>'
+        f'<marker id="{prefix}-arr-lk" viewBox="0 0 8 8" refX="7" refY="4" markerWidth="{size}" '
+        f'markerHeight="{size}" orient="auto-start-reverse">'
+        '<path d="M0,0.6 L7.4,4 L0,7.4 Z" fill="var(--edge-lookup)"/></marker>'
+        "</defs>"
+    )
+
+
+def _node_card(
+    x: int,
+    y: int,
+    w: int,
+    conn_type: str,
+    name: str,
+    sub: str,
+    href: str | None,
+    *,
+    code: bool = False,
+    status: str | None = None,
+) -> str:
+    """One 54px-high node card. ``code=True`` renders the drt-managed sync style
+    (violet accent + doc icon); otherwise a connector card with a brand badge.
+    ``status`` adds a right-aligned dot + word pair colored by the status tokens.
+    """
+    accent = (
+        f'<rect x="{x}" y="{y}" width="3" height="54" rx="1.5" fill="var(--brand-600)"/>'
+        if code
+        else ""
+    )
+    stroke = "var(--zone-drt-line)" if code else "var(--line)"
+    if code:
+        icon = (
+            f'<g transform="translate({x + 13},{y + 14})">'
+            '<path d="M0,1.5 a1.5,1.5 0 0 1 1.5,-1.5 h6 l4,4 v11 a1.5,1.5 0 0 1 -1.5,1.5 '
+            'h-8.5 a1.5,1.5 0 0 1 -1.5,-1.5 z" fill="none" stroke="var(--brand-600)" '
+            'stroke-width="1.4" stroke-linejoin="round"/>'
+            '<path d="M7.5,0 v4 h4" fill="none" stroke="var(--brand-600)" '
+            'stroke-width="1.4" stroke-linejoin="round"/>'
+            '<path d="M3,9 h6 M3,12 h4" stroke="#a78bfa" stroke-width="1.2" '
+            'stroke-linecap="round"/></g>'
+        )
+        tx = x + 34
+    else:
+        icon = _badge_svg(conn_type, x + 12, y + 14)
+        tx = x + 48
+    name = _clip(name)
+    sub = _clip(sub, 16 if status else 30)
+    status_svg = ""
+    if status:
+        var = _STATUS_VARS.get(status, "--muted")
+        sx = x + w - 62
+        status_svg = (
+            f'<circle cx="{sx}" cy="{y + 36.5}" r="3.5" fill="var({var})"/>'
+            f'<text x="{sx + 8}" y="{y + 40}" font-size="10" font-weight="500" '
+            f'fill="var({var})">{escape(_clip(status, 8))}</text>'
+        )
+    body = (
+        f'<rect x="{x}" y="{y}" width="{w}" height="54" rx="8" '
+        f'fill="var(--surface)" stroke="{stroke}"/>'
+        f"{accent}{icon}"
+        f'<text x="{tx}" y="{y + 23}" font-size="12.5" font-weight="600" '
+        f'class="mono" fill="var(--fg)">{escape(name)}</text>'
+        f'<text x="{tx}" y="{y + 40}" font-size="10" letter-spacing="0.6" '
+        f'fill="var(--muted)">{escape(sub.upper())}</text>'
+        f"{status_svg}"
+    )
+    if href:
+        return f'<a href="{escape(href, quote=True)}">{body}</a>'
+    return body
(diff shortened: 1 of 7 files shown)


Muawiya-contact:
@masukai draft PR, any feedback?

masukai:
Reviewed the **code** side end-to-end (checked out locally, ran the suite, read `dag.py` / `_svg.py` / the `html.py` wiring in full). The **visual/design** pass is @masukai's — this comment is just the correctness half so it's not blocking on me.

## ✅ Code looks solid
- **Tests**: `test_docs_dag.py` + the updated `test_docs_html.py` → 27 passed locally; `ruff` + `mypy drt` clean. Good coverage — byte-identity determinism (#697), empty-project fallback, `file://` safety (no `fetch(`/mermaid/CDN), and escaping are all asserted.
- **Injection safety**: every manifest-derived string (`name`, `sub`, `status`, badge initials, `href` with `quote=True`) goes through `escape()` at the point of use — no sync name / connector type / dest label reaches SVG markup or an attribute raw. 👍
- **Refactor is clean**: `_slug` / `_node_card` / `_badge*` / marker defs moved into `_svg.py` and the old copies **removed** from `html.py` (now imported), so the ego-graphs and the full DAG share one path — no drift surface.
- **Wiring**: `render_html()` drops `render_mermaid` and calls `render_dag_svg(manifest) if manifest.syncs else None`; the `render_mermaid` / `import re` imports are gone. Empty projects hit the `{% else %}` "No syncs found" state.
- `_slug_map` fails fast on slug collisions — nice touch, prevents a silent page overwrite.

## ⚠️ The two Copilot comments are both false positives — don't chase them
1. *"nothing calls `render_dag_svg` when building the site"* — it's wired in `render_html()` (`dag_svg=render_dag_svg(...)`), Mermaid removed.
2. *"`html.py` still has its own `_slug`/`_node_card`/marker copies"* — those are deleted from `html.py` and imported from `_svg.py`.

Both look like a read of an intermediate commit. Safe to resolve/dismiss.

## Minor (non-blocking, your call)
- `dag.py` `_f()` uses `format(value, "g")`, which switches to scientific notation at ≥ `1e6` (`format(1000000.0, "g")` → `"1e+06"`) — invalid as an SVG coordinate. Only reachable on an absurdly large DAG (>~999,999px), so realistically never hit; flagging as a latent edge case only. A `f"{value:.2f}".rstrip("0").rstrip(".")`-style format would sidestep it if you ever want to close it.

Nothing here blocks. @masukai has the design/visual pass (zone-band palette, the "managed by drt" pill placement, dropping the "version-controlled YAML" secondary label) — I'm deliberately leaving that to them. Nice work on the determinism + escaping discipline. 🙂

drt/docs/dag.py
```diff
@@ -0,0 +1,193 @@
...
+    dest_by_id = {d.name: d for d in manifest.destinations}
+
+    width = layout.width
+    height = layout.height + _HEADER_H
+    card_w = round(cfg.node_w)
+
+    parts: list[str] = [
+        f'<svg viewBox="0 0 {_f(width)} {_f(height)}" width="{_f(width)}" '
+        f'height="{_f(height)}" role="img" aria-label="Lineage graph: sources feed '
+        'drt-managed syncs which write to destinations">',
```
masukai: **Responsive width — keeps the destination rank on-screen.** The SVG is emitted with a fixed `width="1044"` / `max-width:none`, but the docs content column caps at ~914px even on a 1920px screen, so the whole DESTINATIONS rank (where data lands) sits off-screen behind a horizontal scroll on first view. Dropping the fixed px size and letting `viewBox` + a `max-width` style scale it fixes that — measured 1044px → 894px in the demo column, no scroll, byte output unchanged (style attr only). The existing `.dag { overflow-x:auto }` still catches genuinely tiny viewports.

```suggestion
        # Responsive: viewBox carries the intrinsic geometry; the style scales
        # the SVG down to its column and never upscales past natural size, so
        # the destination rank stays on-screen instead of overflowing a
        # narrower content column (the `.dag` overflow-x:auto only engages on
        # genuinely tiny viewports). No fixed width/height px attributes.
        f'<svg viewBox="0 0 {_f(width)} {_f(height)}" '
        f'style="width:100%;height:auto;max-width:{_f(width)}px" '
        'role="img" aria-label="Lineage graph: sources feed '
        'drt-managed syncs which write to destinations">',
```

drt/docs/dag.py
```diff
@@ -0,0 +1,193 @@
...
+            f'<text x="{_f(left + 14)}" y="32" font-size="11" letter-spacing="1.2" '
+            f'fill="{title_fill}" font-weight="600">{title}</text>'
+        )
+        if is_sync:
+            parts.append(
+                f'<rect x="{_f(left + 66)}" y="19" width="92" height="18" rx="9" '
+                'fill="var(--brand-600)"/>'
+                f'<text x="{_f(left + 112)}" y="32" font-size="10.5" fill="#fff" '
+                'text-anchor="middle" font-weight="600">managed by drt</text>'
+            )
```
masukai, line 91: **Restore the mock's second SYNCS descriptor.** "version-controlled YAML" was dropped because the band can't carry it beside the "managed by drt" pill on one row — placing it as a muted line *beneath* the pill sidesteps the overlap and matches the sources/destinations secondary treatment. Verified it clears both the pill and the first sync card in the demo render.

```suggestion
            # "managed by drt" pill on the title row, then the mock's second
            # descriptor as a muted line beneath it (the band is too narrow to
            # carry both on one row — see the sources/destinations secondary,
            # which is right-aligned on the title row instead).
            parts.append(
                f'<rect x="{_f(left + 66)}" y="19" width="92" height="18" rx="9" '
                'fill="var(--brand-600)"/>'
                f'<text x="{_f(left + 112)}" y="32" font-size="10.5" fill="#fff" '
                'text-anchor="middle" font-weight="600">managed by drt</text>'
                f'<text x="{_f(left + 14)}" y="44" font-size="9.5" letter-spacing="0.3" '
                'fill="var(--muted)">version-controlled YAML</text>'
            )
```

drt/docs/_svg.py
```diff
@@ -0,0 +1,175 @@
...
+    "azure_blob": ("AZ", "#0078d4", "#ffffff"),
+    "slack": ("SL", "#4a154b", "#ffffff"),
+    "discord": ("DC", "#5865f2", "#ffffff"),
+    "hubspot": ("HS", "#ff7a59", "#3e1c00"),
+    "salesforce": ("SF", "#00a1e0", "#ffffff"),
+    "airtable": ("AT", "#fcb400", "#3d2c00"),
+    "klaviyo": ("KL", "#232426", "#ffffff"),
+    "rest_api": ("API", "#5a6068", "#ffffff"),
+    "webhook": ("WH", "#5a6068", "#ffffff"),
+}
```
masukai, line 97: **Expand the badge map to every connector (18 → 40).** Right now ~20 registered connectors (twilio, notion, mixpanel, teams, zendesk, jira, linear, redshift, sqlserver, deltalake, iceberg, …) fall through to the neutral 2-letter badge, and the `salesforce` key never matched (the registered type is `salesforce_bulk`) so even Salesforce was neutral. This gives each a brand-coloured badge; the four genuinely generic ones (file / email_smtp / staged_upload / rest_api) stay neutral on purpose. One design call flagged inline: snowflake moves `SF → SN` so the now-active `salesforce_bulk` keeps the canonical `SF` — revert if you'd rather Snowflake keep SF.

```suggestion
_BADGES: dict[str, tuple[str, str, str]] = {
    # Warehouses / databases
    "bigquery": ("BQ", "#4285f4", "#ffffff"),
    "postgres": ("PG", "#336791", "#ffffff"),
    "redshift": ("RS", "#8c4fff", "#ffffff"),
    "duckdb": ("DK", "#fff100", "#3a3a00"),
    # NB: snowflake moved SF -> SN so the now-active salesforce_bulk keeps the
    # canonical "SF". Revert if design prefers SF for Snowflake instead.
    "snowflake": ("SN", "#29b5e8", "#062d3d"),
    "databricks": ("DX", "#ff3621", "#ffffff"),
    "clickhouse": ("CH", "#faff69", "#3a3d00"),
    "sqlite": ("SQ", "#0f80cc", "#ffffff"),
    "mysql": ("MY", "#00758f", "#ffffff"),
    "sqlserver": ("MS", "#a91d22", "#ffffff"),
    # Lakehouse / files / object stores
    "deltalake": ("DL", "#00add4", "#00303d"),
    "iceberg": ("IB", "#2c88d9", "#ffffff"),
    "parquet": ("PQ", "#50abf1", "#0a2740"),
    "file": ("FL", "#5a6068", "#ffffff"),
    "s3": ("S3", "#ff9900", "#3e2500"),
    "gcs": ("GC", "#4285f4", "#ffffff"),
    "azure_blob": ("AZ", "#0078d4", "#ffffff"),
    "elasticsearch": ("ES", "#00bfb3", "#00312e"),
    # Messaging / collaboration
    "slack": ("SL", "#4a154b", "#ffffff"),
    "discord": ("DC", "#5865f2", "#ffffff"),
    "teams": ("TM", "#6264a7", "#ffffff"),
    "email_smtp": ("EM", "#5a6068", "#ffffff"),
    "twilio": ("TW", "#f22f46", "#ffffff"),
    "sendgrid": ("SG", "#1a82e2", "#ffffff"),
    # CRM / marketing / product analytics
    "hubspot": ("HS", "#ff7a59", "#3e1c00"),
    "salesforce_bulk": ("SF", "#00a1e0", "#ffffff"),
    "intercom": ("IC", "#1f8ded", "#ffffff"),
    "klaviyo": ("KL", "#232426", "#ffffff"),
    "airtable": ("AT", "#fcb400", "#3d2c00"),
    "notion": ("NO", "#191919", "#ffffff"),
    "amplitude": ("AM", "#1f6fff", "#ffffff"),
    "mixpanel": ("MX", "#7856ff", "#ffffff"),
    "google_ads": ("GA", "#4285f4", "#ffffff"),
    "google_sheets": ("GS", "#0f9d58", "#ffffff"),
    # Issue trackers / other
    "jira": ("JR", "#0052cc", "#ffffff"),
    "linear": ("LN", "#5e6ad2", "#ffffff"),
    "github_actions": ("GH", "#2088ff", "#ffffff"),
    "zendesk": ("ZD", "#03363d", "#ffffff"),
    "staged_upload": ("SU", "#5a6068", "#ffffff"),
    "rest_api": ("API", "#5a6068", "#ffffff"),
}
```

masukai (commented):
Paired on the visual pass — rendered your branch against the 10-sync demo (`examples/docs-demo`, pulled from #754) and it holds up well: ownership zones, status dots, lookup lanes and dark mode all read exactly like the mock. Three suggestions below, each implemented + screenshotted locally before posting:

1. **Destination clipping** — the write side is off-screen on first open (fixed 1044px SVG in a ~914px column). Responsive-width fix.
2. **Second SYNCS label** — restore "version-controlled YAML" as a muted line under the pill.
3. **Connector badges** — 18 → 40 so every connector is branded (and `salesforce_bulk` stops falling through).

All three keep determinism (byte-identical output) and I added two guard tests locally — `test_badge_keys_match_registered_connectors` (pins badge keys to real connector types) and `test_svg_is_responsive_not_fixed_width` — happy to push those too, or you can grab them from `design/796-clip-badges`. Copilot's two comments above are both false positives (renderer *is* wired; helpers *are* deduped). Design approval stays yours — shout if you'd rather adjust any of these.

Muawiya-contact:
@masukai Thanks both — really glad the determinism + escaping held up, and @masukai that the zones/lanes/dark-mode read like the mock against the real demo. 🙏

All three of your fixes make sense and I'd love to take them:
1. **Responsive width** — yes, the destination rank being off-screen defeats the point; dropping the fixed px for viewBox + max-width is the right call.
2. **Second SYNCS line** — restore "version-controlled YAML" as the muted line under the pill, matching the sources/dest secondary treatment.
3. **Badges 18 → 40** — agreed, every connector branded, and it fixes salesforce_bulk falling through.

I'll pull them from `design/796-clip-badges` onto this branch (with your two guard tests — `test_badge_keys_match_registered_connectors` + `test_svg_is_responsive_not_fixed_width`), so it all lands in one PR. Also closing the `_f()` scientific-notation edge case the correctness pass flagged (`format(…, "g")` → scientific at ≥1e6) with a `.2f`-style format while I'm in there — latent, but free to harden.

And resolving Copilot's two — both false positives (renderer's wired, helpers deduped), as you both noted. Will push shortly.

masukai:
Perfect — thanks for taking all three, and especially for closing the `_f()` scientific-notation case while you're in there 🙏 Landing them on your branch with the two guard tests keeps it one clean PR. `design/796-clip-badges` is pushed and current if you want to cherry-pick straight from it. Ping me if the guard tests need any tweak once they're on your side — otherwise I'll re-review once you push. 💜

masukai (approved):
**Approve** — re-reviewed the integration (`344142b`), all three suggestions landed faithfully and the coord hardening is a nice bonus. Verified locally:

- **Responsive width ✓** — `viewBox` + `style="width:100%;height:auto;max-width:…px"`, no fixed `width`/`height` px. Destination rank now scales into the column instead of scrolling off-screen.
- **Second SYNCS label ✓** — "version-controlled YAML" restored as the muted line under the pill.
- **Badges 18 → 40 ✓** — every registered connector branded, `salesforce_bulk` no longer falls through; my two guard tests (`test_badge_keys_match_registered_connectors`, `test_svg_is_responsive_not_fixed_width`) are in and passing.
- **`_f()` hardening ✓** — nice, and thanks for adding a test. Confirmed `_f(1000000.0) == "1000000"` (not `1e+06`), `_f(84.0) == "84"`, `_f(84.5) == "84.5"` — fixed-point, byte-identical to the old `"g"` form for normal coords, so determinism holds.

`ruff` + `mypy` clean, 30 docs tests green, determinism assertion intact.

**One tiny nit (non-blocking):** the `_ZONES` comment (around L27–29) still reads "the mock's 'version-controlled YAML' tag is dropped" — now stale, since the label is restored in the `is_sync` branch. Worth a one-line tidy next time you're in there so the comment doesn't contradict the code.

The DAG reads exactly like the mock now against the demo — zones, lanes, dark mode, full badges, and nothing clipped. Great to see it land. 🙏

Issue #721, opened 2026-07-07
refactor(config): split config/models.py (1330L) + extract Profile dataclasses from credentials.py


masukai:
## Summary

Two oversized config modules with clean cohesion seams.

### `drt/config/models.py` (1330 lines, ~34 config classes)
One module holds every destination config (SaaS webhooks, SQL, cloud storage, staged upload) plus all sync-option models (`SyncOptions`, `MirrorConfig`, `MaskRule`, alerts, tests). Natural split:
- `config/destinations_sql.py` — `BaseSqlDestinationConfig` family + Snowflake/Databricks/BigQuery/Elasticsearch
- `config/destinations_saas.py` — Slack/Discord/HubSpot/Zendesk/… webhooks
- `config/destinations_storage.py` — S3/GCS/Azure/Parquet/File
- `config/sync_options.py` — RetryConfig … SyncConfig

Secondary win: **26 of 34 `describe()` methods are the identical one-liner** `return f"{self.type} (...)"` — a small `DescribableConfig` base removes the boilerplate. Keep `models.py` re-exporting to preserve the public import surface. Note the 34-member discriminated union (`models.py:1012-1049`) is hand-maintained in parallel with the registry — a standing drift risk worth a comment/guard.

### `drt/config/credentials.py` (814 lines)
Mixes 13 `*Profile` dataclasses (`55-259`) with the env-resolution + secret-loading logic. Extract the dataclasses into `config/profiles.py` to clarify the boundary; they each have a trivial `describe()` that can share the same base as above.

**Scope note:** the `load_profile`/`save_profile` registry-driven rewrite is already tracked in **#402** — this issue is the *module split + describe() dedup only*, not the dispatch rewrite.

**Effort/risk:** M / LOW (mechanical, but touches import surface — needs re-exports + a coverage pass).

Muawiya-contact:
Taking this — module split + `describe()` dedup only, staying clear of the load/save dispatch rewrite (#402). Planning to do it as two PRs: `models.py` split (the 4 destination modules + sync_options.py, with models.py re-exporting to keep the import surface intact) first, then `credentials.py` → profiles.py as a follow-up. Full test suite green is the acceptance bar for the re-exports. Will flag the hand-maintained discriminated union drift risk you noted while I'm in there.

masukai:
🙌 exactly the right framing — the two-PR split + full-suite-green acceptance bar is the way, and staying clear of the #402 dispatch rewrite keeps it mechanical.

[...]

PR #798, opened 2026-07-12
refactor(config): split models.py into destination/sync-option modules (#721)

Muawiya-contact:
**Phase 1 of 2** off #721 — split `drt/config/models.py`; the credentials→profiles extraction is the follow-up PR. Draft for review (@masukai).

## What changed
Split the ~1300-line `drt/config/models.py` into cohesive modules; `models.py` becomes a **pure re-export facade** with an explicit `__all__` (72 names = the documented public surface). No class behavior, fields, or validation change — a mechanical move plus a `describe()` base.

- `base.py` — `DescribableConfig` (new) + auth, pagination, `RetryConfig`, `LookupConfig`, `SslConfig`, project/source primitives
- `destinations_sql.py` — `BaseSql` family + Snowflake / Databricks / BigQuery / Elasticsearch
- `destinations_saas.py` — webhook / CRM / messaging / issue-tracker / email / staged-upload
- `destinations_storage.py` — S3 / GCS / Azure / Parquet / File
- `sync_options.py` — retry consumers, tests, alerts, the `DestinationConfig` union, `SyncOptions` / `SyncConfig`
- `models.py` — re-exports all 72 + `__all__`

A shared `base.py` was necessary: `RetryConfig` is used by ~19 destination configs while `SyncConfig` needs the `DestinationConfig` union, so keeping `RetryConfig` in `sync_options` would be a circular import. Lifting the shared primitives into `base.py` gives a clean DAG (`base → destinations_* → sync_options → models`).

`DescribableConfig` dedups the **26** trivial `f"{type} (detail)"` `describe()` one-liners via a `_describe_detail` hook; **8** connectors with a hardcoded label keep their own.

## Verification
- **Whole-package `python -m mypy drt/` clean — 0 `no_implicit_reexport` errors across 148 files** (the #729 trap: single-file mypy would pass while the 135 import sites fail without the explicit `__all__`). The only remaining error is the pre-existing `credentials.py:368 [no-any-return]` — that file is **byte-identical to `main`** (`git diff` empty).
- **Structural parity proven:** snapshotted every model's fields / defaults / validators / JSON-schema before vs after — **67/67 schemas identical, all validators identical**. `describe()` output identical across all 34 destination configs.
- **Full `pytest -q`:** green except the known Windows `test_sqlite.py` `os.unlink` `PermissionError` flake (reproduces on clean `main`, passes on CI Linux).
- ruff clean.

## Registry-parity note
A comment at the `DestinationConfig` union points at `tests/unit/test_cli_list_connectors.py::test_DESTINATIONS_matches_registry`, which guards the hand-maintained union/registry drift.

## Two honest notes worth surfacing
1. **Public-surface count.** I enumerated the surface at **72** names; the tracking figure was 74. The gap looks like the new internal `DescribableConfig` (intentionally *not* exported — reachable via `drt.config.base`) plus a counting difference. Every name imported anywhere else is re-exported, and whole-package mypy confirms nothing is dropped — flagging so we can reconcile the 74.
2. **`config_class` output shift.** The split moves classes to real submodules, so `drt destinations --detailed --format json` now reports e.g. `drt.config.destinations_sql.PostgresDestinationConfig` (was `drt.config.models.…`). Both import paths work (re-export). One test asserted the old path; I updated it to the truthful new path rather than pin `__module__`. Called out since it's a (minor) observable output change.
3. **`retry` ForwardRef → eager.** Incidentally, `RetryConfig` was defined *after* its ~19 users in the old file, so their `retry:` field stayed an unresolved `ForwardRef` in `model_fields`; importing `RetryConfig` at the top of each module now resolves it eagerly. **JSON-schema is identical**, so validation is unchanged — reads as a latent tidy, not a behavior change, but noting it.

Acceptance bar met: whole-package mypy clean + full suite green (minus the known sqlite flake) + before/after schema parity.

### DIFF
--- a/drt/config/base.py
+++ b/drt/config/base.py
@@ -1,0 +1,193 @@
+"""Shared config primitives for drt (auth, pagination, retry, lookups, project).
+
+Phase 1 of #721 lifted these out of the former monolithic ``models.py`` so the
+destination and sync-option modules can share them without a circular import
+(destinations need :class:`RetryConfig`; ``sync_options`` needs the destination
+union). ``models.py`` re-exports everything here — import sites are unchanged.
+"""
+
+from __future__ import annotations
+
+from typing import Annotated, Literal
+
+from pydantic import BaseModel, Field, model_validator
+
+
+class DescribableConfig(BaseModel):
+    """Mixin base for destination configs whose ``describe()`` is the canonical
+    ``f"{type} (detail)"`` shape (#721). Trivial subclasses supply
+    ``_describe_detail``; connectors with a non-standard label (a hardcoded name
+    or no parens) override ``describe()`` and inherit :class:`BaseModel` directly.
+    """
+
+    def describe(self) -> str:
+        return f"{self.type} ({self._describe_detail()})"  # type: ignore[attr-defined]
+
+    def _describe_detail(self) -> str:  # pragma: no cover - always overridden
+        raise NotImplementedError
+
+
+class BearerAuth(BaseModel):
+    type: Literal["bearer"]
+    token: str | None = None
+    token_env: str | None = None
+
+
+class ApiKeyAuth(BaseModel):
+    type: Literal["api_key"]
+    header: str = "X-API-Key"
+    value: str | None = None
+    value_env: str | None = None
+
+
+class BasicAuth(BaseModel):
+    type: Literal["basic"]
+    username_env: str
+    password_env: str
+
+
+class OAuth2ClientCredentialsAuth(BaseModel):
+    type: Literal["oauth2_client_credentials"]
+    token_url: str
+    client_id_env: str
+    client_secret_env: str
+    scope: str | None = None
+
+
+AuthConfig = Annotated[
+    BearerAuth | ApiKeyAuth | BasicAuth | OAuth2ClientCredentialsAuth,
+    Field(discriminator="type"),
+]
+
+
+class OffsetPaginationConfig(BaseModel):
+    type: Literal["offset"]
+    limit: int = 100
+    offset_param: str = "offset"
+    limit_param: str = "limit"
+    max_pages: int = 100
+
+
+class CursorPaginationConfig(BaseModel):
+    type: Literal["cursor"]
+    limit: int = 100
+    cursor_param: str = "cursor"
+    limit_param: str = "limit"
+    cursor_field: str
+    max_pages: int = 100
+
+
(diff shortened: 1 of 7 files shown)


masukai (approved):
**Approve.** Reproduced the parity claims independently rather than trusting the description — a 1300-line core split is exactly where a byte-level check earns its keep. All green.

**Verified locally (on `94194cc`):**
- **Schema parity — 67/67 byte-identical.** Dumped `model_json_schema()` for every `BaseModel` in `drt.config.models` on `main` and on this branch, diffed sorted JSON: **zero models missing, zero new, zero differing.** The public model surface is provably unchanged.
- **Whole-package `mypy drt/` clean — 148 files, 0 errors.** Confirms the `__all__` facade resolves the #729 `no_implicit_reexport` trap (single-file mypy would have masked it). The only remaining note is the pre-existing `credentials.py:368`, and `git diff main -- drt/config/credentials.py` is empty — not this PR.
- **Full suite: 2005 passed, 12 skipped** (deselected the known Windows `test_sqlite.py os.unlink` flake, which reproduces on clean `main`). `ruff` clean.
- **My recent config additions survived the move:** `WatermarkConfig.lag` (#788) is in `sync_options.py`, `RestIncrementalConfig` (#789) is in `base.py`, both re-exported — good, since this branch is current with `main`.

**On your three honest notes:**
1. **72 vs 74** — confirmed `len(__all__) == 72` (67 classes + 5 unions/aliases). Since whole-package mypy is clean and 67/67 schemas match, nothing importable was dropped; the 74 was a stale tracking figure. No action.
2. **`config_class` path shift** — agreed the truthful `drt.config.destinations_sql.…` is better than pinning `__module__`; both import paths work via the facade, and `test_cli_list_connectors` is updated. It's the one observable output change (anyone parsing `drt destinations --detailed --format json` for the class path sees the submodule now) — worth a one-line CHANGELOG mention if we ship it in a release, but not a blocker.
3. **retry ForwardRef → eager** — schema-identical (part of the 67/67), so validation is unchanged; reads as a latent tidy, agreed.

The `base → destinations_* → sync_options → models` DAG is the right call for the `RetryConfig`↔`DestinationConfig` cycle, and folding the 26 trivial `describe()` one-liners into `DescribableConfig` while leaving the 8 hard-coded labels alone is a clean line to draw. Nice, careful work — happy to see this land. 🙏

PR #799, opened 2026-07-14
refactor(config): extract Profile dataclasses from credentials.py into profiles.py (#721)

### LINKED ISSUE #721: refactor(config): split config/models.py (1330L) + extract Profile dataclasses from credentials.py

Muawiya-contact:
**Phase 2 of 2** of #721 — `Closes #721`. Extracts the typed source-profile dataclasses out of `credentials.py` into a new `profiles.py`, behind a re-export facade so every existing import keeps working. Draft for review (@masukai). Same discipline as Phase 1 (the models.py split): a mechanical move whose whole job is keeping the public import surface intact.

## Module mapping
- **`drt/config/profiles.py`** (new) — the 13 `@dataclass` profiles (`BigQuery` · `DuckDB` · `SQLite` · `Postgres` · `Redshift` · `ClickHouse` · `MySQL` · `Snowflake` · `SQLServer` · `Databricks` · `RestApi` · `DeltaLake` · `Iceberg`) + the `ProfileConfig` union. Pure data — imports only `dataclasses` + `typing`, nothing from `credentials.py`, so no circular import.
- **`drt/config/credentials.py`** — keeps env resolution / secret loading / `load_profile`·`save_profile` dispatch and `OtelConfig`/`ObservabilityConfig`, unchanged; re-exports the profile names behind an explicit 25-name `__all__`.

`__all__` (the full public surface): the 13 profiles, `ProfileConfig`, `OtelConfig`, `ObservabilityConfig`, and `resolve_env` · `load_snowflake_private_key` · `resolve_env_dict` · `load_observability_config` · `load_profile` · `save_profile` · `load_raw_profiles` · `write_raw_profile` · `remove_profile`.

## Issue 

Closes #721 

## Verification
- **Whole-package `python -m mypy drt/` clean — 0 `no_implicit_reexport` errors across 151 files** (the #729 trap: the explicit `__all__` is what makes the re-exports count under strict mypy; single-file mypy would pass while the ~73 import sites fail without it). The only remaining error is the pre-existing `credentials.py:175 [no-any-return]` in `load_snowflake_private_key` — that function is **byte-identical to `main`** (its line shifted from 368→175 only because the profiles above it moved out); not introduced here.
- **Structural parity proven:** snapshotted every profile's `dataclasses.fields()` (names / types / defaults / factories) + `describe()` before vs after — **0 diffs across 14 entries; `describe()` identical 13/13**. The only change is the module path in the `ProfileConfig` union repr.
- **ruff clean**; full `pytest -q` green — `1876 passed`, the only 6 failures are the known Windows `test_sqlite.py` `os.unlink` `PermissionError` flake (reproduces on clean `main`, passes on CI Linux).

## Honest note — no `describe()` dedup this phase
The issue suggested folding the profiles' trivial `describe()` onto `DescribableConfig` (added in Phase 1's `base.py`). That base is a **pydantic `BaseModel`**, but the profiles are **stdlib `@dataclass`es** — inheriting the pydantic base would require converting them, which changes construction / validation / `__eq__` / `dataclasses.fields()` introspection (the last is used by `save_profile`). That's a behavior change this mechanical PR deliberately avoids, so each profile keeps its own `describe()`. A dataclass-friendly describe dedup can be a separate follow-up if wanted.

## Other notes
- Two private helpers (`_load_secrets`, `_load_profiles_yaml`) intentionally **stay in `credentials.py`** — they're secret-loading, not profiles, and some tests import them directly, so they remain importable (not in `__all__`, which is correct — they're private).
- Caught and fixed a generator gotcha before it shipped: `ast.get_source_segment` silently drops decorators, which first produced non-`@dataclass` classes; regenerated with decorator-inclusive slices and verified all 13 are dataclasses again.

Acceptance bar met (same as Phase 1): whole-package mypy clean + full suite green + before/after structural parity.


Muawiya-contact:
@masukai need your eyes on it.

masukai (dismissed):
Reviewed this properly rather than taking the parity claim on faith — and it holds up. I re-derived it independently rather than reading yours:

- **Structural parity ✅** — snapshotted `dataclasses.fields()` (name / type / default / `default_factory`) for all 13 profiles on `main` and on this branch and diffed: **identical, 13/13**. Nothing silently changed shape.
- **`__all__` ✅** — 25 names, and whole-package mypy is clean, so the `no_implicit_reexport` trap from #729 is genuinely closed. Good instinct to make that explicit; single-file mypy would have lied to you here.
- **Docs import paths ✅** — `from drt.config.credentials import ProfileConfig` / `resolve_env` (docs/guides + docs/plans) still resolve through the facade.
- **The `ast.get_source_segment` decorator gotcha** — nice catch, and thanks for writing it down. That one silently produces plain classes and everything *looks* fine until `dataclasses.fields()` explodes at runtime.

## One thing the PR body doesn't mention

`drt sources --detailed` (text **and** `--output json`) emits a `config_class` string built from `cls.__module__`:

```
drt/cli/_connector_detail.py:206
    config_class=f"{cls.__module__}.{cls.__qualname__}",
```

`SOURCE_CONFIG_CLASSES` points at the profile dataclasses, so moving them changes that string for **all 13 sources**:

```
main       bigquery -> drt.config.credentials.BigQueryProfile
this PR    bigquery -> drt.config.profiles.BigQueryProfile
```

Tests didn't catch it because `test_cli_list_connectors.py` only pins the **destination** side (`drt.config.destinations_sql.PostgresDestinationConfig`) — nothing pins a source's.

**I'm not asking you to undo it.** The new path is the truthful one, and there's precedent: destinations already report `drt.config.destinations_sql.…` after the earlier config split, so `config_class` is designed to track the real module. But it *is* user-visible JSON metadata, and it shouldn't land silently. Could you fold in:

1. a `### Changed` line under `[Unreleased]` in CHANGELOG.md — something like *"`drt sources --detailed` now reports `config_class` as `drt.config.profiles.<X>Profile` (the dataclasses moved; the connectors are unchanged)"*; and
2. a small guard in `test_cli_list_connectors.py` pinning a source's `config_class` prefix, mirroring the existing destination assertion — so the *next* module move is a deliberate decision instead of a surprise.

If you'd rather ship as-is, say so and I'll add both myself before merging; I don't want to make you do another round for two lines.

## On the skipped `describe()` dedup

Agreed, and the reasoning is right — `DescribableConfig` is a pydantic `BaseModel` and these are stdlib dataclasses; inheriting it would change construction, `__eq__`, and `dataclasses.fields()` introspection, which `save_profile` depends on. Leaving 13 trivial `describe()`s in place was the correct call for a mechanical move. Not worth a follow-up in my view — the duplication is inert.

Approving on the substance. Ping me once the two items above are in (or tell me to take them) and I'll merge.

Muawiya-contact:
Thanks for the properly independent re-derivation, @masukai — and great catch on `config_class`. You're right that I should've checked the source side; I snapshotted the dataclass fields + `describe()` but not the CLI's `config_class` output, which is exactly the shift that bit the *destinations* in phase 1. Classic "verified the thing I moved, missed the thing that reflects where it moved." Both items are in as of `6a7144c` — I took them rather than make you do it:

1. **CHANGELOG** — added a `### Changed` line under `[Unreleased]`: `drt sources --detailed` now reports `config_class` as `drt.config.profiles.<X>Profile` (dataclasses moved; connectors unchanged; mirrors the destination-side shift from the phase-1 `models.py` split).
2. **Guard test** — `test_sources_detailed_json_config_class_tracks_real_module` in `test_cli_list_connectors.py`, pinning `postgres` → `drt.config.profiles.PostgresProfile`, mirroring the existing destination assertion. Nothing pinned the source side before, so the next module move is now a deliberate decision instead of a silent surprise.

Verified: the new test + the connector suite + the CHANGELOG-format check all green (31 passed), `ruff` clean. Whole-package `mypy drt/` was already clean and is unaffected.

Appreciate the thorough pass — and agreed on leaving the `describe()` dedup alone; the duplication is inert and folding onto the pydantic base would change dataclass semantics `save_profile` leans on. Ready for merge whenever suits; thanks again. 🙏

masukai:
Both items look great, @Muawiya-contact — verified the guard test (`test_sources_detailed_json_config_class_tracks_real_module`) and the CHANGELOG `Changed` line locally (17 passed in the connector suite, config_class resolves to `drt.config.profiles.<X>Profile` via the facade). Thank you for taking them rather than bouncing it back.

Heads-up on one thing I did: this had picked up a **CHANGELOG conflict** — not from your change, but from the CLI stack I merged in the meantime (#791/#793/#794 all added `[Unreleased]` entries in the same spot). Since it was my merges that caused it, I resolved it for you rather than making you rebase: pushed a single merge-of-main commit to your branch (`maintainerCanModify` was on) — **non-force, no history rewrite**, your commits are untouched. The resolution just keeps the CLI `### Added` entries and appends your `### Changed` section after them. It'll collapse cleanly on squash-merge.

That push reset my approval, so I'll re-approve once CI is green (it's running the full matrix now — first time on this PR, thanks to the #801 fix) and merge. Nothing for you to do.

masukai (approved):
Re-approving — CI green across the full matrix (first real run on this PR, via the #801 fix), conflict resolved non-destructively (CHANGELOG only; your refactor untouched), guard test + CHANGELOG verified locally. Clean phase-2 close of #721. 🙏

Issue #385, opened 2026-04-20
Support ${VAR} environment variable substitution in sync YAML config (not just SQL)


K-Masuda-SL:
## Problem

Currently, \`drt\` supports \`\${VAR}\` environment variable substitution **only in SQL model strings** (via \`drt/engine/resolver.py\`).

However, other parts of sync YAML also need environment-dependent values — particularly \`sync.watermark.bucket\`:

\`\`\`yaml
# syncs/trigger_bq_meeting_gha.yml
sync:
  mode: incremental
  cursor_field: completed_at
  watermark:
    storage: gcs
    bucket: \${PIPES_GCS_BUCKET}  # ← not substituted, passed literally to GCS
    key: watermarks/trigger_bq_meeting_gha.json
\`\`\`

The resulting error:

\`\`\`
"error": "Bucket names must start and end with a number or letter."
\`\`\`

This is because \`\${PIPES_GCS_BUCKET}\` is passed verbatim to the GCS client (\`$\` is rejected as an invalid bucket-name character).

## Use case

In a multi-environment setup (DEV / PRD), each environment has its own GCS bucket for watermarks:

- DEV: \`dev-dagster-pipes\`
- PRD: \`prd-dagster-pipes\`

Without env substitution in YAML, we'd have to maintain separate sync YAML files per environment, or duplicate bucket names across many syncs.

## Expected behavior

Environment variable substitution should work in **any string field of the sync YAML**, not just \`model:\`. Common candidates:

- \`sync.watermark.bucket\` / \`sync.watermark.prefix\`
- \`sync.watermark.table\` (for DB-backed watermarks)
- \`destination.host\`, \`destination.port\`, \`destination.database\` (for multi-env connectors)
- \`destination.key\` / \`destination.prefix\` (for GCS-backed destinations)
- Test config strings (e.g., thresholds referenced via env)

The SQL model substitution logic in \`resolver.py\` already exists; it could be promoted to a **generic pre-processing pass** applied to all string fields right after YAML load (before Pydantic validation) or during profile resolution.

## Proposal

Option A — **Pre-validate pass in parser** (\`drt/config/parser.py\`):
After \`yaml.safe_load()\`, walk the dict tree and substitute \`\${VAR}\` in every string value using \`os.environ\`. Missing vars raise \`ValueError\` (consistent with resolver.py's \`StrictUndefined\` behaviour).

Option B — **Pydantic validator**:
[...]

PR #804, opened 2026-07-15
feat(config): project vars — vars: block, --vars override, var() in SQL/YAML (closes #783)

### LINKED ISSUE #783: feat(config): project vars — vars: block + --vars CLI override, var() in model SQL and YAML
## Problem

The only parameterization mechanism is OS env vars (`${VAR}` in YAML/SQL, #385/#240). Anything project-shaped — a default lookback window, a campaign tag, the HubSpot pipeline ID that differs between sandbox and prod portals — must become an environment variable, which means: no defaults in the repo, no visibility in `drt validate`, values living in CI config instead of reviewed YAML, and a bag of `DRT_FOO=` exports to document per runner.

dbt users reach for `vars:` + `--vars` on day one; drt has no answer.

## Prior art

- **dbt**: project `vars:` with CLI `--vars` override, `var('name', default)` in SQL/YAML.
- **dlt**: layered config values (config.toml → env) with defaults in-repo.

## Proposal

```yaml
# drt_project.yml
vars:
  lookback_days: 7
  hubspot_pipeline: "default"

# syncs/*.yml — model SQL and YAML string fields
model: |
  SELECT * FROM mart_users
  WHERE updated_at > CURRENT_DATE - {{ var('lookback_days') }}
```

```bash
drt run --vars 'lookback_days: 1, hubspot_pipeline: sandbox'
```

- `{{ var('name'[, default]) }}` joins the existing (deliberately tiny) SQL Jinja surface (`cursor_value`, `watermark`) and is available in YAML string fields alongside `${ENV}`
- Precedence: `--vars` > env (`DRT_VAR_*` escape hatch) > project `vars:`; undefined var without default = validate-time error (StrictUndefined discipline)
- SQL-injection posture: same as `${VAR}` today — document that vars interpolate into SQL text; identifier-shaped validation warning for suspicious values

## Scope

**In:** project block, CLI flag, resolver + parser integration, validate coverage, JSON Schema update, docs.
**Out:** full Jinja control flow/macros in SQL (explicit non-goal — keep the surface LLM-predictable), per-sync `vars:` shadowing (add later if demanded).

## Related

- #385/#240 (env substitution — the layer this completes), #424 (hardcoded-secret detection should also scan var values)

## Milestone fit

Candidate: v0.8.x/v0.9 — config-layer feature; unblocks the sandbox-vs-prod story without env-var sprawl.

Muawiya-contact:
Implements #783 — the dbt-familiar `vars:` + `--vars` vocabulary, completing the parameterization layer that `${ENV}` substitution (#385/#240) started. Anything project-shaped — a default lookback window, a campaign tag, the pipeline id that differs between sandbox and prod — now gets a reviewed default in the repo instead of an environment variable documented per runner.

```yaml
# drt_project.yml
vars:
  lookback_days: 7
  hubspot_pipeline: "default"
```
```yaml
# syncs/*.yml — model SQL and YAML string fields
model: |
  SELECT * FROM mart_users
  WHERE updated_at > CURRENT_DATE - {{ var('lookback_days') }}
destination:
  type: hubspot
  pipeline: "{{ var('hubspot_pipeline') }}"
```
```bash
drt run --vars 'lookback_days: 1, hubspot_pipeline: sandbox'
```

## Semantics
- **Precedence:** `--vars` > `DRT_VAR_<NAME>` env > project `vars:`.
- **`{{ var('name') }}` / `{{ var('name', default) }}`** joins the deliberately tiny SQL template surface (`cursor_value` / `watermark`) and composes with it in a **single render pass**; also available in YAML string fields alongside `${ENV}`.
- **Undefined var without a default is a validate-time error** — `drt validate` reports it per file (including vars used inside `model:` SQL) instead of a run failing halfway.
- **`--vars`** accepts the issue's braceless form and a braced YAML flow mapping: `'k: v, k2: v2'` or `'{k: v, k2: v2}'`.
- Full Jinja control flow/macros remain an explicit non-goal; per-sync `vars:` shadowing is out of scope as specified.
- **SQL-injection posture:** unchanged from `${ENV}` — var values interpolate into SQL text and are reviewed project config, not user input. Documented.

## Modules
- **`drt/config/vars.py`** (new) — `parse_cli_vars` · `env_vars` · `resolve_vars` (precedence) · `var_environment`/`render_vars`/`expand_vars` · `VarError`.
- `ProjectConfig.vars` (`drt/config/base.py`); parser gains `project_vars()` + an optional `vars` param on `load_syncs`/`load_syncs_safe`; `resolve_model_ref(..., vars=)`; `run_sync(..., vars=)`; `drt run --vars` plumbed via `_RunContext`.

## Two design decisions worth surfacing
1. **Model SQL is deliberately not var-rendered at YAML load time.** My first cut expanded `var()` across every YAML string — including `model:`. Because model SQL shares its template surface with `{{ cursor_value }}` (unavailable at load time), a model using **both** had its cursor template silently rendered to an empty string (`WHERE u > ''`) — quietly breaking the incremental predicate. I hit this in testing and restructured: the **resolver owns the whole SQL template surface**, the parser owns the rest of the YAML. `load_syncs_safe` trial-renders the model purely to surface undefined vars at validate time (output discarded). This also makes `--vars` genuinely reach model SQL. Guarded by `test_model_sql_is_left_for_the_resolver`.
2. **Strictness lives in `var()`, not `StrictUndefined`.** The issue called for "StrictUndefined discipline". Flipping the Jinja environment to `StrictUndefined` would change how an existing stray `{{ token }}` renders (today: empty string) — a behaviour change for projects unrelated to vars. Instead `var()` itself raises on an unknown name, which achieves the stated goal ("undefined var without default = error") while keeping this layer a **no-op for projects that don't use vars**. Happy to switch if you'd rather take the stricter env.

Minor: `--vars 'bare_word'` is rejected rather than silently becoming `{bare_word: None}` (wrapping a bare word in braces yields a null-valued key).

## Verification
- **27 new tests** (`tests/unit/test_config_vars.py`): `--vars` parsing, precedence (project/env/CLI), `var()` value + default + undefined, YAML and SQL rendering, `var()` in a `ref()`'d `.sql` file (the resolver is the only place those render), validate coverage on both the YAML and model-SQL sides, and back-compat guards — var-only SQL still gets the incremental `WHERE` injection, cursor-template-without-value still raises, and SQL without vars is untouched.
- ruff clean; whole-package `python -m mypy drt/` clean (only the pre-existing `credentials.py:175 [no-any-return]`).
- Full suite green (the only failures are the known Windows `test_sqlite.py` `os.unlink` flake, which reproduces on clean `main` and passes on CI Linux).
- Docs: `docs/llm/API_REFERENCE.md` (new "Project vars" section) + `CONTEXT.md`, README run flag, CHANGELOG.

Closes #783


Muawiya-contact:
Self-review pass on this one (I wrote it, so I audited #783's spec line-by-line rather than trusting my own PR description — that's what turned up most of the below). **CI is green** across 3.10–3.13 + CodeQL + schema-drift + codecov.

## Things that were wrong and are now fixed

- **`schema-drift`** — adding `ProjectConfig.vars` desynced the bundled VS Code schema. This *was* the "JSON Schema update" the issue listed in scope and I'd skipped it. Regenerated + extension 0.1.6 + CHANGELOG.
- **Missing: the suspicious-value warning.** The issue's injection posture asks for an "identifier-shaped validation warning for suspicious values" — I'd documented the posture but never implemented it. Added `suspicious_vars()`; `drt run` warns (never blocks) when a var value carries SQL metacharacters.
- **The `--vars` CLI path had no test.** Codecov caught what I didn't (`run.py` patch 53.8%) — every test hit the resolver, none the headline interface. Broken plumbing would have shipped green. Added end-to-end coverage.
- **My new CLI tests were environment-dependent** — green locally only because I have a real `~/.drt/profiles.yml`; on a clean runner `drt run` died at profile lookup before reaching vars. Stubbed it, then verified against the CI condition (fake empty HOME: without the stub they fail exactly like CI, with it they pass).
- **CodeQL** (`py/jinja2/autoescape-false`) — a `# lgtm[...]` suppression isn't honoured here. Switched to `select_autoescape(default_for_string=False, default=False)`: the sanctioned API, behaviourally identical (we only render strings → no escaping; HTML-escaping SQL would corrupt `O'Brien` → `O&#39;Brien`). Worth noting this environment isn't new risk — the identical one lived at `resolver.py:144` and *moved* here.
- **Jinja env churn** — `expand_vars` built one `Environment` per rendered string (measured 10 fields → 10 envs). Hoisted to one per tree.
- **A real bug caught mid-build:** my first cut var-rendered every YAML string *including* `model:`. Since model SQL shares its surface with `{{ cursor_value }}` (unavailable at load time), a model using both had its cursor template silently rendered to `''` — quietly breaking the incremental predicate. Restructured: the resolver owns the whole SQL surface, the parser owns the rest of the YAML; `load_syncs_safe` trial-renders only to keep validate coverage. Pinned by `test_model_sql_is_left_for_the_resolver`.
- **Cross-PR gap with #795:** `drt build` builds its own `_RunContext` without threading vars, so model-SQL `var()` would have raised `Undefined var` under `drt build` while `drt run` rendered it fine. Rather than depend on a change over there, closed it at the source: **`vars=None` now means "resolve from the project"** in `resolve_model_ref` — the same contract `load_syncs` already had, so any command or library caller gets `vars:` + `DRT_VAR_*` for free; only `--vars` needs threading. Lazy, and only when the SQL contains `var()`.

## Two calls I'd rather you make than me

1. **`suspicious_vars` and apostrophes.** `{"label": "O'Brien's team"}` warns on every run. A bare quote *is* the injection vector the issue wanted surfaced (breaking out of a string literal), so narrowing the rule to statement-breakers (`;`, `--`, `/*`) trades safety for quiet. I left it broad + warning-only, but I don't have a strong claim — happy to narrow it.
2. **StrictUndefined deviation.** The issue says "StrictUndefined discipline"; I put the strictness in `var()` (raises on an unknown name) instead, because flipping the environment to `StrictUndefined` would change how an existing stray `{{ token }}` renders (today: empty) — a behaviour change for projects unrelated to vars. Same stated outcome, zero blast radius, but it *is* a deviation. Say the word and I'll switch it.

Also documented, not fixed: vars are **scalar-shaped** — a list/dict var interpolated into a YAML field renders as its Python repr and fails that field's validation. That matches the shape `${ENV}` already has and the issue's examples, but it wasn't written down; it is now.

masukai (approved):
Reviewed this end-to-end and re-derived the load-bearing claims myself rather than trusting the writeup — it holds up, and the two design calls you surfaced are the right ones. Approving.

## What I verified independently

**Design decision #1 — model SQL is the resolver's, not the parser's.** This is the part that would silently corrupt data if it were wrong, so I probed `resolve_model_ref` directly rather than reading the test names:

| SQL | result |
|---|---|
| `var()` **and** `{{ cursor_value }}` in one statement | both render in one pass, no double `WHERE` — `… CURRENT_DATE - 7 AND u > '2020-01-01'` |
| `var()` only, with `cursor_field`+value | var renders **and** the incremental `WHERE` is still injected (wrapped subquery) |
| `{{ cursor_value }}` only | unchanged from pre-#783 |
| plain SQL | `WHERE` injected as before |
| undefined `var()`, no default | `VarError` |
| cursor template, `last_cursor_value=None` | still `ValueError` (pre-existing guard intact) |

So the `WHERE u > ''` silent-break you caught in testing is genuinely fixed, and `_expand_sync_vars` skipping `key == "model"` while the resolver owns the whole template surface is the correct seam. Nice catch that this needed restructuring rather than a bigger regex.

**Precedence** — `resolve_vars({x: proj, y: proj}, {x: cli})` with `DRT_VAR_X=envval` set → `{x: cli, y: proj}`. CLI > env > project, confirmed empirically.

**Engine purity** — `sync.py` only threads `vars` through as a plain dict; the `project_vars()` read is lazy, inside the resolver, guarded by `has_var_template`, and the resolver already does filesystem I/O for `ref()`. So the pure sync loop stays pure and var-less projects pay nothing. Good.

**validate-time surfacing** — the trial-render + `except (ValidationError, ValueError)` (with `VarError <: ValueError`) lands an undefined var in `result.errors[path.stem]`, including vars used inside `model:` SQL. Confirmed the discard-the-output trial render only raises on `var()`, not on the blanked `{{ cursor_value }}`.

**Not dead code** — `suspicious_vars` is actually surfaced at `run.py:463` (warn-only, text mode). 

32 tests pass (writeup said 27 — even better), `mypy` clean on all five touched modules, 249 passed across the parser/resolver/engine/cli-run/validate slice locally, vscode schema carries `vars`.

**Design decision #2 (strictness in `var()` not `StrictUndefined`)** — agreed, and for the reason you gave: flipping the env to `StrictUndefined` would change how an existing stray `{{ token }}` renders for projects that don't use vars at all. Keeping this layer a no-op unless `var(` appears is the right blast radius.

## Two non-blocking observations

1. **`drt build` won't honor `--vars`** (it has no such flag), so it uses `vars:` + `DRT_VAR_*` but not a CLI override — correct and documented here, just noting that once #777's `build` grows options, `--vars` is a natural one to thread through the same way, since build is the run+test combo.
2. **`DRT_VAR_<NAME>` lowercases the whole name**, so it can't override a `vars:` key that isn't lowercase (`myVar` ≠ `DRT_VAR_MYVAR` → `myvar`). Fine given the lowercase convention, but a one-line note in the docs ("var names are lowercase; `DRT_VAR_*` matches case-insensitively by lowercasing") would save someone a confusing ten minutes. Purely optional.

Neither blocks. This is a clean, well-scoped close of #783 — the `${ENV}` → `var()` symmetry, the single-pass template surface, and the validate-time errors all read exactly right. 🚀

Issue #718, opened 2026-07-07
feat(mcp): parity tools — DLQ/retry, docs manifest, profile (+ run_sync overrides + SSoT-derived inventory)


masukai:
## Summary

MCP parity audit: `drt/mcp/server.py` exposes 9 tools and the ROADMAP parity trio (`drt_doctor`, `drt_run_test`, `compute_diff`) all exist and are wired to real code. But several shipped CLI capabilities have **no MCP equivalent**, so agents can't drive them.

### Gaps (most-impactful first)

- [ ] **DLQ / `drt retry`** — the entire v0.7.9 capability (#278) is invisible to MCP. `drt_get_status`/`drt_get_history` show that rows failed, but there's no tool to inspect queue depth or replay. Add `drt_dlq` (read/`DlqStore.depth()`) and/or `drt_retry` wrapping `drt/cli/commands/retry.py`.
- [ ] **`drt docs` manifest** — `drt docs generate --format json|mermaid` (epic #499) produces exactly the machine-readable lineage/sync-catalog an agent would consume, but there's no `drt_get_manifest`. `drt.docs.builder.build_manifest` is reusable + side-effect-free on the json path.
- [ ] **`drt profile`** (read-only) — `profile list`/`show` (secrets already masked) + `profile test` (connectivity) are natural diagnostics alongside `drt_doctor` (#423). Add `drt_list_profiles` + `drt_test_profile`. (`add`/`remove` are interactive/mutating — out of scope.)
- [ ] **`drt_run_sync` overrides** (low) — CLI `run` supports `--cursor-value` (bounded backfill), `--profile`, `--select`; `drt_run_sync` only takes `sync_name`/`dry_run`/`compute_diff`/`diff_limit`. Add optional `cursor_value` + `profile_name` passthrough.

### Durable follow-up (also fixes a drift class)

- [ ] Derive `drt_list_connectors` from the `drt.config.connectors` SSoT instead of the hand-maintained literal (its docstring already says it's "used by the MCP server", but the server never imports it). This is the root cause of the `salesforce_bulk` inventory miss fixed in the drift-guard PR.

All verified against `drt/cli/commands/` + CHANGELOG. Config-file sync features (mirror tracked/scoped, mask, field_mappings, replace) are correctly covered indirectly via `drt_get_schema` + `drt_validate` and don't need dedicated tools.

PR #805, opened 2026-07-16
feat(docs): default-safe connection labels — describe_safe() seam + hashed node ids + --full-labels (closes #696)

### LINKED ISSUE #696: feat(docs): default-safe connection labels — docs output currently publishes endpoints, phone numbers, emails

masukai:
## Summary

Closes #696 — the last security item on the v0.8 release gate.

`drt docs generate` embedded each destination's `describe()` verbatim in the site (every page + the shipped `manifest.json`): full REST endpoint URLs, SMTP hosts, phone numbers, sender emails, `owner/repo`, account ids. Hosting `target/docs/` — the intended use — published an inventory of internal endpoints and contact identities.

**Default now:** labels keep *object identity*, drop network locations / personal identifiers. **`--full-labels`** restores verbatim output for trusted/internal hosting.

| connector | before (default) | after (default) |
|---|---|---|
| rest_api | `rest_api (https://internal-host/path)` | `rest_api` |
| twilio | `twilio (+818012345678)` | `twilio (+81…)` |
| sendgrid | `sendgrid (ops@corp.example)` | `sendgrid (…@corp.example)` |
| airtable | `airtable (appXXX/Leads)` | `airtable (Leads)` |
| github_actions | `github_actions (owner/repo)` | `github_actions` |
| postgres / s3 / slack / hubspot… | unchanged | unchanged (object identity) |

## Design (the issue's "structured option")

- **`describe_safe()` on `DescribableConfig`, safe by default.** Object-identity connectors opt in with one line (`_detail_is_public = True`); partially-safe ones override (twilio keeps the country code, sendgrid masks the local part but keeps the org's own domain, notion keeps `database` and drops the uuid). The builder duck-types the method with a bare-`type` fallback — **an unclassified future connector renders type-only**, so forgetting is safe, never leaky.
- **Node ids stop leaking too.** The old id was `dest_<type>_<slug(describe())>` — the endpoint was baked into *page filenames* and manifest node names, not just labels. Now `dest_<type>_<sha1(describe())[:8]>`: deterministic (byte-identical regen — the #697 property, still enforced by the existing test) and collision-free when two same-type destinations share a safe label (two `rest_api` endpoints stay separate DAG nodes / pages).
- **MCP inherits the fix** — its docs manifest flows through the same builder.
- **Bonus catch from the E2E scan:** a Twilio `from_number` sailed straight through the raw-YAML tab's redaction (its key matches neither `phone` nor anything else in the #752-era list). The pattern gains `number|sender|recipient`.

## Guard rails (how this stays fixed)

Three-layer test over the **full `DESTINATIONS` registry** (34 connectors):

1. **Completeness** — a new connector fails the build until its label is classified in the golden table.
2. **Golden labels** — the exact safe label per connector; a redaction change is a visible, reviewed diff.
3. **Planted sentinel** — every network/personal field in the samples carries `LEAKME`; no safe label may contain it, *independent of the golden table being right*.

Plus builder tests (label mode, id determinism, id distinctness under label collision) and an E2E check: generated a site from a project with rest_api / twilio / sendgrid / postgres destinations and grepped the whole output — zero hits for the endpoint, phone, or email; the postgres table still shows in the catalog.

## Notes for review

Two judgment calls worth eyes:
- **sendgrid keeps the domain** (`…@corp.example`) — reasoning: on the org's own docs site, the org's own domain isn't a leak; the personal identity is the local part. Happy to go type-only if you'd rather.
- **File/parquet paths stay** (`file (out/users.csv)`) — they're object identity like a table name. An absolute path with a username would show; flag if that bothers you.

2082 unit tests green, ruff + mypy clean. CHANGELOG under `[Unreleased]` → `### Changed`; `docs/llm/CONTEXT.md` documents the flag.


masukai:
@yodakanohoshi review request — security-relevant one, no rush at all. 🙏

The two judgment calls I'd most value your eyes on are in "Notes for review" in the PR body (sendgrid domain retention, file-path labels). The golden table in `tests/unit/test_docs_safe_labels.py` is the whole redaction policy in one place, so reviewing that file basically reviews the PR.

Muawiya-contact (changes requested):
**Request changes** — one high-severity issue that undermines the PR's own goal. Everything else here is strong, and I want to be clear about that up front.

## ① [high] The node id hashes the unredacted string — and it's brute-forceable

`_destination_id` is `dest_<type>_<sha1(describe())[:8]>`, where `describe()` is the **verbatim** string. That id becomes page filenames and manifest node names — i.e. it ships. Hashing a *low-entropy* identifier isn't redaction, so the label is protected and the same value is republished in recoverable form.

I ran this against the PR's own Twilio example rather than theorising:

```
safe label (published):  twilio (+81…)              <- redacted, good
node id    (published):  dest_twilio_8912a4d6
brute force:             RECOVERED +818012345678
                         12,345,679 candidates, 48s, 256k/s -- pure Python
                         full +8180 space = 10^8 -> ~7 min on a laptop, seconds on a GPU
```

**The partial-redaction design compounds it.** The deliberate "keep the country code / keep the org domain" choices shrink the very preimage space the hash needs to protect:

```
sendgrid: safe label publishes the domain   ->  "...@corp.example"
          node id publishes sha1(full)      ->  only the local part is unknown
          RECOVERED ops@corp.example in 18 dictionary guesses, 0.000s
```

Same shape for `github_actions` (`owner/repo`) and guessable internal `rest_api` URLs. The pre-#696 slugged id was *obviously* leaky; this one is subtly leaky, which is worse — it reads as fixed.

**Fix direction:** don't derive the id from the sensitive string at all. Hash the **safe** label plus a deterministic disambiguator (e.g. occurrence index among same-type destinations, ordered by first-referencing sync name). That keeps every property the docstring claims — deterministic/byte-identical (#697), and collision-free when two `rest_api` endpoints share the safe label `rest_api` — while leaking nothing. A salt can't rescue the current shape: it would have to ship in the artifact.

## What's good

- **`describe_safe()` defaulting to safe, with `_detail_is_public = True` as the opt-in, is the right polarity.** A connector added without thinking about docs exposure renders type-only; the builder's duck-typed fallback reinforces it. Forgetting is safe, never leaky — exactly as the summary claims.
- **The three-layer guard is better than what I'd have written.** Registry completeness → golden labels → a planted `LEAKME` sentinel that holds *independent of the golden table being right* is the part that makes this stay fixed; a wrong golden entry still fails.
- **The Twilio `from_number` catch is a genuine find.** That key matched nothing in the #752-era `_SENSITIVE_KEY_RE` I reviewed and signed off on — a real blind spot of mine, caught by your E2E scan.
- Determinism preserved, and MCP inheriting via the shared builder is the right seam.

## Minor
- If ① is fixed by hashing a safe value, consider `blake2b(digest_size=4)` over `sha1` — same properties, avoids future "why sha1?" scanner noise. (Non-crypto id, so sha1 isn't wrong today.)
- `--full-labels` restores verbatim labels *and* keeps ids derived from them — worth one doc line noting ids are identifying in that mode too.

Happy to be wrong on ① if you see a reason the id space isn't reachable — but the two recoveries above were run end-to-end, so I'd want that addressed before this lands as the #696 fix.

Muawiya-contact (changes requested):
@masukai just do that.

masukai:
Fixed, @Muawiya-contact — and thank you for running the recoveries instead of theorising. You were right on every count, and the compounding observation (the partial redactions shrinking exactly the preimage space the hash needed to protect) is the part I flatly missed: I designed the label policy and the id scheme separately and never checked them against each other.

**What changed** (`a16bbe5`):

- Ids are now `dest_<slug(describe_safe())>` + a deterministic `_2`/`_3` suffix when distinct destinations share a safe label — your fix direction, minus the hash entirely. Once the input is the safe label, hashing adds nothing, so **no function of the sensitive string ships at all** — strictly stronger than swapping in blake2b, and it answers your minor #1 by deleting the question.
- Distinctness is tracked by the full `describe()` **in memory only** (it never reaches the output); determinism comes from the sorted sync-file iteration, so #697 holds.
- Ids are **independent of `--full-labels`** — switching label modes never rewires the graph or renames pages. New test pins this, which also resolves your minor #2 for ids (labels in `--full-labels` mode remain identifying by definition; the flag help already says trusted/internal only).
- Side benefit: ids are human-readable again (`dest_postgres_public_users`, not `dest_postgres_0417ef2b`).
- ADR 0001's amendment paragraph + the CHANGELOG entry now record the rejected intermediate design and why — the 48-second recovery is exactly the kind of thing the next person should find written down.

E2E re-run on the leak-test project: site output contains no endpoint/phone/email and no derivative of them; 2083 unit tests green. Re-review when convenient. 🙏

Pawansingh3889:
Funny timing, I had my own #696 branch about ready when this landed. Yours goes further than mine did (the yaml-tab `from_number` catch and the registry-wide guard especially), so I put mine down and ran this branch through the little test project I'd been using instead. Two things came out of it, flagging before merge in case they matter.

**1. Object-store buckets and local file paths still ship in the default labels.** s3/gcs/azure_blob and parquet/file are flagged `_detail_is_public = True`, so a default build gave me `s3 (s3://corp-secret-datalake/crm/)` and `parquet (/home/pawan/exports/data.parquet)`. Reading the issue's exposure table I had expected the bucket to land on the locator side (bucket names are easy to probe once public, and local paths carry home dirs). The sentinel scan can't catch this one because the sample buckets have harmless names, so LEAKME only guards fields that were already classified sensitive. One option: keep the prefix (that's the per-sync routing identity) and put bucket/container plus the directory part of local paths behind `--full-labels`, giving `s3 (crm/)` and `parquet (users.parquet)`. Entirely possible this was a deliberate call though, in which case ignore me.

**2. Same-type node ids swap when a sync file gets renamed.** The `_2` suffix follows sync-file iteration order. With two rest_api endpoints in `a.yml` and `b.yml`, renaming `a.yml` to `z.yml` and rebuilding:

```
before: alerts -> dest_rest_api    digests -> dest_rest_api_2
after:  digests -> dest_rest_api   alerts -> dest_rest_api_2
```

Page filenames swap with the ids, so a bookmarked destination page quietly starts showing the other endpoint's syncs. Since `_by_identity` already holds the full `describe()` in memory, allocating the suffix by sorted rank of that identity instead of encounter order would keep ids stable across file renames; they would then only move when the set of same-label destinations actually changes.

Happy to PR either of these on top if useful. One small extra I noticed while comparing against my branch: `drt_get_manifest` inherits the safe default but didn't get the `full_labels` opt-in the CLI grew. Easy add for MCP parity (#718), can send that one right away.

masukai:
@Pawansingh3889 this is exactly the review this PR needed — thank you, and doubly so for setting your own #696 branch aside and putting the energy into stress-testing this one instead. All three findings were real; both label/id issues are fixed in `24b3531`:

**1. Buckets & paths — adopted your proposal as-is.** `s3 (crm/)` / `gcs (crm/)` / `azure_blob` (empty prefix → type-only), and `parquet (users.parquet)` / `file (users.csv)` — routing prefix and file name stay, bucket/container and directories go. You were right that the sentinel scan couldn't see this one; the golden samples now plant `LEAKME` in buckets, containers, and directory parts, so it's enforced from here on. And the E2E re-run promptly caught a third member of the same species: a raw `bucket:` value sailing through the YAML tab exactly like the Twilio `from_number` did — the redaction key list gains `bucket|container|path`. Your test project earned its keep twice.

**2. Rename-swap — fixed with one refinement on your suggestion.** Sorting by the in-memory `describe()` would be stable, but it *orders by the secret* — with two same-type destinations, `_2` would reveal which endpoint sorts lexicographically later. Suffix rank now follows the smallest **referencing sync name** instead: manifest-public (zero new information), file-rename-independent, and it matches what you'd intuitively expect the ids to track. New test renames `alerts.yml` → `zz_renamed.yml` and pins both ids.

**3. `drt_get_manifest` parity — yes please, it's yours.** A `full_labels` opt-in on the MCP tool mirroring the CLI flag, off by default, is exactly the right shape (#718 lineage). Send it whenever suits — happy to review quickly so it rides the same release.

2084 unit tests green, full E2E leak scan clean across all 30+ generated files.

Pawansingh3889:
The sync-name rank is the better fix. Sorting by describe() ordering on the secret is a side channel I completely missed, that one goes in the notebook. #807 is up for the MCP flag, stacked on this branch so it rides along.

PR #807, opened 2026-07-16
feat(mcp): full_labels opt-in on drt_get_manifest, mirroring the CLI flag

Pawansingh3889:
## What does this PR do?

The follow-up from the #805 thread. `drt_get_manifest` flows through the same builder as `drt docs generate`, so it picked up the docs-safe default automatically but had no way back to verbatim labels. It now takes `full_labels` (off by default) matching `--full-labels`, for callers operating in a trusted context. #718 parity.

Two tests pin both directions: the default payload carries no trace of the endpoint anywhere in the serialized manifest, and the opt-in restores the verbatim label.

Targeting the #805 branch so it rides the same release. Happy to rebase onto main instead if you'd rather land it after.

## Related Issue

#718, follows #696 / #805.

## Checklist

- [x] Tests pass (`make test`) — the one `test_mcp` failure locally is the pre-existing Windows-only doctor HOME test, identical on main
- [x] Updated `CHANGELOG.md` under `[Unreleased]` (extended the #805 entry's MCP sentence)


masukai (dismissed):
Verified locally on top of the CI run — 41 MCP tests green, ruff format agrees with your rejoined signatures, mypy clean. Approving.

What I like here beyond the parity itself: `test_get_manifest_labels_are_docs_safe_by_default` asserts over `json.dumps(result)` of the **whole manifest**, not just the label field — that's the same shape as the E2E leak scan on #805 and exactly the right level to pin this at. And thank you for extending the existing CHANGELOG sentence instead of adding a second entry; the release notes stay one story.

**Merge order note** (so you're not surprised): #805 needs its CHANGES_REQUESTED lifted first. When it lands, I'll retarget this PR to `main` *before* merging #805 — merging a stack parent with branch deletion auto-closes the children, and a force-push after that makes them unrecoverable (we lost a PR to exactly this last week; it's now written up in CONTRIBUTING). Nothing for you to do — the diff will collapse to just your commit once #805 is in. 🚀

masukai (approved):
Re-approving after the conflict resolution (I merged `main` into your branch to clear the CHANGELOG overlap with the now-landed #805 — non-force, your commit is intact; net diff is just the MCP tool + tests + one CHANGELOG clause). CI green, 116 docs/MCP tests pass locally. Merging so it rides the same release. Thanks for the fast parity follow-up. 🚀

Issue #757, opened 2026-07-10
feat(engine): match_policy — update_only / create_only semantics for upsert destinations


masukai:
## Problem

`sync.mode: upsert` always creates records that don't exist in the destination. For SaaS/CRM destinations this is frequently *not* what operators want:

- **Enrichment syncs**: push warehouse-computed scores into HubSpot/Salesforce contacts that sales already created — creating *new* contacts for every warehouse row pollutes the CRM with junk records
- **Create-only campaigns**: seed an audience once, never overwrite fields that reps have since edited by hand

Today drt has no way to express "only touch records that already exist" (or the inverse). `lookups.check_only` (#354) can existence-filter against *SQL* destinations, but SaaS destinations (HubSpot, Salesforce, Intercom, Klaviyo, …) — where this need is strongest — have nothing.

## Prior art

- **Census sync behaviors**: "Update Only" / "Create Only" alongside upsert/append/mirror — a first-class, per-sync choice.
- **Hightouch**: "Update existing rows only" toggle on CRM destinations.
- dbt/dlt have no equivalent — this is reverse-ETL-specific semantics.

## Proposal

```yaml
sync:
  mode: upsert
  match_policy: upsert       # upsert (default) | update_only | create_only
```

- `update_only`: rows with no destination match are skipped (counted + reported in SyncResult as `skipped_no_match`, not errors)
- `create_only`: rows that already exist are skipped
- Implementation is per-destination capability: SaaS APIs that expose create-vs-update endpoints (HubSpot, Salesforce Bulk, Intercom, …) map naturally; SQL destinations map to `INSERT ... ON CONFLICT DO NOTHING` / `UPDATE`-only paths; destinations that can't support a policy fail fast at validate time (same pattern as tracked mirror).

## Scope

**In:** config key + validation, capability flag on the Destination protocol, SaaS legs for HubSpot/Salesforce/Intercom + SQL legs for Postgres/MySQL, skip counters in results/history.
**Out:** field-level "don't overwrite non-null destination values" merge rules (worthwhile, but a separate design), diff-based change detection (#see diff-engine issue).

## Related

- #354 (`lookups.check_only`) — existence *filter* for SQL; `match_policy` is the destination-native generalization
- #340 (mirror) — precedent for mode semantics rolling out destination-by-destination with fail-fast

## Milestone fit

Candidate: v0.8.x follow-up or v0.9 — high user value, moderate per-destination effort; a good "one destination per PR" community track once the engine seam exists.

[...]

PR #812, opened 2026-07-18
feat(engine): sync.match_policy — update_only / create_only on the upsert path (#757)

masukai:
## What does this PR do?

Adds `sync.match_policy` — the reverse-ETL "Update Only" / "Create Only" semantics (Census / Hightouch) — and ships the **engine seam + Postgres reference leg**.

```yaml
sync:
  mode: upsert
  match_policy: update_only   # upsert (default) | update_only | create_only
```

- **`update_only`** — only touch rows that already exist in the destination; no-match rows are **skipped, not inserted**. The enrichment case: push warehouse-computed scores into CRM contacts sales already created, without spraying a junk contact per warehouse row.
- **`create_only`** — only insert rows that don't yet exist; existing rows are left untouched (seed an audience once, never overwrite hand edits).

Skipped rows land in `SyncResult.skipped` (`drt run` prints `… N skipped`), never errors. Default stays `upsert`, so existing syncs are unchanged.

**Engine seam:**
- Config field + validator (`sync_options.py`): rejected for `mode: replace` / `mirror` at validate time.
- A structural `MatchPolicyCapable` capability protocol (`base.py`) + a pure fail-fast guard in `engine/sync.py`: a non-default policy on a destination that doesn't implement it **errors loudly** rather than silently upserting.
- **Postgres leg**: `create_only` → `INSERT … ON CONFLICT (key) DO NOTHING`; `update_only` → `UPDATE … SET <non-key> WHERE <key>`. Both use Postgres's unambiguous `cur.rowcount` (inserted / matched) as the exact skip signal.

## Related Issue

**Part of #757** — this PR delivers the engine seam + the Postgres reference leg. **#757 stays open** as the tracking issue: MySQL and the SaaS/CRM legs (HubSpot / Salesforce / Intercom) are the "one destination per PR on a shared engine seam" follow-ups the issue itself anticipates.

## Checklist

- [x] Tests pass (`make test`) — new `tests/unit/test_match_policy.py` (21 tests: config validation, engine fail-fast, Postgres create/update legs incl. param order & skip counting)

## Judgment points

1. **MySQL deferred on purpose.** MySQL's `UPDATE` reports *changed* rows, not *matched* rows (unless `CLIENT_FOUND_ROWS`), and enabling that flag interacts with `ON DUPLICATE KEY` affected-row counts — so a correct `update_only` skip count needs its own careful PR. Rather than ship an ambiguous MySQL skip counter, this PR does Postgres only. *(Resolved: keeping #757 open as the tracking issue rather than auto-closing.)*
2. **Skip counter reuses `SyncResult.skipped`** rather than adding a dedicated `skipped_no_match` field (the issue mentions the latter). Reusing `skipped` is zero-blast-radius (already summed in the engine, surfaced in CLI + OTel). Happy to add a dedicated counter if you'd rather distinguish match-policy skips from lookup/mask skips — flag it in review.
3. **`update_only` with no non-key columns** (every column in `upsert_key`) raises a clear error — there's nothing to SET. Alternative would be a silent no-op; I chose the loud error.


yodakanohoshi (approved):
LGTM 🎉 The engine seam + Postgres reference leg is cleanly separated — nice work.

**What I checked:**

- **Engine boundary** — `_check_match_policy_supported` is pure (no I/O, no logger/state), so it respects the `SyncObserver` boundary and won't trip the observer boundary test. Failing fast on an unsupported policy instead of silently upserting is consistent with how unsupported `mirror` configs behave. 👍
- **rowcount skip semantics** — Since we `cur.execute` row-by-row (not `executemany`), `rowcount` is reliable per row. `create_only` pins the conflict target with `ON CONFLICT (upsert_key)`, so a collision on *another* unique constraint correctly raises → lands in `failed`, not silently counted as a skip. And `UPDATE` reports matched rows regardless of whether a value changed, so `rowcount == 0` is an exact "no match" signal. Both legs are correct.
- **`skipped` plumbing** — verified end to end: `SyncResult.skipped` → engine aggregation → OTel `load.skipped` → the `N skipped` line in `drt run`.
- validator (rejects `replace`/`mirror`, default always allowed) + the 3-layer / 21-test suite + docs (CHANGELOG / postgres.md / API_REFERENCE / vscode schema) are all in place.

On the judgment points: deferring MySQL over its `changed`-vs-`matched` rowcount ambiguity is the right call, and reusing `SyncResult.skipped` rather than a dedicated field is zero-blast-radius at this scale — no objection.

**Non-blocking, follow-up-later:**

1. `create_only` / `update_only` make `upsert_key` semantically mandatory, but an empty `upsert_key` would surface as an opaque SQL error (`ON CONFLICT ()` / empty `WHERE`). Same assumption as the existing upsert path, so not a regression — but an early check in the spirit of the "needs a non-key column" error would be friendlier.
2. Minor: the `# type: ignore[arg-type]` in the tests could be a `cast(Destination, ...)`, but it's test-only and harmless.

Merging as-is is fine by me. 👍

Issue #808, opened 2026-07-17
docs(ui): drt docs site — post-v0.8.0 follow-ups (successor to #702)


masukai:
Successor to #702, opened as the v0.8.0 `drt docs` gate closes (#499). The **visual design train is complete** — tokens/tabs/badges/KPI (#704), empty-state system (#751), YAML tab + mobile (#752), a11y/print/favicon finishing (#753), static SVG DAG (#796), and default-safe labels (#696/#805, +MCP parity #807). What remains from #702 is post-v0.8.0 work, split by what unblocks it.

## UI polish (no new data needed — schedule for a v0.8.x)
- [ ] **Interactivity layer** (separable JS, degrades to plain links on `file://`): expand-in-place ego view, hover-highlight up/downstream, optional pan-zoom, manual theme toggle
- [ ] **Scale mode** for large projects: compact card density, tag/type group collapse, search-results UX
- [ ] **DAG-page mobile treatment** (the general mobile pass landed via #752; the DAG page specifically still needs it)
- [ ] **Visual-regression screenshot diff** (Playwright) as CI — the byte-identical HTML check (#753) already guards output; this would guard *rendering*

## Blocked on manifest schema v2 (#698)
- [ ] **DLQ badge** (`n → DLQ` on failed/partial syncs) — needs run/DLQ data in the manifest
- [ ] **Drift badge** (`⚠ drift` on destinations) — needs a persisted drift-audit artifact
- [ ] **Column-level lineage (L2)** — expose `field_mappings` / `mask` (write side, declared, no SQL parsing); source-side table extraction from model SQL needs SQL parsing (later still)

## Separate epics
- [ ] **`drt docs serve`** (P4 of #499) — live Web UI; currently a documented `NotImplementedError` pointing at `--format mermaid`. Static site is the v0.8.0 deliverable; serve is its own follow-up.
- [ ] **Public demo/showcase hosting** via drt-web `static/demo/docs` + the `examples/docs-demo` fixture — plan in `drt-marketing/docs-showcase-plan.md`; marketing-side, announce post-launch.

Nothing here gates v0.8.0. Carried over verbatim from #702 so no item is lost.

PR #814, opened 2026-07-18
feat(docs): manifest schema v2 — run history, declared column facts, DLQ depth

### LINKED ISSUE #698: feat(docs): manifest schema v2 — expose run history (duration, per-sync timeline) from .drt/history
## Summary

Run history is already persisted — `.drt/history/<sync>.jsonl` stores one `HistoryEntry` per execution (`started_at`, `completed_at`, `duration_seconds`, `status`, `records_synced`, `records_failed`, `errors[]`) since #276. But **manifest schema v1 only carries a single latest-state snapshot** per sync (no duration, no history), so the docs site's "Recent Runs" panel is actually *latest state per sync* sorted by time — it can't show Duration, can't show two runs of the same sync, and a per-sync run timeline (the ADR's P4 direction) has no data to draw from.

## Proposal

Manifest **schema v2**: expose recent history alongside the state snapshot.

- Per sync: `runs: [ …last N HistoryEntry-shaped records… ]` (newest first; N default ~10, `--history-depth` to tune, `--no-state` continues to omit both).
- Keep the existing `state` snapshot block as-is for compatibility; bump `SCHEMA_VERSION` to 2 with a migration note per VERSIONING.md.
- Docs builder reads `HistoryManager` (already thread-safe JSONL, newest-first reads).

What this unlocks in the HTML site:

- **Recent Runs** becomes a true cross-sync run log (with the Duration column the ADR mock always had).
- Per-sync **run timeline** on the sync detail page (status over time, duration trend).
- Status sparklines / failure-streak affordances later, all statically rendered.

One interlock: `errors[]` strings can contain connection details or data fragments — whatever redaction policy lands for docs labels (#696) should apply to embedded error text too.

Related: #276 (history storage, shipped), #507 (schema v1), #510, #499.

Pawansingh3889:
Closes #698. This is the schema half; the site changes that consume it are ready as a follow-up PR once this merges.

## What the manifest carries now

Three optional blocks per sync, following the plan ADR 0001 recorded ("each is additive in spirit but reshapes `Sync`, so they land together as v2"):

- `runs`: up to `--history-depth` recent executions from `.drt/history` (newest first, default 10, 0 disables). Public shape of `HistoryEntry`, minus `sync_name` (runs nest under their sync) and `dry_run` (always false on disk).
- `fields`: declared write-side column facts from `sync.field_mappings` + `sync.mask`, which is the data the #808 column-lineage item needs. Keyed by the destination-side name, since mask keys already reference the post-rename name. Declared columns only, no SQL parsing.
- `dlq_depth`: current DLQ depth, for the #808 badge.

`runs` and `dlq_depth` are machine-local so they ride the existing `--no-state` switch. `fields` derives from the YAML like the rest of the catalog and is always emitted. v2 is a pure superset of v1: nothing renamed or removed, so a v1 consumer keeps working unchanged. The MCP `drt_get_manifest` tool gets a matching `history_depth` param (#718 parity, same shape as #807 did for `full_labels`).

## The #696 interlock

The issue flagged that `errors[]` strings can carry connection details. While wiring that up I noticed `state.last_error` already ships verbatim in v1, and connector errors routinely embed DSNs, hosts and addresses, so that was a real leak into a hostable artifact. Both now pass through a free-text sweep (URLs, e-mails, phone numbers, credential-shaped `key=value` fragments) unless `--full-labels`. It is deliberately over-eager ("user: 42" masks the 42): for a hosted site over-redaction is cosmetic and under-redaction is a leak. Same trust switch as labels.

Tried it against `examples/docs-demo` with synthesized history: a failed run whose error contained the Slack webhook URL comes out as `POST « redacted » returned 404`.

## Notes

- ADR 0003 records the decisions; ADR 0001's v2-candidates paragraph now points at it. CHANGELOG and `docs/llm/CONTEXT.md` updated.
- New tests cover round-trip, depth limiting, ordering, `--no-state` gating, redaction on both runs and state, DLQ depth, and the CLI/MCP passthrough. One existing assertion pinned `schema_version == 1` and now compares against the constant.
- Model-SQL source-table extraction stays out (needs SQL parsing), as does everything presentational.


masukai (approved):
Excellent work, @Pawansingh3889 🙌 — clean "v2 as a pure superset" design, and the ADR 0003 write-up makes the machine/repo split and the redaction policy easy to follow.

A few things I especially liked:

- **The `state.last_error` catch.** v1 shipped it verbatim, so the #698 ↔ #696 interlock closes a real *pre-existing* leak, not just a new-surface one. Good instinct to fold it in.
- `_declared_fields` keying on the post-rename name (matching `mask`'s own keys) is exactly right for the write-side lineage #808 needs, and the "declared columns only, no SQL parsing" boundary is drawn clearly.
- Coverage is thorough — redaction on both `runs[].errors` and `state.last_error`, the `--full-labels` bypass, depth-limit/`0`/`--no-state` omission, and MCP parity all pinned.

Two tiny, **non-blocking** nits on `_redact_error_text` (the ADR already treats over-redaction as cosmetic, so these are optional — fine to fold into the #808 site follow-up):

1. When a `key=value` value is itself a URL (`dsn=postgres://…`), the URL pass redacts first and the KV pass then re-hits the marker, producing a cosmetic `dsn=« redacted » redacted »`. No leak — just untidy output.
2. `_ERROR_PHONE_RE` requires a leading `+`, so bare-format numbers in free error text slip through. E.164 (the #696 case) is covered; free text stays best-effort.

Approving — the `SCHEMA_VERSION` 1→2 bump is safe within drt-core (the renderers are version-agnostic; the views that consume the new blocks are your stated #808 follow-up, so this lands as data-ahead-of-UI). Thanks for the careful ADR and the bundled security fix! 🚀

Issue #414, opened 2026-04-29
feat: sync failure alert notifications (Slack / webhook)


yodakanohoshi:
## Motivation

`on_error: fail` stops a sync but provides no push notification. In production, operators need to know immediately when a sync fails — without polling `drt status`.

## Proposed config

```yaml
alerts:
  on_failure:
    - type: slack
      webhook_url: ${SLACK_ALERT_WEBHOOK}
      message: "drt sync {{ sync_name }} failed: {{ error }}"
    - type: webhook
      url: ${ALERT_WEBHOOK_URL}
```

- Fires after a sync run ends in error
- Supports Slack incoming webhook and generic HTTP webhook as first targets
- Template variables: `sync_name`, `error`, `rows_processed`, `duration_s`, `started_at`

## Scope

- `alerts:` top-level key in sync YAML (`SyncConfig`)
- Alert dispatch happens in `run_sync()` on exception / non-zero error count
- Separate `AlertConfig` Pydantic model

## Out of scope

- PagerDuty / OpsGenie (can use webhook target)
- On-success notifications
- Alerting for `drt test` failures

PR #815, opened 2026-07-20
feat(config): alert conditions — degraded-sync thresholds (#784)

### LINKED ISSUE #784: feat(config): alert conditions — row-error %, duration SLA, zero-rows, and DLQ-depth thresholds for alerts
## Problem

`alerts.on_failure` (#414) fires on hard failure only — all-or-nothing. Real reverse-ETL degradation is usually *partial*: `on_error: skip` quietly drops 4% of rows into the DLQ every night, a sync's duration creeps from 2 to 20 minutes as volume grows, row counts collapse to zero because an upstream mart went empty (sync "succeeds"). None of these page anyone today; operators discover them from angry CRM users.

## Prior art

- **Census/Hightouch**: alert thresholds (failed-row %, sync duration SLA) and "alert when 0 rows" toggles — their ops teams' most-used knobs.
- **dbt**: `warn_if`/`error_if` thresholds on tests (condition-based signaling).

## Proposal

Extend the existing `AlertsConfig` (keep `on_failure` untouched):

```yaml
alerts:
  on_failure: [slack]
  conditions:                       # mapping keyed by metric (one threshold per metric)
    row_errors_pct: { gt: 1 }       # failed rows / extracted (skips are expected — #757 match_policy, --limit, check_only)
    duration_seconds: { gt: 900 }   # SLA breach
    rows_extracted: { eq: 0 }       # empty-source guard
    dlq_depth: { gt: 500 }          # backlog accumulating across runs
```

- Evaluated post-sync from data already in `SyncResult`/DLQ store — no new collection
- Alert payload states which condition tripped + values; multiple conditions coalesce into one message per sync per run
- `--output json` includes tripped conditions so CI can react without parsing Slack

## Scope

**In:** condition schema (small fixed set above), evaluation + payload, docs.
**Out:** anomaly detection/baselines (v2 — needs history statistics), per-condition channel routing, paging integrations beyond existing slack/webhook (webhook covers PagerDuty et al.).

## Related

- #414 (alerts base), #278 (DLQ depth source), #276 (history — future baseline source for anomaly v2)

## Milestone fit

Candidate: v0.8.x — small config+engine-observer feature, big operational trust win.

Muawiya-contact:
Implements #784 — `alerts.on_degraded`, thresholds that fire on *partial* degradation, complementing `on_failure` (#414) which only fires on hard failure. Real reverse-ETL degradation is usually partial: a creeping error rate, a duration SLA breach, an empty source that "succeeds", a DLQ backlog accumulating across runs — none of which page anyone today. Evaluated post-sync from data already in `SyncResult` + the DLQ store (no new collection).

```yaml
alerts:
  on_failure: [{ type: slack, webhook_url_env: SLACK_WEBHOOK_URL }]   # unchanged
  on_degraded:
    channels: [{ type: slack, webhook_url_env: SLACK_WEBHOOK_URL }]   # empty => JSON-only (CI)
    conditions:                       # mapping metric -> threshold
      row_errors_pct: { gt: 1 }       # failed / rows_extracted, as %
      duration_seconds: { gt: 300 }   # SLA
      rows_extracted: { eq: 0 }       # empty-source guard
      dlq_depth: { gt: 500 }          # backlog, cumulative across runs
```

## Built to the three decisions on #784
- **Syntax → mapping** (`conditions:` metric → threshold, not a list): duplicate metrics are impossible for free and there's no `model_validator` unpack. Threshold is `{gt/lt/gte/lte/eq: N}`, **exactly one** operator (validated).
- **`row_errors_pct = failed / rows_extracted`, failed-only.** `skipped` is deliberately excluded — a skipped row is a *normal* outcome via #757 `match_policy` / `--limit` / `lookups.check_only`, so folding it into an *error* rate would false-fire on healthy syncs. When `rows_extracted == 0` the rate is `0%` (empty-source is the separate `rows_extracted` condition — never a divide-by-zero or a false 100%).
- **Evaluated at the CLI seam** (`_run_one` in `run.py`), so the engine (`sync.py`, the `on_failure` path) is **untouched** — good for the Rust boundary — and `drt build` inherits it for free. `dlq_depth` is read from `DlqStore` at that seam.

## Behaviour
- Multiple tripped conditions **coalesce into one alert per sync per run**.
- `--output json` gains `conditions_tripped: [{metric, operator, threshold, actual}, …]`, in a **deterministic** order (fixed metric order, not dict iteration).
- `on_degraded.channels` reuses the **existing** Slack/webhook dispatcher — I extracted a shared `dispatch_targets()` and had `dispatch_alerts()` delegate to it, so there's no duplicated dispatch logic and `on_failure`'s behaviour is byte-identical. `channels` defaults to empty, so conditions can be JSON-only for CI without wiring a paging target.
- Back-compat: no `on_degraded` block → zero behaviour change (no dispatch, no `conditions_tripped` key). Asserted by a test.

## Files (10; 2 new)
- `drt/config/sync_options.py` — `ConditionThreshold` / `DegradedConditions` (mapping) / `OnDegradedConfig`; `AlertsConfig.on_degraded`
- `drt/config/models.py` — re-export the 3 new types + `__all__`
- `drt/alerts/conditions.py` **(new)** — pure `evaluate_conditions()` + `row_errors_pct()` + `TrippedCondition`
- `drt/alerts/dispatcher.py` — extract `dispatch_targets()`, add `build_degraded_context()`
- `drt/alerts/__init__.py` — exports
- `drt/cli/commands/run.py` — CLI-seam hook in `_run_one`
- `tests/unit/test_alert_conditions.py` **(new)** + `tests/unit/test_alerts.py` — 40 tests: operators, failed-only + zero-guard, per-metric eval, ordering, coalescing, JSON, dispatch reuse, back-compat, CLI-seam end-to-end (with a real DLQ file feeding `dlq_depth`)
- `docs/llm/API_REFERENCE.md` (new Alerts section — also documents the previously-undocumented `on_failure`) + `CHANGELOG.md`

## Verification
- Whole-package `python -m mypy drt/` — clean (only the pre-existing `credentials.py:175 [no-any-return]`, an untouched file).
- `ruff check drt tests` — clean.
- Full `pytest -q` — `2098 passed, 22 skipped`; the only 6 failures are the known Windows `test_sqlite.py` `os.unlink` flake (passes on CI Linux). Zero PR-area failures.

## Two honest flags for review
1. **`on_degraded.channels` is a small extrapolation.** The decision comment showed `on_degraded:` with only `conditions:`; I gave it its own `channels` list (defaulting empty) rather than implicitly reusing `on_failure`'s. Happy to switch to reusing `on_failure`'s channels if that's what you meant — it's a small change.
2. **Default Slack message says "failed".** A degraded alert reuses `SlackAlertConfig.message` (default: ``drt sync `{sync_name}` failed: {error}``), so it renders ``… failed: degraded — row_errors_pct 4.2 (gt 1.0)`` — content-correct but the "failed:" prefix is slightly off. Users can override `message:`. A per-event default message felt out of scope; flagging rather than deciding.

Out of scope, as specified: anomaly detection / baselines (v2), per-condition channel routing, new paging integrations, a `skipped`-based metric.

Closes #784


Muawiya-contact:
@masukai Ready for review.

Muawiya-contact (commented):
DOne review.

masukai (commented):
This is a really clean implementation, @Muawiya-contact 🙌 — and a fast turnaround on #784. It lands every call from that thread faithfully:

- **Mapping form**, with the `_exactly_one_operator` validator using `is not None` — so `eq: 0` stays valid, which is exactly the trap I'd have worried about. 👌
- **`row_errors_pct = failed / rows_extracted` (failed-only)**, with the #757 reasoning in the docstring *and* a dedicated `test_row_errors_pct_excludes_skipped`, plus the `rows_extracted == 0 → 0%` guard so empty-source is only ever the `rows_extracted` condition's job.
- **Evaluated at the CLI seam**, engine untouched, `drt build` inherits it; one coalesced message per sync; `conditions_tripped` in `--output json`; empty `channels` → JSON-only. And you handled the schema-drift gate yourself (regen + ext `0.1.7`). The test coverage is genuinely thorough — the two CLI-seam end-to-end cases are the ones I'd have asked for.

**One thing I'd like before I approve** — a small robustness point, not a design change:

In `_run_one`, the `on_degraded` block runs on the **success path**, and `DlqStore(Path(".")).depth(sync.name)` there isn't wrapped. It sits inside the outer `try` but *outside* the inner run handler, so if that DLQ read ever raised (a permission / OS error on `.drt/dlq/…`), it would take down an otherwise-successful sync — a monitoring feature failing the run it's monitoring. Low probability (`depth()` just counts lines), but worth closing since we already treat alerting as best-effort (`dispatch_targets` swallows + logs).

Could you wrap the block best-effort? Something like:

```python
if not ctx.dry_run and sync.alerts is not None and sync.alerts.on_degraded is not None:
    try:
        from drt.alerts import build_degraded_context, dispatch_targets, evaluate_conditions
        from drt.state.dlq import DlqStore

        dlq_depth = DlqStore(Path(".")).depth(sync.name)
        tripped = evaluate_conditions(result, dlq_depth, sync.alerts.on_degraded.conditions)
        if tripped:
            entry["conditions_tripped"] = [...]
            dispatch_targets(sync.alerts.on_degraded.channels, build_degraded_context(...))
    except Exception as exc:  # noqa: BLE001 — degraded eval must never fail the run
        logger.warning("Degraded-condition evaluation failed for %r: %s", sync.name, exc)
```

Once that's in I'll approve + merge. Thanks again — this is exactly the shape #784 was after. 🚀

Muawiya-contact:
Thanks @masukai — and good catch. You're right that a monitoring read sinking the run it monitors is exactly the wrong failure mode, and it's inconsistent with `dispatch_targets` already being best-effort.

Done in `af0b99d`: wrapped the whole `on_degraded` block (the `DlqStore.depth` read + eval + dispatch) in a best-effort `try/except` that degrades to a logged warning — used your shape, with `logging.getLogger(__name__)` since `run.py` has no module logger. `on_failure` is untouched.

Added a test that pins the guarantee rather than just the wrap: `test_degraded_eval_failure_does_not_sink_the_run` patches `DlqStore.depth` to raise `PermissionError` and asserts the successful sync still exits 0 with no `conditions_tripped`. Full alert + run suites green, ruff + whole-package mypy clean.

(codecov's one missing line is the `ConditionThreshold.operator` property's `raise AssertionError` — unreachable given the exactly-one-operator validator, so I left it rather than write a test for a state the model forbids. Happy to add a `# pragma: no cover` if you'd prefer it off the report.)

Ready for another look whenever suits. 🙏

Muawiya-contact (commented):
DOne.

masukai (approved):
Perfect — this is exactly it, @Muawiya-contact 🙌. Wrapping the whole `on_degraded` block (the `DlqStore.depth` read + eval + dispatch) best-effort is precisely the guarantee I was after, and `test_degraded_eval_failure_does_not_sink_the_run` pins the *property* (a raising DLQ read must not sink an otherwise-successful run) rather than just the presence of the wrapper — nicer than what I'd have written.

On the codecov note: leave the `ConditionThreshold.operator` `raise AssertionError` as-is. It's genuinely unreachable given the exactly-one-operator validator, 98.9% patch clears the gate, and I'd rather keep the honest raise than annotate a can't-happen state — no `# pragma: no cover` needed.

Approving + merging. Thanks for the fast, clean turnaround on #784 — `alerts.on_degraded` is a great addition, and the "monitoring must never sink the run it monitors" posture is exactly right. 🚀

Issue #470, opened 2026-05-06
perf(diff): batch lookup queries for large diff sets (--diff optimization)


masukai:
## Background

The \`--diff\` implementation in #413 (v0.7.1) uses the existing lookup machinery, which fetches the entire destination table into memory and compares in Python:

\`\`\`sql
SELECT * FROM <table>
\`\`\`

For tables in the 10K-1M row range this is fine. For larger tables (~10M+) this becomes a noticeable preview cost on every \`--dry-run --diff\` invocation.

## Proposed optimization

Batch the lookup query by primary key extracted from the source records:

\`\`\`sql
SELECT * FROM <table> WHERE <id_col> IN (?, ?, ?, ...)
\`\`\`

Bound at e.g. 1000 IDs per batch (DB-specific limits). For destinations where source rows are << destination rows, this turns an O(destination_size) memory cost into O(source_size) network round-trips.

## Trigger conditions

This is **benchmark-gated**. We should only invest in this if real users hit the cost. Keep this issue open as a parking spot for the optimization, but don't implement until:

- A user reports slow \`--dry-run --diff\` on a large table, OR
- Internal benchmarking shows >5s preview time becomes common

If neither happens by v0.9, close as won't-fix.

## Out of scope

- Streaming or pagination of diff results — current shape (truncated sample) is intentional
- Cross-destination optimization — each DB has its own \`IN\` limit

Issue #400, opened 2026-04-22
bug: _test_display_name() missing freshness/unique/accepted_values test types


masukai:
## Summary

`_test_display_name()` in `drt/cli/main.py` only handles `row_count` and `not_null` test types. The three test types added in v0.6 (`freshness`, `unique`, `accepted_values`) fall through to `"unknown"`.

## Location

- `drt/cli/main.py:771-786` — `_test_display_name()`

## Expected

```
freshness(column, max_age=7 days)
unique(col1, col2)
accepted_values(status: active, inactive, archived)
```

## Actual

All three display as `unknown`.

PR #830, opened 2026-07-23
feat(test): custom SQL query tests, severity, and --store-failures (#779)

### LINKED ISSUE #779: feat(test): custom SQL query tests, severity: warn, and --store-failures for drt test
## Problem

`drt test` supports 5 fixed validators (row_count / not_null / freshness / unique / accepted_values) with binary semantics: any failure → exit 1. Three gaps as test suites grow real:

1. **No custom assertions**: anything beyond the 5 shapes ("no order rows with negative totals", "synced count within 1% of source count") can't be expressed — dbt's escape hatch (arbitrary SQL returning failing rows) is the feature that made its testing story stick
2. **No severity**: a soft freshness slip pages CI the same as a hard uniqueness violation — teams need `warn` vs `error` to keep suites honest (otherwise noisy tests get deleted)
3. **No failure samples**: a failing test reports a count; debugging requires hand-writing the inverse query

## Prior art

- **dbt**: singular tests (SQL file returning failing rows), `severity: warn|error` + `warn_if/error_if` thresholds, `--store-failures` persisting failing rows.

## Proposal

```yaml
tests:
  - name: no_negative_totals
    query: "SELECT * FROM {{ table }} WHERE total < 0"   # failing-rows contract: 0 rows = pass
    severity: warn                                        # warn | error (default)
  - unique: { columns: [email] }
    severity: error
```

- `query` test type: rendered with `{{ table }}` (qualified destination table), pass = zero rows; runs only on queryable destinations like existing types
- `severity: warn` → reported + counted, exit code unaffected; JSON output gains `warnings`
- `drt test --store-failures` → write up to N failing rows per failed test to `.drt/test_failures/<sync>/<test>.jsonl` (DLQ-style, masked columns respected — reuse `sync.mask` so PII doesn't land in failure files)

## Scope

**In:** query test type, severity on all types, store-failures with mask integration + row cap, `_test_display_name` update (#400's lesson: keep the display map in sync).
**Out:** source-side tests (tests stay destination-side), `warn_if/error_if` numeric thresholds (follow-up), relationships test (SQL `query` type covers it; revisit if demand).

## Related

- #233 (test framework), #366/#371 (test UX), #400 (display-name drift bug — the map to extend), #427 (mask reuse)

## Milestone fit

Candidate: v0.9 — testing depth is an "enterprise trust" theme; `query` type alone is a strong first PR.

Muawiya-contact:
Closes #779.

## Three parts

1. **`query` test type** — arbitrary SQL that returns the FAILING rows (0 rows = pass). `{{ table }}` renders to the qualified destination table via the same `_safe_table()`-validated value every other type uses. Wrapped in `SELECT COUNT(*) FROM (...) AS _drt_query_test` to preserve the existing single-int `execute_test_query()` path unchanged — mirrors the `_drt_base` subquery convention already in `engine/resolver.py`. No new execution path, no new injection surface beyond `model:` SQL's existing trust model. A stray `;` becomes a syntax error inside the parenthesized subquery rather than a second statement. Runs only on queryable destinations — inherited automatically from the existing per-sync gate, no new gate needed.

2. **`severity: warn | error`** (default `error`) on every test type. `warn` runs, is reported + counted, but never fails the exit code. `--output json` gains a top-level `warnings` array. `drt build` and the `drt_run_test` MCP tool share this via the existing `execute_tests_for_sync` seam / a small MCP parity fix, so neither drifts from `drt test` on this feature.

3. **`drt test --store-failures`** (`--store-failures-limit`, default 10) writes up to N failing rows per failed test to `.drt/test_failures/<sync>/<test>.jsonl`. `sync.mask` (#427) is applied immediately after fetch, before anything else touches the rows or reaches disk. `row_count` has no per-row failure concept, so it's an intentional no-op there.

## Quality gates
- `ruff check drt tests` — clean.
- `python -m mypy drt/` (whole package) — clean; only the pre-existing `credentials.py:175 [no-any-return]` (unrelated file, present on clean main).
- `python -m pytest -q` (full suite) — **2273 passed, 15 skipped, 6 failed**. All 6 failures are `test_sqlite.py::TestSQLiteSource` — the known Windows `os.unlink` flake (reproduces on clean `main`, passes on CI Linux). Zero PR-area failures.
- Proactive VS Code schema regen (`0.1.7` → `0.1.8`) for the new `SyncTest` fields, verified idempotent.

## Safety verification (the two things I was asked to prove, not just claim)

**SQL execution path** — `query` goes through the exact same call site as every other type ([`drt/cli/commands/test.py`](drt/cli/commands/test.py)): `query, check = build_test_query(test_def, table)` → `execute_test_query(sync.destination, query)`. No branch, no separate path. `{{ table }}` renders via `_QUERY_TEST_ENV.from_string(query).render(table=table)` in [`drt/engine/test_runner.py`](drt/engine/test_runner.py), using the same `_safe_table()`-validated value the other 5 types already use — table-name injection is rejected identically (`Invalid character in table name`).

**Mask reuse (#427)** — `apply_mask(raw_rows, sync.sync.mask)` runs immediately after `fetch_failing_rows()` returns and before the write; `raw_rows` is never referenced again. Proven end-to-end in `test_store_failures_masks_pii_before_write`: a mocked destination returns a **raw, unmasked** row (real email + SSN), and the test reads the actual bytes written to disk and asserts the raw values appear nowhere in the file while an unmasked field passes through untouched.

## Design decisions flagged for your call

**Need your confirmation or adjustment (a default and a public JSON contract — your calls):**

1. **`--store-failures-limit` default = 10.** The issue didn't specify a default or how to configure it. I picked 10 (a "show me a handful" debugging default) and added a `--store-failures-limit` flag to make it configurable. **Please confirm or adjust the default.**
2. **JSON `warnings` shape**: a flat top-level array, `[{"sync": ..., "test": ..., "value": ...}]`, not nested under `results` — so `jq '.warnings'` works without walking the per-sync/per-test tree. **Please confirm or adjust this shape** — it's a public JSON contract once shipped.

**Deliberate divergences from the issue's literal text (implemented, flagging for visibility):**

3. **`--store-failures` is overwrite-per-run, not accumulate.** Each run replaces `<sync>/<test>.jsonl` with *this* run's failures, and clears the file entirely once the test passes again. This is a deliberate divergence from the DLQ's accumulate-until-`drt retry`-drains-it model (`drt/state/dlq.py`) — a debugging sample should reflect current state, not history, and a stale failing-rows file for a now-passing test would be actively misleading. Flagging so this reasoning is visible rather than assumed.
4. **An explicit `name:` now wins the display label for every test type, not just `query`.** The issue's example only shows `name:` on the `query` test. My own test caught the inconsistency: a `not_null` test with `name: "warn_one"` was showing the auto-generated `not_null(a)` in output while the *file* used `warn_one` for its path. Fixed so `name:` always wins the label too (`not_null(warn_one)`) — a behavior addition beyond the issue's literal example, and it touches the `test_display_name` map #400 was a drift bug in, so flagging it explicitly for review.

(5. `row_count` + `--store-failures` = intentional no-op, not an error — no per-row failure concept for a whole-table aggregate. Not flagging for a decision, just noting it's deliberate.)

## Changeset
16 files modified, 2 new (`drt/state/test_failures.py`, `tests/unit/test_store_failures.py`). Full list and rationale in the commit message.

Opening as **draft** — items 1 and 2 above need @masukai's confirmation before this is merge-ready.


masukai:
Nice — this is a lot of ground covered (#779's three parts in one pass). No rush since it's a draft.

One thing the CI caught that's worth a look before you polish: the 3.13 failure in `test_failing_rows_predicate_matches_count_predicate` is a **real signal, not a flake**. `build_test_query()` and `build_failing_rows_query()` each evaluate the freshness `now()` independently, so the two `WHERE` predicates drift by microseconds:

```
- updated_at < '2026-07-16T19:45:52.389591+00:00'   (count query)
+ updated_at < '2026-07-16T19:45:52.389570+00:00'   (failing-rows query)
```

The test's own premise — "the COUNT check and the failing-rows sample must query the exact same `WHERE`, single source of truth" — is exactly the right invariant. So the fix is to **compute the predicate once and share it** between the `COUNT(*)` and the failing-rows `SELECT`, rather than rebuild it per call (which re-stamps `now()`). That also removes the drift risk permanently, not just in the test.

Really like the `SELECT COUNT(*) FROM (<rows_query>) AS _drt_query_test` wrapping to keep the single-int `execute_test_query()` path unchanged — clean. Ping me when it's ready for a proper review!

Muawiya-contact:
@masukai Ready for review now 😊.

masukai (changes requested):
Thanks for turning this around so fast — and for fixing the drift at the root rather than in the test. I want to be specific about that because it's the best part of this PR: `67c07e49` deleted the four parallel predicate builders and made `COUNT(*)` derive from `build_failing_rows_query`'s own output, so the two predicates *cannot* diverge structurally, not just "don't happen to." I mutation-tested the guards (reverted the reuse so `build_test_query` ignores the provided string) and three tests fail, including an end-to-end CLI one that catches a microsecond delta. That's the real thing. 👏

Two blocking items and one small bug below. Everything else is follow-up material.

---

## 🔴 Blocking

### 1. The `;`-defanging claim isn't true on Postgres or Snowflake

`drt/engine/test_runner.py` (the wrap) says:

> also defangs a stray `;` in `query:` SQL into a syntax error inside the parens rather than a second statement.

The wrap is plain string concatenation, so a `query:` value can close the paren and start a new statement. I ran this:

```
query: "SELECT 1 FROM t) AS x; DROP TABLE users; SELECT COUNT(*) FROM (SELECT 0"
```

which renders to a valid **three-statement** batch:

```sql
SELECT COUNT(*) FROM (SELECT 1 FROM t) AS x; DROP TABLE users; SELECT COUNT(*) FROM (SELECT 0) AS _drt_query_test
```

Per driver, as far as I could verify:

- **MySQL — genuinely safe.** `MySQLDestination._connect` never sets `client_flag`, and pymysql's default is `0`, so `CLIENT.MULTI_STATEMENTS` is off. Your instinct holds here.
- **Snowflake — the worst case.** `_connect` never passes `autocommit`, and the connector defaults to `True` — our own comment in `snowflake.py` ("Snowflake autocommits, so the SWAP commits before the DROP") confirms it. A second statement **commits**.
- **Postgres — protected only by accident.** psycopg2 defaults `autocommit=False`, so uncommitted DDL/DML rolls back on `close()`. That's a side effect of connection defaults, not a designed guard, and it doesn't cover autocommitting DDL or side-effecting functions.
- **ClickHouse — I didn't verify** (couldn't exercise `clickhouse_connect` locally).

I don't think the *feature* is wrong — arbitrary SQL is the point, and `model:` is the same trust model. My concern is narrower: the comment promises a guarantee that isn't there, and a future maintainer (or I, in six months) will believe it. One thing worth naming in the framing though: `model:` runs against the **source**, while `query:` runs against the **destination**, where drt holds write credentials — so the blast radius genuinely is larger.

Either fix is fine by me:
- **(a)** drop the claim: delete that comment clause and reframe as "`query:` executes arbitrary SQL by design, same trust level as `model:`" — honest and defensible; or
- **(b)** make it real: reject `;` outside string literals at the pydantic layer in `sync_options.py`.

I'd lean (a), optionally plus a cheap validator, since (b) done properly means SQL parsing.

### 2. A storage failure can turn a **passing** test red

In `drt/cli/commands/test.py`, `_store_or_clear_failure_sample(...)` sits inside the same `try` as the query, and the `except Exception` records `{"passed": False, "error": ...}` and sets `had_failures = True`. So:

- a test that **passes** (count `0`) but whose `clear_test_failures` hits `PermissionError` → reported `passed: False`, **exit 1**
- a fetch error while sampling → the real `value` is discarded and replaced by the exception text

A debugging-convenience flag shouldn't be able to manufacture a test verdict. Wrapping just the storage call in its own `try/except` — surfacing a warning and maybe `failures_stored: {"error": ...}` — would keep the verdict authoritative. There are currently no tests on this path, so a test where the write raises and the passing verdict survives would be the one to add.

### 3. `--store-failures-limit` has no lower bound

`store_failures_limit: int = typer.Option(...)` has no `min=`, and it's interpolated straight into `... AS _drt_sample LIMIT {int(limit)}`. `--store-failures-limit -5` produces `LIMIT -5`, a syntax error on all three SQL dialects — and because it throws inside the per-test `try`, the user sees "your test failed" rather than "bad flag." `min=1` on the option fixes it.

---

## 🟡 Non-blocking / follow-ups

- **`project_dir` isn't wired.** `execute_tests_for_sync(..., project_dir: Path = Path("."))` is never passed by either caller, so paths are effectively CWD-relative and the parameter implies configurability that doesn't exist. Not broken (the default is correct, and `drt build` simply leaves `store_failures=False`), but worth either wiring or dropping. Related: the PR body says build "shares this via the existing seam" — true for `severity`, not for `--store-failures`; worth stating the `drt test`-only scope explicitly.
- **Driver coverage.** 36 of the 44 uncovered lines are `query.py`'s dispatch: `_fetch_failing_rows_mysql` entirely (including the `isinstance(row, dict)` DictCursor branch), plus the ClickHouse and Snowflake helpers. Only Postgres is exercised. Mitigating: `_query_mysql` / `_fetch_rows_mysql` / `_query_clickhouse` have no driver tests either, so this matches existing practice rather than regressing it — but the MySQL dict branch and ClickHouse's different API (`result.column_names`) are the parts that will actually run against real warehouses.
- **`_test_id` slug collisions.** `"email nn"` and `"email/nn"` both slug to `email-nn`, so two named tests in one sync overwrite each other's `.jsonl`. The `index` prefix that would prevent it is applied only on the fallback path. (Path traversal *is* correctly neutralized — I checked `"../../etc/passwd"` → `"etc-passwd"`.)
- **MCP severity parity** (`mcp/server.py`) has no test. The anti-#400-drift claim is reasonable but currently rests on nothing executable, and #400 was itself a drift bug.
- **`drt build` has no `warnings` array** in its JSON while `drt test` does — asymmetric public contract.
- On your two open questions: **limit default 10** is the right call (dbt stores all; 10 is safer for a tool writing PII locally). **Flat top-level `warnings`** is also right for `jq`. One gap — `_collect_warnings` collapses `value` and `error` into one key, so a consumer can't distinguish a threshold breach from an exception; worth splitting since you're pinning this contract now.
- `test_failing_rows_predicate_matches_count_predicate` still needs the clock freeze because it calls both builders independently — a path production no longer takes. Its docstring is admirably honest about this and points at the test that covers the real contract. Keeping it is fine (I confirmed it catches non-time drift), just noting it verifies predicate *shape*, not the production invariant.

---

## What I liked

- The mask ordering isn't just correct, it's **proven correctly**: `test_store_failures_masks_pii_before_write` returns raw PII from the mock and asserts on the **actual bytes on disk** (whole-file, not the field), while letting an unmasked field through. That's a real privacy property, not a mirror of the implementation.
- `test_build_test_query_reuses_provided_failing_rows_query` deliberately passes a *non-matching* string to prove there's no silent recomputation. Good adversarial instinct.
- **Backward compatibility is pinned, not asserted.** `severity` defaulting to `"error"` plus `test_default_severity_matches_pre_779_behavior` means every pre-#779 config takes the identical path — I looked for a behavior change for non-opted-in configs and found none.
- **Engine purity respected**: all filesystem work lives in `drt/state/test_failures.py`, invoked from the CLI seam; `engine/test_runner.py` gained no I/O.
- **The periphery grep was actually done** — connector docs ×4, `API_REFERENCE.md`, `CONTEXT.md`, the vscode schema (regenerated, matching the pydantic model including `minLength: 1` and the `severity` enum, with a version bump and changelog entry), MCP, CHANGELOG. That's the discipline this repo asks for, done without being asked.
- `test_display_name_covers_all_test_types` diffing `SyncTest.model_fields` against an expected set is a genuine structural guard against the #400 class.

Happy to file the follow-ups (driver coverage, slug collisions, MCP severity test, a `docs/guides/data-tests.md`) as separate issues so they don't grow this PR. Just the three blocking items and I'm glad to merge this.

masukai:
Heads-up on a merge-order side effect, sorry: I just merged #829 (`perf(diff): batch --dry-run --diff destination lookup by source PKs`, closes #470), which lands in **`drt/destinations/query.py`** — the same file your `fetch_failing_rows` + 4 driver helpers go into. So this branch will need a rebase on top of `main`.

Concretely, expect conflicts in two places:
- **`drt/destinations/query.py`** — #829 added `fetch_rows_by_keys` plus `_fetch_rows_by_keys_{postgres,mysql,snowflake}` right after the `fetch_rows` dispatcher. Your `fetch_failing_rows` block goes in the same neighbourhood; it should be a "keep both" resolution, no logic overlap. Worth a look at the new helpers though — `_in_placeholder` / `_flatten_key_params` / `_chunks` are there now, and if any of them saves you code in the failing-rows path, feel free to reuse.
- **`CHANGELOG.md`** — the usual `[Unreleased]` collision; keep both entries.

No rush, and please fold it in with the blocking fixes rather than doing a separate rebase pass — no point paying that cost twice.

Muawiya-contact:
Thanks for the thorough review @masukai — especially for actually mutation-testing the predicate-sharing guard, that's exactly the kind of check I couldn't do myself. All three blocking items addressed in `927185a`:

**1. `;`-defanging claim** — you're right, that was false. Went with your option (a): dropped the claim, replaced the comment with an honest note that the wrap is plain string concatenation (not a sandbox), documented the per-driver behavior you verified (Snowflake autocommits — worst case; Postgres's rollback-on-close is a side effect of connection defaults, not a guard; MySQL is genuinely safe), and named the source-vs-destination trust asymmetry (`model:` reads the source, `query:` writes-credentialed against the destination).

**2. Storage failure could flip a verdict** — moved `_store_or_clear_failure_sample` out of the count-check's `try/except` into its own, so a storage error now lands as `failures_stored: {"error": ...}` without touching `passed`/`value`/exit code. Added both tests you described: a passing test surviving a `clear_test_failures` `PermissionError`, and a failing test's real value surviving a `fetch_failing_rows` error.

**3. `--store-failures-limit` lower bound** — added `min=1`.

Full suite: 2280 passed, 15 skipped, 6 failed (the known Windows `test_sqlite.py` `os.unlink` teardown flake, unrelated). ruff/mypy clean.

Left the non-blocking follow-ups (project_dir wiring, driver coverage, slug collisions, MCP severity test, `drt build`'s missing `warnings` array, splitting `_collect_warnings`' value/error keys) for separate issues as you suggested — happy to file them if useful, or you can when you have the list handy. Ready for another look.

Muawiya-contact (commented):
Done ✅

masukai:
All three blocking items are resolved. I verified each one rather than taking the diff at face value — details below. One thing left before I can merge, and it's mechanical.

## Verified

**1. `;`-defanging claim** — the reframing is accurate. I re-ran the adversarial value against the current branch and it still renders a three-statement batch:

```sql
SELECT COUNT(*) FROM (SELECT 1 FROM t) AS x; DROP TABLE users; SELECT COUNT(*) FROM (SELECT 0) AS _drt_query_test
```

which is exactly what the new comment now says happens. Option (a) was the right call, and you kept the part that matters most — the source-vs-destination trust asymmetry is now written down where the next person will actually find it. A comment that says "this is not a sandbox" is worth more than one that quietly overstates a guarantee.

**2. Storage failure can't flip a verdict** — confirmed by mutation rather than by reading. I collapsed the new dedicated `try/except` back into the surrounding flow (`except Exception as store_err:` → `if True:`) and exactly your two new tests failed:

```
FAILED test_store_failures_clear_error_does_not_fail_a_passing_test
FAILED test_store_failures_fetch_error_preserves_failing_verdict
```

So the guard is load-bearing, not decorative. The `continue` in the verdict `except` is the subtle part of this restructure and it's correct — a test that errors during the count check skips storage entirely rather than trying to sample rows for a query that never ran.

**3. `--store-failures-limit` lower bound** — `min=1` behaves as intended:

```
Invalid value for '--store-failures-limit': -5 is not in the range x>=1.
```

Exit code **2** (usage error) instead of the old exit 1, so a bad flag no longer masquerades as a failing test. That distinction was the actual point.

Also ran the full unit suite on your branch: **2227 passed, 5 skipped**. The 2 failures I saw were `test_cli_version.py` line-wrapping against my own long worktree path — environmental, not yours. CI is green across all 9 checks.

## One thing left: the rebase

`main` has moved 5 commits and this branch hasn't picked them up yet, so GitHub still shows `BLOCKED`. Since my earlier heads-up, **#832** (`#719 phase 2b — mirror-finalize pull-up`) also landed on top of #829.

I checked what actually conflicts, and it's still just the two files I flagged — #832 touches the load path, not yours:

- **`drt/destinations/query.py`** — keep-both with #829's `fetch_rows_by_keys` block, no logic overlap
- **`CHANGELOG.md`** — the usual `[Unreleased]` collision

So the rebase is mechanical. No new surprises from #832.

## On the follow-ups

Agreed on leaving them out of this PR — it's at +1523/-89 already and the blocking items were the merge gate. I'll file the driver coverage, slug collisions, MCP severity test, `drt build` `warnings` asymmetry, and the `value`/`error` split as separate issues so they're tracked rather than remembered.

Rebase it and I'll merge. Nice work on this one — the predicate-sharing refactor in `67c07e4` is the part I'd point someone at, and you got there from a CI red rather than from a spec.

masukai:
Filed the follow-ups so they're tracked rather than remembered — none of these are gates on this PR:

- **#834** — driver coverage for `fetch_failing_rows` (MySQL `DictCursor` branch + ClickHouse's different row API). Noted in the issue that this matches existing practice in `query.py` rather than regressing it.
- **#835** — `--store-failures` slug collisions (`"email nn"` / `"email/nn"` → same `.jsonl`). Also recorded there that path traversal *is* correctly neutralized, so nobody re-audits that later.
- **#836** — MCP `severity: warn` parity has no test.
- **#837** — `warnings[]` conflates a threshold breach with an exception in one `value` key.
- **#838** — `drt build` JSON lacks the `warnings[]` array `drt test` emits.

Two notes on sequencing:

**#837 is the time-sensitive one.** `warnings[]` is new public contract in this PR, so splitting `value`/`error` is free before it ships and a breaking change after. Worth catching in the same release rather than the next one. **#838** should follow whatever shape #837 settles on, so it doesn't copy the current one and inherit the problem.

**#836** overlaps with #723's `mcp/server.py` split — the minimal fix is just a parity test, but if that split lands first, having `drt_run_test` call `execute_tests_for_sync` would remove the duplicated severity rule entirely. Probably worth sequencing after #723 rather than doing both.

All are unassigned — take any of them if you want them, or leave them. #835 and #836 are the most self-contained.

Muawiya-contact (commented):
@masukai done.✅

Muawiya-contact:
@masukai I have some questions regarding the `drt-web` repository. I posted them under the issue; please read and respond when you have a moment.

masukai:
Nice work getting all four #723 parts out — I'll review those next.

Before that, a nudge on this one: it's still the only thing of yours sitting unmerged, and the longer it waits the more `main` moves under it. Since my last comment, **#833** landed and added a third conflicting file:

- **`drt/destinations/query.py`** — keep-both with #829's `fetch_rows_by_keys` block (no logic overlap)
- **`drt/cli/output.py`** — new. #833 changed the `--dry-run --diff` deletion preview; your change is in `print_test_result`. **Different functions in the same file**, so also keep-both.
- **`CHANGELOG.md`** — the usual `[Unreleased]` collision

I checked each one and there's still no real overlap — it's a mechanical resolution, not a redesign. But that list grew from two files to three in about a day, which is the argument for doing it now rather than after #839–#842 land. #839 in particular moves a lot of `mcp/server.py` around, and this branch touches that file too.

Everything else here is done: all three blocking items verified, CI green, 2227 tests passing. **Rebase is the only thing between this and merge.**

If you'd rather land the #723 series first and rebase this afterward, that's a defensible call too — just tell me which order you want, so this doesn't sit waiting on nothing.

masukai:
I owe you an apology on merge ordering here. This PR has been the longest-waiting one and it was already carrying a rebase from #829 — I should have landed it *before* touching `main` again. Instead I merged #839 (the `mcp/server.py` split) first, which made your rebase harder rather than easier. That's on me.

To make up for it, here's the exact map of where your changes now go. The good news is it's mechanical — nothing about your logic changes.

**`drt/mcp/server.py` → `drt/mcp/tools/run_test.py`.** #839 split each tool into its own module, and `drt_run_test` now lives in `tools/run_test.py`. All four of your severity edits move there verbatim:
1. the docstring line documenting `severity` in the `tests` list shape
2. `"severity": test_def.severity` in the pass/fail entry
3. `if not passed and test_def.severity != "warn":`
4. `"severity": test_def.severity` in the `except` entry + the `if test_def.severity != "warn": had_failures = True` guard

**`drt/destinations/query.py`** — still the #829 conflict I flagged earlier: `fetch_rows_by_keys` and friends landed near where your `fetch_failing_rows` goes. Keep-both resolution; no logic overlap. (#843, still open, adds `fetch_all_keys` in the same area — I'll rebase *that* onto yours rather than the other way round, so don't worry about it.)

**`CHANGELOG.md`** — the usual `[Unreleased]` collision, keep both entries.

I've also held off merging #840/#841/#842 — they're approved and green, but three of your four #723 parts now conflict on `CHANGELOG.md` (and #841 on `query.py`) purely because of the order I merged in. If you'd rather I rebase those three myself so you can focus on this one, say the word and I'll do it — you shouldn't have to pay for my sequencing.

On the review state: I confirmed all three blocking items were genuinely fixed back on the 25th. The `CHANGES_REQUESTED` flag is stale — once this is rebased and green I'll flip it and merge straight away.

masukai (approved):
Rebased this onto `main` myself and merging — the conflicts were a consequence of my merge ordering (I landed #829 and then all four #723 parts while this sat waiting), so they were mine to clear.

What the rebase involved, for the record:

- **`drt/mcp/server.py`** was the only real decision. #839 turned the tool bodies into thin delegates, so your severity logic no longer had a home there. I took the delegate shape and moved your four edits into **`drt/mcp/tools/run_test.py`** — the `severity` field on both the pass/fail and `except` entries, and both `!= "warn"` guards on `had_failures`, comment included. Your **docstring** change stayed in `server.py`, since that's the wrapper MCP clients actually read — the public contract belongs with the public surface.
- `drt/destinations/query.py`, `drt/cli/output.py`, `CHANGELOG.md` — keep-both. No logic overlap; git flagged them only because #829/#833/#843 touched other regions of the same files.

Verified after: **2298 passed**, mypy and ruff clean, and the 48 MCP tests green. The blocking items from my review were already confirmed fixed on the 25th — the `CHANGES_REQUESTED` flag was just stale.

Nice piece of work. The predicate-sharing refactor in particular is better than what it replaced, and the periphery discipline (connector docs, `API_REFERENCE`, vscode schema regen with a version bump, MCP parity) was all there without being asked.

Issue #719, opened 2026-07-07
refactor(destinations): extract BaseSqlDestination to DRY Postgres/MySQL load + tracked-mirror (~1700 dup lines)


masukai:
## Summary

`PostgresDestination` (`drt/destinations/postgres.py`, 961 lines) and `MySQLDestination` (`drt/destinations/mysql.py`, 757) are structural clones. No runtime base class exists — `destinations/base.py` only holds Protocols. Method-for-method the bodies are identical except for the quoting primitive (psycopg2 `sql.Identifier` vs backtick `_quote_ident`) and the placeholder-expansion trick.

Verified duplications (line-for-line):
- `load()` mirror.scope missing-column validation (pg `183-195` == my `125-137`, same error string), replace/upsert dispatch, mirror-key + scope accumulation loop (pg `238-270` == my `180-211`)
- `_load_replace` (pg `313-367` / my `260-306`), `_load_replace_swap` (pg `369-434` / my `308-365`), `_load_upsert` (pg `827-872` / my `621-661`) — identical control flow + `RowError(...)` construction
- `finalize_sync` (pg `436-494` / my `367-415`), `_finalize_mirror` (pg `496-592` / my `417-516`), and especially `_finalize_mirror_tracked` (pg `594-700` / my `518-619`, ~85 lines each) — full duplication of the tracked-mirror state-table DDL, previous-key read, `diff_keys`, delete, state rewrite
- `list_orphan_swap_tables` / `drop_orphan_swap_tables`

## Proposal

Extract a `BaseSqlDestination` with a small dialect surface — `quote_ident(table)`, `build_insert_sql`, `build_upsert_sql`, `expand_not_in(cols, keys)`, `connect(config)` — collapsing ~1700 lines to one implementation + two thin adapters. `BaseSqlDestinationConfig` already exists on the config side, so the pattern is established.

**Effort/risk:** L / MED — core write path, but coverage is strong (833 tests + empty-batch contract suite + DuckDB E2E harness).
**Collision check:** none — #699/#707 are Snowflake/Databricks only; #402 is profiles. Safe to start now.
**Note (per repo convention):** file-moving refactor — list affected paths in the PR body and notify on any open PR that touches these files post-merge.

masukai:
**Done** — landed across three PRs, all pure refactors with zero behavior change:

| PR | Phase | What moved into `BaseSqlDestination` |
|---|---|---|
[...]

Issue #836, opened 2026-07-25
test(mcp): severity: warn parity in drt_run_test rests on a comment, not a test (#400-class drift risk)


masukai:
Follow-up from the #830 review (#779). Non-blocking for that PR.

## Problem

#830 adds `severity: warn` handling to `drt_run_test` in `drt/mcp/server.py`, mirroring the rule `drt test` and `drt build` apply: a `warn`-severity failure is reported but must not flip the top-level `status` to `"failed"`.

The implementation looks right, but **nothing executable enforces it**. The parity is currently maintained by a code comment.

That matters here more than usual because **#400 was itself a CLI/MCP drift bug** — the exact failure mode this comment claims to prevent. A comment asserting "this doesn't drift" is the weakest possible guard against a class of bug that has already happened once in this file.

## Why it's easy to drift

`drt/mcp/server.py`'s `drt_run_test` duplicates the test-execution loop rather than calling `execute_tests_for_sync`. So a future change to severity semantics in `drt/cli/commands/test.py` will not propagate here, and no test will notice.

## Suggested fix

Minimum: a test asserting that a `severity: warn` failure leaves `drt_run_test`'s `status` as `"passed"` while still appearing in `results[].tests[]` with `severity: "warn"` and `passed: false`.

Better, if the duplication is worth addressing: have `drt_run_test` call `execute_tests_for_sync` so there is one implementation of the severity rule rather than two. That's a larger change and overlaps with #723's `mcp/server.py` tool split — probably worth sequencing after it rather than doing both at once.

## Acceptance criteria

- [ ] Test pinning MCP severity parity (`warn` failure does not flip `status`)
- [ ] Test covers the exception path too — `severity: warn` on a test that *errors* also must not flip `status`

PR #839, opened 2026-07-26
refactor(mcp): split mcp/server.py into per-tool modules (#723 part 1)

Muawiya-contact:
Closes part 1 of #723 — the `mcp/server.py` tools split masukai flagged (the other three items in #723 — `schema.py` table-driving, `_connect()` normalization, `run.py` cleanup — are separate future PRs, explicitly out of scope here).

## What changed

`drt/mcp/server.py`'s 14 `@mcp.tool()` closures each re-loaded project/sync config inline. Split into:

```
drt/mcp/
    server.py       — thin: builds FastMCP + one shared McpContext, registers
                      each tool via a thin wrapper (exact original name/
                      params/docstring), delegates to tools/ in one line
    _context.py      — McpContext + _load_ctx(project_dir): centralizes
                      load_project() / load_syncs() / load_syncs_safe() /
                      find_sync(name), lazy — nothing loads until asked for
    tools/
        list_syncs.py, run_sync.py, run_test.py, get_status.py,
        get_history.py, validate.py, get_schema.py, list_connectors.py,
        dlq.py, retry.py, get_manifest.py, list_profiles.py,
        test_profile.py, doctor.py
```

One tool per module, mirroring `drt/cli/commands/`'s convention; `_context.py` sits beside `server.py`/`tools/` the way `drt/cli/_helpers.py` sits beside `drt/cli/commands/`.

## Duplication removed

- `load_syncs(project_dir)` / `load_project(project_dir)` / `load_syncs_safe(project_dir)`, each re-imported and re-called inline per tool → `ctx.load_*()`.
- The "find one sync by name or return `{"error": f"No sync named '{sync_name}' found."}"` pattern — byte-identical error message in `drt_run_sync` and `drt_retry` (different implementations, same behavior) → `ctx.find_sync(name)`.
- `drt_run_test`'s sync filtering is structurally different (filters a list it keeps using afterward) — left as-is rather than forced through `find_sync`, to avoid a second `load_syncs()` call.

## No behavior change

- Every `@mcp.tool()` wrapper's name/parameters/defaults/return type/**full docstring** is byte-identical to what's on `main` today — verified with an AST diff, not just eyeballing (it caught and fixed two typos I introduced hand-retyping docstrings — a line reflow and a `,`/`;` slip, neither in the original).
- Function-local imports (e.g. `from drt.cli.main import _get_source`) stayed function-local in the new modules — several existing tests monkeypatch these at their source module, which only works if the import runs fresh per call, not hoisted to module scope.
- `drt_run_test`'s implementation is unchanged from `main` today — it does **not** pick up the `severity`-aware parity fix from the unrelated, not-yet-merged #779 branch.
- 5 tools that never touched `project_dir` before (`get_schema`, `list_connectors`, `list_profiles`, `test_profile`, `doctor`) take no `ctx` param — nothing new gets loaded for them.

## Gates

- `ruff check` clean.
- All 55 tests exercising `create_server`/the MCP tools pass unmodified: `test_mcp.py` (42), `test_cli_status_history.py` (9), `test_cli_commands_smoke.py` (4, mocks `drt.mcp.server` for CLI wiring).

## Design decisions

- `ctx` is only threaded to tool functions that already used `project_dir`/sync config — not uniformly, to keep each function's parameters honest about what it needs.
- All 14 tools get a thin wrapper in `server.py` (including the 5 needing no `ctx`), for one uniform registration pattern, rather than registering those 5 directly with a `name=` override.
- `find_sync` is extracted, but the surrounding `{"error": ...}` dict construction is not — kept each call site building its own error dict to minimize the diff's behavioral surface.
- Tool-function imports in `server.py` are at module top level; each tool's own heavy dependencies stay deferred inside its function body, preserving the original lazy-import behavior (confirmed `drt.mcp.server` is still only imported lazily from `drt/cli/commands/mcp.py`).

CHANGELOG entry added under `### Changed`, matching the `#719` precedent for internal-only refactors.


Muawiya-contact (commented):
DOne.

masukai (approved):
Approving. The thing I most wanted to check on a split like this is whether the **public MCP surface** survived it, since that's what agents actually bind to and a silently-dropped `@mcp.tool()` decorator would be invisible to unit tests.

Built the server on both `main` and this branch and diffed the registry:

```
13 tools on main, 13 here — names and descriptions byte-identical
```

So `create_server()` still registers exactly the same surface. That's the property that mattered, and it holds.

Also ran the full suite (2230 passed), ruff clean, mypy clean over `drt/mcp/`.

What I liked: `_context.py` as a shared loader rather than each tool module re-deriving project state — that's the part that would have rotted fastest if each of the 13 modules had grown its own copy. And keeping `server.py` as the single registration point means there's still one file that answers "what tools exist", which is worth more than distributing the decorators.

This lands the file that #836 is about (MCP `severity: warn` parity has no test). Not a blocker — just noting the two touch the same module, so whichever merges second may want a glance.

PR #840, opened 2026-07-26
refactor(destinations): table-drive schema.py introspection (#723 part 2)

Muawiya-contact:
Part 2 of #723 — the `schema.py` table-drive masukai flagged, including Databricks from the start (unblocked by #707). The other two #723 items (`_connect()` normalization, `run.py` cleanup) are separate future PRs, explicitly out of scope here — as is any MCP work (part 1, #839).

## What changed

`describe_columns` branched across four nearly-identical `_describe_{postgres,mysql,snowflake,databricks}` implementations, each repeating: connect → build query → execute → fetch → empty-check → categorize each row. Only 3 things genuinely varied: the SQL/params, the connect call convention, and the categorization rule — plus one real (not incidental) split the existing tests prove: Postgres/MySQL use a bare `conn.cursor()`, Snowflake/Databricks use `with conn.cursor() as cur:`.

- `_DescribeSpec` (`NamedTuple`): `connect`, `build_query`, `categorize`, `cursor_is_context_manager` — the 4 things that actually differ per dialect.
- `_DESCRIBE_SPECS: dict[str, _DescribeSpec]`, keyed by `config.type` (`"postgres"`/`"mysql"`/`"snowflake"`/`"databricks"`) — matches `drt/connectors/registry.py`'s existing string-keyed convention (no dataclass/NamedTuple dispatch table existed anywhere else in the repo; checked before choosing this shape).
- `describe_columns(config)` is now 8 lines: look up the spec, delegate, catch-all → `None`.
- `_describe(config, spec)`: the single shared pipeline — connect, run the query, fetch, categorize.
- Per-dialect: `_connect_*` / `_query_*` / `_categorize_*`, small and single-purpose.
- `describe_databricks_ddls` is untouched — it's not one of the four duplicated implementations (no other dialect has an equivalent), so out of scope.

## Preserved deliberately, not unified

- **Connect call convention** (Postgres/MySQL `staticmethod`, Snowflake/Databricks instance method) — `_connect()` normalization is tracked separately in #723, not done here.
- **Cursor-acquisition style** — I considered unifying to `with conn.cursor() as cur:` everywhere, but the existing tests' mock helpers (`_conn_returning` vs `_sf_conn_returning`) wire the mock differently per style, proving it's real behavior the tests lock in, not incidental style. Unifying would have meant rewriting those tests, which the task said to avoid unless the refactor forces it — kept as a per-dialect `cursor_is_context_manager` flag instead.

## Safely unified

Row-transform now checks `isinstance(row, dict)` for all four dialects (previously MySQL-only, for pymysql's possible DictCursor rows) — a no-op for the other three (their rows are never dicts), so behavior is unchanged; still exactly preserves MySQL's dict-cursor tolerance.

## No behavior change — verified

- **AST diff**: `describe_columns` / `describe_databricks_ddls` signatures + docstrings, and the `JSON`/`ARRAY`/`SCALAR` constants, are byte-identical old vs. new.
- **Dispatch equivalence**: `config.type` string dispatch is behaviorally identical to the original `isinstance` chain — every `DestinationConfig` variant's Pydantic-validated `type` literal always matches its class; confirmed Postgres/MySQL/Snowflake/Databricks/BigQuery are siblings, not subclasses of each other.
- **No external coupling to internals**: grepped the repo for any reference to the old private `_describe_postgres`/`_describe_mysql`/`_describe_snowflake`/`_describe_databricks` names — none exist; only `describe_columns`/`describe_databricks_ddls` are ever imported elsewhere.
- **No new tests needed**: the existing `test_schema.py` already exercises every dialect × both cursor styles × the dict-cursor edge case × empty-result × connection-failure — full coverage of the new shared pipeline without modification.

## Gates

- `ruff check` clean.
- **161/161 tests pass, zero test edits**: `test_schema.py` (22) + every other file touching this module — `test_databricks_schema_aware.py`, `test_postgres_destination.py`, `test_snowflake_destination.py`, `test_mysql_destination.py`.

## Trade-offs

- Chose string-keyed dict dispatch (`config.type`) over an isinstance-iteration table — simpler and matches existing repo convention, but relies on the Pydantic discriminator invariant holding (documented in a comment).
- Considered a structural drift-guard test (`_DESCRIBE_SPECS` has exactly 4 keys) but judged it framework-building beyond this issue's ask — the existing per-dialect tests already fully protect the shared pipeline.

CHANGELOG entry added under `### Changed`, matching the `#719`/`#839` precedent for internal-only refactors.


Muawiya-contact (commented):
Done.

masukai (dismissed):
Approving — with one change worth recording, because the PR describes this as "no behavior change" and there's a real semantic shift underneath that the description doesn't name.

**The dispatch key changed from type to string.** `describe_columns` used to branch on `isinstance(config, PostgresDestinationConfig)`; it now looks up `getattr(config, "type", None)` in `_DESCRIBE_SPECS`. For the four current configs those are equivalent, and I verified it rather than assuming:

```
spec keys:  ['databricks', 'mysql', 'postgres', 'snowflake']
PostgresDestinationConfig    Literal['postgres']    ✓
MySQLDestinationConfig       Literal['mysql']       ✓
SnowflakeDestinationConfig   Literal['snowflake']   ✓
DatabricksDestinationConfig  Literal['databricks']  ✓
```

`type` is a required `Literal` on each, so every instance carries the right key. Equivalent today.

Where they'd diverge is subclassing: `isinstance` matches a subclass, an exact-string lookup doesn't. I checked — nothing in `drt/` subclasses those four concrete configs (only `BaseSqlDestinationConfig` → the concretes), so there's no live path where this differs. Fine as-is, and it matches `connectors/registry.py`'s convention, which is a good reason to prefer it. I'd just have called it out in the CHANGELOG entry as a dispatch change rather than filing it under "no behavior change" — the public API is byte-identical, but how it routes isn't.

Verified the rest: all 22 `test_schema.py` tests pass, full suite 2230 passed, ruff + mypy clean.

Genuinely good call on `cursor_is_context_manager`. It would have been tempting to "clean up" that inconsistency while table-driving the module, and the comment correctly identifies that the existing mocks lock in the difference — so normalizing it would have been a behavior change smuggled into a refactor. Leaving it as data and naming why is the right instinct.

Issue #720, opened 2026-07-07
refactor(destinations): unify Snowflake/Databricks under a shared warehouse base (after #699/#707)


masukai:
## Summary

`SnowflakeDestination` (`drt/destinations/snowflake.py`, 591) and `DatabricksDestination` (`drt/destinations/databricks.py`, 672) are near-duplicate classes. Verified by diff, these methods are identical apart from docstrings / one renamed local / the `table_fq` assembly (`database.schema.table` vs `catalog.schema.table`):

- `_bind_row` (sf `70-74` == db `121-127`)
- `_insert_rows` (sf `351-388` == db `435-472`) — identical except local name
- `_finalize_mirror` (sf `444-499` == db `525-571`)
- `drop_orphan_swap_tables` (sf `539-564` == db `611-636`); `list_orphan_swap_tables` / `_target_exists` differ only in the `SHOW TABLES` dialect string
- `load()` merge path + the mirror `upsert_key`/`tracked`/`scope` fail-fast guards (sf `133-151` == db `207-225`)

Only `_value_clause` genuinely diverges (Snowflake `PARSE_JSON` vs Databricks `from_json`/`parse_json` + DDL).

## Proposal

Fold both under a shared warehouse base (ideally the same `BaseSqlDestination` from the Postgres/MySQL extraction), leaving only `_value_clause` and the dialect strings per-class.

## ⚠️ Sequencing — blocked

This **collides** with in-flight work and should land **after** both:
- **#699** rewrites `_insert_rows` in *both* files (drops the `list(row.values())` fallback)
- **#707** rewrites Databricks bind sites (`%s`→`?`) and its schema queries

Alternatively, scope a first cut to the non-colliding methods only (`_finalize_mirror`, `list_/drop_orphan_swap_tables`, `_target_exists`, `finalize_sync`).

**Effort/risk:** M-L / MED-HIGH.

masukai:
Re-scoped against the current code before starting (both files have roughly doubled since this issue was filed in July — snowflake.py 1008 lines, databricks.py 1181, vs. the 591/672 quoted above). Two real design decisions surfaced that weren't visible in July; documenting them here before implementation starts.

## What's still accurate from the original scope
[...]

PR #841, opened 2026-07-26
refactor(destinations): normalize _connect() to classmethod (#723 part 3)

Muawiya-contact:
Part 3 of #723 — `_connect()` normalization masukai flagged. Leaves the remaining `run.py` cleanup item for the next PR, as instructed.

## Current state found (before touching anything)

| Destination | Convention | Called via |
|---|---|---|
| Postgres | `@staticmethod` | `PostgresDestination._connect(config)` |
| MySQL | `@staticmethod` | `MySQLDestination._connect(config)` |
| ClickHouse | `@staticmethod` | `ClickHouseDestination._connect(config)` |
| Snowflake | plain instance method | `SnowflakeDestination()._connect(config)` — instantiated just to call it |
| Databricks | plain instance method | `DatabricksDestination()._connect(config)` — same |

**Checked #720 first** (Snowflake/Databricks warehouse-base): not landed on `main` — nothing here was already resolved by it. **#719's `BaseSqlDestination`** (on `main`, Postgres/MySQL only) already has a `_dialect_connect(self, config)` hook each subclass implements as `return self._connect(config)` — the existing base-class seam the task asked me to look for, but it only unifies the instance-facing side for 2 of 5 destinations and doesn't touch `_connect` itself or ClickHouse/Snowflake/Databricks.

## Design

All five `_connect` methods are now `@classmethod`. This was the minimal-diff choice because a classmethod is callable in **both** pre-existing styles — `Class._connect(config)` (matches the old staticmethod calls) and `instance._connect(config)` (matches the old instance-method calls) — so **every internal `self._connect(config)` call site (~25, across all five files) needed zero changes**. I only updated the 6 *external* cross-file call sites (3 in `query.py`, 3 in `schema.py`) that used the now-obsolete `X()._connect(config)` instantiate-to-call pattern, simplifying them to `X._connect(config)` — that's the actual inconsistency being normalized. Two now-stale comments ("`_connect` is an instance method on the Snowflake/Databricks destination") were removed.

## Why it's behavior-preserving

- Read both Snowflake's and Databricks' `_connect` bodies in full before converting — neither ever referenced `self` (only `config`), so `cls` (also unused) changes nothing about what runs.
- `@patch("...ClassName._connect")`-style test mocks patch the class attribute directly — identical mechanism regardless of descriptor type (staticmethod/classmethod/instance method), so no test needed updating.

## Files changed

`drt/destinations/{postgres,mysql,clickhouse,snowflake,databricks}.py` (decorator + signature), `drt/destinations/{query,schema}.py` (6 call-site simplifications + 2 stale-comment removals). **18 insertions, 20 deletions across 7 files** — nothing else touched.

## Gates

- `ruff check` clean.
- **474 relevant tests pass, zero test edits**: every destination test file, `test_schema.py`, `test_destination_query.py`, `test_diff.py`, `test_lookup.py`, plus Postgres/MySQL/ClickHouse/Snowflake mirror-mode tests.

## Follow-ups intentionally left out

- `run.py` cleanup — the remaining #723 item, next PR.
- Giving ClickHouse a `_dialect_connect`-style base-class hook — that's #719/#720 territory (dialect-agnostic orchestration), not `_connect` normalization.
- `drt/sources/*.py`'s separate `_connect` methods — both the issue and this task are specifically scoped to *destinations*.

CHANGELOG entry added under `### Changed`, matching the `#719`/`#839`/`#840` precedent for internal-only refactors.


Muawiya-contact (commented):
Done.

masukai (dismissed):
Approving. Smallest of the four and the one with the highest ratio of clarity gained to lines changed — `SnowflakeDestination()._connect(config)` had accumulated an explanatory comment at every call site precisely because it was odd, and now neither the workaround nor the comments are needed.

The claim I wanted to verify is "no test needed updating", since `@classmethod` changes the descriptor and the suite mocks `_connect` in two different ways:

- `@patch("drt.destinations.clickhouse.ClickHouseDestination._connect")` — class-level, unaffected
- `mock.patch.object(dest, "_connect", ...)` in `test_orphan_cleanup.py` — **instance-level**, which is the one that could plausibly break

Ran both groups: 74 passed. `patch.object` on an instance sets an instance attribute that shadows the classmethod, so it works regardless of descriptor type. Claim holds.

Full suite 2230 passed, ruff + mypy clean.

Making all five `@classmethod` rather than all five `@staticmethod` is the right pick — it's callable both ways, so the internal `self._connect(config)` sites kept working untouched, which is why this is 19 lines instead of a sweep. Worth noting you also updated the two `# _connect is @staticmethod(config)` comments in the #719 dialect hooks; stale comments surviving a rename is the usual way this kind of change leaves debris.

Issue #837, opened 2026-07-25
fix(cli): drt test JSON warnings[] conflates threshold breach with exception in one 'value' key


masukai:
Follow-up from the #830 review (#779). Non-blocking for that PR.

## Problem

`_collect_warnings()` in `drt/cli/commands/test.py` flattens every `severity: warn` failure into a top-level `warnings` array — good, and the right shape for `jq`. But it collapses two semantically different things into one key:

```python
"value": t.get("value", t.get("error")),
```

So a consumer reading `warnings[].value` cannot distinguish:

- **a threshold breach** — the test ran fine and returned `3` (three null rows). `value: "3"`
- **an exception** — the test never ran; the destination was unreachable. `value: "connection refused"`

Those demand different reactions. The first is a data-quality signal to chart or alert on; the second is an infrastructure failure where the data quality is simply *unknown*. Conflating them means a dashboard counting "warnings over time" silently mixes in outages, and a numeric parse of `value` fails on the exception case with no way to have predicted it.

## Why now

The `warnings` array is **new public contract** introduced by #830. Splitting the key later is a breaking change for anyone who has started consuming it; splitting it before it ships to users is free. That's the whole reason to file this rather than let it ride.

## Suggested fix

Emit both keys explicitly, mirroring the per-test entry shape that already distinguishes them:

```json
{"sync": "s", "test": "not_null(email)", "value": "3"}
{"sync": "s", "test": "not_null(email)", "error": "connection refused"}
```

i.e. carry `value` when the test produced one, `error` when it raised — never overload one field. Consumers can then branch on key presence, the same way they already can inside `results[].tests[]`.

## Acceptance criteria

- [ ] `warnings[]` distinguishes a threshold breach from an exception
- [ ] Shape is consistent with the per-test entries in `results[].tests[]`
- [ ] Test covering both cases in one run
- [ ] Lands before/with the release that first ships `warnings` publicly

Issue #834, opened 2026-07-25
test(destinations): driver coverage for fetch_failing_rows — MySQL DictCursor + ClickHouse row shapes untested


masukai:
Follow-up from the #830 review (#779). Not blocking that PR — filing so it's tracked rather than remembered.

## Problem

`drt/destinations/query.py`'s `fetch_failing_rows` dispatches to four per-driver helpers, but only the Postgres path is exercised by tests. 36 of the PR's 44 uncovered lines are in this dispatch.

Two of the uncovered branches are the ones most likely to actually break against a real warehouse:

- **`_fetch_failing_rows_mysql`** — entirely uncovered, including the `isinstance(row, dict)` branch that handles pymysql's `DictCursor`. That branch exists because the cursor type changes the row shape; if it's wrong, every stored sample from MySQL is malformed.
- **`_fetch_failing_rows_clickhouse`** — uses a different API shape from the DB-API drivers (`result.column_names` / `result.result_rows` rather than `cur.description` / `fetchall()`), so it shares no code path with the tested Postgres helper.

Snowflake is also uncovered, though it follows the same DB-API shape as Postgres.

## Context: this matches existing practice, it doesn't regress it

Worth stating plainly so this isn't read as a knock on #830: `_query_mysql`, `_fetch_rows_mysql`, and `_query_clickhouse` have no driver-level tests either. #830 followed the pattern the module already had. This issue is about the module, not that PR.

## Suggested approach

The DWH smoke epic (#654) validated three warehouses live, so the infrastructure question is largely answered. Options, cheapest first:

1. **Fake-cursor unit tests** — assert the row-shaping logic (dict vs tuple zip, ClickHouse's column extraction) without a live connection. Cheap, catches the shape bugs, no CI credentials needed.
2. **Extend the DWH smoke job** to run one `drt test --store-failures` per warehouse and assert the written `.jsonl` parses with expected columns.

(1) is probably the right first step since the failure modes here are shape bugs, not connectivity bugs.

## Acceptance criteria

- [ ] `_fetch_failing_rows_mysql`'s `DictCursor` branch is exercised
- [ ] `_fetch_failing_rows_clickhouse`'s `column_names` / `result_rows` extraction is exercised
- [ ] `_fetch_failing_rows_snowflake` has at least smoke-level coverage

Muawiya-contact:
@masukai, thank you for opening the PR. However, it was mine, as I assigned it to myself.

masukai:
[...]

PR #842, opened 2026-07-26
refactor(cli): extract remaining run.py infra — signal mapping + observer composition (#723 part 4, final)

Muawiya-contact:
Part 4 of #723 — the last remaining item, `cli/commands/run.py` cleanup. JSON logging (`_JsonFormatter`/`_configure_json_logging`) was already extracted to `drt/cli/_logging.py` in #729, confirmed via the module docstring and `main.py` already importing them directly from there. This closes out #723 alongside #839 (MCP tools split), #840 (`schema.py` table-drive), and #841 (`_connect()` normalization).

## Current state found (before touching anything)

- `run.py` is now **796 lines** (the issue's original text said 573 — it's grown with #774/#775/#783/#784 features since).
- No newer PR had already touched this — checked `git log upstream/main`, nothing beyond #729's logging extraction.
- What remained: (1) `_exit_code_for_signal`, a tiny generic `128 + signum` mapping, still defined locally; (2) `_run_one`'s observer-composition block, which masukai specifically flagged.

## What was extracted, and where it now lives

1. **`exit_code_for_signal`** → `drt/cli/_helpers.py`, the existing shared-CLI-utility module (already home to `resolve_profile_name`/`get_source`/`get_destination`/`get_watermark_storage` — exactly the "shared CLI utility module that already exists" the task asked me to check for). `run.py` re-imports it under its original name (`exit_code_for_signal as _exit_code_for_signal`) — the same pattern already used for `_configure_json_logging`.
2. **Observer composition** → `_build_observer(sync, ctx, wm_storage)`, a new *same-file* helper in `run.py` — it's specific to `_RunContext`/this file's DLQ-config shape and isn't reused anywhere else, so a same-file extract-method is the right size, not a new module.

## Why `_run_one()` is simpler

The 15-line block building `observers: list[Any] = [...]`, the DLQ opt-in branch, and `CompositeObserver(observers)` collapsed into one line: `observer = _build_observer(sync, ctx, wm_storage)`. `_run_one`'s opening now reads as four clear steps — destination, watermark storage, observer, start-line — instead of interleaving observer-construction logic into that flow.

## No behavior change — confirmed

- Both extractions are verbatim moves: same conditionals, same order, no logic changes.
- `dest = get_destination(sync)` stayed in `_run_one` (used later for the `run_sync()` call, unrelated to observer wiring); the `_on_signal` closure itself (json_mode/quiet-aware console output, `stop_event` wiring) stayed in `run()` — genuinely run-specific, not generic.
- **Caught and fixed a real mypy strict issue along the way**: `main.py`'s back-compat re-export of `_exit_code_for_signal` (previously imported *through* `run.py`) failed `--no-implicit-reexport` once the name was only aliased, not defined, there. Fixed by having `main.py` import it directly from `drt.cli._helpers` — mirroring the exact precedent already established for `_configure_json_logging`/`_JsonFormatter`, which `main.py` already imports directly from `_logging.py` rather than through `run.py`.

## Gates

- `ruff check` clean.
- **550 relevant tests pass, zero test edits**: every `test_cli_*.py` file, `test_engine_observer.py`, `test_cli_graceful_shutdown.py` (exercises `_exit_code_for_signal` directly via `from drt.cli.main import`), `test_print_watermark_summary.py` (imports directly from `drt.cli.commands.run`), DLQ/alerts/telemetry tests.

## Files changed

`drt/cli/_helpers.py` (+5), `drt/cli/commands/run.py` (extraction + docstring update), `drt/cli/main.py` (1-line re-export fix). **42 insertions, 28 deletions across 3 files.**

## Improvements intentionally left out (exceed #723's scope)

- The `_on_signal` closure itself — genuinely `run`-specific, not generic, per "avoid framework-building."
- No new class hierarchy, no new module — per the explicit "prefer small helpers" instruction.
- MCP code, `schema.py`, destination `_connect()` — all out of scope, untouched (confirmed via `git diff --stat`: only the 3 files above changed).

CHANGELOG entry added under `### Changed`, matching the `#719`/`#839`/`#840`/`#841` precedent for internal-only refactors.


Muawiya-contact:
@masukai Once all four lands successfully reach the main branch, I will start work on #834 and continue from there.

Muawiya-contact (commented):
Done.

masukai (dismissed):
Approving — this closes #723.

Two things I want to credit specifically, because both are about *not* trusting the issue text:

**You checked the ground truth before moving anything.** The issue said `run.py` was 573 lines; it's 796 now, and part of what the issue asked for (`_JsonFormatter` / `_configure_json_logging`) had already been extracted in #729. Re-deriving current state instead of implementing a months-old description is the difference between finishing an issue and re-doing it. The PR body documenting *what you found* before *what you changed* made this reviewable in a couple of minutes.

**`exit_code_for_signal` went to `_helpers.py` rather than a new module.** A generic `128 + signum` helper is exactly the kind of thing that grows a `_signals.py` nobody else ever imports. Putting it next to the existing shared CLI utilities, and re-importing under the original name to keep `drt.cli.main`'s back-compat re-export intact, is the boring correct answer.

`_build_observer` staying in-file is also right — it's not shared infrastructure, it's `_run_one`'s composition step given a name and a docstring. Moving it out would have been motion without benefit.

Full suite 2230 passed, ruff + mypy clean.

---

All four #723 parts approved. They're independent (`base=main`, no stacking), so merge order is free — though **#830 should go first**, since it touches `mcp/server.py` too and #839 moves that file substantially. Rebasing #830 after #839 lands will be meaningfully more work than before.

On picking up #834 next: sounds good, and thanks for volunteering. One suggestion — of the five follow-ups, **#837** is the time-sensitive one (`warnings[]` is new public contract in #830; splitting `value`/`error` is free before it ships, breaking after). #834 has no such deadline. If you'd rather do the interesting one first that's genuinely fine, just don't let #837 slip past the release.

masukai:
All four #723 parts are on `main` now — #839, #840, #841, and this one. You're unblocked for #834.

Note for the record: parts 2–4 needed a rebase before they'd merge, caused by my own sequencing (I merged part 1 while the other three were sitting approved, which put a `CHANGELOG.md` collision on all of them and a `schema.py` collision on part 3). Rather than hand four rebases back to you, I did them — your commits, your authorship, unchanged logic.

The one place I had to make a judgment call was **part 3 (#841) × part 2 (#840)** in `schema.py`. Part 2 had already table-driven `_describe_*` so those functions now just `return sql, params` and the shared `_describe` pipeline runs them — which meant the exact lines part 3 was rewriting no longer existed. I took part 2's structure and applied part 3's *intent* to the two spots where it now lives: `_connect_snowflake` and `_connect_databricks` both dropped their instantiation (`SnowflakeDestination()._connect(config)` → `SnowflakeDestination._connect(config)`), and the "instance method" comments came out with them since they're no longer true. Verified with a repo-wide grep that no `Destination()._connect` call sites remain. Full suite green at each step.

If you'd rather I hadn't touched them, say so and I'll route conflicts back to you next time — but given the collisions were mine to begin with, this felt like the right call.

Issue #769, opened 2026-07-10
feat(destinations): rate limiting v2 — honor Retry-After, per-destination rate_limit override, shared bucket across threads


masukai:
## Problem

Rate limiting today is a single token bucket driven by `sync.rate_limit.requests_per_second` (default 10) and retry is blind exponential backoff over `retryable_status_codes`. Three gaps for SaaS destinations — the core reverse-ETL workload:

1. **`Retry-After` is ignored**: on 429, `with_retry` sleeps its computed backoff even when the API explicitly says how long to wait (Slack, HubSpot, Intercom, Zendesk all send it) — we retry too early (burning attempts) or too late (wasting minutes)
2. **No per-destination override**: `requests_per_second` lives on `sync.rate_limit` only; a sync fanning into a strict API (Klaviyo burst limits) and one bucket setting must be tuned per sync-file, and multi-sync runs into the *same* destination share nothing — `--threads 4` can quadruple effective pressure on one API
3. **No burst/concurrency semantics**: token bucket has no burst capacity or max-in-flight knob

## Prior art

- **Census/Hightouch**: per-destination API budget management with header-aware throttling — table stakes for their reliability story.
- **dlt**: honors `Retry-After` in its HTTP client helpers.

## Proposal

```yaml
destination:
  type: hubspot
  rate_limit:                    # overrides sync.rate_limit for this destination
    requests_per_second: 8
    burst: 16
```

1. Parse `Retry-After` (seconds + HTTP-date forms) in `with_retry`; sleep `max(retry_after, computed_backoff)` capped by `max_backoff`
2. Accept `rate_limit` on destination configs (precedence: destination > sync > default — same pattern as `resolve_retry`)
3. Process-wide limiter registry keyed by destination identity so `--threads` shares one bucket per destination endpoint

## Scope

**In:** Retry-After parsing, destination-level `rate_limit` + precedence, shared limiter across threads, burst capacity, docs per SaaS connector with the vendor's published limits.
**Out:** adaptive rate discovery (respond to headers like `X-RateLimit-Remaining` — nice follow-up), per-endpoint budgets within one destination.

## Related

- #48 (Slack 429 → retry, the original incident), #277 (retry precedence pattern to mirror), #50/#101 (RateLimiter hardening history)

## Milestone fit

Candidate: v0.8.x — pure reliability polish on existing seams (`retry.py`, `rate_limiter.py`), no new protocols.

masukai:
[...]

Issue #756, opened 2026-07-10
feat(state): remote state backend — object storage (GCS/S3), CI-safe, team-shared


masukai:
> **Scope note (2026-08-06):** This issue originally covered both an
> object-storage backend and a warehouse-backed one under one config knob.
> [ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md)
> found that bundling wrong — different justification (durability vs.
> observability), different cost (no warehouse write vs. requires one) — and
> split them. **This issue is now the object-storage half only, and ships
> first.** The warehouse half is [#920](https://github.com/drt-hub/drt/issues/920).

## Problem

Watermarks already support remote storage (`sync.watermark.storage: gcs | bigquery`), but the rest of drt's state is local-only files under `.drt/`: run state (`state.json`), history (`history/*.jsonl`), and the DLQ (`dlq/*.jsonl`). In the deployment model we actively promote — GitHub Actions cron via drt-action, ephemeral containers, Dagster/Airflow workers — that state evaporates after every run unless users hand-roll cache/artifact persistence. Concretely:

- `drt status` / `drt status --history` show nothing in a fresh CI checkout
- the DLQ (and therefore `drt retry`) is unusable across CI runs — failed records are lost with the runner
- two operators (or laptop + CI) can't share a view of "what ran, what failed"

`HistoryConfig` already hints at this ("Future: s3, sqlite" comment in `drt/state/`).

## Prior art

- **dlt**: pipeline state is persisted *in the destination* as a table — survives any runner.
- **Hightouch Warehouse Sync Logs**: writes per-run/per-row sync outcomes back to warehouse tables so teams can SQL-query their reverse-ETL observability. (Requires the Lightning/write-grant engine — see ADR 0005's Evidence section for why that doesn't transfer directly to this half.)

## Proposal

Extend the existing `WatermarkConfig` storage pattern into a project-level state backend covering state + history + DLQ, on object storage:

```yaml
# drt_project.yml
state:
  backend: gcs                  # local (default) | gcs | s3 (this issue) | warehouse (#920)
  bucket: my-bucket
  prefix: drt-state/             # optional; default project-scoped path
```

- No warehouse write grant required — same permission posture as `sync.watermark.storage`
- `StateStore`/`HistoryStore`/`DlqBackend` Protocols already exist (#900) — this issue implements them against GCS/S3, wired through the factory (step 1, this issue)
[...]

Issue #755, opened 2026-07-10
feat(engine): diff-based incremental — warehouse-side snapshot diff, no cursor column required


masukai:
## Problem

drt's only incremental strategy is cursor-based (`mode: incremental` + `cursor_field`). This requires the source model to expose a reliable `updated_at`-style column — which many curated marts simply don't have — and it can only detect *new/updated* rows, never removals or arbitrary changes. Our own comparison page ([docs/guides/comparison.md](https://github.com/drt-hub/drt/blob/main/docs/guides/comparison.md)) acknowledges this: drt = "Cursor-based", Census/Hightouch = "Multiple strategies".

For reverse ETL this is the core engine gap: every run of a non-incremental sync re-sends the full result set to the destination API, burning rate limits and quota on rows that haven't changed.

## Prior art

- **Hightouch Lightning sync engine**: computes difference-based CDC *inside the customer warehouse* — each run snapshots the model query result into a managed schema (`hightouch_planner`), diffs it against the previous snapshot, and syncs only adds/changes/removals. Only the two most recent snapshots are kept.
- **dbt snapshots** are the closer structural reference, and a verified one: `strategy: check` + `check_cols: all` is the same feature as this proposal's `incremental_strategy: diff` + `hash_columns: all`, and dbt requires warehouse write for the same reason this proposal does — because it is a transformation tool operating on data it owns. ([ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md#evidence))
- **dlt**: `merge` write disposition dedupes/upserts against destination state.
- ~~Census basic diff engine: same concept; state tables live in the source warehouse.~~ **Unverified, dropped as prior art** — `docs.getcensus.com` now 301s to Fivetran's Activations documentation, which states neither the permission requirement nor the state location, following Census's absorption into Fivetran. ([ADR 0005 correction](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md#evidence))

## Proposal

A new incremental strategy that requires no cursor column:

```yaml
sync:
  mode: upsert
  incremental_strategy: diff        # default: cursor (current behavior)
  diff:
    state: source                   # snapshot tables in the source warehouse
    schema: _drt_planner            # drt-managed schema/dataset
    hash_columns: all               # or explicit list
```

[...]

PR #845, opened 2026-07-27
docs(adr): ADR 0004 — streaming / event-triggered syncs (refs #786)

Pawansingh3889:
Refs #786 — the runtime-shape / prerequisites / build-no-build half of the research spike, per the split agreed in the issue thread (@Muawiya-contact owns the trigger matrix).

## What this ADR decides

- **No native watcher.** Every cheap change signal across the shipped sources is a poll, and every push signal is a message bus the user already runs — a sensor handles that shape well and a daemon duplicates it badly. A [falsification condition](https://github.com/drt-hub/drt/blob/b858e6723c7a3165165cd404caad014f1f639da1/docs/adr/0004-streaming-and-event-triggered-syncs.md#falsification-condition) states exactly what in the finished trigger matrix would overturn this.
- **Three tiers instead:** warehouse-native scheduling → `drt-action` (ungated, documentable today); Dagster sensors as the real event-driven path (a `dagster-drt` *package addition* — the integration has no sensors today); hardened `drt serve` for push sources.
- **Gates, not footnotes:** #756 remote state blocks promoting Tier 2 (`.drt/state.json` is local disk — a sensor and a CI run cannot share watermarks). The open half of #769 (per-destination override + shared cross-thread bucket; `Retry-After` shipped in v0.8.1) blocks Tier 2/3. Gates block *publishing* the guidance, not writing it.
- **Open-core line:** freshness is core (OPEN_CORE.md puts the sync engine incl. rate limits/retry/cursors under "always free"); hosting the always-on component is the existing "Cloud hosting" enterprise item. Nothing moves the line.

## Provisional pieces

The trigger-matrix table is my reading of the connector code, included so the ADR is self-contained — it is explicitly superseded by @Muawiya-contact's dedicated matrix on landing, and the recommendation is written to be falsifiable by it.

Every code claim is cited to file:line (server lock/423-drop/sync-in-handler/bearer-only auth, local state, dagster-drt's seven exports).

## Follow-ups this ADR sanctions (to be opened as separate issues after review)

- [ ] `drt serve` concurrency contract — framed as a design decision (bounded queue vs `429` + `Retry-After` vs per-sync locks), plus `202` + run id and HMAC verification
- [ ] `dagster-drt` sensors — generic cheap-signal sensor + Delta/Iceberg version + Snowflake `STREAM` variants; blocked-by #756
- [ ] "Event-driven syncs" docs guide covering all three tiers; blocked-by #756 + #769

Docs-only change — no code touched, no CHANGELOG entry per the guard's scope.

### DIFF
--- a/docs/adr/0004-streaming-and-event-triggered-syncs.md
+++ b/docs/adr/0004-streaming-and-event-triggered-syncs.md
@@ -1,0 +1,217 @@
+# ADR 0004 — Streaming / event-triggered syncs
+
+- **Status:** Proposed — the recommendation below is provisional until the
+  per-warehouse trigger matrix ([#786](https://github.com/drt-hub/drt/issues/786),
+  @Muawiya-contact) lands, and is written to be falsified by it. See
+  [Falsification condition](#falsification-condition).
+- **Issue:** [#786](https://github.com/drt-hub/drt/issues/786)
+- **Implementation:** none — this ADR recommends **not** building a native
+  watcher. The work it does sanction is listed under
+  [Follow-up issues](#follow-up-issues).
+
+## Context
+
+drt runs when invoked. Census Live Syncs and Hightouch Streaming/CDC ship
+"row lands in the warehouse → seconds later it is in the destination", and the
+warehouses now feed them: Snowpipe Streaming, BigQuery's Storage Write API and
+change history, Databricks DLT. The gap this opens is not throughput but
+*staleness*: the operational use cases reverse ETL exists for — abandoned cart
+to ads, fraud flag to the support tool, churn score to the CRM — are the ones
+where a nightly or hourly sync is visibly wrong.
+
+The question this ADR answers is what an OSS, CLI-first tool should do about
+that. It is deliberately a build/no-build decision, not a design for a daemon.
+
+Three facts about the code as it stands (v0.8.2) constrain the answer more than
+the competitive framing does.
+
+**`drt serve` is a trigger endpoint, not a trigger runtime.** It exists
+(#218) and works for its designed cadence — a dbt job finishes, POST
+`/sync/<name>`. Its module docstring says the quiet part outright: *"One sync
+at a time — concurrent requests get 423 Locked"* (`drt/cli/server.py:15`). At
+sub-minute cadence three properties stop being acceptable:
+
+- A single global `_SyncLock` (`drt/cli/server.py:28`) serialises *every* sync,
+  not merely concurrent runs of the same one, and a second trigger is answered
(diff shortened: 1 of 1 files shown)


masukai (approved):
Approving and merging. This is better than what I asked for, and the parts that go beyond the brief are the parts that make it useful.

I verified every code citation rather than reading past them — all accurate:

| Claim | Checked |
|---|---|
| `server.py:15` docstring, `:28` `_SyncLock`, `:61` bearer auth, `:90` the 423, `:100` synchronous execution | ✅ each line is what you say it is |
| `state/manager.py:43` → `.drt/state.json` | ✅ |
| `dagster_drt` exports **seven** symbols, none a sensor | ✅ `__all__` has exactly 7 |
| `deltalake.py:67` already calls `.version()` | ✅ |
| `iceberg.py:52` reachable via the loaded table | ✅ |
| `rest_api.py:31` — polling *is* the extract | ✅ |
| OPEN_CORE's "rate limits, retry logic, cursor management" and the Cloud-hosting boundary item | ✅ both quoted verbatim |

Three things I want to call out specifically, because they're the difference between an ADR that ages well and one that gets quietly ignored:

**The falsification condition is doing real work.** Two concrete, checkable conditions — a cheap push signal with no broker in between, or a signal whose cost only amortises over a long-lived connection — either of which overturns the recommendation. That's what makes "provisional until the matrix lands" a real commitment rather than a hedge, and "if the matrix contradicts this, the ADR should change rather than be defended" is exactly the posture I wanted.

**The gates are scoped correctly.** *Blocks promotion, not authorship* is the right cut, and the reason you give for it — that guidance published before #756 documents a topology "whose two halves silently disagree about what has already synced" — is the sharpest statement of the risk anyone has made, including me. The distinction between #755 (enabling) and #756/#769 (blocking) is right too.

**The attribution note on the trigger matrix.** Marking your table as provisional raw material, explicitly superseded by @Muawiya-contact's deliverable and losing to it on disagreement, is how two people work the same problem without stepping on each other. The REST API row as a *stated non-option* is a good call — a documented "nothing here" saves the next person the same investigation.

One small thing, no change needed now: `## Context` opens with "the code as it stands (v0.8.2)" and we tagged v0.8.3 about an hour ago. Nothing in your analysis changes — none of #693/#470/#784/#779 touch `serve`, state locality, or the dagster integration — so I'd rather merge as-is than churn the diff. Worth bumping when the matrix lands and the ADR moves from Proposed to Accepted.

Please do open the three follow-ups. #1 especially — framing the `drt serve` concurrency contract as a design decision (bounded queue vs `429` + `Retry-After` vs per-sync locks) rather than a bug fix is right, and `202` + run id plus HMAC belong in the same conversation.

PR #847, opened 2026-07-27
docs(research): per-warehouse trigger matrix for event-driven syncs (refs #786)

Muawiya-contact:
Completes the remaining half of #786.

## What this is

#786 was intentionally split. The ADR half shipped in #845 and stated its own trigger matrix was **provisional**, that it would be superseded by this dedicated deliverable, and that where the two disagree, the matrix wins. Masukai's framing — "your trigger matrix is the load-bearing input; the recommendation above is provisional until it exists" — is what this PR supplies.

**This PR delivers only the trigger matrix.** ADR 0004's architectural recommendation is not rewritten and not changed.

## The deliverable

`docs/research/warehouse-trigger-matrix.md` — all **thirteen** supported sources researched individually against official vendor documentation and the shipped connector code:

PostgreSQL · MySQL · Snowflake · BigQuery · Databricks · Delta Lake · Apache Iceberg · ClickHouse · DuckDB · SQLite · Redshift · SQL Server · REST API

Each row answers: cheapest change signal, push vs poll, notification mechanism, polling mechanism, latency, infrastructure needed, cost implications, whether event-driven is practical for a CLI-first tool, **one preferred recommendation** with ranked alternatives, and limitations.

Per the "need a recommendation, not just a list" requirement, every source names a single preferred mechanism — e.g. Snowflake is *Preferred: `STREAM` + `SYSTEM$STREAM_HAS_DATA()` in a `TASK` `WHEN` clause*, with Alerts+webhook and Snowpipe Streaming as explicitly-rejected alternatives (Snowpipe Streaming being the wrong layer entirely — an ingestion API, not a change signal).

Every claim carries a citation: a vendor documentation URL or a repo `file:line`. No values invented. Latency figures are documented mechanism floors, not benchmarks, and the two weakest (Databricks' ~1–2 min, BigQuery's audit-log delivery) are flagged in place as community-reported / no published SLA.

## No architectural recommendations changed

ADR 0004 commits to two falsification conditions. Both are checked explicitly in the new document, and **neither is met**:

1. *"A cheap push signal that is not already a message bus the user runs"* — not found. Of four push mechanisms, BigQuery's is Pub/Sub (a bus), and Postgres `NOTIFY` / SQL Server Query Notifications both need a long-lived consumer draining a queue. The nearest counter-example, **Databricks' table update trigger**, turns out to be the platform's own managed ~1-minute polling loop that then invokes the CLI — Tier 1 working as designed, not a broker-free push.
2. *"A signal whose cost only makes sense amortised across a long-lived connection"* — not found. Every preferred signal is cheap per-poll. The one family where per-connection setup dominates (logical decoding, binlog) is CDC, already excluded by the ADR on separate grounds.

So the no-watcher recommendation stands. If anything it's slightly stronger: the cheapest signal for 11 of 13 sources is a poll, and the two purpose-built signals (Snowflake `STREAM`, SQL Server Change Tracking) are *designed* to be polled cheaply — exactly the shape a Dagster sensor provides.

## Where the research refined the provisional matrix

Mechanism-level corrections only; a full table is in the document.

| Source | Change |
|---|---|
| **PostgreSQL** | Cheapest signal is a **poll** (`max(updated_at)`), not the two push options listed — `NOTIFY` isn't durable, so a poll is needed behind it regardless |
| **SQL Server** | **Change Tracking earns its own row** rather than the "generic and slow" bucket — delete-aware, version-cursored, all editions incl. Express; closer in kind to Snowflake `STREAM` |
| **ClickHouse** | **`system.parts` probe** beats `max(updated_at)`; also rules out `system.tables.metadata_modification_time`, which looks correct but **does not fire on `INSERT`** |
| **SQLite** | **`PRAGMA data_version`** beats file mtime — in WAL mode the main file's mtime can stay flat while writes accumulate in `-wal`. DuckDB has no equivalent, so the two engines split |
| **Redshift** | EventBridge **explicitly ruled out** — lifecycle/job events only, no table-level signal (recorded so the next reader doesn't re-investigate) |
| **BigQuery** | `APPENDS` TVF named preferred; the push path identified as Data Access audit logs via Log Router → Pub/Sub (there's no dedicated "table change notification" feature) |
| **Snowflake, Delta Lake, Iceberg, MySQL, DuckDB** | **Confirmed as written**, with added caveats (Snowflake's warehouse-cost trap and false positives; `deltalake.py:67` / `iceberg.py:51-52` citations verified) |

## ADR now has the factual foundation it referenced

Two minimal edits, no conclusions touched:

1. The provisional table is **replaced by a reference** to the new document — chosen over inlining to avoid duplicating a large table, per the repo's no-duplication expectation.
2. The status line records that the matrix landed and did not falsify the recommendation.

No tier, gate, consequence, or follow-up text changed. `git diff --stat` on the ADR: **22 insertions, 26 deletions** — net smaller.

## Quality checks

- All 13 required sources researched individually.
- Every internal link and anchor verified programmatically (0 broken); both markdown tables verified column-consistent.
- Every repo `file:line` citation verified against the actual files — two line ranges corrected during review (`sqlserver.py` is 72 lines, dagster `__init__.py` is 14).
- Recommendations consistent with ADR 0004's tiers; Tier 2/Tier 3 guidance is **not** promoted, respecting the #756/#769 gates.
- Open-core boundary re-checked against `OPEN_CORE.md` — every preferred signal is a query or metadata read by connector code that already ships, so freshness stays a property of the sync engine (Always Free); only *hosting* an always-on component remains enterprise. Nothing requires a daemon or GUI, per CLAUDE.md's CLI-first rule.
- Docs-only: no code, no CHANGELOG (repo convention reserves it for user-facing changes; ADR 0004 itself shipped without one). ruff/mypy/pytest not applicable.

### DIFF
--- a/docs/adr/0004-streaming-and-event-triggered-syncs.md
+++ b/docs/adr/0004-streaming-and-event-triggered-syncs.md
@@ -130,10 +130,10 @@ The structural pattern that reading suggested holds: **every cheap signal is a
 poll, and every push signal is already a message bus the user runs.** That is the
 strongest argument for sensors and against a native watcher — a sensor is a
 scheduled cheap poll with durable cursors, which is exactly the shape of the
-signals that actually exist. The finished matrix confirms it for eleven of
-thirteen sources, and the nearest counter-example (Databricks' table update
-trigger) turns out to be the platform's own managed polling loop invoking the
-CLI — Tier 1 working as designed, not a broker-free push. See
+signals that actually exist. The finished matrix confirms it for twelve of the
+thirteen sources, and the sole exception (Databricks' table update trigger)
+turns out to be the platform's own managed polling loop invoking the CLI —
+Tier 1 working as designed, not a broker-free push. See
 [What this means for ADR 0004](../research/warehouse-trigger-matrix.md#what-this-means-for-adr-0004)
 for the falsification check against both conditions below.
 
--- a/docs/adr/0004-streaming-and-event-triggered-syncs.md
+++ b/docs/adr/0004-streaming-and-event-triggered-syncs.md
@@ -1,9 +1,11 @@
 # ADR 0004 — Streaming / event-triggered syncs
 
-- **Status:** Proposed — the recommendation below is provisional until the
+- **Status:** Proposed. The recommendation was written to be falsified by the
   per-warehouse trigger matrix ([#786](https://github.com/drt-hub/drt/issues/786),
-  @Muawiya-contact) lands, and is written to be falsified by it. See
-  [Falsification condition](#falsification-condition).
+  @Muawiya-contact); that matrix has since landed as
+  [docs/research/warehouse-trigger-matrix.md](../research/warehouse-trigger-matrix.md)
+  and meets neither [falsification condition](#falsification-condition), so the
+  recommendation stands unchanged.
 - **Issue:** [#786](https://github.com/drt-hub/drt/issues/786)
 - **Implementation:** none — this ADR recommends **not** building a native
   watcher. The work it does sanction is listed under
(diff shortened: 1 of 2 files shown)


Muawiya-contact (commented):
Done

masukai (approved):
Approving and merging. This is the deliverable the ADR was written to be judged against, and it did the judging properly — including correcting me where I was wrong.

## The falsification check is the part that matters

I asked for the matrix to be able to overturn the ADR, and you actually ran that check rather than concluding what was convenient. Both conditions evaluated explicitly, both found unmet, with the reasoning shown. **Databricks named as the closest counter-example** rather than quietly omitted is the right instinct — and the resolution ("its own managed polling loop that then starts a job — the platform's scheduler, not a broker-free event stream") is more precise than anything in my provisional reading.

The refinement to the structural claim is a genuine improvement: *nine* push mechanisms exist, and the claim holds not because push is rare but because every one falls into a category that keeps drt out of the consumer business. That four-way categorisation is better than the flat "every push signal is a bus" I wrote.

## Corrections to my provisional table — all verified, all mine to own

I checked each rather than accepting them:

| Correction | Verified |
|---|---|
| REST API cited `rest_api.py:31` — that's the `extract()` signature, not the HTTP call | ✅ `:31` is `def extract(...)`; the request is at `:132-137`. My citation was wrong. |
| Iceberg `:52` → `:51-52` (the `load_catalog` line matters too) | ✅ `:51` is `load_catalog(...)`, `:52` the scan |
| Delta `deltalake.py:67` `.version()` | ✅ unchanged, confirmed |
| BigQuery has no "table change notification" feature — the push path is Data Access audit logs via Log Router | Accepted; I'd stated a feature name that doesn't exist |
| SQL Server Change Tracking deserves its own row, not a grouping with "generic and slow" | Accepted — delete-aware and version-cursored is much closer to Snowflake `STREAM` than to a scan, and I undersold it |
| SQLite `PRAGMA data_version` over file mtime | Good catch, and the reason is the convincing part: in WAL mode the main file's mtime can stay flat while writes accumulate in `-wal`. Grouping SQLite with DuckDB hid a real difference. |
| ClickHouse `system.parts` over `max(updated_at)` | Accepted, and ruling out `system.tables.metadata_modification_time` (looks right, doesn't fire on `INSERT`) is exactly the kind of dead end worth recording |

The Snowflake caveats you added — stream consumption requires DML, `SYSTEM$STREAM_HAS_DATA` has documented false positives, triggered tasks are tighter than scheduled — are the sort of thing that only shows up in real vendor docs, not in a code reading. Same for Redshift's EventBridge events being lifecycle-level rather than table-level.

## ADR edit is scoped correctly

Replacing the provisional table with a pointer, updating the status line to say the matrix landed and neither condition is met, and **leaving the recommendation untouched** is exactly right. The ADR stays the decision record; this stays the research input.

## One note, no change needed

`docs/research/` is new as a directory. That seems right to me — it isn't an ADR (no decision) and isn't a user guide, and the header saying so explicitly ("Status: Research… it does not make an architectural decision") keeps the distinction legible. Worth knowing it now exists as a category.

Thanks for taking the time to research thirteen sources individually instead of thirteen rows of the same sentence. The "Preferred + ranked alternatives" shape makes it usable when the gated tiers open up rather than something we'd re-derive.

Issue #856, opened 2026-07-27
docs(guides): event-driven syncs — the three tiers from ADR 0004


masukai:
Sanctioned by [ADR 0004](https://github.com/drt-hub/drt/blob/main/docs/adr/0004-streaming-and-event-triggered-syncs.md) (refs #786). The user-facing half of the decision.

## What

`docs/guides/event-driven-syncs.md` covering the three tiers the ADR settles on:

1. **Warehouse-native scheduling → `drt-action` / the CLI.** The trigger lives where the data lands; drt stays a process that starts, syncs, exits. Covers most "fresh enough" (1–15 min) requirements with zero drt-side runtime.
2. **Dagster sensors** — the recommended path for genuine event-driven activation (#855).
3. **Hardened `drt serve`** for push sources (#854).

The [trigger matrix](https://github.com/drt-hub/drt/blob/main/docs/research/warehouse-trigger-matrix.md) supplies the per-source mechanism, latency, infrastructure and cost for each — the guide should point at it rather than restate it.

Also worth stating plainly: drt does **not** ship a native watcher, and why (the ADR's reasoning, compressed). That's a question people arrive with; a documented posture beats silence in a comparison table.

## ⚠️ Blocked by #756 and #769 — this is a gate on *publishing*, not writing

ADR 0004 is explicit that Tier 2 and Tier 3 must not be **promoted** — in docs, the README, or a comparison table — until both prerequisites clear:

- **#756 (remote state)** blocks Tier 2: a sensor and a CI run can't share a watermark while state is local disk. Guidance published before it lands documents a configuration that breaks quietly rather than loudly.
- **#769 (rate limiting v2)** blocks Tiers 2 and 3: frequent small runs multiply bursts against SaaS destinations in a way a nightly batch never touches, and only the `Retry-After` half has shipped.

**Tier 1 can be documented immediately** — it's ungated. A reasonable first cut is a Tier-1-only guide that says the other two are coming, rather than waiting for both gates.

## Related

- Blocked by #756, #769 (for Tiers 2/3)
- #854 (`drt serve` hardening), #855 (dagster sensors)
- ADR 0004

masukai:
Written: [docs/guides/event-driven-syncs.md](https://github.com/drt-hub/drt/blob/main/docs/guides/event-driven-syncs.md), covering all three tiers with a decision table and worked Dagster-sensor example.

[...]

Issue #855, opened 2026-07-27
feat(dagster-drt): sensors for event-driven activation — the Tier 2 path ADR 0004 recommends


masukai:
Sanctioned by [ADR 0004](https://github.com/drt-hub/drt/blob/main/docs/adr/0004-streaming-and-event-triggered-syncs.md) (refs #786). This is Tier 2 — the ADR's recommended path for genuine event-driven activation.

## Why this is a package addition, not a write-up

#786 supposed `dagster-drt` "may already be 80% of the answer". It isn't: the package ships `assets.py`, `resource.py`, `specs.py`, `translator.py` and exports seven symbols — **none of them a sensor** (`integrations/dagster-drt/dagster_drt/__init__.py`). The asset and resource plumbing to build on exists; the sensor doesn't.

## Scope

A sensor that evaluates a cheap change signal and yields one `RunRequest` per changed sync. Dagster supplies the durability, cursoring and backfill semantics drt would otherwise have to invent — which is the whole reason the ADR routes event-driven activation here rather than into a drt-owned daemon.

Two nearly-free starting points, per the [trigger matrix](https://github.com/drt-hub/drt/blob/main/docs/research/warehouse-trigger-matrix.md):

- **Delta Lake** — `DeltaTable(...).version()` is already called in shipped code (`drt/sources/deltalake.py:67`); a monotonic integer is an ideal sensor cursor.
- **Iceberg** — snapshot id, reachable from the loaded pyiceberg table (`drt/sources/iceberg.py:51-52`), metadata-only read.

Then the purpose-built signals: **Snowflake `STREAM` + `SYSTEM$STREAM_HAS_DATA()`** and **SQL Server Change Tracking** — both designed to be polled cheaply, which is exactly the shape a sensor provides. See the matrix for per-source caveats (notably that `SYSTEM$STREAM_HAS_DATA` has documented false positives and stream consumption requires DML).

## ⚠️ Blocked by #756

Not a soft dependency. `.drt/state.json` is local disk (`drt/state/manager.py`), so a sensor running in an orchestrator and a CI run **cannot share a watermark today**. Shipping this before remote state lands would hand users a topology whose two halves silently disagree about what has already synced — the failure mode ADR 0004 explicitly gates against.

## Related

- Blocked by #756 (remote state backend)
- Also gated on #769 for the Tier 2 promotion per ADR 0004
- ADR 0004 Tier 2

masukai:
[...]

Issue #854, opened 2026-07-27
feat(cli): drt serve concurrency contract — a dropped trigger is silent event loss


masukai:
Sanctioned by [ADR 0004](https://github.com/drt-hub/drt/blob/main/docs/adr/0004-streaming-and-event-triggered-syncs.md) (refs #786). Tier 3 of that ADR is "hardened `drt serve` for push sources"; this is the hardening.

## Problem

`drt/cli/server.py`'s module docstring states it outright: *"One sync at a time — concurrent requests get 423 Locked."* That's a reasonable simplification at the cadence `drt serve` was designed for (a dbt job finishes, POST `/sync/<name>`). At the sub-minute cadence a real push subscription delivers, three properties stop being acceptable:

1. **A dropped trigger is silent event loss, not backpressure.** `_SyncLock` (`drt/cli/server.py:28`) is global across *every* sync, not per-sync, and a second trigger is answered `423` and dropped (`:90`) — no queue, no retry contract. The sender believes it delivered; nothing ran.
2. **The sync runs synchronously inside the request handler** (`:100`), so the HTTP response is held for the sync's full duration. Any sender with a timeout — Pub/Sub push, GitHub, EventBridge — records a failure for a sync that actually succeeded, and retries it.
3. **Auth is an optional static bearer token** (`:61`). Real push sources sign their deliveries; without HMAC verification `drt serve` can't be exposed to one without a proxy in front.

(1) is the load-bearing defect. The others are correctness-of-contract issues that compound it.

## This is a design decision, not a bug fix

The open question is **what a concurrent trigger should get**, and the options have different contracts:

- a **bounded queue** (delivery guaranteed up to depth N, then shed)
- **`429` + `Retry-After`** (push the retry back to the sender, which Pub/Sub and friends already know how to do)
- **per-sync locks** replacing the global one (concurrent *different* syncs just run; only same-sync overlap is rejected)

Worth deciding deliberately rather than picking whichever is easiest to implement — it's the public behaviour of an endpoint people wire webhooks to.

## Also in scope

- Return **`202` with a run id** instead of holding the request open; expose the run's outcome separately (`drt status`, or a `GET /runs/<id>`).
- **HMAC signature verification** alongside the bearer token.

## Related

- Gated by #769 per ADR 0004 — frequent small runs multiply request bursts against SaaS destinations, and only the `Retry-After` half of #769 has shipped.
- ADR 0004's Tier 3.

Pawansingh3889:
[...]

Issue #693, opened 2026-07-02
feat(diff): --dry-run --diff should preview mirror deletions (destination/tracked/scoped)


masukai:
Follow-up to #686 / #687, fits the v0.8.1 Diff Polish theme.

`drt run --dry-run --diff` previews record-level upserts but not the mirror DELETE pass — for all three delete strategies (destination diff #340, tracked #686, scoped #687). Tracked is the most valuable preview: it can be computed read-only (read `_drt_synced_keys`, diff against the would-be key set) without writing state.

Scope sketch:
- destination strategy: preview = destination keys − source keys (needs a destination read — same machinery as the #470-era diff work)
- tracked: preview = state keys − source keys (cheap, read-only)
- scoped: apply the scope predicate to either
- Render in the existing `--diff` output shape with a `DELETE` marker

PR #857, opened 2026-07-27
docs(adr): accept ADR 0004 now the trigger matrix has landed (refs #786)

Pawansingh3889:
Follow-up to #845, picking up the three things left open when it merged. Refs #786 (closed), builds on #847.

## Proposed → Accepted

@Muawiya-contact's matrix checked both falsification conditions explicitly and met neither, so the no-watcher recommendation stands **on evidence rather than on having been asserted first**. That was the whole point of writing it falsifiable, so the status move belongs to the matrix landing, not to the ADR merging.

## The review notes from #845

**`v0.8.2` → `v0.8.3`** — flagged at merge as "worth bumping when the matrix lands and the ADR moves to Accepted". Both now true. I re-verified every line citation against the v0.8.3 tree individually rather than assuming: `server.py:15/28/61/90/100`, `state/manager.py:43`, `dagster_drt/__init__.py:6`, `deltalake.py:67`, `iceberg.py:52` all still point at what the ADR says they do, and none of #693/#470/#784/#779/#719/#723 touch `serve`, state locality, or the Dagster integration.

**The push/poll claim is now stated precisely.** This is the substantive change. The ADR asserted "every push signal is already a message bus the user runs," which the matrix showed understates what exists — nine push mechanisms across thirteen sources, and only one of four categories is actually "already a broker." Left as written, the ADR's headline reasoning invited a fair objection that the matrix had already answered. It now carries the four-category table, with Databricks' table update trigger named as the closest counter-example and why it resolves to Tier 1 (managed polling that starts a job).

**Follow-ups wired to real numbers** — #854, #855, #856, replacing the "to be opened separately" placeholder, plus the matrix's note that Delta `version()` and Iceberg's snapshot id are the cheapest first sensors for #855.

## Not changed

My four provisional matrix rows that the research corrected — the `rest_api.py` citation pointing at `extract()`'s signature rather than the HTTP call, BigQuery's non-existent "table change notification" feature, SQL Server Change Tracking undersold, SQLite's WAL-mode `PRAGMA data_version` — needed no edit here. They left with the provisional table in #847, and I confirmed none survive elsewhere in the ADR.

Docs-only.

### DIFF
--- a/docs/adr/0004-streaming-and-event-triggered-syncs.md
+++ b/docs/adr/0004-streaming-and-event-triggered-syncs.md
@@ -1,11 +1,13 @@
 # ADR 0004 — Streaming / event-triggered syncs
 
-- **Status:** Proposed. The recommendation was written to be falsified by the
+- **Status:** Accepted. The recommendation was written to be falsified by the
   per-warehouse trigger matrix ([#786](https://github.com/drt-hub/drt/issues/786),
   @Muawiya-contact); that matrix has since landed as
   [docs/research/warehouse-trigger-matrix.md](../research/warehouse-trigger-matrix.md)
   and meets neither [falsification condition](#falsification-condition), so the
-  recommendation stands unchanged.
+  recommendation stands unchanged. Accepted on that evidence rather than on
+  having been asserted first — the matrix checked both conditions explicitly
+  and found neither met.
 - **Issue:** [#786](https://github.com/drt-hub/drt/issues/786)
 - **Implementation:** none — this ADR recommends **not** building a native
   watcher. The work it does sanction is listed under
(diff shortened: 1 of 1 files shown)


masukai (approved):
Approving and merging. This picks up all three of the things I left open at #845, including the one I explicitly said didn't need fixing.

**Proposed → Accepted is correctly earned.** Tying the status move to the matrix landing rather than to the ADR merging is the right reading — the ADR was written to be falsifiable, the matrix checked both conditions and met neither, so acceptance rests on evidence. That's the distinction worth preserving in the record.

**The `v0.8.2` → `v0.8.3` bump.** I said at merge this was worth doing "when the matrix lands and the ADR moves to Accepted", and you did it at exactly that point rather than earlier or never. I re-verified the citations myself against current `main` rather than taking the claim: `server.py:15/28/61/90/100`, `state/manager.py:43`, `dagster_drt/__init__.py:6` all still land on what the ADR says they do. Your added note — that none of v0.8.3's changes touch `serve`, state locality, or the Dagster integration — is the part that makes the bump meaningful rather than cosmetic.

**The refinement to the structural claim is the best change here.** Splitting "every cheap signal is a poll" (holds, twelve of thirteen) from the push half, and stating the push half as the four-category argument rather than "every push signal is a bus", removes a fair objection: nine push mechanisms *do* exist, and the loose phrasing invited someone to point at one and think the ADR hadn't looked. The precise version survives that.

Merging. #786's follow-ups are tracked in #854 / #855 / #856, and #854 is yours.

Issue #851, opened 2026-07-27
refactor(mcp): drt_run_test should call execute_tests_for_sync, not duplicate its loop


masukai:
Follow-up from #836 (the "Better" fix it named, split out for independent review).

## Problem

`drt/mcp/tools/run_test.py`'s `run_test()` duplicates the test-execution loop that `drt/cli/commands/test.py`'s `execute_tests_for_sync()` already implements — connect, build query, execute, check, severity handling, error handling. Two implementations of the same rule means a future change to severity semantics (or anything else in the loop) can silently drift between the CLI and the MCP tool, which is exactly what #400 was.

#836 named this as the "Better" fix but explicitly deferred it, since it was better sequenced after #723 (the `mcp/server.py` split into `mcp/tools/*.py`) — which has since landed (v0.8.3).

## Proposal

Have `run_test()` call `execute_tests_for_sync()` per sync, translating its `(_SyncTestResult, had_failures)` return into `drt_run_test`'s existing JSON shape. This should collapse to close to a thin adapter — `execute_tests_for_sync` already accepts `dry_run` / `json_mode` / `quiet` and returns the same per-test entry shape (`name`, `passed`, `value`/`error`, `severity`) that `run_test()` builds by hand today.

Things to check before landing:
- `execute_tests_for_sync` takes `store_failures` / `store_failures_limit` / `project_dir` — `drt_run_test` doesn't expose those params today; decide whether to thread them through or hardcode `store_failures=False` for the MCP tool (probably the latter — no obvious way to retrieve a written sample over MCP yet).
- `execute_tests_for_sync` prints to console when `show=True` (`not json_mode and not quiet`) — the MCP tool has no console output today, so this needs `quiet=True` or an equivalent no-print path, not just the default.
- Diff the current `drt_run_test` return shape against what `execute_tests_for_sync`'s result dict provides field-for-field before landing, since the current shape is public MCP contract.

## Acceptance criteria

- [ ] `run_test()` no longer duplicates the test-execution loop
- [ ] Existing `test_mcp.py` tests (including #836's severity parity tests) pass unmodified against the new implementation — same contract, different plumbing
- [ ] No console output leaks through the MCP tool call

Muawiya-contact:
Hi @masukai, I have one clarification before I start working on this.

[...]

PR #858, opened 2026-07-28
feat(destinations): rate limiting v2 — destination override + shared per-endpoint bucket (closes #769)

### LINKED ISSUE #769: feat(destinations): rate limiting v2 — honor Retry-After, per-destination rate_limit override, shared bucket across threads

masukai:
Closes #769 (items 2 and 3 — `Retry-After` shipped in v0.8.1). Unblocks #855 / #856, which ADR 0004 gates on this.

## The bug, concretely

`drt run --threads 4` against one HubSpot portal was putting **4× the configured rate** on it. Each sync built its own `RateLimiter` inside `load()`, so N parallel syncs meant N independent buckets against one quota. A nightly batch never noticed; event-driven activation (ADR 0004's Tier 2/3) would hit it immediately, which is why that ADR named this a gate.

## What's in it

**1. Per-destination `rate_limit` override** — precedence `destination > sync > default`, a direct mirror of `resolve_retry`. Added to 19 configs next to their existing `retry`.

**2. Process-wide limiter registry** — one bucket per *endpoint*, shared across threads. `RateLimiter.acquire()` gained a lock (it read-then-wrote `_last` with nothing in between; safe only while every instance was thread-local, which is exactly what this PR changes).

**3. `burst`** — opt-in accumulation. `burst: null` (the default) is byte-identical to the previous behaviour.

## Why `describe()` couldn't be the registry key

This is the non-obvious part. Both existing identity methods are wrong, in opposite directions:

| Config | `_describe_detail()` | Effect if used as a key |
|---|---|---|
| Slack | the literal `"webhook"` | **Different webhooks collide** into one bucket — unrelated workspaces throttle each other |
| HubSpot | `object_type` | `contacts` and `deals` get **different** buckets despite being one portal with one quota — i.e. the bug we're fixing |

`describe_safe()` is additionally lossy *by design* (#696 hardening — Twilio returns only a country-code prefix). So there's a new `rate_limit_key()`, overridden per connector to name the actual quota holder: HubSpot → portal token (**not** `object_type`), Slack → webhook, Zendesk → subdomain, Airtable → `base_id` (**not** table), REST API → `netloc`, GitHub Actions → **token, not `owner/repo`** (GitHub's REST quota is per authenticated user across repos). Keys prefer env-var *names* over resolved secrets; literals are hashed; nothing is logged.

## Vendor ceilings are preserved

HubSpot 9/s, Zendesk 11/s, GitHub Actions 5/s, Notion 3/s are published hard limits. The clamp applies to the **resolved** rps *before* the registry lookup — clamping the returned instance would let each caller overwrite the shared rate, last-writer-wins. Verified: an override of 500/s against HubSpot resolves to 9; an override of 2/s stays 2 (ceiling, not target); `rps=0` ("pacing disabled") survives the clamp.

## Same endpoint, different rates → min-wins

The stricter limit is the one the endpoint actually requires; running at the looser one is how you get a 429. Applied by **mutating the shared instance** rather than replacing it — a replacement wouldn't be visible to threads already holding a reference, and would reset `_last`, leaking one slot.

## A bug found during review, not by the tests

The first implementation used `_last == 0.0` as the "never used" sentinel. With burst, `_last` deliberately runs *behind* the wall clock while credit is spent — at `monotonic() == 0.0` it walks `-2.0 → -1.0 → 0.0`, re-hits the sentinel, and **refills the allowance every third call**. It never slept: an unbounded rate leak in the one class that must not have one. Caught by running the plan's own worked example (`rps=1, burst=3` → expected 3 free calls, got 6 of 6). Replaced with an explicit `_used: bool`, with a regression test pinned at `monotonic() == 0.0` that I verified fails against the old check.

## Notes for the reviewer

- **`limiter_factory` seam:** three existing tests (`zendesk`, `linear`, `sendgrid`) patch `drt.destinations.<name>.RateLimiter` and count `acquire()`. Delegating construction to the registry would have made those patches target a name `load()` no longer calls — they'd have passed against nothing. The seam keeps them live; confirmed by removing it and watching the zendesk test fail.
- **Registry state between tests:** an autouse fixture in `conftest.py` calls the existing `_reset_limiter_registry()`. No existing test was modified.
- **`RateLimitConfig` moved** from `sync_options.py` to `config/base.py` (beside `RetryConfig`) — referencing it from destination modules otherwise closed an import cycle. Every public import path is unchanged.
- The class docstring claimed "token-bucket" and never was one. Corrected, and the Rust-migration note now says the lock is a Python-concurrency artifact rather than part of the algorithm.

## Verification

- **2446 passed, 5 skipped** (baseline 2390 → +56 new tests). **Zero existing tests modified** — `git diff` over `tests/` is insertions-only.
- `make lint` (ruff + mypy, 172 files) clean, **no `type: ignore` added**
- `make check-drift` 0 · `make check-skills` ✓ · `make check-changelog` ✓
- VS Code schema **regenerated, not hand-edited** — re-running the generator produces zero diff. Extension version + CHANGELOG bumped per the convention of prior regens.
- 13 connector docs gained a `## Rate limiting` section stating each vendor's published limit and whether drt enforces it.


masukai:
@Muawiya-contact @Pawansingh3889 @yodakanohoshi — requesting review from all three of you, but **one of you is plenty**. Please don't all read it, and please don't drop what you're on for it: @Muawiya-contact has #851 and @Pawansingh3889 has #854, and both of those are more interesting than this. If nobody has room this week, say so (or say nothing) and I'll merge it — it's green and the invariants are pinned by tests.

It's 49 files, which sounds worse than it is: ~40 of those are one-line config additions and connector docs. **The whole review is two decisions**, both in `drt/config/base.py` and `drt/destinations/rate_limiter.py`:

**1. The registry key.** Two syncs share a rate limiter exactly when they share a vendor quota. I couldn't use `describe()` — it's wrong in both directions: Slack describes every webhook as the literal `"webhook"` (unrelated workspaces would throttle each other), and HubSpot includes `object_type` (one portal's contacts and deals would each get a bucket, which *is* the bug this PR fixes). So there's a new `rate_limit_key()`, overridden per connector to name the real quota holder. **Is any connector's key wrong?** That's the question where being close to the connector code pays off — @Muawiya-contact, you rewrote a lot of these `load()` methods in #723, so if one of them looks off to you it probably is.

**2. Inline credentials collapse to one bucket.** CodeQL flagged hashing a credential-shaped value (correctly — it can't see "but it's only a dict key", and the rule fires on any fast hash, not just SHA-256). Rather than reach for a slow KDF on a hot path, I stopped deriving the key from the secret at all: name an env var and you get a per-account bucket; inline the literal and you share one bucket with every other inline-literal config of that type. That errs *toward* throttling, so the failure mode is a slower sync rather than a 429 — but it **is** user-visible behaviour, and I decided it mid-PR while chasing the scanner. **A second opinion on that trade-off is the thing I'd most like.**

@Pawansingh3889 — if you do look, the ADR-adjacent question is whether this actually clears the #769 gate you wrote into ADR 0004 for Tiers 2/3, or whether the shared bucket needs to survive across processes (it doesn't today — the registry is per-process, so two `drt run` invocations still don't coordinate). I think per-process is the right scope for now and #756 is where cross-process would live, but you framed the gate so you'd know if that's a stretch.

Everything else — the burst credit, the thread-safety lock, the vendor ceilings (HubSpot 9/s etc., which the override cannot exceed) — has tests that fail when you break them; I checked by breaking them.

Muawiya-contact (commented):
## Review

@masukai Overall, I think this is a solid implementation. I went through the changes and even reproduced a few scenarios locally. The shared limiter, registry behavior, and the overall architecture all look good.

I only found one thing that I think is worth addressing before merge.

### High: `tighten_to()` breaks the "min-wins" rule when `requests_per_second = 0`

`RateLimiter.tighten_to()` currently compares rates numerically:

```python
if requests_per_second < self.requests_per_second:
    self.requests_per_second = requests_per_second
```

The problem is that `0` is not just another number in this implementation—it is the documented sentinel meaning **"rate limiting disabled"**.

Because of the numeric comparison, the disabled value actually wins.

For example:

```
Sync A -> requests_per_second = 10
Sync B -> requests_per_second = 0
```

The shared limiter becomes:

```
0
```

instead of:

```
10
```

That means a single sync configured with `0` disables rate limiting for every sync sharing that endpoint.

This seems to contradict the goal of this PR, which is to ensure that multiple syncs sharing the same vendor quota always use the strictest effective limit.

I also checked the tests.

`test_cap_preserves_the_zero_means_unlimited_sentinel` verifies that the clamp preserves the sentinel, but it never exercises `tighten_to()`, which is where this behavior occurs.

One possible approach would be to treat positive rates as stricter than the sentinel, so:

* positive rate + positive rate → use the minimum
* positive rate + 0 → use the positive rate
* 0 + 0 → remain disabled

That would preserve the existing sentinel semantics while still maintaining the "min-wins" behavior for shared limiters.

Other than that, the implementation looks good to me.

Pawansingh3889:
@masukai — good question to route to me, and the honest answer is that it clears for one tier and not the other.

The gate as I wrote it names two things, and this PR ships both: the per-destination `rate_limit` override, and the shared bucket across threads. On the requirement text alone, #769 is done.

But the rationale I attached to it was:

> a per-process bucket that resets every run will trip limits nightly batches never saw

and that half is not fixed. So the answer splits per tier:

**Tier 3 — cleared.** `drt serve` is one long-lived process. Every trigger lands in it, so a process-wide registry isn't a partial fix there, it's the whole scope. The bucket lives for the life of the server rather than resetting per run, which is precisely what that sentence was worried about. #854 will make serve concurrent via per-sync locks or a queue, and that concurrency is in-process, so the registry covers it as-is. Tier 3's remaining blocker is #854 itself, not #769.

**Tier 2 — not cleared, but not by this PR's fault.** A Dagster sensor yields one `RunRequest` per changed sync, and Dagster launches each run as its own process. Four changed syncs into one HubSpot portal is four processes and four buckets; the registry can't see across them. That's the exact topology the rationale describes.

So your instinct is right on both counts: per-process is the right scope here, and cross-process belongs with #756. It also changes no ordering, since #756 already blocks Tier 2 on its own and is the longer pole. A cross-process bucket without shared state isn't really achievable anyway — it *is* #756.

If you agree, I'd rather fix this on my side than yours:

- close #769 on this PR (after @Muawiya-contact's `tighten_to` finding — that one should land first)
- amend the ADR gate table: #769 blocks Tier 3 only and clears here; Tier 2's residual rate-limit concern folds into the #756 row that already blocks it
- note in #856 that the Tier 3 section becomes publishable once #854 lands

One caveat I'll write into the ADR: several `drt serve` replicas behind a load balancer puts us back at N buckets. That's a hosting topology, so it sits on the drt Cloud side of the open-core line — consistent with the rest of the ADR.

---

On the inline-credential trade-off, since you asked for a second opinion on it specifically:

I think the call is right. Refusing to derive the key from the secret at all is the correct response to that scanner finding — reaching for a slow KDF on a hot path to satisfy a rule that exists for password storage would have been the worse fix, and #696 already settled that credential-derived material shouldn't be published. Collapsing to one bucket also fails in the safe direction: stricter pacing, so the worst case is a slow sync rather than a 429.

The gap is that it's silent. Two inline-literal HubSpot configs quietly sharing a bucket is invisible at runtime, and the person who hits it will be debugging throughput, which is the hardest symptom to trace back to a key-derivation decision. The docstring in `base.py` covers it well for anyone reading the source, but that's not where a user looks. A line in the connector docs — name the env var if you want a per-account bucket — would close it, and I don't think it needs to block this PR.

masukai:
@Muawiya-contact — confirmed, fixed, and thank you. That was a real bug and a good catch: it defeats the entire point of the PR, and it would have shipped.

Reproduced before touching anything:

```
A=10 then B=0  -> effective rps = 0
```

`acquire()` returns immediately at `rps<=0`, so pacing was genuinely off — not merely slow. One sync opting out with `requests_per_second: 0` removed the shared-quota protection from every other sync on that endpoint. Exactly inverted.

**It was worse than reported, in a way worth recording.** The reverse order fails too:

```
A=0 then B=10  -> effective rps = 0
```

Registering the sentinel *first* leaves the bucket at 0 permanently, because no positive rate is ever numerically smaller than one already in the registry. Since registration order is thread-scheduling order, only fixing 10-then-0 would have left a bug that reproduces on roughly half of runs — the worst kind to chase. Your framing ("positive rates are stricter than the sentinel") is what makes both directions fall out of one rule, which is why I took it as written:

```python
if requests_per_second > 0 and (
    self.requests_per_second <= 0 or requests_per_second < self.requests_per_second
):
    self.requests_per_second = requests_per_second
```

Your read of the test gap was also exactly right — `test_cap_preserves_the_zero_means_unlimited_sentinel` pins the *clamp* and never reaches `tighten_to`. Three tests now cover the registry path: sentinel-then-rate, rate-then-sentinel, and 0+0 staying disabled.

---

@Pawansingh3889 — the tier split is more precise than what I asked for, and I'm taking all three of your proposed actions. Please go ahead and make the ADR edits yourself, as you offered; you wrote the gate and your wording for it is better than mine would be.

Two things I want to confirm back explicitly:

**Tier 2.** You're right, and I'd have gotten this wrong on my own. I was thinking of "cross-process" as a hardening step for later, but a Dagster sensor launching one process per `RunRequest` isn't an edge case — it's the normal topology, and four changed syncs into one portal is four buckets today. Folding the residual into the #756 row is correct precisely because a cross-process bucket without shared state *is* #756; there's no smaller intermediate fix worth inventing.

**The load-balancer caveat.** Please do write it in. Several `drt serve` replicas putting us back at N buckets is the kind of thing that's obvious once stated and invisible otherwise, and putting it on the Cloud side of the open-core line is consistent with the rest of the ADR.

On the inline-credential trade-off — thank you for engaging with the actual question. Your point about *silence* is the part I'd underweighted: I had reasoned "fails safe, so it's acceptable" and stopped there, but "fails safe **and** is undiagnosable" is a different claim, and throughput really is the symptom nobody traces back to key derivation. So I didn't defer it — it's in this push, on the two connector docs that already describe the sharing rule:

- **hubspot.md** — use `auth.token_env` rather than an inline `auth.token` when syncing to more than one portal, with the reason (drt keys on the env var *name*, never the token value) and a YAML example.
- **slack.md** — this one was actively wrong, not merely silent. It claimed "Different webhooks never throttle each other," which is false for inline literals. Now corrected to say distinct `webhook_url_env` names never collide but two inline `webhook_url` literals share one bucket.

I verified every claim in that prose against `rate_limit_key()` rather than reasoning about it, since docs that describe behaviour are exactly where I'd expect to be confidently wrong:

```
env EU vs US    : slack:HOOK_EU | slack:HOOK_US   -> distinct
literal A vs B  : slack:literal | slack:literal   -> shared
contacts/deals, same portal                       -> shared
different portals                                 -> distinct
secret value present in key?                      -> no
```

While there I also fixed three docstrings still saying a literal is "hashed" / "only its digest goes in the key" — left over from before the CodeQL fix replaced derivation with `LITERAL_CREDENTIAL_KEY`. Stale in a bad direction: they'd tell a future reader secrets are still digested.

---

Sequencing: this is @Muawiya-contact's fix first, as you suggested. I'll close #769 on merge; @Pawansingh3889, the ADR amendment and the #856 note are yours whenever suits — no rush, and neither blocks this.

masukai:
Heads-up: I rebased this branch onto `main` just now (force-push, `57d5ea8` → `4df7c87`). #859 merged, which put a CHANGELOG conflict on this PR, and I would rather resolve that myself than hand you a conflicted branch mid-review.

**No code changed.** The rebase touched `CHANGELOG.md` only — both entries now sit under `[Unreleased]`, and the #769 wording is the corrected one (the earlier text still claimed literals were hashed, which stopped being true when the CodeQL fix replaced derivation with a constant).

Your `tighten_to` fix and its three tests are untouched and still green: 2550 passed with both PRs' changes integrated, ruff and mypy clean.

If you had the old commits checked out locally, you'll want a fresh fetch. Sorry for the churn — no rush on the review.

PR #862, opened 2026-07-29
test(sources): pin driver doubles against the real packages

masukai:
Follow-up to #861, which found a real bug that a green unit suite could not have caught. This generalises the fix for the *cause*, and records two things only live servers revealed.

Tests only — no production change.

## The pattern behind #861

The Databricks double had `RequestError(Error)`; the real driver has `RequestError(OperationalError)`. That single wrong base meant a 401 never reached the `isinstance` path the auth exclusion guards, so the bug was structurally untestable. The suite was green and the code was broken.

Every optional-driver source here is tested against a hand-written double, so the same drift can hide the same class of bug anywhere. These tests compare each double against the real package **class by class, full MRO**, and fail when they disagree.

The subtlety worth reviewing: asserting the *real* hierarchy on its own does not work. I wrote that version first and it passed happily with a deliberately broken double — the double is what every other test actually runs against, so the two have to be compared **directly**. Confirmed by mutation:

| mutation | before | after |
|---|---|---|
| pymssql double: `OperationalError(DatabaseError)` → `(Error)` | passed | **fails** |
| ClickHouse double: `StreamClosedError(ProgrammingError)` → `(OperationalError)` | passed | **fails** |

They `importorskip`, so they skip in CI's default matrix — which is precisely where #861 survived. A backstop, not the primary guard.

## What live servers showed

**SQL Server: the error number cannot classify a login failure.** pymssql packs the whole DB-Lib conversation into `args`, and 18456 rides along even on a connection that was simply refused. Captured from SQL Server 2022:

```
auth failure : (18456, b"Login failed for user 'sa'.DB-Lib error message 20018...")
refused      : (18456, b"...Unable to connect: Adaptive Server is unavailable...
                Connection refused (61)")
```

Both are `OperationalError`, both carry 18456. So the number-based check I considered and rejected in #859 would have been unable to mark the real auth failure without *also* killing the retry for a server that is merely down — it fails in both directions at once. The message match is the only thing separating them. There is now a test saying so, with the real payloads in it, so nobody "simplifies" it to a number later.

**The rest of the audit came back clean**, this time against live servers rather than by reading source:

- **MySQL 8** — 1045 (bad password) not retried, 1049 (unknown database) not retried, 2003 (connection refused) retried. Correct as it stands.
- **ClickHouse 24** — a real authentication failure arrives as a bare `DatabaseError` (code 516), already classified permanent. No auth exclusion needed.

I'd flag that I had *also* convinced myself ClickHouse was broken, by constructing an `OperationalError` by hand — a configuration the server never actually produces. The live check is what settled it in both directions, and that is the general lesson: for these drivers, reading the source tells you what *can* exist, not what *does*.

## Not covered

Snowflake and Databricks have no live-server verification here — no free local server exists for either. The live home is [`dwh-smoke.yml`](https://github.com/drt-hub/drt/blob/main/.github/workflows/dwh-smoke.yml) (#654 itself is closed), which gained source-side legs for both warehouses in #865. Worth naming its own caveat: that workflow **no-ops without `DRT_SMOKE_*` secrets** — the same skip property one layer up, and precisely the thing this PR exists to make visible rather than assume away. Their doubles are pinned by #861's equivalent test, but "what does the server actually send" stays open for both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,13 +38,16 @@ jobs:
         run: pip install uv
 
       - name: Install dependencies
-        # SQL extras (postgres / mysql / clickhouse) ensure the destination
-        # test suites actually run in CI rather than being silently skipped
-        # by their `pytest.importorskip(...)` guards (#596 / #340 follow-up).
+        # SQL extras (postgres / mysql / clickhouse / sqlserver) ensure the
+        # destination test suites actually run in CI rather than being silently
+        # skipped by their `pytest.importorskip(...)` guards (#596 / #340
+        # follow-up). sqlserver joined them in #862: the driver-double fidelity
+        # pins are worthless while they skip, and pymssql ships manylinux
+        # wheels so it costs no toolchain.
         # snowflake is omitted — its tests use sys.modules injection so the
         # heavy snowflake-connector-python install would only slow CI.
         # bigquery / gcs / parquet remain opt-in for the same reason.
-        run: uv pip install --system -e ".[dev,mcp,duckdb,postgres,mysql,clickhouse]"
+        run: uv pip install --system -e ".[dev,mcp,duckdb,postgres,mysql,clickhouse,sqlserver]"
 
       - name: Dependency vulnerability scan (pip-audit)
         # Pinned version for reproducibility; bump in lockstep with Dependabot.
--- a/drt/sources/sqlserver.py
+++ b/drt/sources/sqlserver.py
@@ -142,7 +142,13 @@ class SQLServerSource:
 
         return pymssql.connect(
             server=config.host,
-            port=config.port,
+            # pymssql's own stub types `port` as `str`, though the driver
+            # accepts an int at runtime (verified: an int port fails on the
+            # connection, not on the argument). Passing the string satisfies
+            # the stub without changing behaviour — the alternative was a
+            # `type: ignore` for a third-party stub inaccuracy, which is
+            # noisier and would outlive the stub being fixed.
+            port=str(config.port),
             user=config.user,
             password=password,
             database=config.database,
--- a/tests/unit/test_clickhouse.py
+++ b/tests/unit/test_clickhouse.py
@@ -132,6 +132,7 @@ def _install_fake_ch_exceptions(monkeypatch: pytest.MonkeyPatch) -> Any:
     InternalError / NotSupportedError as siblings under DatabaseError, and
     StreamClosedError under ProgrammingError.
     """
+    import builtins
     import sys
     import types
 
@@ -162,6 +163,30 @@ def _install_fake_ch_exceptions(monkeypatch: pytest.MonkeyPatch) -> Any:
     class StreamClosedError(ProgrammingError):
         pass
 
+    # Added in #862 once the pin enumerated the real module instead of a
+    # hand-kept list, which is exactly how their absence surfaced.
+    # StreamFailureError and StreamCompleteException descend from plain
+    # Exception rather than DatabaseError — worth mirroring precisely, since
+    # re-parenting either under OperationalError is what would silently start
+    # retrying a stream failure.
+    class InternalError(DatabaseError):
+        pass
+
+    class NotSupportedError(DatabaseError):
+        pass
+
+    class StreamCompleteException(Exception):
+        pass
+
+    class StreamFailureError(Exception):
+        pass
+
+    # The driver's Warning subclasses the *builtin* Warning as well as its own
+    # base. Referenced via `builtins` because defining a class named Warning in
+    # this scope makes the bare name local.
+    class Warning(builtins.Warning, ClickHouseError):  # noqa: A001 - mirrors the driver
+        pass
+
     exc_mod = types.ModuleType("clickhouse_connect.driver.exceptions")
     for cls in (
         ClickHouseError,
@@ -173,6 +198,11 @@ def _install_fake_ch_exceptions(monkeypatch: pytest.MonkeyPatch) -> Any:
         DataError,
         IntegrityError,
         StreamClosedError,
+        InternalError,
+        NotSupportedError,
+        StreamCompleteException,
+        StreamFailureError,
+        Warning,
     ):
         setattr(exc_mod, cls.__name__, cls)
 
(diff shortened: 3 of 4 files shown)


Muawiya-contact:
@masukai I will review it. I think there is a conflict. Can I resolve it, or will you?

masukai:
@Muawiya-contact — I resolved it, and thank you for offering. The conflict was mine to clean up: I merged the streaming stack around this PR while it sat waiting for you, so the mess was created by my merge order rather than by anything in your court.

Rebased onto `main` and pushed. **No code changed** — the conflict was two appended test classes landing at the same spot in `test_clickhouse.py` and `test_sqlserver_source.py`, so both sides are kept. 2616 passed, ruff clean.

I also mutation-tested your guard after the rebase, to be sure it still bites:

```
変異: ClickHouse フェイクの StreamClosedError を OperationalError 下へ
  → FAILED test_the_fake_exception_hierarchy_matches_the_real_driver
復元 → 24 passed
```

One thing I fixed while in there, unrelated to the conflict: `TestStreamingExtraction`'s docstring in `test_clickhouse.py` still claimed `query_rows_stream()` peaked at **+0.0 MB** and was "the largest gap of any source". That was the measurement error I corrected in #864 — `ru_maxrss` is a process-wide high-water mark, so measuring the buffered variant first made the streaming one look free. I updated the source docstring there but missed this one, so `main` has been carrying both the corrected figure (`+224 → +149 MB`, in `drt/sources/clickhouse.py`) and the wrong one (here) since #864 landed. Now consistent.

**The review is still genuinely open** — nothing about the rebase pre-empts it. The question I care about is the approach, not the code: every optional-driver test double here is hand-written, and #861 showed one with `RequestError` under the wrong base made a real bug structurally untestable while the suite stayed green. This pins each double against the real package, and skips when the extra is absent — which is exactly where the original bug survived, so it is a backstop rather than the primary guard.

Given how #40 went, that "green suite, wrong environment" failure mode is clearly not hypothetical for either of us. If you think the pinning approach is right, say so and I'll merge. If you think it papers over something better solved another way, that is the more useful answer.

No rush, and no obligation — you have already spent real time on drt-web this week.

Muawiya-contact (approved):
@masukai Reviewed properly — ran it, mutated it, and checked the CI claim rather than trusting it. **The pinning approach is right**, and I'll say why before the two things I found.

It's right because it attacks the causal mechanism, not the symptom: #861 wasn't a missing test, it was a fixture that lied — so the assertion space couldn't contain the bug, and every behavioural test inherited the lie. Pinning the double to reality is the only intervention at that layer, and it's falsifiable: I reproduced both mutation rows (ClickHouse `StreamClosedError` re-parent → the pin fails with the exact MRO diff; pymssql likewise; restore → green). I also wrote the counterfactual you described but didn't show — the assert-the-real-hierarchy-only variant — and confirmed it passes happily against a deliberately broken double. Your subtlety is load-bearing. On the alternatives: auto-generated doubles fail in the wrong direction (they *silently self-heal* to match a driver change, destroying exactly the signal you want loud), and protocols can't express "is a subclass of the driver's OperationalError" without restating the hierarchy. The 18456 test settles attribute-based classification for SQL Server permanently — good that the real payloads are pinned in it.

Two findings, both in the PR's favor:

**1. The "skips in CI's default matrix" framing is wrong for ClickHouse — the good kind of wrong.** `ci.yml` already installs the `clickhouse` extra (deliberately — the comment on that line cites exactly this silent-importorskip concern), so that pin executes on every PR across all four Pythons; codecov corroborates. Only SQL Server actually skips — and adding `sqlserver` to that same extras list closes it. `pymssql>=2.3` ships manylinux wheels (installed here with no toolchain), and the precedent is already written on that line. I'd make that one-word change in this PR rather than defer it: half of this is already a live guard, and the other half shouldn't wait on a smoke programme when it's one word from parity.

**2. The hardcoded class list is the same drift surface, one level up.** The pins iterate a literal list of names — six real ClickHouse classes and four pymssql classes are unpinned, including `InternalError`/`NotSupportedError` which the doubles' docstrings claim to mirror. Second-order today (we only isinstance two classes), but if a driver re-parents an unpinned class *under* `OperationalError` — `StreamFailureError` is plausible, and ClickHouse streaming is live since #765 — we'd silently start retrying it and the pin can't see it. Enumerating the real module's exception classes instead of hardcoding removes the hand-maintained list entirely — the PR's own principle applied to itself.

Small pointer fix: #654 is closed — the live home is `dwh-smoke.yml`, worth citing directly (noting it currently no-ops without secrets, the same skip property one layer up).

Approve — happy for this to merge as-is with the two above as immediate follow-ups, or with the `sqlserver` extra folded in here. Either way the approach is the right one.

masukai:
Both findings applied in `5fa149e`, and finding 2 paid for itself immediately.

**1. `sqlserver` extra added.** You were right and I was wrong in the direction that matters: I wrote "skips in CI's default matrix" into both docstrings without checking, and `ci.yml` has been installing the `clickhouse` extra all along — with a comment saying why. So half of this was already a live guard and my own description said otherwise. `pymssql` resolves a manylinux wheel (`pymssql-2.3.13-...manylinux_2_28_x86_64.whl`), so it costs no toolchain. Folded in here rather than deferred, as you suggested.

**2. Enumeration — this one found real gaps.** Replacing the literal list with "every exception class the module defines" failed on the first run:

```
ClickHouse: InternalError, NotSupportedError, StreamCompleteException,
            StreamFailureError, Warning
pymssql:    InternalError, NotSupportedError, ColumnsWithoutNamesError, Warning
```

Nine unmodelled classes. **Your `StreamFailureError` hypothesis was not hypothetical** — it exists, it descends from plain `Exception`, and it was invisible to the double. So the exact scenario you described (re-parented under `OperationalError`, silently retried, pin can't see it) was reachable, and ClickHouse streaming has been live since #765.

`ColumnsWithoutNamesError` is the pymssql equivalent: it sits under `InterfaceError`, which `_is_transient` already treats as **retryable**. Its shape was load-bearing and unpinned.

Mutation-tested both new properties:

```
StreamFailureError → OperationalError   : pin fails
double から InternalError を削除          : pin fails
復元                                     : 24 passed
```

Two things the doubles had to mirror exactly, which I'd have got wrong by guessing: both drivers' `Warning` subclasses the **builtin** `Warning` (and ClickHouse's also its own base), and referencing it needs `builtins` — defining a class named `Warning` in that scope makes the bare name local, which cost me a confusing `UnboundLocalError`.

**On the #654 pointer** — you're right, it's closed. Corrected to `dwh-smoke.yml` in the description, with the no-op-without-secrets caveat you noted. That property is worth stating plainly: it is the same skip behaviour one layer up, which is exactly the thing this PR exists to make visible.

---

Thank you for the depth here — specifically for reproducing the mutation rows and for writing the counterfactual I described but didn't demonstrate. "The fixture lied, so the assertion space couldn't contain the bug" is a better articulation of #861 than mine, and I'm going to reuse it. Your point about auto-generated doubles failing in the wrong direction — silently self-healing and destroying the signal — is the argument I should have made against that alternative myself.

2629 passed, ruff + mypy clean. Merging once CI is green.

Issue #766, opened 2026-07-10
feat(sources): source-side retry for transient extract failures (connect/execute, driver-aware allowlists)


masukai:
## Problem

Retry exists only on the load side (`RetryConfig` → `with_retry` in `drt/destinations/retry.py`). The extract side has none: a transient warehouse hiccup — Snowflake `390114` session expiry, BigQuery 503, Postgres connection reset during a long read, Databricks warehouse cold-start timeouts (we saw these classes during the #654 smoke program) — fails the whole sync immediately, even though the query is a read and perfectly safe to retry.

In scheduled/CI contexts this turns routine warehouse blips into failed runs and pager noise.

## Prior art

- **dlt**: source-side retries with backoff are built into extraction helpers.
- **dbt**: adapter-level connection retries (`connect_retries`, `retry_all` on several adapters).
- **Census/Hightouch**: managed retries on both legs.

## Proposal

```yaml
# profiles.yml (per source profile)
retry:
  max_attempts: 3
  initial_backoff: 2.0
  backoff_multiplier: 2.0
```

- Reuse the existing `RetryConfig` model; wrap connection establishment + query execution per source
- Retry only on transient error classes per driver (connection/timeout/service-unavailable allowlists — mirror the destination `retryable_status_codes` philosophy), never on SQL errors
- Emit retry events through `SyncObserver` so attempts are visible in logs/OTel

**Caveat to design around:** with streaming extraction, a mid-stream failure can't blindly re-run the query for at-least-once semantics into non-idempotent modes — first leg can retry connection+initial execution only (covers the dominant cold-start/connect class), mid-stream resume folds into the windowed-backfill design.

## Scope

**In:** connect/execute retry for all SQL sources + REST source, per-profile config, transient-error classification table per driver, observer events.
**Out:** mid-stream resume (windowed backfill), destination retries (exist).

## Related

- #277 (destination retry config — the model to mirror), #654 (real-warehouse smoke — the evidence base), #243 (destination retry alignment precedent)

## Milestone fit

Candidate: v0.8.2 (warehouse hardening & security) — directly serves the nightly smoke reliability story.

PR #865, opened 2026-07-29
perf(sources): stream Snowflake/Databricks extracts + source-side smoke legs (refs #765)

masukai:
Slice 4 of #765, completing the warehouse set. Stacked on #860 → #863 → #864, so the diff here is Snowflake/Databricks only.

**This one needs your credentials to be fully verified.** See "What I need from you" at the bottom.

## What

| | mechanism | `fetch_size` |
|---|---|---|
| Snowflake | iterate cursor, `arraysize` | yes |
| Databricks | iterate cursor | **no** |

Snowflake's cursor is iterable and honours `arraysize`, so it matches the Postgres leg's shape. Databricks' `__iter__` delegates to the result set, whose own `__iter__` is a `fetchone()` loop — verified against the driver source, so iteration genuinely streams. It takes no `fetch_size` because its cursor exposes no `arraysize` and `fetchmany(size)` requires an explicit argument: there would be nothing for the field to set.

## The real content of this PR: source-side smoke legs

Every other slice of #765 was verified against a live server, and **every one turned up behaviour no mock predicted**:

- psycopg2's `description` staying `None` on a named cursor (#860)
- pymysql needing `cursor.close()` before `conn.close()` (#863)
- clickhouse-connect buffering HTTP internally, which made my first ClickHouse measurement off by ~150 MB (#864)

Snowflake and Databricks have no local server. Shipping them on mocks alone would be the one case with no way to catch the equivalent — so both gained legs in the dwh-smoke harness.

**A gap worth naming:** until now that harness drove Snowflake and Databricks only as *destinations*, with DuckDB always the source. The extract path had **no real-service coverage at all**, for retry (#766) or anything else. These are the first tests that exercise it.

Each leg extracts 50k generated rows (`GENERATOR` / `range()`, so no fixture table) and asserts exact count, first and last row, and no duplicates — a batch-boundary bug surfaces as a short read or a repeat. Plus an empty result, a small `fetch_size` that must not change the rows, and a mid-stream abandonment that would **hang rather than fail** if the generator's `finally` did not run.

They skip without `DRT_SMOKE_*` credentials, so forks and local runs are unaffected (verified: 18 skipped, 0 attempted connections).

## What I need from you

I cannot run these — the credentials are repo secrets and I have no local Snowflake or Databricks. **Please trigger the `dwh-smoke` workflow on this branch before merging.** If a leg fails, that failure is the whole point of the PR and I'd rather see it than not.

Specifically worth watching:
- **`test_*_abandoned_mid_stream_does_not_hang`** — if closing a partially-read cursor blocks on either warehouse, this hangs the job rather than failing cleanly. That is the outcome I consider most likely, and it is exactly what the mocks cannot tell us.
- **Snowflake column casing** — I assert `ID` / `PAYLOAD` uppercase, matching the unquoted-identifier folding the existing destination tests rely on. If the source path returns them differently, these fail on the assertion rather than the mechanism.

## Verification I could do

2482 passed, ruff + mypy + drift clean. Guards mutation-tested with `__pycache__` cleared (dropping `arraysize` fails 1 test; reversing either close order fails 2 — I cleared the cache explicitly because a stale `.pyc` silently faked a passing mutation test in #863).

## Note on stacking

This branch predates #861, so `test_databricks_source.py` here still has the old `RequestError(Error)` double. #861 fixes that hierarchy and adds the real-driver pin. No conflict expected — different functions — but whichever merges second should confirm the fake is the corrected one.

## Remaining after this

SQL Server (`fetchmany`, docker-verifiable) and DuckDB/SQLite (local files — arguably out of scope for #765, as the issue itself suggests).


masukai:
`dwh-smoke` run on this branch: **all green, and all of it actually ran** — no credential-skip.

| job | result |
|---|---|
| Snowflake | **9 passed** (59.01s) |
| Databricks | **8 passed** (107.26s) |
| BigQuery | 4 passed (untouched) |

[run 30425157029](https://github.com/drt-hub/drt/actions/runs/30425157029)

**Both of the things I flagged as most likely to fail, passed:**

```
test_snowflake_source_abandoned_mid_stream_does_not_hang    PASSED
test_databricks_source_abandoned_mid_stream_does_not_hang   PASSED
```

Closing a partially-read cursor does not block on either warehouse, so `--limit` / `--fail-fast` mid-stream is safe. That was the outcome I could not have established from mocks — it would have hung the job rather than failed a check.

Snowflake's uppercase column folding on the source path also held (`ID` / `PAYLOAD`), and `test_snowflake_source_respects_a_small_fetch_size` passed with `fetch_size=100` over 1000 rows, so batch boundaries are handled correctly against the real service rather than only against a mock's `__iter__`.

That completes real-service verification for every streaming source in #765:

| source | verified against |
|---|---|
| Postgres / Redshift | Postgres 16 (docker) |
| MySQL | MySQL 8 (docker) |
| ClickHouse | ClickHouse 24 (docker) |
| Snowflake | live account |
| Databricks | live account |

Worth noting the side effect: the smoke harness previously drove both warehouses only as destinations, so the **extract path had no real-service coverage at all** — including the #766 retry work. It does now.

Ready to merge from my side. The stacking note in the description still applies: this predates #861, so whichever of the two merges second should confirm `test_databricks_source.py` ends up with the corrected `RequestError(OperationalError)` double.

Issue #39, opened 2026-03-30
good first issue: add PostgreSQL / MySQL destination (upsert rows)


masukai:
## Summary

Add database destinations (PostgreSQL and/or MySQL) so users can sync DWH rows into an operational database — e.g. push analytics results back to an app database, maintain a reporting table.

## Example sync YAML

```yaml
destination:
  type: postgres
  connection_string_env: TARGET_DATABASE_URL   # postgres://user:pass@host/db
  table: public.synced_users
  upsert_key: id          # column used for ON CONFLICT DO UPDATE
  columns_template: |     # optional: map DWH columns to destination columns
    {
      "id": "{{ row.user_id }}",
      "email": "{{ row.email }}",
      "updated_at": "{{ row.updated_at }}"
    }
```

## Implementation hints

1. Look at `drt/sources/postgres.py` for psycopg2 connection pattern
2. Add `PostgresDestinationConfig` to `drt/config/models.py`
3. Create `drt/destinations/postgres.py`
4. Use `INSERT ... ON CONFLICT (upsert_key) DO UPDATE SET ...` for upsert
5. Register in `DestinationConfig` union and `_get_destination()` in CLI

## Acceptance criteria

- [ ] `PostgresDestinationConfig` (and optionally `MySQLDestinationConfig`) added
- [ ] Upsert implemented with `ON CONFLICT DO UPDATE`
- [ ] At least 5 unit tests using a fake connection (no real DB)
- [ ] `make lint` and `make test` pass

Issue #38, opened 2026-03-30
good first issue: add Notion destination (append database rows)


masukai:
## Summary

Add a Notion destination so users can append rows to a Notion database from DWH data — e.g. sync customer metrics to a Notion dashboard, push weekly reports automatically.

## Example sync YAML

```yaml
destination:
  type: notion
  database_id_env: NOTION_DATABASE_ID
  properties_template: |
    {
      "Name": {"title": [{"text": {"content": "{{ row.name }}"}}]},
      "Email": {"email": "{{ row.email }}"},
      "Revenue": {"number": {{ row.revenue }}},
      "Status": {"select": {"name": "{{ row.status }}"}}
    }
  auth:
    type: bearer
    token_env: NOTION_TOKEN
```

## Implementation hints

1. Look at `drt/destinations/hubspot.py` as a reference
2. Add `NotionDestinationConfig` to `drt/config/models.py`
3. Create `drt/destinations/notion.py`
4. Register in `DestinationConfig` union and `_get_destination()` in CLI

## Notion API reference

- Endpoint: `POST https://api.notion.com/v1/pages`
- Auth: `Authorization: Bearer <NOTION_TOKEN>` + `Notion-Version: 2022-06-28` header
- Docs: https://developers.notion.com/reference/post-page
- Payload:
  ```json
  {
    "parent": {"database_id": "<DATABASE_ID>"},
    "properties": { ... }
  }
  ```

## Acceptance criteria

- [ ] `NotionDestinationConfig` added to `config/models.py`
- [ ] `drt/destinations/notion.py` implemented with retry + rate limiting
- [ ] At least 5 unit tests
- [ ] `make lint` and `make test` pass

PFCAaron12:
This is completed.
feat: add Notion destination [#326]

PFCAaron12:
I see that the test failed. I forgot this part that you mentioned to me:
Import sorting — ruff check --fix will sort these automatically. Tip: running make fmt locally before pushing catches these.

masukai:
**Status:** This issue is being handled by #328 (@armorbreak001). Previous PRs #326 and #327 have been closed in favor of #328 due to its more complete scope (includes tests, MCP server update, and a rich_text fallback). Please avoid opening additional PRs on this issue.

armorbreak001:
Hi @masukai, I'd like to formally request assignment for issue tracking.

**Completed:**
- ✅ PR #328 (Notion destination) — submitted, pending CLA sign
[...]

PR #866, opened 2026-07-29
perf(sources): stream SQL Server, DuckDB and SQLite extracts (refs #765)

masukai:
Slice 5 of #765, completing every SQL source. Stacked on #860 → #863 → #864 → #865.

## The local-file question: I was wrong

I argued on the issue that DuckDB/SQLite were out of scope because "it's a local file". You asked whether that was really true, so I measured instead of assuming:

| source | `fetchall()` | batched |
|---|---|---|
| DuckDB | +150.9 MB | **+42.2 MB** |
| SQLite | +110.6 MB | **+4.4 MB** |
| SQL Server | +151 MB | **+39 MB** |

The reasoning was wrong. What this change removes is **holding every row as a Python object** — a cost a local file pays exactly like a remote warehouse. Where the bytes came from is irrelevant to an OOM on a 7 GB runner. So both are in, not deferred.

(SQL Server and DuckDB numbers are through `extract()` itself; the SQLite/DuckDB pair above is the driver-level comparison. All measured in fresh processes, per the `ru_maxrss` correction in #864.)

## Mechanism differs per driver

- **SQL Server / SQLite** — iterate the cursor with `arraysize` set.
- **DuckDB** — explicit `fetchmany` loop. Its result object has **no `__iter__`**, unlike sqlite3 / pymssql / snowflake / databricks. Worth knowing before anyone "simplifies" it to a `for` loop.

## ⚠️ One deliberate semantic weakening — please read this bit

SQL Server previously materialised the result set with `list(cur.fetchall())` **inside** the retried closure. A driver failing part-way through producing rows therefore failed *before anything was yielded*, so the retry could safely re-run: no row had reached the destination, nothing could duplicate.

That made SQL Server strictly stronger than every other SQL source — but as an **accident of buffering**, not by design. Streaming removes the accident. A mid-stream failure now escapes with earlier rows already loaded downstream, which is the documented #766 boundary that Postgres/MySQL/Snowflake/etc. have always had.

The old test asserted the stronger guarantee. I rewrote it to assert the new one, with the reasoning in the docstring, rather than deleting it — so nobody reads the old shape and assumes it still holds. **If you'd rather keep SQL Server buffered to preserve that guarantee, say so** — it is a defensible trade and the only source where it applies.

## Mutation testing caught a real gap

First pass: reversing SQL Server's close order **passed**, because that file had no streaming tests at all. Added them; the mutation now fails 2 tests. Same for `arraysize` on both new sources.

Run with `__pycache__` cleared, after a stale `.pyc` faked a passing mutation test in #863.

## Verification

2495 passed, ruff + mypy + drift clean.

One note: `mypy drt` reports a pre-existing `call-overload` error in `sqlserver.py:143` (`_connect`, untouched by this PR) **when pymssql is installed**. It reproduces on unmodified `main` and does not appear in CI, which has no extras. Flagging rather than fixing, to keep this PR's scope.

## #765 is now complete

| source | verified against |
|---|---|
| Postgres / Redshift | Postgres 16 (docker) |
| MySQL | MySQL 8 (docker) |
| ClickHouse | ClickHouse 24 (docker) |
| Snowflake / Databricks | live accounts, dwh-smoke green |
| SQL Server | SQL Server 2022 (docker) |
| DuckDB / SQLite | real files |

Every source measured against something real, not a mock.


Muawiya-contact:
@masukai I noticed that there are many open pull requests waiting for review. I'm here, so feel free to ask me to review for you.

masukai:
@Muawiya-contact thank you — and yes, that pile is real. It is eight PRs, which is entirely my doing: I kept finding adjacent problems and kept opening branches rather than stopping to get them merged.

Before you take any of it, two things you should know, because they change whether this is worth your time.

**They are not blocked on review.** They are blocked on *me*. I authored all eight, so I cannot self-approve, and the merge itself needs an admin push past branch protection. Your approval would not unblock them on its own — I would still have to do the merge. So please don't feel that the queue is waiting on you, because it isn't.

**Five of them are a stack**, in this order: #860 → #863 → #864 → #865 → #866. Each builds on the last, so reviewing #866 in isolation would show you a diff whose context lives four PRs upstream. Reviewing them as a set is a much bigger ask than the count suggests.

If you do want to spend time here, the two I would genuinely value your eyes on are the standalone ones:

- **#861** — Databricks retries 401/403. The driver itself classes those as never-retryable and then surfaces them as `RequestError`, which subclasses `OperationalError`, so drt retried what the driver had already given up on. You rewrote a lot of these connectors in #723, so if the fix looks wrong at the edges you would spot it faster than I would.
- **#862** — the generalisation of *why* #861 was invisible: the test double had `RequestError` under the wrong base, so the bug was structurally untestable and the suite stayed green. It pins every driver double against the real package. That one is more about whether the approach is right than whether the code is.

Either of those is self-contained, ~15 minutes, and genuinely useful. The streaming stack I would rather you skipped — it is measurement-heavy and I verified each leg against a live server (including the Snowflake/Databricks smoke run in #865), so a review would mostly be re-reading numbers.

Also: you have #851 open, and you have been carrying drt-web almost single-handed this week — #38, #39, #40, plus welcoming @laurentketterle-hub on #41, which was a genuinely nice piece of maintainership. That is already more than a fair share. If the honest answer is "not this week", that is completely fine and I would rather you said it.

masukai:
@Muawiya-contact — correction to what I wrote a few hours ago, and an apology.

I told you #861 and #862 were the two worth your time. Then I merged **#861 twelve minutes later**, before you had any realistic chance to look at it. That is the exact thing I asked you *not* to feel pressured by, done by me, in the same thread. Sorry — asking for a review and then acting before it arrives is worse than not asking.

What happened, plainly: @masukai gave me the go-ahead to work through the merge queue, and I did that without re-reading what I had just written to you twenty minutes earlier. No excuse beyond that.

**Where things stand now.** #861, #860 and #863 are merged. I have stopped there. Still open:

- **#862** — the one I actually most wanted your opinion on, and the one still open. It is the generalisation: every optional-driver test double is hand-written, and #861's had `RequestError` under the wrong base, which made a real bug structurally untestable while the suite stayed green. #862 pins each double against the real package. The question is whether that approach is right, not whether the code works — and you are the person here who has written against the most of these drivers.
- **#864 / #865 / #866** — the rest of the streaming stack. Still the ones I would rather you skipped; nothing has changed there.
- **#858** — your `tighten_to` finding. I read your 👍 on the rebase note as "fine to proceed", so I plan to merge it last unless you say otherwise. If that 👍 meant something narrower, say so and I will hold it.

**#862 is genuinely paused for you.** No deadline, and "not this week" is still a completely fine answer — you have #851 open and you have been carrying drt-web on your own. But this time I will actually wait rather than merging out from under you.

PR #869, opened 2026-07-31
docs(adr): rescope the #769 gate to Tier 3 and mark it cleared (refs #786)

Pawansingh3889:
## What does this PR do?

Amends ADR 0004 to rescope the **#769 gate to Tier 3 only** and mark it **cleared**. Docs-only — one file, no code.

As first written, that gate row ran two distinct harms together:

1. per-destination pacing being unavailable at all, and
2. a per-process bucket resetting every run.

[#858](https://github.com/drt-hub/drt/pull/858) fixes the first completely and the second **only within a process**. So the gate resolves by tier rather than yes/no:

- **Tier 3 — cleared.** `drt serve` is one long-lived process, so a process-wide registry (`_limiter_registry`, `drt/destinations/rate_limiter.py:162`) is the whole scope there, and the bucket outlives individual runs. Tier 3's remaining blocker becomes **#854**.
- **Tier 2 — not cleared.** A Dagster sensor yields one `RunRequest` per changed sync and Dagster launches each as its own process, so N changed syncs against one endpoint remain N buckets.

The Tier 2 residual needs shared state — i.e. **#756** — so it is folded into that row rather than left standing as a gate that cannot be closed on its own terms. **Net effect on ordering: none**; #756 already blocked Tier 2 and remains the longer pole.

Also records one topology the fix does not cover: several `drt serve` replicas behind a load balancer are several processes, so the shared bucket degrades to one per replica. That is a hosting concern, and sits on the drt Cloud side of the open-core line the ADR already draws.

## Related Issue

Refs #786 (ADR issue, closed). The gate this amends is #769, closed by #858.

Gate/blocker states asserted by the ADR, verified against GitHub at time of writing: #756 open, #769 closed, #854 open, #855 open, #856 open, #858 merged.

## Notes for review

This branch was written before #858 merged, so its central claim was a prediction at authoring time. It has since been checked against what actually landed: #858's description confirms both named requirements shipped (per-destination `rate_limit` override; process-wide registry keyed per endpoint, shared across threads).

Rebasing onto current `main` moved one citation — [#868](https://github.com/drt-hub/drt/pull/868) shifted `deltalake.py`'s `version()` call from `:67` to `:91`. That is refreshed here, and the verification note now states that line numbers track `main` rather than the v0.8.3 tag. All other cited lines (`server.py:15/28/61/90/100`, `state/manager.py:43`, `iceberg.py:51-52`, `dagster_drt/__init__.py:6`) are unchanged since the tag and were re-verified against the merge target.

## Checklist

- [x] Tests pass (`make test`) — docs-only; no code touched. Full CI matrix runs on this PR.
- [x] Linter passes (`make lint`) — docs-only; no Python changed.
- [x] Updated `CHANGELOG.md` — n/a. No change under `drt/` or `pyproject.toml`, so `check_changelog_required.py` reports no code change. An ADR amendment ships no user-facing behaviour.

### DIFF
--- a/docs/adr/0004-streaming-and-event-triggered-syncs.md
+++ b/docs/adr/0004-streaming-and-event-triggered-syncs.md
@@ -8,6 +8,10 @@
   recommendation stands unchanged. Accepted on that evidence rather than on
   having been asserted first — the matrix checked both conditions explicitly
   and found neither met.
+- **Amended:** 2026-07-29 — the #769 gate is rescoped to Tier 3 only and marked
+  cleared; its cross-process residual folds into the #756 row. The decision and
+  the tiers are unchanged; only the gate table and the ordering it implies move.
+  See [Gates](#gates-two-prerequisites-block-promotion-not-authorship).
 - **Issue:** [#786](https://github.com/drt-hub/drt/issues/786)
 - **Implementation:** none — this ADR recommends **not** building a native
   watcher. The work it does sanction is listed under
@@ -29,7 +33,9 @@ that. It is deliberately a build/no-build decision, not a design for a daemon.
 Three facts about the code as it stands (v0.8.3) constrain the answer more than
 the competitive framing does. Every citation below was re-verified against the
 v0.8.3 tree; none of that release's changes touch `serve`, state locality, or
-the Dagster integration.
+the Dagster integration. Line numbers track `main` rather than the tag: nothing
+load-bearing has moved since, but post-v0.8.3 work shifted one citation
+(`deltalake.py`, [#868](https://github.com/drt-hub/drt/pull/868)).
 
 **`drt serve` is a trigger endpoint, not a trigger runtime.** It exists
 (#218) and works for its designed cadence — a dbt job finishes, POST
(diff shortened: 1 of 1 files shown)


masukai (approved):
Approving. This is the amendment you proposed on #858, and it holds up to checking.

I verified the claims against live data rather than reading them:

```
#756 OPEN   #769 CLOSED   #854 OPEN   #855 OPEN   #856 OPEN   #858 MERGED
deltalake.py version()  → :91   (was :67)
manager.py:43           → self._state_file = ... "state.json"
iceberg.py:51-52        → load_catalog / load_table(...).scan().to_arrow()
rate_limiter.py:162     → _limiter_registry
```

Every gate state the table asserts is true today, and every citation resolves.

**The `:67 → :91` catch is the part I want to name.** That line moved because of my own #868, and you found it by re-verifying against the merge target rather than trusting a branch written before it. An ADR whose citations rot is worse than one without citations — it reads as precise while pointing at the wrong code. Thank you for treating the rebase as a re-verification rather than a formality.

**The rescope is right, and for the reason you gave rather than because #858 shipped.** The original row bundled two harms that turned out to have different fixes: "no per-destination pacing at all" and "the bucket resets every run". A gate that can only ever be half-cleared isn't a gate, it's a standing note — so folding the cross-process residual into #756 rather than leaving a permanently-open second gate is the correct shape. That it changes no ordering is the tell that it's bookkeeping rather than a decision being quietly revisited.

Two things I'd have missed and am glad are recorded:

**The load-balancer topology.** Several `drt serve` replicas are several processes, so the shared bucket degrades to one per replica. Putting it on the Cloud side of the open-core line is consistent with the rest of the ADR, and stating it now means nobody later reads "#769 cleared for Tier 3" as a stronger guarantee than it is.

**"Line numbers track `main` rather than the tag."** Small, but it converts an unstated assumption into a checkable one. The next person to re-verify knows what to check against.

One thing worth flagging for whoever picks up #856: Tier 3 becoming publishable now hinges on #854, which is yours. No action here — just noting the guide's unblocking is now a dependency on your own open work rather than on rate limiting.

Merging.

Issue #679, opened 2026-06-28
perf(sources): predicate pushdown for Delta/Iceberg (avoid full-table materialize)


masukai:
Follow-up to #675 (Delta Lake + Iceberg sources), surfaced by its second-opinion review.

**Problem:** both sources currently read the *whole* table into Arrow (`DeltaTable.to_pyarrow_table()` / `table.scan().to_arrow()`) and then filter via DuckDB. So a model's incremental cursor `WHERE` (and column selection) does **not** reduce what's read — every run materializes the full table, and very large tables risk OOM.

**Fix:** push the filter + projection into the read:
- [ ] Delta: use DuckDB's `delta_scan('<location>')` (delta extension) so DuckDB does predicate/column pushdown, or use deltalake's `to_pyarrow_dataset()` + filters/columns.
- [ ] Iceberg: use DuckDB's `iceberg_scan(...)` or pyiceberg `scan(row_filter=…, selected_fields=…)`.
- [ ] Keep the same profile shape + the in-memory-DuckDB SQL contract.

Relevant to the large-table leg of #654 (live-warehouse validation will make the memory cost visible).

masukai:
Picked this up to do the Iceberg leg (#868 did Delta only, deliberately). Result: it's not actionable as scoped, for a correctness reason rather than a missing-feature one — writing this up rather than silently shipping something wrong.

**Both options in the issue checklist are foreclosed:**

1. **`iceberg_scan(...)`** — DuckDB's iceberg extension isn't bundled; it's fetched from DuckDB's extension repository on first use. That's the exact implicit-network-call-at-sync-time problem `delta_scan()` was already rejected for in #868 (see `drt/sources/deltalake.py`'s comment and `docs/research/extraction-memory.md`). Same call applies here for consistency.

2. **`pyiceberg` `scan(row_filter=…, selected_fields=…)`** — this needs the filter/columns known *before* the scan runs. drt registers the table in DuckDB and lets the user's arbitrary model SQL supply the `WHERE`/`SELECT` afterward, so there's nothing to hand `row_filter`/`selected_fields` — getting one would mean parsing arbitrary SQL, which is its own fragile feature, not a two-line pushdown.

**The fallback I checked instead — `to_arrow_batch_reader()` for at least memory streaming — is actively unsafe, not just unhelpful.** It's single-pass. Verified directly:

```python
conn.register('t', reader)
conn.execute('SELECT * FROM t').fetchall()   # [(0,), (1,), (2,)]
[...]

Issue #280, opened 2026-04-13
feat: benchmark suite for sync performance profiling


masukai:
## Summary

Create a benchmark suite that measures sync performance across different source/destination combinations and data volumes.

## Motivation

Two purposes:
1. **Production readiness**: users need to know expected throughput
2. **Rust migration**: provides baseline measurements to identify actual bottlenecks (I/O vs CPU) before investing in Rust rewrite

## Scope

- Benchmark framework using pytest-benchmark or custom harness
- Scenarios: small (100 rows), medium (10K), large (100K)
- Measure: total duration, rows/sec, memory peak, API calls
- CI integration: run benchmarks on release branches, track regressions
- Output: JSON results for comparison across versions

masukai:
Closed via #999 (merged). `make benchmark` — fixed 100/10,000/100,000-row local SQLite → real engine → file-destination scenarios, versioned JSON results (wall time, throughput, peak memory, destination-call count). Codex review caught and fixed a real batch-truncation bug in the harness plus OTel measurement contamination. Local/manual tooling only — CI automation deferred until there's measured need. #301 builds directly on this harness.

PR #873, opened 2026-07-31
docs: make the #765 extraction-memory figures a single source of truth

Pawansingh3889:
## What does this PR do?

The #765 memory figures were written into **five places independently** — source docstrings, `docs/connectors/*.md`, `docs/llm/API_REFERENCE.md`, `CHANGELOG.md`, and the [closing comment on #765](https://github.com/drt-hub/drt/issues/765#issuecomment-5132427067). Nothing generates them, so they drifted apart.

This adds `docs/research/extraction-memory.md` as the single table (following the `warehouse-trigger-matrix.md` convention) and reduces every other surface to the *shape* of the improvement plus a link. One edit next time instead of five.

## What was actually wrong

| Source | docstring | connector doc | API_REFERENCE | CHANGELOG | #765 comment |
|---|---|---|---|---|---|
| Postgres | +182 → **+16** | +182 → **+18** | +182 → **+18** | +182 → **+18** | **+196** → +16 |
| MySQL | +109 → **+0.8** | — | — | +109 → **+8** | **+112** → **+3** |
| SQL Server | **+110.8 → +4.3** | +151 → +39 | — | +151 → +39 | +151 → +39 |
| ClickHouse | +224 → +149 | — | — | +224 → +149 | +224 → +149 ✅ |
| DuckDB | +150.9 → +42.2 | +150.9 → +42.2 | — | +150.9 → +42.2 | +151 → +42 ✅ |
| SQLite | +110.6 → +4.4 | +110.6 → +4.4 | — | +110.6 → +4.4 | +111 → +4 ✅ |

**Two are provably wrong without re-measuring anything:**

1. **`sqlserver.py`** claimed `+110.8 → +4.3`. Its own connector doc, the CHANGELOG and the issue all say `+151 → +39`. The discarded pair sits within **0.2 MB of SQLite's** (`+110.6 → +4.4`) — implausible as an independent measurement of a different engine on a different driver, and #866 landed SQL Server, DuckDB and SQLite in one PR.

2. **`mysql.py` contradicted itself** — `+0.8 MB` at line 84, *"iterating peaked at +3.6 MB"* at line 104, describing the same `SSCursor` path twenty lines apart.

## How the rest was resolved

**A fresh-process re-measurement supersedes an in-process one.** The #765 comment documents the methodology correction: `ru_maxrss` is a process-wide high-water mark, so measuring the buffered variant and then the streaming one in a single process makes the second look free. Where the comment genuinely differs it wins; where it is only a rounding of a repo figure (DuckDB, SQLite) the repo's decimals stand.

Two independent checks agree, which is why this isn't a coin-toss:

- **MySQL** — the undisputed `+3.6` at line 104 matches the comment's `+3`, not `+0.8`. CHANGELOG's `+8` reads as a lost decimal point from `+0.8`.
- **Postgres** — CHANGELOG's own prose says *"the Postgres (12x) ... leg"*. 196/16 = 12.25. The `182/18` it quotes in the same sentence is 10x. The multiplier corroborates the figures it wasn't written next to.

## Deliberately kept inline

Two numbers are *arguments*, not benchmarks, and stay next to the code they justify (each now pointing at the research page): MySQL's `+3.6` vs `+61.9` comparison (why there is no `fetch_size` knob) and Delta's `delta_scan()` `+19 MB` note (why that dependency was rejected).

## Honest limitation

These numbers are now **consistent and attributed, not independently re-verified.** No harness is committed that can regenerate them — `docker-compose.yml` defines only the `drt` app service, so the Postgres/MySQL/ClickHouse/SQL Server instances behind these runs were ad hoc. That is **#280**, and is out of scope here. The research page says exactly this rather than implying the digits are a regression gate.

## Related Issue

Refs #765, #679. Groundwork for #280.

## Checklist

- [x] Tests pass (`make test`) — **2721 passed, 18 skipped**, run with the full CI extras set (`dev,mcp,duckdb,postgres,mysql,clickhouse,sqlserver`).
- [x] Linter passes (`make lint`) — `ruff check` clean; the 7 touched Python files report *already formatted*. The one `mypy` error (`drt/config/credentials.py:175`) is **pre-existing on `main`** and in a file this PR does not touch — verified by re-running against a clean tree.
- [x] Updated `CHANGELOG.md` — `[Unreleased]` Postgres/MySQL figures corrected onto the reconciled set and linked to the research page. Released sections untouched.

Changes are docstrings, comments and markdown only — no executable line moves.


masukai (approved):
Approving, and thank you — the drift is real and it is mine. I wrote those figures into five surfaces by hand across six PRs and never once checked them against each other.

**One correction, on information you couldn't have had.**

You flagged `sqlserver.py`'s `+110.8 → +4.3` as provably wrong, reasoning that it sits within 0.2 MB of SQLite's `+110.6 → +4.4` and is therefore implausible as an independent measurement. That inference is sound, and the conclusion happens to be wrong: **both pairs are real measurements of different things.**

```
sqlserver, pymssql directly : +110.8 → +4.3 MB
sqlserver, through extract(): +151.2 → +39.3 MB
```

I measured the driver in isolation first, then re-measured through `extract()` once the implementation existed — and then put the *first* pair in the docstring and the *second* everywhere else, without noting which was which. The near-match with SQLite is a coincidence of both being ~200B × 300k rows through a Python driver, not a copy-paste.

So the defect is not a wrong number, it is **two numbers measuring different boundaries with nothing saying so** — which is worse, because both survive scrutiny individually.

**Your fix is right anyway, and for a better reason than the one in the description.** The page standardises on "measured through `extract()`, so the number includes drt's own per-row dict construction". That is the figure a user can act on — nobody runs pymssql directly. Fixing the methodology line settles the discrepancy properly rather than by discarding a datum. Worth amending the Corrections section to say the driver-level pair was *superseded* rather than *wrong*, so a future reader doesn't go looking for a measurement error that isn't there.

**The `ru_maxrss` note is the most valuable paragraph here.** I hit exactly that trap on ClickHouse and reported `+0.0 MB` before catching it — a figure that would have been extraordinary if true. Having the reason written down where the numbers live, rather than buried in a PR comment, is what stops the next person repeating it.

**On Postgres `+182` vs `+196`:** the issue comment's `+196` is the re-measured value and the `+182` scattered elsewhere is the original. Same story — the later number is the one taken with the corrected methodology. Your table has `+196`, which is correct.

The single-source-plus-link approach matches `warehouse-trigger-matrix.md`, so this is now the established pattern for measured figures rather than a one-off.

Merging. The Corrections wording is worth a follow-up but not a blocker — I'll do it rather than send this back.

Issue #838, opened 2026-07-25
fix(cli): drt build JSON lacks the warnings[] array that drt test emits — asymmetric public contract


masukai:
Follow-up from the #830 review (#779). Non-blocking for that PR.

## Problem

#830 gives `drt test --output json` a top-level `warnings` array collecting every `severity: warn` failure. `drt build` runs the same tests through the same `execute_tests_for_sync` seam and honours the same severity rule — but its JSON has **no `warnings` array**.

So the same failing test is machine-discoverable from one command and not the other:

```
drt test  --output json   ->  .warnings[]        ✅
drt build --output json   ->  (walk .results[].tests[] and filter yourself)
```

A CI pipeline that runs `drt build` (the common case — sync then validate) has to reimplement `_collect_warnings` in `jq` to get what `drt test` hands it directly. That asymmetry isn't intentional design, it's just where the feature stopped.

## Why it's worth fixing

`severity: warn` exists precisely so CI can surface degradation *without* failing the build. `drt build` is the command most likely to be the CI entrypoint, so it's the one where a warn-severity signal most needs to be easy to find. Right now it's the harder of the two.

## Suggested fix

`_collect_warnings()` is already a standalone function taking `list[_SyncTestResult]`. `drt build` accumulates the same per-sync result dicts, so this should be close to reusing it as-is rather than writing anything new.

Coordinate with #837 — if the `value`/`error` split lands there, `drt build` should adopt the corrected shape rather than copying the current one and inheriting the same problem.

## Acceptance criteria

- [ ] `drt build --output json` emits `warnings[]` with the same shape as `drt test`
- [ ] Shape stays consistent with whatever #837 settles on
- [ ] Test asserting a `severity: warn` failure appears in `drt build`'s JSON without affecting its exit code

PR #874, opened 2026-07-31
docs(claude): record the post-v0.8.3 work on main

Pawansingh3889:
## What does this PR do?

`CLAUDE.md`'s status section stopped at **v0.8.3**, so an agent reading this repo today has no idea the extract path was rewritten underneath it — nine sources switched to streaming, seven gained retry, and the destination rate limiter became process-wide and shareable.

Adds one `Unreleased (on main, since v0.8.3)` bullet at the top of `## Current Status`. One file, one line.

## What it records

The nine issues closed since the v0.8.3 tag:

| | |
|---|---|
| [#765](https://github.com/drt-hub/drt/issues/765) | streaming extraction, all nine SQL sources |
| [#766](https://github.com/drt-hub/drt/issues/766) | source-side retry — REST API + seven SQL sources |
| [#769](https://github.com/drt-hub/drt/issues/769) | rate limiting v2 — per-destination override + process-wide registry |
| [#786](https://github.com/drt-hub/drt/issues/786) | ADR 0004 — the three-tier posture and the no-watcher decision |
| [#834](https://github.com/drt-hub/drt/issues/834)–[#838](https://github.com/drt-hub/drt/issues/838) | the `drt test` / `drt build` JSON + test-coverage tranche |

Plus the two that landed **without** closing their issue: the Delta leg of [#679](https://github.com/drt-hub/drt/issues/679) (Iceberg deliberately still materialising) and the Databricks 401/403 no-retry fix.

## Two judgment calls

**Behaviour changes are carried, not just features.** Streaming sources now hold the connection open for the whole load, and SQL Server mid-stream failures are no longer retried. Those are what would otherwise surprise an agent, so they are in the entry rather than left to the CHANGELOG.

**#723 and #844 are deliberately excluded.** Both closed *after* the v0.8.3 tag timestamp (04:53 and 03:50 UTC against the tag at 04:15 UTC), so a naive "closed since the tag" query picks them up — but their code shipped *in* v0.8.3 and the existing v0.8.3 bullet already credits both. Listing them would double-count.

## Verification

Every claim was checked against the source tree rather than against PR titles or CHANGELOG prose:

- `DEFAULT_FETCH_SIZE = 10000` (`drt/config/profiles.py:26`)
- Exactly ClickHouse / MySQL / Databricks lack the `fetch_size` field — 6 have it, 3 don't, 9 total
- `with_retry` appears in 7 SQL sources plus `rest_api.py`, matching "the REST API source and seven SQL sources"
- No `fetchall()` remains in any of the nine SQL sources; `iceberg.py:59` still has one, matching the claim that Iceberg stays materialising
- The limiter registry is module-level (`drt/destinations/rate_limiter.py:162`), supporting "process-wide"

All thirteen issue references were re-checked against live GitHub state immediately before pushing.

## Note on scope

No memory figures are quoted here — those are being reconciled separately in #873, and this entry is deliberately independent of it. A cross-link to `docs/research/extraction-memory.md` can follow once that lands, rather than being a dangling reference now.

## Related Issue

Documentation of already-merged work; closes nothing.

## Checklist

- [x] Tests pass (`make test`) — no code touched; `CLAUDE.md` only.
- [x] Linter passes (`make lint`) — no Python changed.
- [x] Updated `CHANGELOG.md` — n/a. `check_changelog_required.py` reports *no code changes in origin/main..HEAD*; `CLAUDE.md` is agent context, not a user-facing behaviour change.


masukai (approved):
Approving. Verified every state it asserts:

```
#765 #766 #769 #786 #834 #835 #836 #837 #838  → all CLOSED
#679                                          → OPEN  (correct: Delta leg only)
iceberg.py to_arrow()                          → still present
```

The `#679 OPEN` / "Iceberg deliberately still materialising" pair is the detail I'd most expect to get written down wrong, and it's right. Same for framing the Databricks 401/403 fix as landing without closing its issue.

**Why this is worth more than its diff suggests.** `CLAUDE.md` is what an agent reads to orient before touching anything, and it had stopped at v0.8.3 while nine issues closed underneath it. An agent starting today would have assumed `fetchall()` semantics, no source-side retry, and per-process rate limiting — then written code against a shape that no longer exists. That is a quieter failure than a stale README, because nothing surfaces the mismatch until the work is already wrong.

Two things I'd have been tempted to leave out and shouldn't have:

- **The two ⚠️ behaviour changes** (connections held for the whole load; SQL Server mid-stream failures no longer retried). Both are the kind of thing that reads as a bug later if nobody wrote down that it was chosen.
- **The three "no `fetch_size`" reasons stated separately.** They look like one decision and are three unrelated ones — MySQL's cursor already streams, Databricks exposes no `arraysize`, ClickHouse's residual is inside the HTTP client. Collapsing them would invite someone to "fix the inconsistency".

Merging.

Issue #871, opened 2026-07-31
ci: check_drift.sh doesn't cover CLI options — flags reach docs but silently miss MCP


masukai:
Root cause behind #870, and the reason it went unnoticed for three releases.

## What the check covers today

`scripts/check_drift.sh` audits eight things — all of them about **connectors** and **MCP tool existence**:

1–5. every destination/source has docs, appears in the README tables, and is mentioned in the `/drt-create-sync` and `/drt-init` skills
6–8. every MCP tool is in `server.py`'s docstring, the README table, and the `drt_list_connectors` inventory

It has caught real drift, exactly as designed (the header records `/drt-debug` going three months without `drt doctor`).

## What it does not cover

**Nothing checks that a CLI command or option reached any user-facing surface.** Adding a connector is caught; adding a flag is not.

Measured, for the v0.8.0 flag wave:

```
feature        skills  docs   mcp
--fail-fast      3      3      0
--limit          3      8      0
match_policy     1      5      0
on_degraded      1      2      0
```

docs and skills keep up because writing them is part of the feature work. MCP does not, because nothing fails when it is skipped.

## Proposal

Extend the script with a check along the lines of:

> every behaviour-changing option on a CLI command that has a corresponding MCP tool is a parameter of that tool

Sketch: enumerate options from the Click tree (the #40 CLI-reference generator in drt-web already walks it — same technique), map command → tool, and diff against the tool signature.

**It needs an explicit exclusion list**, because some options legitimately do not belong in MCP:

- `--select` / `--exclude` — expressed by the `sync_name` argument
- `--output` / `--log-format` / `--verbose` / `--quiet` — presentation; MCP returns structured data

Better as a named list in the script than a heuristic, so each exclusion is a decision someone wrote down rather than a silent pass.

## Also worth adding

A **release-time** check. Per @masukai: the maintainer-facing release skill should verify that what shipped is reachable from the user-facing surfaces (skills / MCP / docs) before the tag, not after. `check_drift.sh` is state-based so it can run at release as-is — it just needs to be *in* the release checklist, and to cover options for that to be meaningful.

Related: #870 (the flags themselves).

masukai:
Correcting my own framing above, because I got the diagnosis wrong in a way that changes the fix.

I wrote that a release-time check "needs to be *in* the release checklist". **It already is.** `.claude/commands/drt-release-check.md`, Phase 0, item 12:

[...]

Issue #870, opened 2026-07-31
feat(mcp): behaviour-changing CLI flags are missing from the MCP tools


masukai:
Found by auditing every CLI command and option against the MCP tool surface, prompted by @masukai asking whether shipped features actually reach users through MCP and skills. For `run` / `test` / `validate`, they largely do not.

## The gap

MCP tools expose a small, older slice of each command. Flags that **change what a run does** are absent:

| CLI | missing from MCP | why it matters |
|---|---|---|
| `drt run` | `--limit` | sampled first send — the safe way to try a sync |
| | `--fail-fast` | stop scheduling after the first failure |
| | `--failed` | re-run only what failed last time |
| | `--vars` | project var overrides |
| | `--threads` | parallelism |
| `drt test` | `--store-failures`, `--fail-fast` | |
| `drt validate` | `--check-connection` | validate *and* prove the profile connects |
| | `--strict` | warnings become errors |

`drt_run_sync` currently takes `dry_run`, `compute_diff`, `diff_limit`, `cursor_value`, `profile_name` — the #775/#774/#773/#783 wave (v0.8.0) never reached it.

**Not gaps** (checked, listing so nobody "fixes" them): `--select`/`--exclude` are expressed by the `sync_name` argument; `--output`/`--log-format`/`--verbose`/`--quiet` are meaningless when the tool returns structured data.

## Why it matters beyond convenience

An agent driving drt through MCP cannot do the safe thing. `--limit` exists precisely so a first send is sampled rather than firing every row at a production CRM; without it the MCP path only offers all-or-nothing. Same for `--check-connection`: an agent can validate config but cannot verify the credential actually works.

## Root cause — this will happen again

`scripts/check_drift.sh` audits **connectors** and **MCP tool existence**, but nothing checks that a CLI *option* reached MCP. So adding a connector is caught, and adding a flag is not. That is why the v0.8.0 flag wave went unnoticed for three releases.

Fixing the flags without fixing the check just resets the clock. Tracked separately in the drift-check issue.

## Scope

**In:** the behaviour-changing flags above, on the existing tools.
**Out:** new tools for `init` / `deploy` / `serve` / `clean` — those scaffold files or run daemons and are deliberately not MCP surfaces.

Issue #776, opened 2026-07-10
feat(cli): --full-refresh + drt state show/reset — sanctioned watermark & tracked-mirror reset


masukai:
## Problem

Incremental/stateful syncs accumulate three kinds of durable state — watermarks (`.drt/watermarks.json` or gcs/bigquery), run state (`state.json`), tracked-mirror key tables (`_drt_synced_keys` in the destination) — and drt offers no sanctioned way to say "start this sync over". Today's reality when a mart is rebuilt or a cursor is poisoned (e.g. the #475 tz bug era):

- users hand-edit/delete JSON files (breaks with remote watermark backends; does nothing for tracked-mirror tables)
- `--cursor-value` overrides *one run's* predicate but leaves stored state as-is — it's a probe, not a reset

dbt/dlt users specifically look for `--full-refresh` and find nothing (it's a FAQ-shaped gap).

## Prior art

- **dbt `--full-refresh`**: rebuild incremental models from scratch — universally understood flag.
- **dlt `refresh` modes**: `drop_sources` / `drop_resources` / `drop_data` — graduated state-reset semantics.

## Proposal

Two complementary surfaces:

```bash
# 1. one-shot: ignore + reset watermark for this run, then persist the new one
drt run --select users_sync --full-refresh

# 2. inspectable state management (works for local AND remote backends)
drt state show [sync]            # watermark, last status, tracked-mirror baseline info
drt state reset <sync> [--watermark] [--tracked-mirror] [--dry-run]
```

- `--full-refresh` on a tracked-mirror sync re-baselines `_drt_synced_keys` (documented loudly: next mirror pass treats current destination as baseline — same semantics as first run / lost-state re-baseline in #686)
- `drt state reset` refuses without `--yes` in non-TTY (CI) contexts

## Scope

**In:** flag + `drt state` subcommand over WatermarkStorage/StateManager/tracked-mirror reset hooks, remote-backend support, guards + confirmations, docs.
**Out:** destination data deletion (never), DLQ purge (`drt retry --clear` exists), history pruning (retention config exists).

## Related

- #390/#391 (`--cursor-value` — the probe this complements), #686 (tracked-mirror re-baseline semantics), #47 (corrupted state — `state reset` is the recovery path)

## Milestone fit

Candidate: v0.8.x — ops hygiene that every incremental user eventually needs; small-medium.

PR #875, opened 2026-07-31
docs(skills): add the task-relevant commands three skills were missing (closes #872)

### LINKED ISSUE #872: docs(skills): task-relevant commands missing from three skills
Third finding from the surface audit (see #870, #871). Smallest of the three, and the one closest to users.

## Method

A raw "does every command appear in some skill" grep says 15 of 33 commands are missing — but that is the wrong bar. The skills are **task-oriented** (create-sync, debug, init, migrate, troubleshoot), not a command reference, so `drt cloud push` has no business in any of them.

The right question is whether each skill mentions the commands **its own task needs**. Audited that way:

| skill | missing |
|---|---|
| `drt-init` | `drt profile add`, `drt profile test` |
| `drt-create-sync` | `drt test`, `drt list` |
| `drt-migrate` | `drt init`, `--dry-run` |
| `drt-debug` | — none |
| `drt-troubleshoot` | — none |

Six real gaps, not fifteen.

## Why these specific ones matter

- **`drt-init` without `drt profile add`/`profile test`** — the skill walks a user through setting up a project but not through creating and verifying the warehouse credential, which is the step most likely to fail and the reason `drt profile` exists.
- **`drt-create-sync` without `--dry-run` follow-through and `drt test`** — the skill produces a sync YAML and stops short of "now preview it before it writes to production", which is drt's main safety affordance.
- **`drt-migrate` without `--dry-run`** — most acute here. Someone moving off Census/Hightouch has live syncs; the first drt run should be a preview, and the skill does not say so.

## Note

`drt-debug` and `drt-troubleshoot` are complete against their tasks — worth stating, since they are the two that were explicitly repaired before (the `/drt-debug` + `drt doctor` drift in `check_drift.sh`'s header). The fix held.

## Scope

Editing three `SKILL.md` files. No code. Should land with, or before, whatever adds `--full-refresh` (#776), so the new flag arrives in the skills rather than being another retrofit.

Pawansingh3889:
## What does this PR do?

Closes the six gaps from #872's surface audit. Three `SKILL.md` files, no code.

I kept the issue's framing: the skills are **task-oriented**, so the bar is whether each names the commands *its own task* needs — not whether every CLI command appears somewhere. `drt-debug` and `drt-troubleshoot` are unchanged, as the issue found them complete.

| Skill | Added | Why |
|---|---|---|
| `drt-init` | `drt profile add`, `drt profile test` (+ `profile list` / `show`) | The skill set up a project but never created or verified the warehouse credential |
| `drt-create-sync` | `drt list`, `drt test` (+ `drt build`, `--limit 10`) | Discovery is glob-based; a misplaced file validates cleanly then silently never runs |
| `drt-migrate` | `drt init`, a first-run `--dry-run` step | Migrating users have live syncs and a populated destination |

## The three judgement calls

**`drt profile test` goes first in `drt-init`'s validation block, not last.** `drt doctor` and `drt validate` inspect configuration; only `profile test` proves the warehouse accepts the credential. The skill now says that explicitly, so a bad password surfaces there rather than mid-run.

**`drt list` earns its line in `drt-create-sync`.** Sync discovery is glob-based, so a file saved outside `syncs/` or with a mismatched `name:` passes `drt validate` and then never runs — a silent failure the skill previously walked users straight past.

**`drt-migrate` gets the most.** It's the sharpest case: the old tool is still running, the destination has live records, and a mismapped `sync.mode` or `upsert_key` writes to production on the first invocation. Added a scaffold step and a preview step, and stated the sampling boundary — `mirror` and `replace` delete or truncate, and `--limit` is **refused for both** ([`drt/cli/commands/run.py:599`](https://github.com/drt-hub/drt/blob/main/drt/cli/commands/run.py#L599)), so those two go straight from `--dry-run --diff` to a full run with no sampled rung in between. That asymmetry is exactly the kind of thing a migrating user should not discover by hitting it.

## Verification

Every command and flag written here was checked against `--help` and the CLI source rather than copied from existing skill prose:

- `drt profile test {name}` / `profile add {name}` — both take a required `name` argument
- `drt test --select` and `drt build --select` — both exist
- `--limit` refusal — `run.py:599` reads *"--limit is not allowed for mode=mirror/replace syncs"*, and `drt run --help` confirms *"refused for mirror/replace syncs"*

Suite and checks:

- `pytest` — **2721 passed, 18 skipped**
- `ruff check` — clean
- `scripts/check_drift.sh` — **0 new drift items** (it audits skill↔connector coverage, so this confirms the edits didn't disturb the connector mentions it tracks)

## Note

Per #871, `check_drift.sh` covers connectors and MCP tool existence but not CLI options, so it cannot catch this class of gap — it passed both before and after these edits. That check is #871's scope and is untouched here.

## Related Issue

Closes #872. Third finding of the audit trio (#870, #871, #872).

## Checklist

- [x] Tests pass (`make test`) — 2721 passed, 18 skipped
- [x] Linter passes (`make lint`) — `ruff check` clean; no Python changed
- [x] Updated `CHANGELOG.md` — n/a. No change under `drt/` or `pyproject.toml`; `check_changelog_required.py` reports no code change. Skill text ships no user-facing behaviour change.

### DIFF
--- a/.claude/commands/drt-create-sync.md
+++ b/.claude/commands/drt-create-sync.md
@@ -17,11 +17,22 @@ Create a drt sync YAML configuration file for the user.
 
 3. Output the YAML in a code block and suggest where to save it: `syncs/<name>.yml`
 
-4. Show the command to validate and run it:
+4. Show the commands to check, preview, then run it:
    ```bash
-   drt validate
-   drt run --select <name> --dry-run
-   drt run --select <name>
+   drt validate                          # YAML schema check
+   drt list                              # confirm the new sync is discovered
+   drt run --select <name> --dry-run     # preview — no data written
+   drt run --select <name> --limit 10    # real send, capped at 10 rows
+   drt run --select <name>               # full run
+   ```
+   `drt list` is worth the extra line: sync discovery is glob-based, so a file
+   saved outside `syncs/` or with a mismatched `name:` validates cleanly and then
+   silently never runs.
+
+5. If the sync declares a `tests:` block, show how to run it after the sync:
+   ```bash
+   drt test --select <name>    # post-sync validation (row counts, freshness, unique, custom SQL)
+   drt build --select <name>   # run + test in one pass
    ```
 
 ## Rules
--- a/.claude/commands/drt-init.md
+++ b/.claude/commands/drt-init.md
@@ -55,12 +55,24 @@ Guide the user through initializing a new drt project.
    - **DuckDB / SQLite**: nothing — file path only
    - **REST API source**: set the bearer token env var the profile references (e.g. `export REST_API_TOKEN=...`)
 
-4. Validate the setup:
+   Both `drt init` flows already write a profile to `~/.drt/profiles.yml`. To add
+   another later (a second warehouse, a prod/dev split) without re-running init:
    ```bash
-   drt doctor       # env-level triage — catches missing env vars, malformed profile, etc.
-   drt validate     # YAML schema check
-   drt list         # confirm syncs are discovered
+   drt profile add <name>     # interactive — prompts for source type + connection fields
+   drt profile list           # names + source types already configured
+   drt profile show <name>    # inspect one, inline secrets masked
    ```
+
+4. Validate the setup — **credential first**, since that is the step most likely to fail:
+   ```bash
+   drt profile test <name>   # actually connects, using the source's own connection check
+   drt doctor                # env-level triage — catches missing env vars, malformed profile, etc.
+   drt validate              # YAML schema check
+   drt list                  # confirm syncs are discovered
+   ```
+   `drt doctor` and `drt validate` inspect configuration; only `drt profile test`
+   proves the warehouse accepts the credential. Run it before the first sync so a
+   bad password surfaces here rather than mid-run.
 
 5. Offer to create a first sync using the `/drt-create-sync` skill.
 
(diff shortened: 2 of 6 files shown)


masukai (approved):
Approving. All six gaps closed, verified against the audit rather than the description:

```
drt-init           欠落: — なし   (profile add, profile test)
drt-create-sync    欠落: — なし   (drt test, drt list)
drt-migrate        欠落: — なし   (drt init, --dry-run)
make check-skills  ✓ All skills in sync
```

**You synced both copies. I did not, on #776, twenty minutes ago** — CI caught `.claude/commands/` out of step with `skills/` and I had to push a fix. Worth saying out loud since it's the failure mode this repo makes easy: the duplication is invisible until the gate fires.

**On keeping the task-oriented framing** — thank you for holding that line. The raw grep said 15 of 33 commands were "missing" and that number was wrong in a way that would have produced worse skills: stuffing `drt cloud push` into a debugging guide to satisfy a checker makes the guide harder to use. Measuring against what each task needs is the only version of this audit worth acting on.

**The three judgement calls are all right, and two are better than what I'd have written:**

**`drt profile test` first, not last.** This is the one I'd have got wrong. `doctor` and `validate` inspect configuration; only `profile test` proves the credential is accepted. Ordering it first means a bad password fails in two seconds instead of mid-run, and saying *why* in the skill is what stops someone reordering it later.

**`drt list` in `drt-create-sync`.** Discovery is glob-based, so a misplaced file validates cleanly and then silently never runs — a failure with no error message anywhere. `drt list` is the only thing that surfaces it, and it isn't obvious that you'd need it.

**A first-run `--dry-run` in `drt-migrate`** is the one I flagged as most acute and I'll restate why: someone moving off Census or Hightouch has live syncs against a populated destination. Their first drt run is the highest-stakes one they will ever do, and the skill previously walked them straight into it.

No overlap with #776, which touches `drt-debug` and `drt-troubleshoot` — the two your audit found complete. Between them the five skills are now current.

Merging, and closing #872.

PR #876, opened 2026-07-31
feat(cli): --full-refresh + drt state show/reset (closes #776)

### LINKED ISSUE #776: feat(cli): --full-refresh + drt state show/reset — sanctioned watermark & tracked-mirror reset

masukai:
Closes #776. Seven commits, bottom-up: three state primitives, a shared confirmation helper, the two commands, then all four user-facing surfaces.

## What

drt keeps **three** kinds of durable state and offered no sanctioned way to clear any of them. Users hand-edited `.drt/*.json` — which does nothing when a remote watermark backend is configured, and nothing at all for `_drt_synced_keys`, which lives in the *destination*.

```bash
drt run --select users --full-refresh      # re-read everything, persist a fresh watermark
drt state show [sync]                      # stored watermark + last run
drt state reset users --watermark --runs   # levels combine explicitly
drt state reset users --tracked-mirror     # see the warning
```

MCP: `drt_state_show`, `drt_state_reset`, `full_refresh` on `drt_run_sync`.

## The design decision worth reviewing: no `--all`

@masukai asked whether dlt-style graduated levels were possible. They are, and the three state kinds map onto them — but dlt's levels start at `drop_data`, which deletes destination data. #776 puts that permanently out of scope, so the axis is different: here the levels differ by **how hard the damage is to notice**.

| level | clears | risk |
|---|---|---|
| `--watermark` | stored watermark | low — re-reads |
| `--runs` | run state + fallback cursor | low |
| `--tracked-mirror` | `_drt_synced_keys` in the destination | **high** |

Only the third is qualitatively different: after re-baselining, rows the *application* wrote count as drt's own and become deletion candidates on the next mirror pass — the exact scenario tracked mirror (#686) exists to prevent.

So there is no `--all`. Adding one would re-create precisely the failure dlt split its levels to avoid, and "reset all three" is rare enough that three commands is not a burden. **`--full-refresh` is watermark-only**, matching what dbt users mean by it.

## Two traps found while implementing

**1. The watermark lives in two places.** `engine/sync.py` resolves a cursor from `watermark_storage` first and from `StateManager.last_cursor_value` only when no storage is configured — an `elif`, verified in source. Clearing one alone makes the reset **silently ineffective** in the other configuration, and "no watermark backend" is the default. Mutation-tested: dropping the fallback clear fails a test.

**2. `isatty()` is the wrong check for "non-interactive".** My first confirmation helper gated on it and broke `test_remove_confirm_declined` — because a pipe carrying a real answer is not a TTY either, so it refused input that *was* supplied. `CliRunner(input="n")` and `echo n | drt ...` are both that shape. The helper now attempts the prompt and translates only `click.Abort`, which is precisely "nothing to read".

## Included: fixing the precedent

`drt profile remove` called `typer.confirm` directly. Measured before:

```
$ drt profile remove x < /dev/null
Remove profile 'x'? [y/N]: Aborted.
(exit 1)
```

It fails correctly, but names neither cause nor fix — and `[y/N]` suggests piping `y` would work, when the answer is `--yes`. Both commands now share one helper. @masukai chose to fix the precedent rather than copy it; folded in here so the helper and both call sites land together.

## Surfaces — the reason this PR is shaped this way

The #870 audit found the v0.8.0 flag wave (`--limit`, `--fail-fast`, `--failed`, `--vars`) reached docs and skills but **never MCP**, across three releases. So this ships to all four at once:

```
               skills  mcp  docs  readme
--full-refresh    2     2     2      -
state show        2     2     2      1
state reset       1     2     2      1
```

`check_drift.sh` caught both new MCP tools missing from the README table mid-task — working as designed. It would *not* have caught a missing flag, which is #871.

## Verification

2655 passed, ruff + mypy + drift clean.

Verified against the real CLI and a real MCP server, not only the test harness:

```
no level          → exit 2 + lists the levels
CI without --yes  → exit 1 + names --yes
--dry-run         → .drt/state.json unchanged
--runs            → ✓ cleared
MCP unqualified   → {"error": ...}, state intact
MCP runs=true     → {"reset": ["runs"]}, state cleared
```

Every guard mutation-tested: removing the level requirement, the sync-name scope on the state table, the existence check, the fallback clear, and `--full-refresh` itself all fail a test when broken.

## Note on one test correction

I first asserted the cursor was *absent* after `--full-refresh`. It isn't — a real run re-reads from the start and then persists a fresh watermark, which is the point. The test now asserts the *stale* value is gone, since that is what would have limited the re-read.


Pawansingh3889:
Not a review, and please ignore if you've already got this — I was watching this PR because #877 overlaps it, saw `codecov/patch` go red, and ran the numbers locally out of curiosity. Passing them on in case it saves you a round trip.

`73.68% of diff hit (target 80.00%)`, and it isn't the `--full-refresh` work — that part is fine. It's the two new files:

```
drt/mcp/tools/state.py       58%   missing 39, 96-103, 109-121
drt/cli/commands/state.py    65%   missing 48-66, 164-175, 191-192
drt/mcp/tools/run_sync.py    82%   missing 45, 65-70
drt/destinations/sql_base.py 100%
drt/destinations/_mirror_state.py 100%
```

All 2759 tests pass — it's purely the patch gate. The uncovered ranges look like error/edge paths rather than the main flows, so it's probably a handful of tests on `state show`/`state reset` to clear 80%.

Mildly pleasing that the gate is doing the exact job its config comment claims: *"what would have caught the `--fail-fast` deadlock at review time — the cancellation branch was unhit and nothing acted on it."*

One thing genuinely worth flagging, since it's an interaction rather than a nitpick: **#877 will ask this PR for one more line.** It adds a check that every CLI command is either mapped to an MCP tool or explicitly listed as deliberately unmapped, so `drt state show` and `drt state reset` will need an entry in `COMMAND_TO_TOOLS` or `UNMAPPED_COMMANDS`. Whichever of the two merges second picks up that line. Given you've added `drt/mcp/tools/state.py`, my guess is they map to it and it's a two-line addition, but it's your call and I didn't want it to arrive as a surprise red build.

Happy to send the classification as a patch to whichever branch merges second — or to just fix it in #877 after this lands, which is probably simpler. No rush from my side; #877 isn't blocked on it.

masukai:
Thanks for running the numbers — saved me a round trip. Fixed: added targeted tests for the uncovered `state show`/`state reset` edge paths (no-backend watermark fallback, declined-prompt, tracked-mirror warn/no-op branches) plus the `run_sync` `full_refresh`/`cursor_value` exclusivity case. `codecov/patch` is green now (73.68% → 90%).

On the #877 interaction: you called it — it does need a mapping, and it wasn't as clean as a two-line addition. Took it on myself since you said it wasn't blocking you. Found two real bugs in `check_cli_mcp_parity.py` beyond the missing mapping: `collect_mcp_parameters()` silently finds nothing for `state.py` (it's the first module with more than one tool function, and the lookup was an exact-name match), and the sanity-check canary had no gate on whether the command is visible in `cli` yet, so registering the mapping ahead of this PR merging would've red-built #877 on its own branch. Both fixed, verified in both merge orders. Full writeup + the fix commit are on the #877 thread — didn't want to duplicate it here.

Net: this PR is ready to merge as-is (no dependency on #877 landing first). #877 needs that fix incorporated before it merges after this one, which is Pawansingh's call on how to land — flagged there, not blocking here.

PR #877, opened 2026-07-31
ci(drift): check that CLI options reach their MCP tool (closes #871)

### LINKED ISSUE #871: ci: check_drift.sh doesn't cover CLI options — flags reach docs but silently miss MCP

Pawansingh3889:
## What does this PR do?

Adds **check 9** to the drift audit: every behaviour-changing CLI option must reach the MCP tool covering the same command. This is the root cause behind #870 — checks 1–8 audit *connectors* and *MCP tool existence*, so adding a connector was caught and adding a **flag** was not.

Per your follow-up comment, this is automation rather than another checklist line: item 12 already asks for it, and asking a human to enumerate "new CLI capabilities" from memory across a thirty-PR release is exactly the task that degrades.

## What it finds today

11 gaps — precisely #870's backlog:

```
run:      --limit  --fail-fast  --failed  --vars  --threads
test:     --dry-run  --fail-fast  --store-failures  --store-failures-limit
validate: --check-connection  --strict
```

**Two of those are not in #870's table**: `test --dry-run` and `test --store-failures-limit`. Found by enumeration rather than by hand, which is the argument for the check in one line.

They're recorded in `drift_baseline.txt` as tracked follow-ups — the automated form of item 12's *"or a tracked follow-up issue"* hatch — so the audit reports a true state today: the detection is new, the debt is not. Delete each line as #870 lands it; a test fails if the two files disagree.

## ⚠️ The trap this check nearly fell into

**Typer 0.27 vendors its own Click.** `TyperOption` inherits `typer._click.core.Parameter`, *not* the installed `click.Option`. So the obvious implementation:

```python
opts = [p for p in cmd.params if isinstance(p, click.Option)]   # matches NOTHING
```

...enumerates zero options for every command, finds zero gaps, and reports a clean audit **forever** while the gap it exists to catch grows. I hit this while writing it — the first working prototype printed `run: (no options)`.

Two defences:

1. Options are identified by duck-typing on `param_type_name`, never `isinstance`.
2. `_assert_enumeration_sane()` **exits 2 instead of 0** unless known-present options still enumerate and every mapped tool still resolves. A parity check that silently finds nothing is worse than no check.

Related: the script imports **no** `click` at all. Typer vendors it, so standalone Click is absent from a core install — importing it would work locally (dev extras pull it in) and fail in the workflow. Verified against a clean core-only venv.

## Two workflow fixes, without which this is decorative

- **`drift-check.yml` never installed `drt`.** Checks 1–8 are pure bash + grep; check 9 introspects the Typer tree and tool signatures, so it needs the package importable. Added a Python setup + core `pip install -e .` (no connector extras — it only reads signatures).
- **Its `paths:` filter omitted `drt/cli/**`.** A PR adding a flag wouldn't have triggered the post-merge audit at all; only the weekly cron would catch it, reintroducing the very lag this fixes. Added `drt/cli/**`, `drt/mcp/tools/**`, and the new script.

**Trigger design is unchanged** — post-merge + weekly, reporting to the tracking issue, never commenting on contributor PRs. I read that as deliberate from the header comment and did not touch it. Promoting this to a PR-blocking gate is a one-line `ci.yml` change if you'd rather have it fail at authorship time; that felt like your call, not mine.

## Judgement calls worth reviewing

**Renames are declared, not guessed** — `--diff`→`compute_diff`, `--sync`→`sync_name`, and `--no-state`→`include_state` (inverted; the capability is reachable, which is what parity means). Without these the check false-positives on three flags that are in fact present.

**Structural exclusions carry their reason inline** — presentation flags (`--output`, `--verbose`, `--quiet`, `--log-format`, `--format`) and selection flags (`--select`, `--exclude`) per your "not gaps" list, plus a few per-command ones (`status --history` is split into `drt_get_history`; `docs generate --inline` is HTML packaging).

**An unclassified command is itself reported.** A command that is neither in `COMMAND_TO_TOOLS` nor `UNMAPPED_COMMANDS` fails the check, so a new command forces a decision instead of contributing no gaps — #870's failure mode one level up. Note this means **#876's `drt state show/reset` will need one line** when it merges.

**`drt build` I classified as unmapped** (composed by calling `drt_run_sync` then `drt_run_test`). It wasn't in #870's scope list either way — easy to flip if you disagree.

## Verification

Every CI gate, in `ci.yml`'s order:

- `ruff check drt tests` — clean (also ran it over `scripts/`, which CI doesn't lint)
- `mypy drt` — clean, 172 files
- `make check-skills` — in sync
- `make check-changelog` — 37 versions covered
- `pytest` — **2736 passed, 18 skipped** (15 new)
- `bash scripts/check_drift.sh` — 0 new drift, 11 baselined, exit 0
- Also run against a **core-only venv** to prove the workflow's install is sufficient, and `drift-check.yml` parsed as YAML

The 15 new tests cover the guard firing on both failure modes, alias resolution, the unclassified-command path, and a consistency test binding the reported gaps to the baseline file.

## Related Issue

Closes #871. Root cause of #870; does not fix those gaps, only detects and records them.


masukai:
Thanks for the flag — and for running the numbers on #876's coverage too, that saved a round trip on my end (already fixed and pushed, 73.68% → 90%).

Took the mapping myself since you said "no rush from my side" and I wanted to verify the interaction against a real merged tree rather than guess at it. Two things turned up beyond the two-line mapping you predicted:

**1. `collect_mcp_parameters()` doesn't find `state.py`'s tools at all**, mapping or no mapping. It's `getattr(module, module_name)` — exact match. `state.py` is the first module with more than one tool function (`state_show`, `state_reset`, not `state`), so that lookup finds nothing and both tools silently vanish from the `mcp` dict regardless of what `COMMAND_TO_TOOLS` says. Widened it to collect every function named after the module or `f"{module}_"`-prefixed, matching how `server.py` already imports both.

**2. Registering the mapping ahead of the command landing broke the sanity guard on your branch alone.** `_assert_enumeration_sane()`'s canary loop checks every tool named anywhere in `COMMAND_TO_TOOLS`, unconditionally — no gate on whether the *command* is visible in `cli` yet. `find_gaps()` already has that gate (`cli.get(command, set())`); the canary didn't. So merging my two-line mapping into #877 as-is would red-build #877's own CI the moment it merges, independent of #776 — which is your exact "one more line, whichever merges second" concern, just relocated one level down. Applied the same gate to the canary.

Verified both merge orders for real rather than reasoning about it: on your branch alone (`drt state` doesn't exist), script exits 0 with only the known #870 backlog; merged with #776, all 16 tests pass. The new regression test `importorskip`s on `drt.mcp.tools.state`, so it's inert now and activates on its own once #776 lands — no follow-up commit needed either way you merge first.

Mutation-tested the `collect_mcp_parameters` fix by reverting to the exact-match version: 3 tests fail, including `test_guard_passes_on_the_real_surfaces` — your own sanity guard would have caught this in CI. Also re-ran `test_guard_rejects_a_renamed_mcp_tool` (deletes `run_sync` from `mcp`) to make sure the ordering gate doesn't quietly weaken real detection — it still fails as designed, since `run` stays in `cli` regardless.

Pushed the single commit to `pr-877-state-mapping` on `drt-hub/drt` (not your fork — I don't have write access there) rather than pasting a patch inline:

```
git fetch origin pr-877-state-mapping
git cherry-pick eb2d9a800e66b41cdfe60a97dda4afd480dcb122
```

Commit: eb2d9a800e66b41cdfe60a97dda4afd480dcb122 — `git show` on it has the full diff.

Genuinely your call on how to land it — cherry-pick into this branch, take it as a follow-up PR once #877 merges, or tell me to just open a normal PR against `main` myself if that's cleaner. No rush either way.

Pawansingh3889:
Took both commits — cherry-picked `5a95d39` and `eb2d9a8` onto a rebase against current `main`. Worth flagging for anyone reading later: the cherry-pick line in your comment names only `eb2d9a8`, but the `COMMAND_TO_TOOLS` entries live in its parent, so taking just the fix would have landed the ordering-tolerance without any actual `drt state` coverage.

Timing moved underneath this in the meantime — #776 merged in `40a0b4b`, so your `importorskip` on `drt.mcp.tools.state` is live rather than inert, and the merged-tree case you verified by hand is now just what CI runs. Confirmed the mapping resolves positively rather than trusting a passing check: `state_show -> {sync_name}` and `state_reset -> {dry_run, runs, sync_name, tracked_mirror, watermark}`, no gaps reported for either.

Also reproduced your mutation independently — reverting to the exact-match `getattr` fails exactly 3 tests including `test_guard_passes_on_the_real_surfaces`. Good catch that `state.py` was the first multi-tool module; the check would have gone green over a blind spot indefinitely.

Authorship on both commits is yours. Thanks for verifying against a real merged tree instead of guessing at it — that's what caught the canary gate problem.

masukai:
Confirmed, and it deserves more than a footnote.

`5a95d39` is the three-line `COMMAND_TO_TOOLS` addition (`state show` → `state_show`, `state reset` → `state_reset`); `eb2d9a8` is only the ordering-tolerance fix. Cherry-picking just the sha I named would have landed a check that can finally *enumerate* `drt state`'s tools and never actually *asks* about them — a green audit with `drt state` uncovered. That is precisely the failure mode this PR exists to prevent, nearly introduced by the comment introducing it. I wrote the cherry-pick line to match what the prose was about (the fix) rather than the range the work actually spanned.

Good catch reading the parent instead of trusting the instruction.

On the timing: agreed, #776 landing in `40a0b4b` makes the `importorskip` live, so the merged-tree case I checked by hand is now just what CI runs — the better arrangement. And thanks for resolving the mapping **positively** (`state_show -> {sync_name}`, `state_reset -> {dry_run, runs, sync_name, tracked_mirror, watermark}`) rather than reading "no gaps reported" as proof. Those two states are indistinguishable when the lookup silently finds nothing, which is how this whole thread started.

Likewise for reproducing the mutation independently instead of taking my word for the 3 failures.

CI is green across the matrix on the rebase. From my side this is ready — I'll do a final read of the merged state and come back on merging.

masukai (approved):
Did the final read I promised, on the rebased head rather than on the description of it.

Both commits are present on top of current `main` (`3ff8d36` mapping, `b5bc409` fix) — thanks for taking the parent as well as the one I named.

Ran the check itself: exits 0 with exactly the 11 known gaps from #870's backlog and nothing else. Then confirmed the two tools resolve **positively** rather than by absence, independently of your report:

```
state_show  -> ['sync_name']
state_reset -> ['dry_run', 'runs', 'sync_name', 'tracked_mirror', 'watermark']
```

Same values you got. Worth having checked separately given that "resolves to nothing" and "resolves and matches" both print as a clean run — which is the bug that started this thread.

One more thing I went looking for: `drt state reset --yes` does not appear as a gap, and I wanted to know whether that was a deliberate exclusion or another silent miss. It's explicit, with the reason written down:

> `--yes`: skips a TTY confirmation prompt (#776); MCP has no TTY and `drt_state_reset` never prompts, so there is nothing to skip

That is the right treatment — the whole value of this check is that absence is always accounted for somewhere rather than inferred.

Approving. Green across the matrix.

PR #880, opened 2026-08-01
feat(destinations): query tagging — comment header + native tags on the write path

### LINKED ISSUE #768: feat(sources): query tagging — BigQuery job labels, Snowflake QUERY_TAG, SQL comment header for cost attribution
## Problem

drt's extract queries arrive at the warehouse anonymously: no BigQuery job labels, no Snowflake `QUERY_TAG`, no Databricks session tags, no SQL comment header (verified across `drt/sources/`). Consequences:

- **Cost attribution**: warehouse admins can't answer "what does reverse ETL cost us?" or "which sync is the expensive one?" — our own smoke-account cost digest (#710/#738) had to attribute spend at the *account* level because per-query attribution doesn't exist
- **Ops**: DBAs seeing a heavy query can't trace it back to a sync/team without guesswork
- **Governance**: query-history audits can't distinguish drt from ad-hoc humans

## Prior art

- **dbt `query_comment`**: JSON comment header on every query (`{"app": "dbt", "node_id": ...}`) + adapter-native tagging (Snowflake `QUERY_TAG`, BigQuery job labels).
- **Hightouch/Census**: tag their planner/diff queries so warehouse spend is attributable.

## Proposal

On by default (like dbt), configurable in `drt_project.yml`:

```yaml
query_tagging:
  enabled: true
  extra: { team: growth }        # optional user labels
```

Per connector:
- **BigQuery**: `QueryJobConfig(labels={"app": "drt", "drt_sync": <name>, "drt_run_id": ...})` (label-safe normalization)
- **Snowflake**: `ALTER SESSION SET QUERY_TAG = '{"app":"drt","sync":...}'` (or session parameter at connect)
- **Databricks**: session/user-agent tags
- **All SQL sources**: leading comment `/* drt sync=<name> run_id=<id> */` as the universal fallback (also applies to destination-side queries: lookups, diff, tests)

## Scope

**In:** the three DWH-native mechanisms + universal comment header, project-level config, docs for cost-attribution queries (INFORMATION_SCHEMA.JOBS / QUERY_HISTORY examples).
**Out:** cost *collection* (that's the user's warehouse tooling), OTel metrics (separate epic #429).

## Related

- #710 / #738 (smoke-account cost monitoring — first consumer: the digest could group by label), run_id issue (tag payload carrier)

## Milestone fit

Candidate: v0.8.2 (warehouse hardening) — small per-connector diffs, immediately useful to exactly the operators v0.8.2 targets.

masukai:
Stacked on #879 (source-side leg). Closes #768 once both land — this is the destination-side half.

## What

Same cost-attribution problem as #879, on the write path: every INSERT/UPSERT/MERGE/DELETE drt issues against a destination arrived at the warehouse anonymously. `sync_options._query_tags` — a new private attr on `SyncOptions`, injected by `engine/sync.py` the same way `_sync_name` already is for tracked mirror (#686) — carries the same tag payload #879's source side computed, so the `Destination` Protocol doesn't need widening.

- **Postgres / MySQL** (`BaseSqlDestination`, `sql_base.py`): a `TaggedCursor` wrapper around the *one* `cur = conn.cursor()` in `load()` / `finalize_sync()` / `_finalize_mirror*()` tags every dialect hook's SQL without touching each individual `cur.execute()` call site (~20 of them across the two files). Handles both plain strings and `psycopg2.sql.Composable` objects — Postgres's DDL (TRUNCATE/CREATE/DROP/shadow-rename) is composed via `psycopg2.sql`, which raises `TypeError` on `+` against a plain string.
- **ClickHouse**: `client.command()` / `client.query()` (TRUNCATE, swap DDL, the mirror `ALTER TABLE ... DELETE` mutation) get the comment. `client.insert()` — the actual row-write path — is a streaming API call with no SQL text to tag; documented as a real gap, not silently skipped.
- **Snowflake**: native `QUERY_TAG` session parameter at connect (JSON payload, same convention as the source side) *and* the comment. Session tags cover every query on that connection, but the comment stays useful wherever someone's looking at query text directly rather than `QUERY_HISTORY`.
- **Databricks**: the driver already has a native `query_tags` connect kwarg (found via source inspection, same discovery as #879's source-side leg) plus the comment.
- **BigQuery**: `mode: merge`'s load-into-temp-table and the `MERGE` are both jobs, so both get `labels`. `mode: insert`'s streaming insert (`insert_rows_json`) has no job to label — labels are job-scoped, and a streaming insert isn't a job.

## Two things worth flagging explicitly

**`TaggedCursor` needed its own `__enter__`/`__exit__`.** Snowflake and Databricks both use `with conn.cursor() as cur:`. Python looks up dunder methods used by implicit protocols (`with`, `len()`, ...) on the *type*, bypassing instance-level `__getattr__` — so without explicit `__enter__`/`__exit__`, wrapping their cursor would raise `TypeError: ... does not support the context manager protocol`. Found by a test written specifically to exercise the `with` path (`test_supports_with_statement`), not by any happy-path assertion — worth calling out since it's the kind of thing that's easy to miss if the wrapper is only ever exercised through Postgres/MySQL's non-context-manager `cur = conn.cursor()`.

**`tagged_cursor()` / `tag_query()` use `getattr(sync_options, "_query_tags", None)`, not direct attribute access.** A fair number of existing white-box tests (`test_sql_base.py`, `test_databricks_schema_aware.py`) build a bare `SimpleNamespace(mode=..., mirror=...)` in place of a real `SyncOptions`, since they only need the fields their own dialect hook reads. Direct attribute access broke 21 of those on first pass; `getattr` with a `None` default keeps every existing fake working unchanged instead of requiring each one to grow a field it doesn't otherwise care about.

`TaggedCursor` ended up living in a new `drt/destinations/sql_utils.py` addition rather than `sql_base.py` once it became clear Snowflake and Databricks needed the identical wrapper — one shared implementation instead of three near-copies.

## What's NOT in this PR

`drt/destinations/query.py` / `lookup.py` — the lookup (`lookups:` FK resolution), `--dry-run --diff` record-level diff, and `drt test` custom SQL queries. These currently take no `sync_name`/`run_id` at all (verified against the code, not assumed), so tagging them means widening those function signatures *and* every caller (`engine/diff.py`, `cli/commands/test.py`, `mcp/tools/run_test.py`) — a comparably-sized change in its own right, and a genuinely different call shape than the write path this PR covers. Follow-up.

## Testing

- New `tests/unit/test_sql_utils_tagging.py` — `TaggedCursor`/`tagged_cursor`/`tag_query` in isolation: plain-string and `Composable` prepending, the context-manager fix (regression-guarded), attribute delegation, the `getattr` fallback.
- Every destination test file (`test_postgres_destination.py`, `test_mysql_destination.py`, `test_clickhouse_destination.py`, `test_snowflake_destination.py`, `test_databricks_destination.py`, `test_bigquery_destination.py`) gained a tagged/untagged pair asserting the comment (and native mechanism, where applicable) actually lands on the executed query — plus a "no tags → byte-identical to before" test per destination, since `tagged_cursor` returning the cursor unchanged when tagging is off is the property that keeps every *pre-existing* exact-query-text assertion passing without modification.
- Full suite: 2654 passed, 5 skipped (pre-existing driver-not-installed skips) — including the 21 pre-existing `SimpleNamespace`-based tests the `getattr` fix keeps green.

Connector docs (`docs/connectors/{postgres,mysql,clickhouse,snowflake,databricks,bigquery}.md`) each got a `## Notes` bullet describing exactly what is and isn't tagged for that destination.

## Test plan

- [x] `pytest tests/unit` — 2654 passed
- [x] `mypy drt` — clean, 173 files

### DIFF
--- a/docs/connectors/bigquery.md
+++ b/docs/connectors/bigquery.md
@@ -107,6 +107,7 @@ then drops the temp table. Composite keys are supported (`upsert_key: [tenant_id
 ## Notes
 
 - Requires `pip install drt-core[bigquery]` (`google-cloud-bigquery`).
+- **Query tagging** ([#768](https://github.com/drt-hub/drt/issues/768)): `mode: merge`'s load + `MERGE` jobs get `labels` (BigQuery's native cost-attribution mechanism, queryable via `INFORMATION_SCHEMA.JOBS`) by default. `mode: insert`'s streaming insert (`insert_rows_json`) is a REST call, not a job, so it isn't labeled — labels are job-scoped. See `query_tagging` in `docs/llm/API_REFERENCE.md`.
 - Tables are addressed fully-qualified as `<project>.<dataset>.<table>`.
 - The target table must already exist with a compatible schema — drt writes into it, it does not create it.
 - `--dry-run` is honoured — `destination.load()` is never called when dry_run is on.
--- a/docs/connectors/clickhouse.md
+++ b/docs/connectors/clickhouse.md
@@ -170,6 +170,7 @@ into the destination and cannot be un-sent. See
 ## Notes
 
 - Requires `pip install drt-core[clickhouse]` (uses `clickhouse-connect`)
+- **Query tagging** ([#768](https://github.com/drt-hub/drt/issues/768)): `TRUNCATE`/DDL/mirror-`DELETE` statements get a leading `/* drt app=drt sync=<name> run_id=<id> ... */` comment by default. The row-write path itself (`client.insert()`) is a streaming API call, not SQL text, so it isn't tagged — see `query_tagging` in `docs/llm/API_REFERENCE.md`.
 - Each record is inserted individually to enable row-level error tracking (consistent with PostgreSQL and MySQL destination patterns)
 - For deduplication on the INSERT path, **create the destination table with `ReplacingMergeTree`** — `upsert_key` on the config is informational only for non-mirror modes
 - `drt test` validators (row_count, not_null, freshness, unique, accepted_values, query) work with ClickHouse
--- a/docs/connectors/databricks.md
+++ b/docs/connectors/databricks.md
@@ -243,6 +243,7 @@ not retried.
 ## Notes
 
 - Requires `pip install drt-core[databricks]` (depends on `databricks-sql-connector>=3.0`).
+- **Query tagging** ([#768](https://github.com/drt-hub/drt/issues/768)): every write query gets both the driver's native `query_tags` connect kwarg (a `QUERY_TAGS` session config applied to every query the session runs) and a leading `/* drt app=drt sync=<name> run_id=<id> ... */` comment, by default — see `query_tagging` in `docs/llm/API_REFERENCE.md`.
 - Target table must be a **Delta Lake table**. Non-Delta formats fail at `MERGE INTO` time with a Databricks server error.
 - Empty batches short-circuit before any `databricks.sql` import or warehouse call — the same "no driver was imported" contract used by the SQL destinations (#595). A run with zero source rows produces zero warehouse statements.
 - Errors during the connect step (missing env vars, bad token, network) raise immediately; errors during row INSERT are captured per-row and surface in `result.row_errors`.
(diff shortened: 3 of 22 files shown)


masukai:
This PR's base branch (`feat/768-query-tagging`, #879) was deleted after #879 merged, and GitHub auto-closed this PR instead of retargeting it to `main` (the head branch `feat/768-query-tagging-destinations` was untouched, so no content was lost). Recreated as #891 against `main` directly — same title/body/commits.

Issue #694, opened 2026-07-02
feat(engine): tracked mirror follow-ups — scope composition + SQL-JOIN state diff


masukai:
Two deferred pieces from #686 / #687 (v0.7.10):

1. **tracked + scope composition** — currently rejected by config validation. Semantics per #687: scope prunes both the state read and the delete to observed parents, cheapening the diff on large state tables. Requires storing scope values (or deriving them from key columns when scope ⊆ upsert_key) alongside tracked keys.
2. **SQL-JOIN state diff** — the v0.7.10 implementation loads the previous key set into memory and deletes via an IN list (parity with #340's memory bound). For key sets beyond a few million, compute `previous − current` destination-side: stage current keys to a temp table and `DELETE ... USING/JOIN` against `_drt_synced_keys`. Same follow-up shape as #340's temp-table note.

Issue #692, opened 2026-07-02
feat(destinations): tracked / scoped mirror for ClickHouse, Snowflake, Databricks


masukai:
Follow-up to #686 / #687 (shipped v0.7.10 for Postgres + MySQL).

Extend `sync.mirror.strategy: tracked` and `sync.mirror.scope` to the remaining mirror-capable destinations. All three currently **fail fast** with a clear error, so this is additive.

Per destination:
- **ClickHouse** — state table as `ReplacingMergeTree` ORDER BY (sync_name, key_hash); deletes via the existing `ALTER TABLE ... DELETE` mutation path (`mutations_sync=1`). Scope = extra predicate on the mutation.
- **Snowflake** — plain state table; tracked read/rewrite through the existing `query.py` machinery; deletes through the MERGE-forced mirror path's finalize DELETE.
- **Databricks** — Delta target required (as for mirror generally); same shape as Snowflake.

The shared pieces (`drt/destinations/_mirror_state.py`: key_json / key_hash / diff_keys, config surface, guards + tests to flip) already exist — each leg is state-table DDL + read + rewrite in that destination's dialect, mirroring `postgres.py::_finalize_mirror_tracked`.

masukai:
All three destinations landed: Snowflake (#882), ClickHouse (#883), Databricks (#884) — all merged to main.

PR #882, opened 2026-08-01
feat(destinations): tracked + scoped mirror on Snowflake

masukai:
Refs #692 (1 of 3 destinations — ClickHouse and Databricks follow separately). Stacked on #881 (#694's tracked+scope composition), since this ships that composition for Snowflake from day one rather than doing tracked-only now and scope-composition later. Not closing #692.

## What

`mirror.strategy: tracked` and `mirror.scope` both currently reject outright on Snowflake ("not yet supported... postgres, mysql"). This PR adds both, plus their #694 composition.

Turned out to be a bigger scope than "tracked only" — #692's own issue text covers `mirror.scope` too, and that's the *default* destination-strategy diff, not just the tracked one, so:

- **`mirror.scope` (#687)** on the default destination-strategy DELETE — wasn't implemented on Snowflake at all before this.
- **`mirror.strategy: tracked` (#686)** — new `_finalize_mirror_tracked`, state table `<database>.<schema>._drt_synced_keys` (identical `sync_name`/`key_hash`/`key_json` shape to Postgres/MySQL — reuses their dialect-agnostic `_mirror_state.py` helpers unchanged), existence probe via `SHOW TABLES LIKE ... IN SCHEMA` (Snowflake has no `to_regclass`/`information_schema.tables` as cheap; reuses the exact pattern `_target_exists` already established for the replace-swap path).
- **`scope` + `tracked` composition (#694)** — identical algorithm to what #881 just shipped for Postgres/MySQL: the state read/rewrite prunes to the observed scope, out-of-scope tracked keys survive the rewrite unchanged, and "no prior state at all" vs. "no prior state for this particular scope" get the right baseline-warning treatment (only the former warns).

## Shared-helper refactor

`check_mirror_supported` (`sql_utils.py`) gained a `supports_tracked_scope` kwarg (default `False`, so ClickHouse/Databricks are unaffected) instead of unconditionally rejecting tracked/scope on every non-Postgres/MySQL dialect. The scope-⊆-upsert_key check (#694) became a shared `check_scope_subset_of_upsert_key`, called from both `BaseSqlDestination._validate_mirror_scope` (refactored to call it instead of duplicating the logic inline) and Snowflake's `check_mirror_supported` — so the constraint and its wording can't drift as more of #692 lands.

## Testing

- `test_sql_utils.py` — the new `supports_tracked_scope` kwarg and `check_scope_subset_of_upsert_key` in isolation.
- `test_snowflake_mirror_mode.py` — the two old "rejected" tests flipped to "accepted"; ~18 new tests covering destination-strategy scope (missing-column fail-fast, scoped delete shape), the scope+tracked subset constraint, and the full tracked-mirror lifecycle (baseline, second-run stale-key delete, empty-source no-op) crossed with scope (stale-within-scope delete, out-of-scope state preservation, both baseline-warning cases).

Full suite: 2645 passed, 5 skipped (pre-existing driver-not-installed skips). `ruff check` / `mypy` clean. No VS Code schema drift this time (no Pydantic model docstring touched).

## Remaining: ClickHouse, Databricks

Follow-ups, same overall shape — but flagging now rather than assuming the Snowflake implementation transfers unchanged: ClickHouse's plain mirror DELETE already runs through an `ALTER TABLE ... DELETE` *mutation* (not a plain `DELETE`), and Databricks' existing mirror already uses a staging+anti-join pattern (#707) for a driver-specific reason (native `?` paramstyle's per-statement parameter limit). Each needs its own look at how tracked-mirror's DELETE fits that existing shape, not a copy-paste of this PR's Snowflake version.

## Test plan

- [x] `pytest tests/unit` — 2645 passed, 5 skipped (pre-existing)
- [x] `mypy drt` — clean, 172 files
- [x] VS Code schema regeneration — no diff


masukai:
**Follow-up from independent review of the #694 part 2 stack**: this PR's \`_finalize_mirror_tracked\` docstring implicitly borrows \`BaseSqlDestination\`'s "target delete and state rewrite share one transaction" guarantee via its "Snowflake counterpart of..." framing, but that guarantee doesn't actually hold here — this connection autocommits by default (already true in this PR, not introduced later), so the target DELETE / state DELETE / state INSERT are three independently-committed statements. Usually degrades safely (partial state reads back as "no prior state" next run → baseline WARN), but not a code-level guarantee.

Rather than rebasing a fix into this branch (would cascade through #883/#884), the docstring correction landed as a follow-up commit on #889 (the tip of this stack) — see [585735e](https://github.com/drt-hub/drt/pull/889/commits/585735e). No behavior change, just accurate documentation. Flagging here so the history is traceable back to this PR.

PR #886, opened 2026-08-01
feat(destinations): SQL-side tracked-mirror diff on Postgres/MySQL (#694 part 2, 1 of N)

masukai:
## Summary

Part 2 of #694 (part 1 — scope composition — shipped in #881). The v0.7.10 tracked-mirror implementation `SELECT`ed every key a sync has ever tracked into Python and diffed it against the current run with a Python `set` — fine at hundreds of keys, but it means a state table with millions of rows gets fully read into memory just to compute a diff that's typically small (most runs only change a handful of rows).

This replaces that with a SQL-side diff on the shared `BaseSqlDestination` path (**Postgres + MySQL** — the first of four dialects; Snowflake/ClickHouse/Databricks follow in stacked PRs, same as #692's split). This run's keys are staged into a scratch table and the database computes `previous - current` via a `NOT EXISTS` join against `_drt_synced_keys`, so only the (small) diff — and the genuinely-new keys to insert — ever reach Python.

## Design notes

- **The old "read every untouched row so it can be reinserted unchanged" step for scope-preserved rows is gone.** Untouched rows (scopes/parents this run never observed) are simply never selected by either the diff query or the new-keys query — not reading and not touching a row has the same effect as reading it and reinserting it unchanged, without ever having to.
- **Scope-filtering the (small) diff in Python after the SQL-side subtraction is mathematically equivalent** to filtering the full previous set by scope before diffing (the #694 part 1 approach): scope membership doesn't depend on `current`, and `current`-membership doesn't depend on scope, so `(previous ∩ scope) - current == (previous - current) ∩ scope`. That's what lets scope-filtering stay in Python without ever materialising the full previous state.
- The baseline-vs-silent distinction (first run ever vs. first run to touch a particular scope) now comes from a cheap `EXISTS ... LIMIT 1` probe instead of the previous full-table-truthiness check.
- Staging uses `CREATE TEMPORARY TABLE` (the one spelling both Postgres and MySQL accept), unqualified/unquoted since it's a fixed constant, never a user-configured identifier.
- `diff_keys`/`_mirror_state.py` stays in use by the dry-run `--diff` preview (`engine/diff.py`, #693) — a read-only, human-triggered, best-effort path where the old Python-side cost is an acceptable tradeoff. Only the real execution path moved.

## Testing

Every existing tracked-mirror test in `test_sql_base.py` / `test_postgres_mirror_mode.py` / `test_mysql_mirror_mode.py` needed rewriting: the mock cursors now answer three distinct reads (baseline probe, diff, new-keys) instead of one, dispatched by the executed SQL's text. Most visibly, the scope-preservation tests no longer assert a single `executemany` call's full row set (there often isn't one when nothing changed) — instead they assert which keys the state-delete/insert calls reference and, symmetrically, which keys never appear in any of them.

- `pytest tests/unit`: 2688 passed, 3 skipped (full suite, no regressions).
- `mypy drt`: clean.
- Coverage: `_mirror_state.py` 100%, `sql_base.py` 99% (one missing line is pre-existing `_resolve_schema` code untouched by this diff).
- No VS Code schema drift (no Pydantic model changed).

Stacked on #884 (Databricks, closing #692) since both touch the same tracked-mirror code paths.


masukai:
**Follow-up from independent review**: this PR's "only the (small) diff ever reaches Python" claim only holds for **unscoped** tracked mirror. The diff query filters only by \`sync_name\`, not by scope, so for \`mirror.scope\` + \`strategy: tracked\` — the primary motivating use case for scope — a run touching one scope out of many historically-tracked ones gets a diff close to the entire previous state, not a small one. The final \`to_delete\` set is still correct (verified the scope-filtering-after-diff algebra), so this is a missed performance win, not a correctness bug.

Doc/docstring corrections landed as a follow-up commit on #889 (tip of the stack) — see [3f6ccd2](https://github.com/drt-hub/drt/pull/889/commits/3f6ccd2), which also files #890 to track a real fix. Flagging here since the base implementation (this PR) is where the pattern originates.

PR #887, opened 2026-08-01
feat(destinations): SQL-side tracked-mirror diff on Snowflake (#694 part 2, 2 of N)

masukai:
## Summary

Second of the #694 part 2 dialect legs (Postgres/MySQL landed in #886). Same change, ported to Snowflake's own connection/cursor and explicit-`%s`-placeholder DELETE shape: this run's keys are staged into a scratch `TEMPORARY` table and `previous - current` is computed with a `NOT EXISTS` join against `_drt_synced_keys` in SQL, instead of `SELECT`ing every tracked key for this sync into Python and diffing with a Python `set`. Only the (small) diff and the genuinely-new keys to insert ever reach Python.

## Design notes

Same reasoning as #886 (the base Postgres/MySQL implementation), just against Snowflake's primitives:

- Scope-filtering the diff in Python after the SQL-side subtraction is mathematically equivalent to filtering the full previous set by scope first (`(previous ∩ scope) - current == (previous - current) ∩ scope`).
- The old "read every untouched row so it can be reinserted unchanged" step for scope-preserved rows is gone — untouched rows are simply never selected by either the diff query or the new-keys query.
- The baseline-vs-silent distinction now comes from a cheap `EXISTS ... LIMIT 1` probe.
- Snowflake's existing `SHOW TABLES` existence probe (`fetchall`-based, unlike Postgres/MySQL's `fetchone`-based `to_regclass`/`information_schema` checks) is untouched.

## Testing

Every existing tracked-mirror test needed the same kind of rewrite as #886: mock cursors now answer three distinct reads (baseline probe, diff, new-keys) instead of one, dispatched by the executed SQL's text. Added two new tests not exercised by any adapted test:
- staging-table lifecycle (create-before-diff, drop-after)
- insert-only-genuinely-new-keys behavior

- `pytest tests/unit`: 2690 passed, 3 skipped (full suite, no regressions).
- `mypy drt`: clean.
- Coverage: `snowflake.py` 99% (3 missing lines are pre-existing, well outside this diff's touched range).
- No VS Code schema drift.
- `docs/connectors/snowflake.md` already cross-references `postgres.md` for the full tracked-mirror write-up (updated in #886) — no doc change needed here.

ClickHouse and Databricks follow in stacked PRs. Stacked on #886.


masukai:
See the follow-up comment on #886 and #882 — same scope-diff doc caveat and Snowflake autocommit/transaction-boundary correction apply here; both landed as a follow-up commit on #889 (tip of the stack) rather than rebased back into this branch, to avoid cascading through #888/#889.

PR #888, opened 2026-08-01
feat(destinations): SQL-side tracked-mirror diff on ClickHouse (#694 part 2, 3 of N)

masukai:
## Summary

Third of #694 part 2's four dialect legs (Postgres/MySQL in #886, Snowflake in #887). Same goal — stop `SELECT`ing every key a sync has ever tracked into Python before diffing — but ClickHouse gets a genuinely different mechanism: **no scratch/staging table at all**.

ClickHouse's `Array(String)` named parameters already hold this run's entire key-hash set as a *single* bound value (no per-element placeholder the way Databricks' native paramstyle has a 255-marker limit), so both directions of the diff are plain `NOT IN` / `arrayJoin` queries against that one parameter instead of a table join:

- `previous - current` (the diff): `key_hash NOT IN {current_hashes:Array(String)}`
- `current - previous` (genuinely-new keys): `arrayJoin({current_hashes:Array(String)}) ... WHERE key_hash NOT IN (SELECT key_hash FROM state ...)`

Only each query's (typically small) result reaches Python — never the full tracked-key history.

## Design notes

- Same scope-filtering proof as the other two legs: filtering the diff in Python after the SQL-side subtraction is mathematically equivalent to filtering the full previous set by scope first.
- Same "untouched rows are never selected, so they're never touched" replacement for the old preserved-rows carry-through.
- The existing "no cross-statement transaction, target DELETE runs first" failure-mode ordering (ClickHouse-specific, since mutations here don't share a transaction) is preserved — and actually **narrows** one failure window: splitting the state rewrite into delete-diffed-hashes + insert-new-hashes means a failure between the two only loses tracking for this run's *new* keys, not (as with the old blanket delete-all-then-insert-current) every key in the run.

## Testing

Every existing tracked-mirror test needed the same kind of rewrite as #886/#887 — `client.query()` is now dispatched by SQL text across up to four distinct queries (`EXISTS TABLE`, baseline probe, diff, new-hashes) instead of two.

- `pytest tests/unit`: 2690 passed, 3 skipped (full suite, no regressions).
- `mypy drt`: clean.
- Coverage: `clickhouse.py` 89% — all 25 missing lines are pre-existing `get_row_count`/`test_connection`/`_connect` code well outside this diff's touched range (confirmed line-by-line via `git diff`).
- No VS Code schema drift.
- `docs/connectors/clickhouse.md` already cross-references `postgres.md` for the full tracked-mirror write-up — no doc change needed.

Databricks is the last of #694 part 2's four dialects, in the next stacked PR. Stacked on #887.


masukai:
Same follow-up as #886/#887 — the unscoped-vs-scoped diff-size doc caveat applies here too (ClickHouse's diff query also has no scope predicate). Landed as a follow-up commit on #889 (tip of the stack) — see [3f6ccd2](https://github.com/drt-hub/drt/pull/889/commits/3f6ccd2).

Issue #778, opened 2026-07-10
feat(cli): run artifacts — versioned target/run_results.json per invocation


masukai:
## Problem

`drt run --output json` prints a summary to stdout and vanishes. There is no durable, machine-readable record of an invocation: CI systems can't upload a run artifact, observability pipelines can't ingest one, and drt-action reconstructs its step summary from parsing stdout. dbt established the pattern: every invocation writes `target/run_results.json` with a versioned schema, and an ecosystem (dbt-artifacts loaders, Elementary, Datafold CI bots) grew on top of it. drt has the docs-side manifest (#507) but no run-results counterpart.

## Prior art

- **dbt `run_results.json`**: schema-versioned per-invocation artifact (status/timing/rows per node + invocation metadata).
- **dlt**: load-info/trace objects persistable per pipeline run.

## Proposal

Every `drt run` / `drt test` / (future `drt build`) writes:

```
target/run_results.json
{
  "schema_version": 1,
  "invocation": { "run_id": ..., "started_at": ..., "argv": [...], "drt_version": ... },
  "results": [ { "sync": "users_to_hubspot", "sync_run_id": ..., "status": "success",
                 "rows_extracted": 1200, "rows_synced": 1198, "row_errors": 2,
                 "duration_seconds": 12.4, "watermark": {"before": ..., "after": ...},
                 "error": null, "tests": [...] } ]
}
```

- Written atomically at end of invocation (and best-effort on SIGTERM via the #279 shutdown path); `--target-path` to relocate
- Schema documented + versioned like manifest v1 (#507); JSON Schema published for tooling
- drt-action switches its outputs/step-summary to read the artifact; `drt run --failed` and future invocation-replay read it too

## Scope

**In:** artifact writer, schema doc + JSON Schema, drt-action consumption follow-up, history dedup note (history remains the *rolling* log; run_results is the *per-invocation* snapshot).
**Out:** artifact upload/storage (CI's job), metrics export (OTel #429).

## Related

- #507 (manifest v1 — sibling artifact + schema-versioning precedent), #698 (manifest v2 run history), #142 (original `--output json`), run_id issue (identifiers), drt-action (#292)

## Milestone fit

Candidate: v0.9 — foundation for CI/observability integrations; cheap once run_id lands.

PR #899, opened 2026-08-03
docs(adr): ADR 0005 — where drt's state lives, and what it costs the operator (#755, #756)

masukai:
## Why

#755 and #756 were both written assuming drt writes managed tables into the **source warehouse**. This ADR examines that assumption and finds it only half right — which changes the implementation order for v0.9's first two issues.

## What the research found

**Hightouch ships two engines split on exactly this line:**

| | Basic | Lightning |
|---|---|---|
| Warehouse permission | **read-only** | read + write, schema creation |
| Diff computed | on Hightouch infrastructure | in the customer's warehouse |
| State lives | Hightouch's servers | `hightouch_planner`, `hightouch_audit` |
| Reversible | upgrade any time | **cannot return to Basic** |

Segment's Reverse ETL is the same shape (`_segment_reverse_etl`, read+write).

**The Census prior-art claim in #755 could not be verified.** `docs.getcensus.com` now 301s to Fivetran's Activations docs, which state neither the permission requirement nor the state location. That claim should be dropped or qualified — tracked as follow-up 2 below.

## The asymmetry that decides it

Every SaaS vendor above has a **third location** for state — their own infrastructure. That is what lets Hightouch offer a read-only tier at all. **drt has none**, so drt's local tier is strictly weaker than Hightouch Basic:

| | read-only source | survives CI | team-shared |
|---|---|---|---|
| Hightouch Basic | ✅ | ✅ | ✅ |
| **drt local (today)** | ✅ | ❌ | ❌ |

drt's equivalent third location is **object storage the operator already owns** — which drt already speaks via `GCSWatermarkStorage`, and which the #756 field report hand-rolled by persisting `.drt/history` to a GCS object.

That splits #756's motivation in two, with different costs:

1. **Durability and sharing** — object storage is sufficient, **no warehouse grant needed**
2. **SQL-queryable observability** — genuinely requires the warehouse

#755 has no such split: a multi-million-row diff cannot run against a JSON blob, so it is structurally a Lightning-shaped capability.

## Decisions

1. State location is **tiered**, and the tier is a config choice — not a product tier or a paywall
2. **#756 splits along that seam**; the object-storage half ships first
3. **#755 requires warehouse write and says so**, with Hightouch's pre-created-schema escape hatch; **cursor-based incremental stays a first-class read-only path**
4. **Tiers must be reversible** — Hightouch explicitly cannot go back; drt can, because config lives in Git rather than a vendor control plane
5. **Borrow dbt snapshot vocabulary** (`check_cols`) rather than inventing names

No differentiation is warranted on this axis — state and diff plumbing is table stakes. The novelty budget belongs on the Git-native/artifacts-as-files axis (#772, #778, drt-action).

## ⚠️ This also corrects ADR 0004

ADR 0004's 2026-07-29 amendment folded the **#769 cross-process rate-limit residual** into #756 on the reasoning that it needs shared state. It does — but not this kind. A token bucket needs low-latency atomic increment; object storage has no cheap compare-and-swap and a warehouse table is worse on both counts. **Neither half of #756 closes it.** It should be lifted back out into its own issue and ADR 0004's gate table amended (follow-up 4).

Conversely, **ADR 0004's Tier 2 gate clears earlier than its table implies** — a Dagster sensor and a CI run sharing a watermark is a durability problem, satisfied by the object-storage half alone.

## Falsification condition

This ADR is wrong if object storage cannot carry history and DLQ at realistic volume — specifically write amplification, if the backend rewrites a whole object per run the way `GCSWatermarkStorage` does today. The measurement belongs in the implementation, not after it. @K-Masuda-SL offered in #756 to test a prototype against a real GCP / Cloud Run Jobs + Dagster deployment, which is the topology this failure mode targets.

## Follow-up issues

Not in this PR — listed in the ADR and to be filed separately:

1. Split #756 into the object-storage durability half and the warehouse-observability half
2. Correct #755's prior-art section (Census → dbt snapshots)
3. New issue: read-only source support + tier reversibility as design goals
4. Re-scope the #769 residual out of #756, amend ADR 0004's gate table

### DIFF
--- a/docs/adr/0005-state-location-and-write-grants.md
+++ b/docs/adr/0005-state-location-and-write-grants.md
@@ -2,11 +2,11 @@
 
 - **Status:** Accepted 2026-08-03. Both #755 and #756 were written assuming an
   answer — "drt writes managed tables into the source warehouse" — that this ADR
-  finds only half right, and the implementation plan
-  ([docs/plans/2026-07-22-756-remote-state-backend.md](../plans/2026-07-22-756-remote-state-backend.md))
-  had already been drafted on that assumption. That plan has been revised to
-  this ADR's ordering; the issues themselves still need the corrections listed
-  under [Follow-up issues](#follow-up-issues).
+  finds only half right, and an implementation plan for #756 (dated
+  2026-07-22, kept locally under the gitignored `docs/plans/`, not tracked in
+  this repo) had already been drafted on that assumption. That plan has been
+  revised to this ADR's ordering; the issues themselves still need the
+  corrections listed under [Follow-up issues](#follow-up-issues).
 - **Issues:** [#755](https://github.com/drt-hub/drt/issues/755),
   [#756](https://github.com/drt-hub/drt/issues/756)
 - **Relates to:** [ADR 0004](0004-streaming-and-event-triggered-syncs.md) — whose
@@ -181,10 +181,16 @@ The operator-visible payoff of #756 lands at step 2, before any permission
 conversation. Step 1 is a prerequisite regardless of this ADR's outcome: the
 three managers are constructed directly at roughly fourteen call sites with no
 factory, so no backend selection can be honoured until that is centralised.
+*(Half-landed already: #900, merged the day after this ADR was opened,
+extracted the `StateStore` / `HistoryStore` / `DlqBackend` Protocols with
+back-compat aliases and a set-equality drift test against each local
+implementation's public API. The factory half — routing a backend choice to
+a concrete implementation at the roughly fourteen call sites above — is still
+open; `drt/state/manager.py:150` carries the placeholder comment for it.)*
 
-**The Protocol freeze (#304 / v0.10) inherits whatever step 1 produces**, and
-#297's third-party plugin system may put external implementations behind it —
-so the Protocol shape wants review before it lands, not after.
+**The Protocol freeze (#304 / v0.10) inherits whatever step 1 produces.**
+#900's Protocols are what it inherits from; #297's third-party plugin system
+may put external implementations behind them.
 
 ## Falsification condition
 
@@ -192,7 +198,20 @@ This ADR is wrong if object storage cannot carry history and DLQ at realistic
 volume. The specific risk is write amplification: if the backend rewrites a
 whole object per run (the shape `GCSWatermarkStorage` uses today), a project
 with many syncs and frequent runs pays a full read-modify-write per sync per
-run, and concurrent runners will clobber each other with last-write-wins.
+run.
+
+Concurrent runners clobbering each other with last-write-wins is a real,
+currently-shipped bug in that same shape — `GCSWatermarkStorage.save()`
+(`drt/state/watermark.py:111-117`) loads, mutates, and calls
+`upload_from_string()` with no generation precondition, so two saves that
+race lose one silently — but it is not evidence against this ADR. GCS
+supports preconditioned writes (`if_generation_match`) and S3 now has
+conditional writes too, so the fix is a retry-on-precondition-failure loop in
+the client, not a change of tier. If anything this strengthens Decision 2:
+the object-storage tier *can* be made safe for concurrent writers, which is
+what "team-shared" in Decision 1 requires, it just isn't yet. Tracked as
+[#919](https://github.com/drt-hub/drt/issues/919), independent of #756 —
+today's bug, not a #756 design question.
 
 If a prototype shows that per-run cost or contention makes the object-storage
 tier unusable at, say, 50 syncs on a 15-minute cadence, then durability must
--- a/docs/adr/0005-state-location-and-write-grants.md
+++ b/docs/adr/0005-state-location-and-write-grants.md
@@ -1,0 +1,216 @@
+# ADR 0005 — Where drt's state lives, and what it costs the operator
+
+- **Status:** Accepted 2026-08-03. Both #755 and #756 were written assuming an
+  answer — "drt writes managed tables into the source warehouse" — that this ADR
+  finds only half right, and the implementation plan
+  ([docs/plans/2026-07-22-756-remote-state-backend.md](../plans/2026-07-22-756-remote-state-backend.md))
+  had already been drafted on that assumption. That plan has been revised to
+  this ADR's ordering; the issues themselves still need the corrections listed
+  under [Follow-up issues](#follow-up-issues).
+- **Issues:** [#755](https://github.com/drt-hub/drt/issues/755),
+  [#756](https://github.com/drt-hub/drt/issues/756)
+- **Relates to:** [ADR 0004](0004-streaming-and-event-triggered-syncs.md) — whose
+  Tier 2 gate is #756, and whose #769 amendment this ADR corrects.
+- **Implementation:** none directly. This ADR sets the ordering and the
+  permission posture that #755/#756 implement; see
+  [Follow-up issues](#follow-up-issues).
+
+## Context
+
+Everything drt remembers between runs lives on local disk under `.drt/` — run
+state (`drt/state/manager.py`), history (`drt/state/history.py`), and the DLQ
+(`drt/state/dlq.py`). Watermarks are the single exception, with `gcs` and
+`bigquery` storage backends already shipped (`drt/state/watermark.py`).
+
+The deployment model drt actively promotes — GitHub Actions cron via
+drt-action, ephemeral containers, Dagster/Airflow workers, Cloud Run Jobs —
+destroys that disk after every run. The symptoms are concrete and reported
+from production (#756, @K-Masuda-SL): `drt status` is blank in a fresh
+checkout, `drt retry` cannot see yesterday's failures, and a laptop and a
+runner hold disjoint histories.
+
+v0.9 puts two features on the table that both need somewhere durable to write:
+#756 (run state, history, DLQ) and #755 (snapshot-diff incremental). As
+written, **both assume drt creates and writes managed tables in the source
+warehouse.** That assumption is what this ADR examines, because it is not free:
+it requires a write grant on the source, and "the data team hands you a
+read-only service account" is a common starting condition for exactly the
+teams drt targets.
+
+## Evidence
+
+**Hightouch** ships two engines, and the split is along this exact line:
+
+| | Basic | Lightning |
+|---|---|---|
+| Warehouse permission | **read-only** | read + write, schema creation |
+| Diff computed | on Hightouch infrastructure | in the customer's warehouse |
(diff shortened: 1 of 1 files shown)


masukai:
@yodakanohoshi — thanks in advance for taking a look, and no pressure to review the whole thing.

Flagging upfront that this is a different kind of review from your ClickHouse identifier fix in #610: it's an architecture decision (an ADR), so there's no code to check for correctness. Partial feedback on any one section is genuinely useful — you don't need an opinion on all of it.

Three places where a second pair of eyes would help most:

1. **Is the read of Hightouch's two-engine model fair?** The whole argument rests on the Basic/Lightning split being a deliberate permission tier rather than an accident of their architecture. If that reading is wrong, decisions 2 and 3 fall over.

2. **Does the object-storage-first ordering have a hole?** The claim is that CI durability (state surviving an ephemeral runner) needs no warehouse write grant, and only SQL-queryable observability does. If there's a durability use case that secretly needs the warehouse, I'd rather find it now than after Part A ships.

3. **ClickHouse's absence from the warehouse backend list — this one is squarely your area.** Part D targets BigQuery / Postgres / Snowflake and deliberately leaves ClickHouse out, on the reasoning that a state store wants cheap row-level updates and deletes, and ClickHouse's `ALTER TABLE ... DELETE` mutation path (the one we use for `mode: mirror`, with `mutations_sync=1`) is the wrong shape for that. You know those semantics better than I do — does that match your read, or is there a table engine choice that would make it workable?

The implementation follow-ups are listed at the bottom of the PR body; none of them are in this PR.

masukai:
@Pawansingh3889 — adding you as a reviewer here, but genuinely as an *if you're up for it*, not an ask. You've got #902 in flight and that's the bigger piece of work; this can sit.

The reason you came to mind: in #902 you hit the `run_drt_sync()` / `StateManager` instance-lock race from inside the process. This ADR argues the same underlying thing from the other side — that drt has no state tier which survives concurrency or an ephemeral runner, and that fixing it does *not* have to start by asking operators for a warehouse write grant. You're the person who most recently ran into the problem for real, so if any part of the reasoning looks wrong from where you were standing, that's worth more than a careful read of the whole thing.

Two caveats on what kind of review this is:

- It's an **ADR** — a decision, no code to check for correctness. Partial feedback on one section is completely fine.
- I got a fact wrong once already while writing it. I originally recommended hashing resolved config in a sibling PR (#901) on the grounds that "dbt resolves env vars", read the docs properly, and found dbt documents the opposite as a known blind spot. So the competitor readings in here are the part I'd most like someone to push on — specifically that Hightouch's Basic/Lightning split is a deliberate **permission tier** (read-only vs read+write) rather than an artefact of their architecture. Decisions 2 and 3 fall over if that reading is wrong.

@yodakanohoshi is also on it with a ClickHouse-specific question in the earlier comment — different angle, no overlap.

No timeline on this from my side.

Pawansingh3889 (approved):
Read this against the code rather than just for argument, since the parts I could check are the parts a future reader will trust without checking. They hold up, and the asymmetry section is the piece I think earns the ADR: "every vendor here has a third location and drt has none" reframes the local tier from "the simple option" to "strictly weaker than the free tier of the thing we are compared against", which is a much better reason to build the object-storage half first than "durability is nice".

The #769 correction is right and I would have missed it. Folding it into #756 was reasonable on "needs shared state" and wrong on what kind, and catching that an issue structurally cannot close a residual assigned to it is worth more than most of what a review finds.

## One thing I would fix before you merge

The Status line links to `../plans/2026-07-22-756-remote-state-backend.md`, but `docs/plans/` is gitignored (`.gitignore:55`) and no plan file is tracked. So that link 404s on GitHub, in the opening paragraph of a document whose job is to be read. Either describe the plan inline without the link, or un-ignore that one file if it is meant to be public. My guess is the first, given the directory is ignored deliberately.

## Two accuracy notes, both from timing rather than from you

**Step 1 is half-landed now.** #900 merged the day after you opened this and extracted the `StateStore` / `HistoryStore` / `DlqBackend` Protocols with back-compat aliases. The factory half is still open, and `drt/state/manager.py:150` carries the placeholder comment for it. So the ordering table reads as entirely-ahead work when the first row is half behind, and the Consequences line about the Protocol shape wanting review "before it lands, not after" was overtaken by events. A parenthetical on that row would keep the document accurate on the day it merges. (For what it is worth, the set-equality drift test in #900 seems to me to do what that sentence was asking for.)

**"Object storage offers no cheap compare-and-swap" is stronger than the argument needs, and it collides with your own falsification condition.** GCS has generation preconditions, and `if_generation_match` on the upload is exactly a compare-and-swap; S3 has since added conditional writes as well. The #769 conclusion survives untouched, because the load-bearing half of that sentence is the per-acquire round trip, and a token bucket that pays a network round trip plus CAS retries per acquire is dead on latency whether or not the CAS exists.

Where it matters is the falsification section. You flag that concurrent runners "will clobber each other with last-write-wins", and that is not hypothetical: `GCSWatermarkStorage.save()` (`drt/state/watermark.py:111-117`) is a load, mutate, `upload_from_string` with no precondition, so the shape the ADR proposes to generalise carries the bug today. But it is fixable in the client library drt already depends on, which means the object-storage tier is not last-write-wins by nature, only by current implementation. That reads to me as strengthening Decision 2 rather than threatening it, and it is worth saying so, because as written the falsification condition leaves clobbering looking like an inherent property of the tier and the "team-shared" claim in Decision 1 sitting slightly uneasily next to it.

If you would rather keep the ADR's line and let step 2 discover this, that is defensible too. It is a design note either way, not a disagreement with the decision.

## Checked and correct

- Watermarks as the single exception, with `gcs` and `bigquery` shipped: `LocalWatermarkStorage` / `GCSWatermarkStorage` / `BigQueryWatermarkStorage` are the three in `drt/state/watermark.py`.
- `_drt_synced_keys` living in the destination when the destination is SQL.
- "Roughly fourteen call sites with no factory": I count fourteen direct constructions outside `drt/state/`, plus a few more in the MCP tools. Fair.
- ADR 0004's 2026-07-29 amendment does fold the #769 residual into the #756 row, so the correction in Consequences is aimed at the right text.
- #304 and #297 are what the ADR says they are.
- The relative link to ADR 0004 and the internal anchors resolve.

No CHANGELOG entry, which I think is right: ADR 0003 and 0004 are mentioned inside the feature entries whose issues they closed, and this one closes nothing on its own.

Approving. The link is the only item I would hold merge for, and if you would rather not respin for it, say so and I will fix it in a follow-up.

masukai:
Thank you for this — genuinely the kind of review that makes a document better than the person who wrote it could manage alone. All three items landed in `65b8c1c`.

**The link.** Fixed as you guessed — described inline, since `docs/plans/` staying gitignored is deliberate and I'd rather the ADR not depend on a file that isn't tracked.

**Step 1 half-landed.** Added the parenthetical, and pulled the "wants review before it lands" line since you're right that it was overtaken by #900 landing. And yes — the set-equality drift test in #900 (`test_state_store_protocol_covers_local_public_api` and its `HistoryStore`/`DlqBackend` siblings) is exactly that: it fails loudly if a Protocol declares fewer methods than the local implementation actually has, which is the concrete form of "review the shape before it lands."

**The CAS paragraph.** Took your framing, and checked the claim rather than just rewording around it. You're right on both counts: `GCSWatermarkStorage.save()` really does race today (verified the exact lines), and it's fixable in the client rather than inherent to the tier — `BigQueryWatermarkStorage.save()` uses `MERGE`, so it doesn't share the bug at all. Filed the GCS fix as #919, independent of #756, and reframed the falsification section along the line you suggested: this strengthens Decision 2 rather than threatens it, since "team-shared" only requires that the tier *can* be made safe, not that it already is.

Re-requesting your review on the diff, mostly so the fix is visible rather than because I think there's more to discuss — say if there is.

One structural note, not about the review itself: your approval doesn't satisfy this repo's CODEOWNERS gate (`* @drt-hub`, org-team-only), so the merge is still blocked on @yodakanohoshi regardless of the fix. Flagging so it doesn't look like I'm sitting on an approved PR — I'm not, the gate is doing that.

Pawansingh3889:
`65b8c1c` looks right to me on all three, and #919 is a better writeup than the review comment it came from. Finding that `BigQueryWatermarkStorage.save()` uses `MERGE` and so never had the bug is the part I should have checked and didn't.

Two corrections to my own review, though, one of which is now load-bearing in the document.

## "Roughly fourteen call sites" is wrong, and I called it fair

I said I counted fourteen. I had eyeballed a grep rather than counted it. The real number on current `main` is **24**, and `65b8c1c` now states the estimate a second time ("the roughly fourteen call sites above"), so my endorsement helped it survive a revision pass.

```
git grep -nE "\b(StateManager|LocalStateManager|HistoryManager|LocalHistoryManager|DlqStore|LocalDlqStore)\(" \
  origin/main -- 'drt/**/*.py' | grep -v ":drt/state/"
```

24 lines, one construction each, no imports or annotations among them. By class: `StateManager` 13, `DlqStore` 6, `HistoryManager` 5. By file the concentration is `drt/cli/commands/run.py` 6, then `drt/docs/builder.py` 3 and `drt/cli/commands/status.py` 3. Six of the 24 are in `drt/mcp/tools/`. No `Local*` variant is constructed outside `drt/state/` at all. It becomes 25 once #902 lands, which adds the shared one in `serve()`.

This cuts your way: "no backend selection can be honoured until that is centralised" is more true at 24 sites than at 14, and step 1 is a bigger prerequisite than the ADR budgets for. But the figure is in the document twice now and it is off by ten.

## `#304 / v0.10` pairs the wrong issue with the milestone

#304 is milestoned **v1.0** and sits in ROADMAP.md's v1.0 section (lines 206-211). The v0.10 entry is **#300**, "Protocol stability review and freeze preparation", listed in v0.10's scope. So it is either "#304 / v1.0" or "#300 / v0.10" depending on which you meant. The freeze-*preparation* reading seems the better fit for "inherits whatever step 1 produces", since that is the v0.10 work and the actual freeze is a release later.

I put #304 in my "checked and correct" list. I had confirmed the issue existed and was about the Protocol freeze, then vouched for a milestone I never opened.

## One more, which doesn't touch the ADR

My closing line said ADR 0003 and 0004 are both mentioned inside the feature entries whose issues they closed. ADR 0003 is. ADR 0004's only appearance in CHANGELOG.md is inside the `drt serve` entry I had edited myself an hour earlier, and #786 has no entry at all. So I quoted my own text back as project convention. The conclusion survives and is better supported than I argued it: ADRs get no entries of their own here, so one that closes nothing needs none.

Both mistakes are the same mistake. I put things in a section headed "checked and correct" that I had reasoned about rather than measured, which is the one section where that shouldn't happen.

Agreed on CODEOWNERS, and thanks for saying so plainly. `* @drt-hub` is org-team-only, so my approval can't clear the gate however it reads.

PR #901, opened 2026-08-03
feat(config): per-sync file fingerprint for state:modified (#772)

masukai:
First piece of #772 — the comparison unit CI needs in order to answer *"which syncs did this PR change?"*. No selector or CLI flag yet; this is the fingerprint alone, so it can be argued with on its own.

## What it hashes

The sync file **as written on disk** — before `${VAR}` substitution and `var()` rendering — plus the raw bytes of the `syncs/models/<name>.sql` a `ref()` points at.

```python
sync_fingerprints(project_dir) -> dict[str, str]   # resolved sync name -> digest
```

## Why not hash the resolved config

An earlier draft did, and it forced the question *"which destination fields are secret?"*. Reusing `config/secrets.py`'s name-suffix heuristic for that would have been a correctness bug — run against every destination config it flags:

```
SnowflakeDestinationConfig: ['upsert_key']
AirtableDestinationConfig:  ['access_token', 'primary_key']
JiraDestinationConfig:      ['project_key']
```

`upsert_key` is a column list, Airtable's `primary_key` is the field passed to `performUpsert.fieldsToMergeOn`, Jira's `project_key` is the `ENG` in `ENG-123`. Redacting `upsert_key` would mean **changing a sync's upsert key does not change its fingerprint** — write semantics change completely and CI does not select it.

Hashing the file **removes the question instead of answering it carefully**, and matches what comparable tools do: dbt keeps credentials in `profiles.yml` outside the project, which is why [`manifest.json` is documented as safe to share as a CI artifact](https://docs.getdbt.com/reference/artifacts/dbt-artifacts); Hightouch Git Sync and Census Git-backed Models keep them in the vendor workspace, referenced by id. **None of them classify secret fields — they are built so the question cannot arise.**

The accepted cost is dbt's documented blind spot: an env var or `var()` *value* change is invisible ([state comparison caveats](https://docs.getdbt.com/reference/node-selection/state-comparison-caveats)). In exchange, a baseline generated in one environment compares meaningfully against another, since file bytes do not depend on the environment.

## Two traps this deliberately avoids, each with a regression test

**The watermark must not reach the hash.** `engine/resolver.py`'s `resolve_model_ref` takes `last_cursor_value` and substitutes the **current watermark** into the SQL it returns. Fingerprinting that would move the hash after every incremental run — `state:modified` would select every sync, forever. So this reads `.sql` files off disk and never calls the resolver. Pinned by a test that writes `state.json` + `watermarks.json` and asserts the fingerprint holds.

**An undefined env var must not hide a sync.** Expanding the whole document to find `name:` meant one unresolvable `${VAR}` *anywhere* in the file dropped the sync out of the map — so a PR adding a sync whose secret is not provisioned yet would silently not be selected, which is exactly the PR CI should be looking at. Name expansion is now isolated to the name value. Found by strengthening the secret test rather than by review.

## Behaviour pinned on purpose

A **comment-only edit marks the sync modified**. That is what raw-byte hashing means and what dbt does with file checksums; being conservative costs one extra dry-run, the opposite is a silent miss. There is a test asserting it, so changing it later has to be deliberate.

## Not covered (documented, not silently missing)

A change to `drt_project.yml` (project `vars`, profile) affects every sync but no sync file. This is a per-sync-file change detector, not a whole-project one — same split dbt makes.

## Verification

- 11 new tests, including cross-process stability under two different `PYTHONHASHSEED` values (a same-process re-call would pass even if the implementation leaned on `hash()`)
- Full suite **2914 passed, 21 skipped**; `ruff` and `mypy` clean

Also renames `parser._expand_sync_vars` → `expand_sync_vars` (two internal call sites), symmetric with the already-public `expand_env_vars`.

### DIFF
--- a/drt/config/fingerprint.py
+++ b/drt/config/fingerprint.py
@@ -81,15 +81,31 @@ def sync_fingerprints(project_dir: Path | str = Path(".")) -> dict[str, str]:
     # project's syncs are.
     for path in sorted(syncs_dir.glob("*.yml")):
         raw = path.read_bytes()
-        name = _resolved_name(raw, variables)
+        data = _parse(raw)
+        if data is None:
+            continue
+        name = _resolved_name(data, variables)
         if name is None:
             continue
-        fingerprints[name] = _digest(raw, _model_bytes(raw, root))
+        fingerprints[name] = _digest(raw, _model_bytes(data, root))
 
     return fingerprints
 
 
-def _resolved_name(raw: bytes, variables: dict[str, Any]) -> str | None:
+def _parse(raw: bytes) -> dict[str, Any] | None:
+    """Parse one sync file, or None if it is not a YAML mapping.
+
+    The single parse point for the module: everything downstream takes the
+    parsed mapping, so no helper has to re-check that it got one.
+    """
+    try:
+        data = yaml.safe_load(raw)
+    except Exception:  # noqa: BLE001 — malformed files are drt validate's job
+        return None
+    return data if isinstance(data, dict) else None
+
+
+def _resolved_name(data: dict[str, Any], variables: dict[str, Any]) -> str | None:
     """The sync's ``name:`` with env vars and ``var()`` applied, or None.
 
     Expansion is applied to the **name value alone**, never to the whole
@@ -102,16 +118,10 @@ def _resolved_name(raw: bytes, variables: dict[str, Any]) -> str | None:
     to avoid. Isolating the name keeps an unresolvable field elsewhere from
     deciding whether the sync is visible at all.
 
-    Only a file that is not parseable YAML, or has no ``name:``, is skipped —
-    reporting those is ``drt validate``'s job, and a selector that raised on an
-    unrelated malformed file would be worse than one that cannot see it.
+    Only a sync with no ``name:`` is skipped here — reporting a broken sync is
+    ``drt validate``'s job, and a selector that raised on an unrelated malformed
+    file would be worse than one that cannot see it.
     """
-    try:
-        data = yaml.safe_load(raw)
-    except Exception:  # noqa: BLE001 — malformed files are drt validate's job
-        return None
-    if not isinstance(data, dict):
-        return None
     name = data.get("name")
     if not name:
         return None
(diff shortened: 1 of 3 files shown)


masukai:
Self-review pass (adversarial, requested explicitly rather than a merge-blocking review).

**One real finding, minor:** the module docstring and the PR description both state "secrets cannot enter the input" as an absolute. That's true given the *recommended* pattern (`password_env: NAME` / `${VAR}` indirection) — but nothing in `sync_fingerprints()` prevents a sync YAML from holding a literal hardcoded value instead (`password: "actual-secret"`). `config/secrets.py`'s scanner flags that as a `drt validate` warning, but it's advisory, not a block — so a misconfigured file *would* have its literal secret hashed into the fingerprint. The digest itself doesn't reveal the value, but it's not the airtight "cannot" the docstring claims — "does not, given the `_env`/`${VAR}` pattern this assumes" is the more accurate framing. Wording fix, not a functional bug; low priority.

Checked one thing I suspected might be a real blind spot and it wasn't: whether drt's templating supports Jinja `{% include %}` / macro files beyond the single `ref()`'d model — if it did, a shared macro edit wouldn't move any fingerprint. Grepped `drt/templates/` — no `FileSystemLoader` or cross-file include mechanism exists, so this isn't a gap.

Everything else held up on a fresh read: the watermark-exclusion and undefined-var traps are exactly the kind of thing that's easy to regress silently, and both have a named regression test tied to the specific failure mode rather than just a happy-path assertion. The second commit's coverage fix (12 uncovered lines, all decision points for sync visibility rather than incidental) is the right thing to have prioritized over the happy path.

masukai (commented):

Traced the two claimed traps:

- **Watermark never reaches the hash**: confirmed — `_model_bytes` reads the `.sql` file straight off disk via `Path.read_bytes()`, never through `resolve_model_ref` (the only function that substitutes `last_cursor_value`). No path in `fingerprint.py` imports or calls the resolver's watermark-substituting entry point.
- **Undefined env var doesn't hide a sync**: confirmed — `expand_env_vars`/`expand_sync_vars` are applied only to `data.get("name")`, never to the full parsed document, so an unresolvable `${VAR}` in, say, a destination's `token_env` field can't take the whole sync out of the map the way whole-document expansion would.

Also checked that `project_vars()` is imported from the real parser rather than reimplemented — so variable-resolution semantics can't silently drift between `fingerprint.py` and `load_syncs()` over time.

No issues found.

Issue #762, opened 2026-07-10
feat(engine): first-class run_id + opt-in metadata columns (_drt_synced_at, _drt_run_id) on destinations


masukai:
## Problem

drt writes destination rows byte-for-byte from the source: no marker of *when* a row was synced, *which run* produced it, or *which sync* owns it. Consequences for operating reverse ETL:

- debugging "why is this CRM row stale?" requires cross-referencing local `.drt/history` files that CI runners throw away
- in DWH-to-DWH syncs (BigQuery → Snowflake/Databricks) consumers can't distinguish drt-loaded rows from other writers, and there's no lineage anchor
- DLQ records, history entries, OTel spans, and destination rows have no shared correlation ID — each failure investigation starts from scratch

## Prior art

- **dlt**: adds `_dlt_load_id` + `_dlt_id` to every loaded row; the load_id links rows ↔ load packages ↔ pipeline state.
- **Hightouch/Census**: per-run sync IDs in logs and warehouse writeback tables.
- **dbt**: `invocation_id` in all artifacts.

## Proposal

1. **First-class `run_id`** (UUID v7 per `drt run` invocation, per-sync `sync_run_id`): threaded through `SyncResult`, history entries, DLQ records, `--output json`, OTel span attributes, and failure alerts. Purely additive to the engine via `SyncObserver`.
2. **Opt-in metadata columns** for SQL/DWH destinations:

```yaml
sync:
  metadata_columns:
    synced_at: _drt_synced_at      # UTC load timestamp
    run_id: _drt_run_id
    sync_name: _drt_sync_name      # optional, off by default
```

Columns are appended at load time (engine-side, destination-agnostic dict enrichment), so they also flow into file/blob destinations. Requires the columns to exist on the target — or composes with `manage_table`/`on_schema_change: append_new_columns` from the managed-tables issue.

## Scope

**In:** run_id plumbing (state/history/DLQ/OTel/alerts), `metadata_columns` config + engine enrichment, docs.
**Out:** hash/row-version columns (diff-engine territory), automatic DDL (managed-tables issue).

## Related

- #429 / #619 (OTel spans — natural carrier for run_id), #278 (DLQ), #276 (history), #698 (docs manifest run history)
- dlt `_dlt_load_id` is the ecosystem-familiar naming precedent

## Milestone fit

run_id plumbing is small and unlocks observability work — candidate for v0.8.x; metadata columns ride with the managed-tables leg (v0.9).

masukai:
[...]

PR #902, opened 2026-08-03
feat(cli): drt serve delivery contract — coalescing, 202 + run id, pluggable auth (closes #854)

### LINKED ISSUE #854: feat(cli): drt serve concurrency contract — a dropped trigger is silent event loss

Pawansingh3889:
Closes #854. Implements ADR 0004 Tier 3 — the `drt serve` hardening — along the lines we settled in the issue thread.

## The contract

Written into the endpoint docs as an explicit promise, per your "make it legible to whoever wires a webhook to it":

- **A trigger accepted with `202` will run.** Nothing accepted is silently dropped; `423` is gone.
- **Different syncs run concurrently.** The old `_SyncLock` was global across all syncs; only same-sync overlap is serialized now.
- **Same-sync triggers coalesce to depth 1.** While a sync runs, the first new trigger creates one pending run; further triggers are answered with the *same* run id (`"coalesced": true`). Went with the coalescing idea from the thread rather than a deeper queue or `429`: drt syncs are watermark-driven, so one pending run picks up everything that accumulated — a queue of N would do the same work as a queue of 1. That reasoning holds *because* triggers are parameterless, and the commit message + CHANGELOG both say so explicitly, so if a trigger ever selects rows the decision gets revisited rather than silently outgrown.
- **Dry-run and real triggers coalesce separately** — folding a real trigger into a pending dry-run preview would silently discard the write.

## 202 + run id (your Q1 answer, applied)

`POST /sync/<name>` returns `202` + a run id immediately; `GET /runs/<id>` returns state (`pending`/`running`/`success`/`partial`/`failed`/`error`) and the result once finished. The id is **in-memory, valid for the lifetime of the process** — documented in the guide with #762 linked as the upgrade path, per your "note the limitation explicitly". `HistoryEntry` and ADR 0003's `SyncRun` are untouched.

`?wait=true` preserves the pre-#854 synchronous `200`/`207` semantics — existing callers (dbt Cloud post-job webhooks that read the response body) migrate with one query param. A `wait=true` request against a busy sync coalesces and blocks until the pending run completes, so the no-loss promise holds for synchronous callers too.

## Auth: pluggable, with the OIDC leg split out

`--auth none|bearer|hmac` (`auto` keeps the historical bearer-if-token-set default), dispatched per your framing so a later verifier is an addition, not a rewrite:

- **bearer** now compares via `secrets.compare_digest` (it was `==`)
- **hmac** verifies an HMAC-SHA256 body signature, accepting GitHub's `sha256=<hex>` (default header `X-Hub-Signature-256`), bare hex, and base64 — so Shopify works via `--hmac-header X-Shopify-Hmac-Sha256`. Stripe's timestamped scheme is documented as not covered.
- **OIDC** (the Pub/Sub JWT path) is split out as offered — filed as a follow-up issue with the dependency question laid out (PyJWT vs google-auth, extras placement). Until it lands the guide keeps the "Pub/Sub needs a verifying proxy" caveat.

## One bug found underneath

`run_drt_sync()` builds a fresh `StateManager` per call, and its thread-safety is an **instance** lock — so the moment different syncs run concurrently, per-request instances would race the load-modify-save on `state.json`. Added an optional `state_manager` param (default unchanged for Airflow/Prefect, which are one-run-per-process) and the server passes one shared instance.

## Also in the diff

- ADR 0004 gains a dated amendment (matching the 2026-07-29 precedent): Tier 3's blocker clears, OIDC named as the residual. The Context section's pre-#854 citations are deliberately left as-is — they're the evidence the decision was made on.
- `tests/unit/test_server.py` rewritten around the new contract: 25 tests covering coalescing (three triggers → exactly two executions, third returns the second's run id), per-sync independence, the `wait=true` block-until-done path, all three HMAC encodings + rejections, and the run lifecycle.
- Guide rewritten around the contract + status-code table; `docs/llm/CONTEXT.md` line updated; CHANGELOG entry flags the two behaviour changes (⚠️ `423` removed, ⚠️ async by default).

## Verification

- Full suite 2894 passed / 18 skipped; ruff + mypy clean.
- Live end-to-end smoke, not just unit-level: real `drt serve` process, DuckDB warehouse, file destination — 401 without token, 202 + run id, poll to success, `wait=true` → 200, rows actually written.


masukai (commented):
Read this properly rather than quickly — it's a lot of PR and most of it holds up well. Two things I'd like settled before approving, one of which is just a decision rather than work.

First, credit where it's due: the `run_drt_sync()` find is real and it is the *third* independent data point for the same underlying problem. `StateManager`'s thread-safety is an **instance** lock, so it only ever worked because every caller was one-run-per-process; the moment anything runs syncs concurrently the load-modify-save on `state.json` races. #756's field report hit this from the Cloud Run Jobs side, ADR 0005 (#899) argues it from the "drt has no durable state tier" side, and you hit it from inside the process. Passing a shared instance is exactly the right minimal fix here.

## `GET /runs/<id>` is unauthenticated

`do_GET` never calls `_check_auth` — it handles `/health`, then `/runs/<id>`, then 404s. So the full `result` (the SyncResult) and `error` (the exception string) are readable by anyone holding the run id, with no token.

I don't think this is a hole: the id is uuid4 and only ever handed to a caller who already passed auth on the POST, so it works as a capability URL. My problem is that it is **implicit**, in the one PR whose stated purpose is making the contract explicit:

- the guide's example is a bare `curl http://localhost:8080/runs/3f2a9c...` with no note that the absence of a token is deliberate
- the status table reads `401 — auth missing or wrong (when auth enabled)`, which a reader will apply to the whole endpoint surface
- drt already treats connector error text as sensitive by default elsewhere — #698 redacts embedded errors in the docs manifest unless asked otherwise

Worth saying plainly: the GET-is-unauthenticated convention was fine when `do_GET` returned a version string. It changes character now that GET returns sync results. Either add the auth check to `do_GET` (a couple of lines, and consistent with POST) or state the capability-URL reasoning in the guide next to that curl — I don't mind which, but I'd rather it were chosen than inherited.

## `coalesced` means something slightly different in the docstring and in the code

The module docstring:

> the first new trigger creates one pending run; **further triggers** while that run is pending are answered with the *same* run id (`"coalesced": true`)

That reads as: `coalesced: true` marks a trigger that joined an existing pending run. But `trigger()` also returns `coalesced=True` for the trigger that *creates* the pending run — which gets a fresh run id — and `test_same_sync_triggers_coalesce_to_depth_one` pins that (`second` and `third` both true, with `second` being the creator).

Both definitions are defensible; "you were not started immediately" is a perfectly good meaning for the flag. The issue is only that a client can't currently distinguish "my trigger queued work" from "my trigger joined queued work", while the docstring implies it can. Tightening the sentence is probably enough.

## Follow-up, not a blocker

`state_manager: StateManager | None` types the concrete local class. #900 (in flight) extracts a `StateStore` Protocol, and #756's remote backends implement it — so once that lands this parameter wants to be `StateStore`, or `drt serve` will be the one caller that structurally cannot use a remote state backend. The back-compat alias means nothing breaks in the meantime; worth a note in the code so it isn't missed when #900 merges.

## Minor

`run.state = result.get("status", "success")` — a status outside `_TERMINAL_STATES` would leave the run permanently un-evictable in `_runs`. Not reachable from today's runner, just brittle if the result contract grows a state.

Everything else I went looking for was already handled: the coalescing lock discipline (state transitions all happen under `self._lock`, so a concurrent `GET /runs/<id>` can't see a half-written run), separate dry-run/real pending slots, oldest-first pending selection by `seq`, `compare_digest` on both auth paths, and the eviction slice taking the oldest finished runs rather than the newest. The three-triggers-two-executions test is the right shape for proving coalescing rather than just asserting it.

masukai:
@Pawansingh3889 — heads up: this PR had picked up a conflict, so I merged `main` in for you rather than leaving you to untangle it (`d269c00`).

The conflict was **only in `CHANGELOG.md`**, and only because both sides added entries under `## [Unreleased]` — the v0.8.5 patch release ([#908](https://github.com/drt-hub/drt/issues/908)) and the drift-audit entry ([#871](https://github.com/drt-hub/drt/issues/871)) landed while this was open. Nothing of yours was rewritten: your `drt serve` entry is intact under `### Changed`, and I verified the resolution against both parents — 2 lines added vs `main`, 14 added vs your branch, **0 removed either way**. Your three commits are untouched; the merge sits on top.

Full suite passes on the merge result (2952 passed / 22 skipped).

Nothing to do on your side for the conflict. The [review comments](https://github.com/drt-hub/drt/pull/902#pullrequestreview-4849810259) are still the open items — mainly the two I flagged: `GET /runs/<id>` being reachable without auth, and the `coalesced` flag's meaning drifting between the docstring and the implementation. No rush, and no obligation to agree with either — push back if you read them differently.

One procedural note so it doesn't surprise you: because I was the last person to push here, branch protection won't let *my* approval count. That resolves itself as soon as you push your next commit — which you'll be doing anyway for the review points — so it changes nothing about the order of work.

Pawansingh3889:
Thanks for merging `main` in, and for reading it at that depth. All four points are addressed in 7124f4c.

## `GET /runs/<id>`

Went with the auth check rather than documenting the capability-URL reasoning. What settled it was a detail your write-up points at: the run id is in the URL *path*, so it is written to every reverse-proxy access log in front of drt. A capability that is logged by design is a weak one, and it left the read path weaker than the write path. Every route except `GET /health` now authenticates; the probe stays open so a load balancer or `docker healthcheck` needs no credential. The `401` is answered before the run lookup, so an unauthenticated caller cannot use the endpoint to probe which ids exist.

One wrinkle worth flagging, since it is a consequence rather than a design I would have chosen. Under `hmac` a GET has no body, so the signature is over the empty body, which makes it a constant per secret rather than a per-request signature. It still proves knowledge of the secret without putting the secret on the wire, but it is replayable in a way a body signature is not. The guide documents it with the `openssl` one-liner that computes it, and says plainly to prefer `--auth bearer` for the polling path if you would not hand out a replayable credential. If you would rather that case were rejected outright than supported with a caveat, I am happy to change it.

## `coalesced`

Took your read and tightened the sentence rather than splitting the flag. It now says "your trigger did not start a run of its own" in the module docstring, in `trigger()`, and in the guide, with the two cases called explicitly indistinguishable so nobody tries to infer which happened. Splitting it into created-vs-joined seemed like inviting clients to build on a distinction that does not mean anything to them, given the pending run covers both cases identically.

## `StateStore`

Did it rather than left a note, since #900 arrived with your merge. Widened through `_runner.py`, `engine/sync.py` and `engine/observer.py`. `LocalStateManager` satisfies the Protocol, so no call site moved and nothing changes at runtime.

## Eviction

Keyed on `finished_at` instead of membership of `_TERMINAL_STATES`, which retires the constant. `finished_at` is set under the scheduler lock for every run that completes, so it does not care what the runner's result contract calls the state. I checked the new test against the old predicate rather than assuming it discriminates: 5 runs retained where the cap allows 3, so it does fail without the fix.

ruff and mypy clean. Full suite green here at 2934 passed / 18 skipped, which is 6 tests up on the branch point. Your 2952/22 is presumably a wider set of extras installed than I have locally.

On the approval: pushed, so that should be unblocked whenever you next have time.

masukai:
Verified all four independently rather than taking the diff at face value — checked out `7124f4c`, ran the suite, and for the eviction fix specifically, reverted just that one predicate back to `_TERMINAL_STATES` and re-ran `test_finished_runs_evict_on_a_status_outside_the_known_set` to confirm it actually catches the regression it claims to:

```
AssertionError: assert 5 <= 3
```

Matches your "5 runs retained where the cap allows 3" exactly. Good test — a status string the eviction set doesn't know about is exactly the failure mode a whitelist invites, and pinning it with an unrecognized status (`"quiesced"`) rather than a real one is the right way to prove the fix doesn't secretly still depend on the state vocabulary.

`test_hmac_get_signs_the_empty_body` is the test I'd have asked for if you hadn't already written it — proving the POST's real-payload signature does *not* authenticate the GET is the part that actually matters, not just that some signature does.

## On the open question — keep it as documented, don't reject outright

Three reasons, roughly in order of weight:

1. It strictly improves on what shipped before (no auth at all on GET). A caveat on a real improvement isn't a reason to hold the improvement back.
2. Rejecting polling under `hmac` would force operators who chose HMAC to match GitHub/Shopify's webhook verification model into running a *second*, different auth scheme just to poll status — a mixed-scheme deployment is a worse footgun than a documented one-line caveat on a single scheme.
3. The actual blast radius is narrower than "the read API is open": exploiting it needs a header-level capture (the reverse-proxy-access-log threat model you were defending against only logs the URL path, not headers — this needs something more invasive, like a MITM or a logging proxy that dumps headers too) *and* separately knowing a target run id (still an unguessable uuid4). That's a meaningfully smaller attack surface than the thing you already fixed.

A real fix — timestamp + nonce in the signed payload, the way SigV4 or most webhook-with-polling designs do it — is legitimate but is a bigger change (new headers, clock-skew tolerance, a nonce cache) than this PR or ADR 0004's "hardened enough" Tier 3 scope calls for. Worth its own issue if polling-under-hmac sees real use and someone wants to close the gap properly; not worth blocking this on.

Approving.

masukai (approved):
Formal approval — see the comment above for verification detail and the HMAC decision. `require_last_push_approval` means my earlier COMMENTED review is stale regardless of content, since I merged `main` in on `d269c00`; this is the fresh one against `7124f4c`.

Issue #890, opened 2026-08-02
perf(destinations): scope-aware SQL diff for tracked+scoped mirror (#694 part 2 follow-up)


masukai:
## Problem

#694 part 2 (#886/#887/#888/#889) replaced the tracked-mirror diff's full-table Python read with a SQL-side `previous - current` computation (`NOT EXISTS`/`NOT IN` against a staged/parameterized set of this run's keys). This works as intended for **unscoped** tracked mirror: the diff is typically small (most runs only change a handful of rows), so only a small result set ever reaches Python.

For **`mirror.scope` + `strategy: tracked`** specifically — the composition #694 part 1 added, and the primary motivating "1:N parent/child regeneration" use case for scope in the first place — the SQL diff query has no scope predicate (it only filters by `sync_name`). Scope-filtering happens in Python *after* the SQL diff, on `raw_diff`. This is mathematically correct (the final `to_delete` set is right — `(previous ∩ scope) - current == (previous - current) ∩ scope`), but when `current` covers only one scope out of many historically-tracked scopes, `raw_diff` ends up being nearly the *entire* previous state table minus a small overlap — essentially the same volume the old (pre-#694-part-2) implementation read into Python. The optimization doesn't deliver its stated benefit for the scoped case.

Caught in review before #886-889 merged (not a data-corruption bug — final deletes are always correct — but the "large table's full key history is never read regardless of scope" claim in `docs/connectors/postgres.md` was false and has been corrected to describe this limitation honestly).

## Why it's hard

Filtering `previous` by scope *in SQL* would need scope values extracted from `key_json` (a JSON array string), since scope isn't stored as a separate state-table column — that constraint was deliberate (#694 part 1): storing scope as a column would need a migration story for `_drt_synced_keys` tables `CREATE TABLE IF NOT EXISTS`-created before #694. Extracting scope positions from `key_json` in a `WHERE` clause is possible but dialect-specific and non-trivial:

- Postgres: `(key_json::json->>N)`
- MySQL: `JSON_UNQUOTE(JSON_EXTRACT(key_json, '$[N]'))`
- Snowflake: `PARSE_JSON(key_json)[N]::STRING`
- ClickHouse: `JSONExtractString(key_json, N+1)` (1-indexed)
- Databricks: `get_json_object(key_json, '$[N]')`

[...]

Issue #695, opened 2026-07-03
docs(destinations): document destination privileges required by tracked mirror (+ consider existence-check before CREATE TABLE IF NOT EXISTS)


K-Masuda-SL:
## Field report

First production rollout of `mirror.strategy: tracked` (#686, MySQL destination) failed on a least-privilege setup with:

```
(1142, "CREATE command denied to user 'analytics-retl-stg'@'...' for table '_drt_synced_keys'")
```

The destination user had exactly the table-level grants the sync needed under `mode: full` (SELECT/INSERT/UPDATE on the target tables + SELECT on lookup tables) — a common hardening pattern when writing back into a product's operational database. Switching to tracked mirror silently added two privilege requirements that weren't obvious from the docs:

1. **DELETE on the target table** (the mirror delete pass) — easy to miss because the first run baselines with no deletes, so the missing grant only detonates on the *first real generation change*, potentially weeks later.
2. **CREATE (+ SELECT/INSERT/DELETE) on `_drt_synced_keys`** — and because the destination executes `CREATE TABLE IF NOT EXISTS` on every run, **pre-creating the state table does not avoid the CREATE requirement**: MySQL checks the CREATE privilege before the existence check.

## Ask 1 (docs)

Add a "required destination privileges" note to the tracked mirror docs, e.g. for MySQL:

```sql
GRANT SELECT, INSERT, UPDATE, DELETE ON `<db>`.`<target_table>` TO '<user>'@'%';
GRANT CREATE, SELECT, INSERT, DELETE ON `<db>`.`_drt_synced_keys` TO '<user>'@'%';
-- table-level grants work even before _drt_synced_keys exists
```

and call out the two traps above (delayed DELETE failure; IF NOT EXISTS still needs CREATE). Same story presumably applies to the Postgres leg.

## Ask 2 (optional enhancement)

Consider checking table existence (information_schema) before issuing the CREATE, so locked-down environments can pre-provision `_drt_synced_keys` by an admin and run the sync user without any DDL privilege. Fallback to the current behaviour when the table is absent. This would make tracked mirror usable under strict "no DDL for app users" policies without weakening its lazy-create default.

## Context

- Verified end-to-end after granting: first run logged `tracked mirror: no prior state for sync '<name>' in _drt_synced_keys — baselining this run's 46 key(s); no deletes this run.` as designed.
- drt-core 0.7.10, MySQL 8.0 (Cloud SQL).

masukai:
[...]

PR #904, opened 2026-08-04
perf(destinations): scope-aware SQL diff for tracked+scoped mirror — Postgres/MySQL (#890, 1 of 4)

masukai:
Closes the Postgres/MySQL leg of #890, following #694 part 2's one-PR-per-dialect split (#886–#889). Design discussion is in the [issue thread](https://github.com/drt-hub/drt/issues/890#issuecomment-5174573551) and its [correction](https://github.com/drt-hub/drt/issues/890#issuecomment-5174649294).

## The measurement

The issue asks for a number rather than an argument. Real Postgres 16, 500 scopes × 200 keys = 100,000 tracked rows, a run touching one scope and dropping 5 of its keys:

| | rows returned to Python | best of 3 |
|---|---|---|
| before (#694 part 2) | **99,805** | 44.4 ms |
| after (#890) | **5** | 6.6 ms |

~20,000× fewer rows crossing the wire. This is the case #694 part 2 was supposed to fix and didn't.

Third row of the same run, with half the state table left un-backfilled (`scope_spec`/`scope_key` NULL, i.e. written before this change):

| | | |
|---|---|---|
| after, un-backfilled rows mixed in | 22,205 | 14.6 ms |

Still correct — the NULL rows simply fall through to the Python filter, which is exactly today's behaviour for them.

<details><summary>Reproduction script</summary>

```python
# docker run -d --rm --name pg -e POSTGRES_PASSWORD=drt -e POSTGRES_DB=drt -p 55432:5432 postgres:16-alpine
import time, psycopg2
from drt.destinations._mirror_state import key_hash, key_json, scope_key_json, scope_spec_json

SCOPES, KEYS_PER_SCOPE, STALE = 500, 200, 5
SPEC = scope_spec_json(["parent_id"])
conn = psycopg2.connect(host="localhost", port=55432, dbname="drt", user="postgres", password="drt")
conn.autocommit = True
cur = conn.cursor()
cur.execute("CREATE TABLE _drt_synced_keys (sync_name VARCHAR(255) NOT NULL, key_hash CHAR(64) NOT NULL,"
            " key_json TEXT NOT NULL, scope_spec TEXT, scope_key TEXT, PRIMARY KEY (sync_name, key_hash))")
rows = [("scores_sync", key_hash((p, f"child-{k}")), key_json((p, f"child-{k}")), SPEC,
         scope_key_json((p, f"child-{k}"), [0]))
        for p in range(SCOPES) for k in range(KEYS_PER_SCOPE)]
cur.executemany("INSERT INTO _drt_synced_keys VALUES (%s,%s,%s,%s,%s)", rows)
cur.execute("CREATE INDEX ON _drt_synced_keys (sync_name, scope_key)"); cur.execute("ANALYZE _drt_synced_keys")

current = [(0, f"child-{k}") for k in range(KEYS_PER_SCOPE - STALE)]
cur.execute("CREATE TABLE __drt_mirror_diff_keys (key_hash VARCHAR(64), key_json TEXT)")
cur.executemany("INSERT INTO __drt_mirror_diff_keys VALUES (%s,%s)", [(key_hash(k), key_json(k)) for k in current])
cur.execute("CREATE INDEX ON __drt_mirror_diff_keys (key_hash)"); cur.execute("ANALYZE __drt_mirror_diff_keys")

BASE = ("SELECT s.key_hash, s.key_json FROM _drt_synced_keys s WHERE s.sync_name = %s "
        "AND NOT EXISTS (SELECT 1 FROM __drt_mirror_diff_keys c WHERE c.key_hash = s.key_hash)")
SCOPED = BASE + " AND (s.scope_key IS NULL OR s.scope_spec <> %s OR s.scope_key IN (%s))"
for sql, params, label in ((BASE, ("scores_sync",), "before"),
                           (SCOPED, ("scores_sync", SPEC, scope_key_json((0,"x"), [0])), "after")):
    t = time.perf_counter(); cur.execute(sql, params); got = cur.fetchall()
    print(f"{label}: rows={len(got):,} {(time.perf_counter()-t)*1000:.1f} ms")
```
</details>

## What changed

Two nullable columns on `_drt_synced_keys`, and the diff narrowed server-side:

```sql
AND (s.scope_key IS NULL          -- written before the columns existed
     OR s.scope_spec <> %s        -- written under a different mirror.scope
     OR s.scope_key IN (...))     -- spec matches, filter here
```

**The Python scope filter is unchanged and stays authoritative.** The SQL predicate is purely *coarse*: it may hand Python more rows than necessary, never fewer. So getting it wrong costs a slower run, not a wrong deletion — which is the property option 1 in the issue (per-dialect `JSON_EXTRACT` out of `key_json`) does not have, since a mis-written extraction on one dialect silently changes which rows are deletion *candidates*.

## Why two columns and not one

This is the part the first design comment got wrong, and it is worth reading before reviewing the predicate.

Today's code re-derives `scope_positions` from config on **every run**, so editing `mirror.scope` in YAML re-derives automatically. Freezing only the *value* into a column removes that: a row written under the old definition holds a `scope_key` no current observed scope can match, so it drops out of the diff and **stops being a deletion candidate — silently, and forever**. Not a slow run; a stale row that never goes away.

Storing the spec next to the value restores the invariant. Rows written under any other spec fall through to Python exactly as they do now.

> **Corrected after this PR was opened.** This paragraph originally ended "and the table self-heals after one run." **It does not** — #694 part 2 deliberately never rewrites an already-tracked row, so rows from before the columns existed keep their NULL scope forever and the filter would never engage on an upgraded state table. Caught by running against a real server, not by any mock. A later commit adds a **lazy, capped backfill** that heals those rows from the diff (they are already fetched and decoded, so it costs one UPDATE), bounded per run because expand/contract says to backfill in batches rather than inside the hot path — and a sync run is the hot path. Full write-up: [#890 comment](https://github.com/drt-hub/drt/issues/890#issuecomment-5175384610).

## Pre-#890 tables

Expand/contract, with the #695 lesson applied verbatim:

```
probe for the columns
  ├─ present             → filter in SQL
  ├─ absent, ALTER ok    → ADD COLUMN, then filter in SQL
  └─ absent, ALTER fails → stay on the Python-only filter, permanently, no error
```

The `ALTER` runs inside a `SAVEPOINT` so a refusal doesn't poison the transaction. Both columns are nullable and default-free — metadata-only on every engine drt targets, so no rewrite of a large existing state table.

## Verification

- 5 new tests, including the two that guard the failure modes above: the `scope_spec` escape hatch, and the ALTER-refused fallback
- Existing 107 mirror tests pass **unmodified** — no behaviour change for unscoped tracked mirror
- An unscoped sync never probes and never issues DDL (asserted)
- Full suite **2933 passed, 21 skipped**; `ruff` and `mypy` clean

## Follow-up

Snowflake, ClickHouse and Databricks each carry their own `_finalize_mirror_tracked` and follow as separate PRs. ClickHouse additionally needs `system.columns` for the probe — it is the one dialect `drt/destinations/schema.py` does not already cover.

The `⚠️` caveat currently in `docs/connectors/postgres.md` and the four dialect docstrings should be narrowed to "the remaining dialects" as each leg lands, and dropped when the last one does.

### DIFF
--- a/drt/destinations/_mirror_state.py
+++ b/drt/destinations/_mirror_state.py
@@ -60,6 +60,29 @@ def key_hash(key: tuple[Any, ...]) -> str:
     return hashlib.sha256(key_json(key).encode()).hexdigest()
 
 
+def scope_spec_json(scope_cols: list[str]) -> str:
+    """Canonical JSON of the scope *column names* (#890).
+
+    Stored next to ``scope_key`` so a run can tell whether a persisted scope
+    value was computed under the scope definition currently configured.
+    Without it, editing ``mirror.scope`` in YAML would strand every previously
+    written row: its frozen ``scope_key`` matches no observed scope, so it
+    drops out of the diff and is never a deletion candidate again — silently,
+    and forever. Today's code re-derives scope positions from config on every
+    run, so it has no such failure; the spec column is what keeps that true.
+    """
+    return json.dumps(list(scope_cols), separators=(",", ":"))
+
+
+def scope_key_json(key: tuple[Any, ...], scope_positions: list[int]) -> str:
+    """Canonical JSON of the scope *values* pulled out of a key tuple (#890).
+
+    Same encoding as :func:`key_json`, so the value stored in the state table
+    is byte-comparable with the one built from a run's observed scopes.
+    """
+    return key_json(tuple(key[p] for p in scope_positions))
+
+
 def diff_keys(
     previous: dict[str, str], current: list[tuple[Any, ...]]
 ) -> list[tuple[Any, ...]]:
--- a/drt/destinations/_mirror_state.py
+++ b/drt/destinations/_mirror_state.py
@@ -38,6 +38,13 @@ STATE_TABLE = "_drt_synced_keys"
 # user-configured identifier, so it never needs Composable-safe quoting.
 DIFF_STAGING_TABLE = "__drt_mirror_diff_keys"
 
+# How many pre-#890 state rows one run may heal (see the backfill note in
+# ``_finalize_mirror_tracked``). Bounded on purpose: the expand/contract
+# guidance is to backfill in batches rather than in one pass inside the hot
+# path, and a sync run *is* the hot path here. A state table converges over a
+# few runs instead of one run paying for the whole history.
+SCOPE_BACKFILL_PER_RUN = 5_000
+
 
 def decode_key(key_json_str: str) -> tuple[Any, ...]:
     """Inverse of ``key_json`` — a diffed state row's JSON back to a key tuple."""
--- a/drt/destinations/mysql.py
+++ b/drt/destinations/mysql.py
@@ -332,8 +332,28 @@ class MySQLDestination(BaseSqlDestination):
             "sync_name VARCHAR(255) NOT NULL, "
             "key_hash CHAR(64) NOT NULL, "
             "key_json TEXT NOT NULL, "
+            "scope_spec TEXT, "
+            "scope_key TEXT, "
             "PRIMARY KEY (sync_name, key_hash))"
         )
+
+    def _state_scope_columns_exist(self, cur: Any, scope: Any, raw: str) -> bool:
+        from drt.destinations._mirror_state import STATE_TABLE
+
+        schema_clause = "table_schema = %s" if scope is not None else "table_schema = DATABASE()"
+        params = (scope, STATE_TABLE) if scope is not None else (STATE_TABLE,)
+        cur.execute(
+            "SELECT COUNT(*) FROM information_schema.columns "
+            f"WHERE {schema_clause} AND table_name = %s "
+            "AND column_name IN ('scope_spec', 'scope_key')",
+            params,
+        )
+        return bool(cur.fetchone()[0] == 2)
+
+    def _add_state_scope_columns(self, cur: Any, ident: Any) -> None:
+        # MySQL has no ADD COLUMN IF NOT EXISTS; the caller only reaches this
+        # after the probe reported both columns absent.
+        cur.execute(f"ALTER TABLE {ident} ADD COLUMN scope_spec TEXT, ADD COLUMN scope_key TEXT")
 
     def _state_sql(self, template: str, ident: Any) -> Any:
         return template.format(ident)
--- a/drt/destinations/postgres.py
+++ b/drt/destinations/postgres.py
@@ -393,18 +393,15 @@ class PostgresDestination(BaseSqlDestination):
         )
 
     def _state_scope_columns_exist(self, cur: Any, scope: Any, raw: str) -> bool:
-        from drt.destinations._mirror_state import STATE_TABLE
-
-        sql = (
-            "SELECT COUNT(*) FROM information_schema.columns "
-            "WHERE table_name = %s AND column_name IN ('scope_spec', 'scope_key')"
+        cur.execute(
+            "SELECT COUNT(*) FROM pg_attribute "
+            "WHERE attrelid = to_regclass(%s) "
+            "AND attname IN ('scope_spec', 'scope_key') "
+            "AND NOT attisdropped",
+            (raw,),
         )
-        params: tuple[Any, ...] = (STATE_TABLE,)
-        if scope is not None:
-            sql += " AND table_schema = %s"
-            params = (STATE_TABLE, scope)
-        cur.execute(sql, params)
-        return bool(cur.fetchone()[0] == 2)
+        row = cur.fetchone()
+        return bool(row is not None and row[0] == 2)
 
     def _add_state_scope_columns(self, cur: Any, ident: Any) -> None:
         from psycopg2 import sql as _pgsql
--- a/drt/destinations/postgres.py
+++ b/drt/destinations/postgres.py
@@ -386,7 +386,33 @@ class PostgresDestination(BaseSqlDestination):
                 "sync_name VARCHAR(255) NOT NULL, "
                 "key_hash CHAR(64) NOT NULL, "
                 "key_json TEXT NOT NULL, "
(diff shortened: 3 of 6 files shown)


masukai:
Self-review pass (adversarial, requested explicitly — not a merge-blocking review).

Traced two things carefully enough to be confident, both held up:

**Concurrent \`ALTER TABLE ... ADD COLUMN\` under \`--threads N\`.** If two threads race \`_state_scope_columns_exist\` → both see \`False\` → both attempt the ALTER, the loser gets a duplicate-column error, which the existing \`except Exception\` catches and rolls back to the savepoint — that thread's *current* run just skips the SQL-side optimization once, but the table isn't corrupted and the next run sees the columns via the probe. Handled correctly, same fallback path as the no-privilege case.

**Whether the backfill (2nd commit) actually reaches the case it's meant for.** Almost flagged this as a real gap — my first pass through the SQL thought the backfill misses continuously-present ("stable") rows, since \`raw_diff\`'s \`NOT EXISTS\` clause excludes anything still in the current source regardless of \`scope_key\`. That's true, but it's also *irrelevant*: a stable row never becomes a delete-candidate, so it never needs its scope backfilled to shrink \`raw_diff\` — nothing about it is slow either way. What the backfill actually targets is the *absent-but-out-of-scope* row (present sometimes, protected by the Python scope check when absent), which is exactly the population that keeps re-appearing in \`raw_diff\` every run without it. Re-traced the SQL construction line by line to be sure before writing this — the design is correct, I'd just misread which row population it operates on.

Nothing else survived scrutiny as worth raising. The \`scope_spec <>\` escape hatch, the SAVEPOINT-per-ALTER pattern, and the "rows about to be deleted are skipped" backfill guard (\`doomed = {key_hash(k) for k in to_delete}\`) all check out against what the comments claim they do.

masukai (commented):
Self-review (adversarial pass, requested explicitly). The measurement and the design writeup (two-column freeze, coarse-filter invariant, corrected backfill note) all check out — traced the `to_delete`/`stale_scope`/backfill interaction line by line and it's correct, including the empty-`current` edge case (falls through to a full unscoped-shaped read, same as the documented #890 caveat, never an incorrect delete).

One real gap: `_state_scope_columns_exist` for Postgres, when `config.table` is unqualified (`scope is None` — the common case, e.g. a `public`-schema table), doesn't filter by `table_schema` at all:

```python
sql = "SELECT COUNT(*) FROM information_schema.columns WHERE table_name = %s AND column_name IN ('scope_spec', 'scope_key')"
params: tuple[Any, ...] = (STATE_TABLE,)
if scope is not None:
    sql += " AND table_schema = %s"
    ...
```

`information_schema.columns` is privilege-filtered, not search-path-scoped — it lists matching columns from every schema the connected role has visibility into, not just the one that would actually resolve for an unqualified reference. Compare `_state_table_exists`, two functions above it, which gets this right via `to_regclass(raw)` — and `_join_qualified(None, STATE_TABLE)` confirms `raw` is the same bare, unqualified name in this case, so `to_regclass` and every DML statement here resolve through `search_path` identically. The columns probe is the one place that doesn't reuse that resolution.

Concretely: a role with grants on more than one schema (tenant-per-schema is the realistic case), where `_drt_synced_keys` exists in more than one of them and only one is already migrated — the probe can report "columns exist" for a schema whose *actual* search-path-resolved table doesn't have them. `scope_sql=True` gets set with no ALTER attempted, and the next `SELECT s.scope_key FROM ...` fails with `UndefinedColumn` — a crash, not the graceful "stay on the Python-only filter" degradation the ALTER-refused branch is built for. MySQL's version of this probe doesn't have the gap — it explicitly falls back to `table_schema = DATABASE()` when `scope is None` — which is the tell that this is a one-line miss on the Postgres side rather than an intentional asymmetry.

Suggest resolving via the same OID `_state_table_exists` already computes, e.g. joining `pg_attribute` on `attrelid = to_regclass(%s)` with `raw`, rather than a schema-name string match.

No test currently exercises `scope=None` against a real multi-schema Postgres — the existing tests patch `_state_scope_columns_exist` directly, so this class of bug wouldn't show up there (mocks can't validate the SQL, only the code paths around it).

Everything else in this PR — the SAVEPOINT/rollback handling, the INSERT/backfill branching, the `scope_key_json`/`key_json` encoding consistency between the SQL predicate and the write paths — traced clean.

masukai:
**Codex review:** (adversarial review requested on the schema-probe fix commit)

> Verdict: needs-attention
>
> Do not ship without a real regression test. The implementation appears correct, but the new test does not exercise the reported unqualified-table, multi-schema failure.
>
> **[medium] Regression test never reproduces the schema-resolution bug** (`tests/unit/test_postgres_mirror_mode.py:940-951`)
> The original false-positive occurs when `scope is None` and the unqualified state-table name must resolve through PostgreSQL's `search_path`. This test instead passes `scope="tenant_a"` and a schema-qualified `raw`, a case where the old implementation already constrained `information_schema.columns` to the correct schema. Moreover, `fetchone()` is hard-coded to `(2,)`, so the assertion cannot demonstrate that a same-named table in another schema is ignored; it only checks for SQL substrings. A broken query containing those substrings could pass while still counting the wrong relation, leaving scoped tracked-mirror runs vulnerable to the original `UndefinedColumn` failure.
>
> Recommendation: Add a PostgreSQL-backed regression test with two schemas containing `_drt_synced_keys`, only the non-resolved table having `scope_spec` and `scope_key`. Set `search_path` to resolve the legacy table, call the probe with `scope=None` and `raw="_drt_synced_keys"`, and assert it returns false; then reverse the search path and assert true.

Agreed — the mocked test proves the SQL text changed, not that the resolution semantics did. Following up with a real multi-schema regression test (likely via `testcontainers-python` for an ephemeral local Postgres, matching the fix's own reproduction script) rather than the mock-only version currently in place.

PR #905, opened 2026-08-04
perf(destinations): scope-aware SQL diff — Snowflake (#890, 2 of 4)

masukai:
**Draft, stacked on #904.** Base is `perf/890-scope-aware-sql-diff`, so the diff here is the Snowflake delta only. Retarget to `main` before merging, per CONTRIBUTING's stacked-merge rule.

The design argument, the two-column rationale, and the measurement (99,805 → 5 rows returned to Python, ~20,000×) are all on **#904** — please review them there rather than here. This PR is the dialect port, and what is worth reviewing *here* is whether the port is faithful and whether the two Snowflake-specific decisions below are right.

## Two dialect differences

**No `BaseSqlDestination`.** `SnowflakeDestination` is a standalone class, so the `_state_scope_columns_exist` / `_add_state_scope_columns` seams #904 adds to the base are not reachable. The probe and the `ALTER` are inline, against `<database>.information_schema.columns` with **upper-cased** identifiers, since Snowflake folds unquoted names.

**No `SAVEPOINT`.** #904 wraps the `ALTER` in one because a refused DDL aborts a Postgres transaction and would poison everything after it. Snowflake has no savepoints — and does not need one here: this connection autocommits (the existing note in `_finalize_mirror_tracked` already says so), so a refused `ALTER` fails on its own with nothing half-applied. Plain `try`/`except` is the correct shape rather than a weaker substitute for something unavailable.

## Unchanged from #904

- The Python scope filter is untouched and remains authoritative
- The predicate keeps both escape branches — `scope_key IS NULL` (pre-#890 rows) and `scope_spec <> %s` (rows written under a different `mirror.scope`) — so it can only ever return too many rows, never too few
- Pre-existing state tables are handled by expand/contract, with a refused `ALTER` meaning "stay on the Python-only filter, permanently, no error"

## Verification

- 3 new tests: SQL narrowing with both escape branches asserted, the ALTER-refused fallback, and an unscoped run never probing
- **113 existing Snowflake tests pass unmodified**
- Full suite **2936 passed, 21 skipped**; `ruff` and `mypy` clean

No live-warehouse run yet. #904's measurement was on real Postgres; the equivalent against the smoke Snowflake account is worth doing before this leaves draft, since Snowflake is where the row-transfer cost actually shows up in credits.

## Remaining

ClickHouse (3 of 4) and Databricks (4 of 4) follow. ClickHouse needs `system.columns` rather than `information_schema`, and its diff has a different shape to begin with (`Array(String)` bound parameter, no staging table), so it is the least mechanical of the three.

### DIFF
--- a/drt/destinations/snowflake.py
+++ b/drt/destinations/snowflake.py
@@ -638,6 +638,7 @@ class SnowflakeDestination:
 
         from drt.destinations._mirror_state import (
             DIFF_STAGING_TABLE,
+            SCOPE_BACKFILL_PER_RUN,
             STATE_TABLE,
             decode_key,
             key_hash,
@@ -742,8 +743,13 @@ class SnowflakeDestination:
                 #   scope_key IS NULL → written before the columns existed
                 #   scope_spec <> ... → written under a different mirror.scope,
                 #                       so its frozen scope_key means nothing
+                projection = (
+                    "s.key_hash, s.key_json, s.scope_key"
+                    if scope_sql
+                    else "s.key_hash, s.key_json"
+                )
                 diff_sql = (
-                    f"SELECT s.key_hash, s.key_json FROM {state_fq} s WHERE s.sync_name = %s "
+                    f"SELECT {projection} FROM {state_fq} s WHERE s.sync_name = %s "
                     f"AND NOT EXISTS (SELECT 1 FROM {diff_table} c WHERE c.key_hash = s.key_hash)"
                 )
                 diff_params: list[Any] = [sync_name]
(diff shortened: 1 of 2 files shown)


masukai:

Core logic already traced carefully on #904 (Postgres/MySQL) — same scope predicate, same backfill targeting (absent-but-out-of-scope rows, not stable ones), same "rows about to be deleted are skipped" guard. This leg replicates it correctly; not re-deriving that reasoning here.

Checked the one thing specific to this dialect: the \`information_schema.columns\` probe uppercases both \`config.schema_\` and \`STATE_TABLE\` to match Snowflake's case-folding of unquoted identifiers. \`state_fq\` (\`f"{config.database}.{config.schema_}.{STATE_TABLE}"\`) is itself built unquoted, so the table genuinely lands in Snowflake under its uppercased name — the assumption is internally consistent with how the identifier is constructed elsewhere in this same file, not a new risk this PR introduces.

No SAVEPOINT on this leg (unlike Postgres) because the connection autocommits — correctly noted in the comment, and matches how the rest of this destination already talks to Snowflake outside of #890.

Nothing else stood out.

masukai (commented):

No issues found.

PR #906, opened 2026-08-04
perf(destinations): scope-aware SQL diff — ClickHouse (#890, 3 of 4)

masukai:
**Draft, stacked on #905 → #904.** Retarget to `main` before merging, per CONTRIBUTING's stacked-merge rule.

Design argument, two-column rationale and the measurement are on **#904** — review them there. What is worth reviewing *here* are the three dialect differences, the first of which is a correctness decision rather than a style one.

## `Nullable(String)`, not `String`

ClickHouse's default `String` has no NULL. The predicate needs to distinguish:

- **written before these columns existed** → must fall through to the Python filter
- **written with an empty scope value** → a real scope value that happens to be `""`

A plain `String` collapses both to `''`, and the `scope_key IS NULL` branch stops firing for pre-#890 rows — which would quietly exclude rows the Python filter would have kept, i.e. exactly the failure this design exists to prevent. `Nullable(String)` is load-bearing here, not a default choice.

## `system.columns`, not `information_schema`

Modern ClickHouse ships an `information_schema` compatibility shim, but `system.columns` is the native catalogue — the same reason `drt/destinations/schema.py` covers Postgres/MySQL/Snowflake/Databricks and not this dialect. Scoped with `currentDatabase()`, matching how `state_q` is resolved.

## No `SAVEPOINT`

As on the Snowflake leg, and unlike Postgres/MySQL: there is no open transaction here to poison, so a refused `ALTER` fails on its own and plain `try`/`except` is the right shape.

## The diff keeps its ClickHouse form

No staging table — `Array(String)` bound parameters already carry the whole current-key-hash set as one value. The scope predicate is three more parameterised terms (`{scope_spec:String}`, `{scope_keys:Array(String)}`), never interpolation.

## ⚠️ Deliberately WITHOUT the scope backfill — decided

The other three legs gained one after the live pass on #904 found that rows tracked *before* the columns existed stay NULL forever, so the filter never engages on an upgraded state table. There the fix is cheap: the rows are already fetched and decoded, so healing them costs one `UPDATE`.

**On ClickHouse that argument does not carry** — the same statement is `ALTER TABLE ... UPDATE`, a mutation that rewrites parts, and ClickHouse's own idiom is that `ADD COLUMN` stays metadata-only precisely so nothing gets rewritten. Asked @yodakanohoshi [in the comment below](https://github.com/drt-hub/drt/pull/906#issuecomment-5175425100) — no obligation, and none arrived. Settling on **no backfill** (option 2 from that comment): correct by inference from #694 part 2's never-rewrite behaviour, which is engine-agnostic, even though this specific leg hasn't had its live pass yet. **ClickHouse benefits only for keys added after the upgrade**; older keys keep falling through to the Python filter until an operator re-baselines. Now written down at both ends — the rationale in-code next to the diff (`drt/destinations/clickhouse.py`) and the operator-facing consequence in `docs/connectors/clickhouse.md`'s scoped-mirror section — so this doesn't need to live only in this thread. Reopenable if @yodakanohoshi's operational read differs once seen.

## Verification

- 3 new tests: the narrowed SQL with both escape branches and the bound parameters asserted, the ALTER-refused fallback, and an unscoped run never probing
- **101 existing ClickHouse tests pass unmodified**
- Full suite **2939 passed, 21 skipped**; `ruff` and `mypy` clean

Not yet run against a real ClickHouse — batched with the other dialects (see #904's thread).

## Remaining

Databricks (4 of 4).


masukai:
@yodakanohoshi — one narrow ClickHouse question, if you have a view. No obligation, and it does not block anything today.

Running this leg against a real ClickHouse turned up something the mocks could not: the `ALTER` applies and new keys get their scope columns, but **every row tracked before the upgrade stays NULL forever**. That is #694 part 2's own optimisation doing its job — an already-tracked row is never rewritten, so there is no write for the new columns to ride along with. On an existing state table the filter would then almost never engage, which is the whole scenario this issue exists for. Full write-up in [the issue](https://github.com/drt-hub/drt/issues/890#issuecomment-5175384610).

The fix on the other three dialects is a **lazy, capped backfill**: the rows needing it are already in Python (they come back in `raw_diff`, already decoded, with their scope already computed for the Python-side filter), so healing them costs an `UPDATE` and nothing else, spread over a few runs rather than one big write.

**On ClickHouse that reasoning does not obviously carry**, which is where I would value your read:

- `ALTER TABLE ... UPDATE` is a mutation that rewrites parts, so "cheap because the rows are already in hand" stops being true — the cost lands on the write side, not the read side
- ClickHouse's own idiom points the other way: `ADD COLUMN` is metadata-only and existing rows take the default at read time, precisely so nothing gets rewritten

So the options I see:

1. **Backfill anyway**, capped per run, and accept mutations on the state table
2. **Do not backfill on ClickHouse** — new keys benefit, pre-existing rows keep falling through to Python — and document it, pointing operators who want it at a one-off `ALTER TABLE ... UPDATE` they run themselves
3. Something I have not thought of

I lean to **2**: a per-dialect difference is uncomfortable, but it is grounded in a real engine difference rather than in convenience, and forcing uniformity would mean picking the un-ClickHouse-like option for uniformity's sake.

There is a third route I looked at and rejected — adding the column with a `DEFAULT` JSON-extraction expression, which would give existing rows a correct value at read time with no rewrite at all. Elegant, but the expression becomes load-bearing for correctness: get it wrong and rows get a wrong non-NULL scope value, drop out of the diff, and silently stop being deletion candidates. That breaks the invariant the whole design rests on — the SQL predicate must only ever be a *coarse* filter — so I would rather pay the mutation or skip the backfill than buy it back with a JSON expression.

If your instinct differs on the mutation cost, that is the input I am after — you know the operational side of ClickHouse mutations better than I do.

masukai:
Two days, no obligation stated and none arrived — closing this out per the fallback we'd agreed: **option 2, no backfill**.

Written down in `25e8581` so it doesn't only live in this thread:
- the engine-rationale (mutation vs. metadata-only `ADD COLUMN`) as a comment right next to the diff in `drt/destinations/clickhouse.py`
- the operator-facing consequence (older keys keep falling through to Python; re-baseline if you want them narrowed too) in `docs/connectors/clickhouse.md`'s scoped-mirror section

Worth being precise about what "decided" means here: this is inferred from #694 part 2's never-rewrite behavior (engine-agnostic Python), not measured — this leg still hasn't had its live-ClickHouse pass, which stays open and tracked separately (batched with the rest of the stack, see #904's thread). Not printing the manual `ALTER TABLE ... UPDATE` an operator would run by hand for the same reason: haven't run it against a real server, and getting the JSON-extraction shape wrong there is exactly the kind of thing a mock can't catch.

@yodakanohoshi if you land on this later and your operational read differs, this is a comment away from being reopened — nothing about the current shape forecloses adding the backfill.

masukai:
Self-review pass (adversarial, requested explicitly — not merge-blocking) — noting this one's already had substantial scrutiny earlier in the same session (the no-backfill decision + rationale is in the thread above, and both the code comment and \`docs/connectors/clickhouse.md\` were updated as part of it). Not re-deriving that here.

One thing worth cross-checking now that I've traced the PG/MySQL/Snowflake backfill logic carefully (#904/#905): those legs only backfill rows that are *absent-but-out-of-scope* (delete-candidates the Python filter protects), never continuously-present ones — a distinction I initially got wrong on #904 before re-tracing the SQL. That distinction doesn't matter here since ClickHouse has no backfill at all, by design — every NULL-\`scope_key\` row, stable or not, keeps falling through to the Python filter until an operator rebaselines. Simpler than the other three legs, and consistent with what's already documented.

No new findings.

masukai (commented):

No issues found.

Issue #734, opened 2026-07-10
perf(databricks): row-per-statement writes — 300-key mirror smoke took ~19 min (one HTTP round trip per row)


masukai:
## Observation (live smoke, #732 validation run)

The [#674 smoke harness run validating #707](https://github.com/drt-hub/drt/actions/runs/29062256383) surfaced a real-world perf gap: the Databricks mirror smoke with **300 observed keys took ~18.7 minutes** (complex-type test passed at 01:24:52, mirror at 01:43:34). The committed smoke was shrunk to 20 keys (#732) to keep the harness fast — with 20 keys the whole 5-test job is ~3 min — but the underlying cost hits production syncs proportionally.

## Cause

Every Databricks write path issues **one `cur.execute` per row**, and on a SQL warehouse each statement is a full HTTP round trip (~1.9 s/statement observed):

- insert mode — `drt/destinations/databricks.py:249-251`
- merge staging load — `databricks.py:298-300`
- replace-swap shadow load — `databricks.py:446-448`
- mirror key staging (new in #707) — `databricks.py:559-560`

The 300-key mirror smoke ran ~600 single-row statements (300 merge staging + 300 mirror keys) → ~19 min. A 10k-row production sync in `mode: merge` + `mode: mirror` would be ~20k round trips.

## Proposal

Chunk the row loops into **multi-row `INSERT INTO … VALUES (?, ?), (?, ?), …`** statements, sized to the native paramstyle's per-statement parameter limit (255 markers): `chunk_rows = max(1, 255 // n_cols)`. For the mirror key staging (1–2 key columns) that's a 127–255× reduction in round trips.

Caveats:
- The schema-aware `from_json`/`parse_json` paths (#317) use the `INSERT INTO … SELECT` form, which doesn't multi-row trivially — either `UNION ALL` chaining or keep row-by-row for json-wrapped tables in a first pass. The all-scalar `VALUES` path (the common case) and the mirror key staging (always scalar keys) are the high-value, low-risk targets.
- Keep per-row failure attribution semantics in mind: today a failing row is identifiable by loop index; chunking coarsens that to chunk granularity (same trade-off other SQL destinations already accept via `executemany`).

## Evidence

- 300-key run: [run 29062256383](https://github.com/drt-hub/drt/actions/runs/29062256383) — `5 passed in 1195.73s (0:19:55)`
[...]

PR #907, opened 2026-08-04
perf(destinations): scope-aware SQL diff — Databricks (#890, 4 of 4)

masukai:
**Draft, stacked on #906 → #905 → #904.** Retarget to `main` before merging, per CONTRIBUTING's stacked-merge rule. Completes the dialect set.

Design argument, two-column rationale and the measurement are on **#904** — review them there. What is specific to this leg:

## Delta's DDL spelling is `ADD COLUMNS (...)`

Plural and parenthesised, unlike `ADD COLUMN` on all three other legs. There is a test pinning the exact statement, not just the fallback behaviour — because getting it wrong would be **invisible from the outside**: the `ALTER` throws, the `except` swallows it as "no privilege", the run stays on the Python-only filter, and the optimisation would silently never engage on any Databricks deployment. That failure mode looks exactly like a correctly-handled permission refusal.

## Unity Catalog probe

`{catalog}.information_schema.columns`, matching the shape `drt/destinations/schema.py` already uses for this dialect.

## No `SAVEPOINT`

As on Snowflake and ClickHouse; only Postgres/MySQL need one.

## Multi-row `VALUES` batching

The #734 batching had a hardcoded width of 3. It is now driven by which column set is being written, so `_rows_per_chunk` still bounds the statement correctly when five columns are in play.

## ⚠️ One existing test changed

The only one across all four legs. A scoped tracked run now records `scope_spec`/`scope_key`, so its state INSERT carries five values rather than three, and `test_tracked_scoped_first_touch_of_a_scope_is_not_a_baseline_warning_databricks` asserted the exact three-value list.

The assertion was **widened, not relaxed** — the two new values are asserted as well, so the change is visible in the test rather than absorbed by loosening it.

## Also carries the scope backfill

Added after the live pass on #904 found that rows tracked *before* the columns existed stay NULL forever — #694 part 2 never rewrites an already-tracked row, so on an upgraded state table the filter would never engage. Delta takes a plain `UPDATE`, so the "the rows are already in hand" argument holds here as it does on Postgres/MySQL/Snowflake. Capped per run; rows about to be deleted are skipped.

⚠️ That narrows a #694 part 2 guarantee: an out-of-scope row may now be touched **once**, by the backfill, while its scope is still NULL. It is still never deleted and never re-inserted. The test that pinned the old property was tightened rather than dropped.

## Verification

- 3 new tests, including the `ADD COLUMNS` spelling pin
- Full suite **2942 passed, 21 skipped**; `ruff` and `mypy` clean

## Live verification (batched, still owed)

Per the discussion on #904, the four legs get checked against real engines together rather than one at a time:

| dialect | how | status |
|---|---|---|
| Postgres | docker | ✅ done — the measurement on #904 |
| MySQL | docker | pending |
| ClickHouse | docker | pending |
| Snowflake | #654 smoke account | pending — the one place row-transfer cost shows up in credits |
| Databricks | #654 smoke (Free Edition) | pending |

Unit tests prove the code runs; they do not prove the SQL parses on a real engine. None of these should leave draft before that pass.

### DIFF
--- a/drt/destinations/databricks.py
+++ b/drt/destinations/databricks.py
@@ -835,6 +835,7 @@ class DatabricksDestination:
 
         from drt.destinations._mirror_state import (
             DIFF_STAGING_TABLE,
+            SCOPE_BACKFILL_PER_RUN,
             STATE_TABLE,
             decode_key,
             key_hash,
@@ -945,8 +946,11 @@ class DatabricksDestination:
                 # ever return too many rows, never too few:
                 #   scope_key IS NULL → written before the columns existed
                 #   scope_spec <> ... → written under a different mirror.scope
+                projection = (
+                    "s.key_hash, s.key_json, s.scope_key" if scope_sql else "s.key_hash, s.key_json"
+                )
                 diff_sql = (
-                    f"SELECT s.key_hash, s.key_json FROM {state_fq} s WHERE s.sync_name = ? "
+                    f"SELECT {projection} FROM {state_fq} s WHERE s.sync_name = ? "
                     f"AND NOT EXISTS (SELECT 1 FROM {diff_table} c WHERE c.key_hash = s.key_hash)"
                 )
                 diff_params: list[Any] = [sync_name]
@@ -959,7 +963,9 @@ class DatabricksDestination:
                     )
                     diff_params = [sync_name, scope_spec, *observed_json]
                 cur.execute(diff_sql, diff_params)
-                raw_diff = cur.fetchall()
+                fetched = cur.fetchall()
+                stale_scope = [(h, kj) for h, kj, sk in fetched if sk is None] if scope_sql else []
+                raw_diff = [tuple(row)[:2] for row in fetched]
 
                 if scope_positions is not None and observed_scopes is not None:
                     to_delete = [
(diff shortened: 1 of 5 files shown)


masukai:
Force-pushed (`5472057` → `9cb0a2f`) to remove a repo-wide format sweep that was baked into this branch.

`ed917a9` ("backfill scope columns for pre-#890 rows, Databricks") carried **117 files**, of which **115 were pure `ruff format` output** on files this PR has nothing to do with — `drt/alerts/*`, `drt/cli/*`, `drt/docs/*` and so on. The same sweep reached `main` through the v0.8.5 release PR ([#909](https://github.com/drt-hub/drt/pull/909)) and was reverted there ([#910](https://github.com/drt-hub/drt/pull/910)) — but reverting `main` did nothing about the copy living in this branch's history. Rebasing onto the post-revert `main` simply made it a live diff again, so **merging this PR would have re-applied the entire sweep**. This repo is deliberately not format-clean, so that is a real regression, not a cosmetic one.

The classification wasn't eyeballed: for each of the 117 files I ran `ruff format` over the *parent* blob and compared it to the child blob. 115 matched exactly (⇒ the commit's only change to them was formatting) and 2 did not — `drt/destinations/databricks.py` and `tests/unit/test_databricks_destination.py`, which are the actual backfill. `ed917a9` was rebuilt with those two files only, message and authorship reused, and the two smoke commits cherry-picked on top.

Verification: the diff between old and new tips is exactly those 115 format-only files and nothing else — no `#890`-relevant file differs. Against `main`, this branch touches the 15 files it should. Every commit in the stack is now 1–5 files. `ruff check` + `mypy` clean, full suite **2944 passed / 24 skipped**. The other three PRs in the stack were audited the same way and were never affected.

No re-verification against the live warehouses is needed: `drt/destinations/databricks.py` is byte-identical to what the smoke run passed against, since the sweep never touched it.

masukai:

Checked the Delta-specific syntax claim rather than trusting the comment: \`ADD COLUMNS (scope_spec STRING, scope_key STRING)\` — plural, parenthesized — is correct Delta/Databricks DDL (unlike the singular \`ADD COLUMN\` every other leg uses), and the test pins the exact string rather than just asserting "an ALTER happened," so a future edit that silently reverts to the wrong-for-Delta singular form would actually fail.

The thing that matters most for this whole stack isn't a code-level finding, though — it's already stated plainly in this PR's own body: **"Live verification (batched, still owed)"**, all three remaining dialects marked pending. Worth restating because the smoke-test commits' own messages (3/4, 4/4) read confidently ("asserts the whole chain," "the pre-existing state rows are backfilled") in a way that could be misread as "this ran and passed" — it hasn't, yet; the tests exist and skip without credentials, same as every other test in that file. The PR body is honest about this in a dedicated section, so nothing is actually overclaimed, but it's the one gate that dominates over everything else in #904/#905/#906/#907 right now.

masukai (commented):

No issues found.

Issue #763, opened 2026-07-10
feat(engine): computed_fields — declarative derived columns via Jinja expressions (rename→derive→mask family)


masukai:
## Problem

Between extract and load, drt can rename (`field_mappings`) and mask (`mask`) — but not *derive*. Real destination payloads constantly need small shape adjustments the warehouse model shouldn't own:

- concat/format: `full_name` from `first + last`, E.164 phone normalization for Twilio
- destination-specific casts: bool → `"true"/"false"` strings for ad platforms, epoch millis for Amplitude/Mixpanel
- constants/environment stamps: `source: "drt"`, `${ENV}`-derived campaign tags

Today the answer is "edit the SQL model" — which forces destination-specific formatting into shared marts (one mart per destination = mart sprawl), or "use REST `body_template`" — which only exists for the REST destination and per-record Jinja, not for the 33 structured destinations.

## Prior art

- **dlt**: `add_map` / `@dlt.transformer` — per-row derivation between extract and load.
- **Census/Hightouch**: mapper-level computed/liquid-template fields per destination field.
- **dbt**: would model this in SQL — valid for warehouse-shaped logic, wrong layer for destination-payload shaping.

## Proposal

```yaml
sync:
  computed_fields:
    full_name: "{{ row.first_name }} {{ row.last_name }}"
    phone_e164: "+81{{ row.phone | replace('-', '') }}"
    source_system: "drt-${ENV}"
  field_mappings: { ... }   # order: computed_fields -> mask -> field_mappings
```

- Same Jinja environment as `body_template` (StrictUndefined, `tojson_safe`) — one templating story across the tool
- Pure engine-stage transform (dict-in/dict-out), so it composes with every destination and stays inside the Rust-rewrite boundary (MiniJinja exists for the future engine, per `drt/templates/` design note)
- Errors follow `on_error` semantics row-by-row

## Scope

**In:** config key, engine stage + ordering guarantees vs `mask`/`field_mappings`, filter parity with `body_template`, docs + skill updates.
**Out:** arbitrary Python hooks (that's the #297 plugin-system boundary), aggregations/cross-row logic (belongs in SQL), lookup joins (exists via `lookups`).

## Related

- #580 (`tojson_safe`) — the filter set becomes shared surface
- #415 (`field_mappings`) — this completes the rename→derive→mask transform family
- #427 / #660 (`mask`) — same engine seam

## Milestone fit

[...]

PR #913, opened 2026-08-05
feat(templates): render templates to values, and compile them once (#763, 1 of 3)

masukai:
First of three for #763 (`computed_fields`). Self-contained: no config key and no engine wiring yet, and one of the two changes is a straight win for code already shipped.

## Compile once

`render_template` built a fresh `Environment` and re-parsed the template **on every call**, and all five of its callers — HubSpot, Klaviyo, Mixpanel, Zendesk, Linear — call it *per record*. Now: two module-level environments and an LRU cache of compiled templates. Jinja templates are safe to render concurrently (each render builds its own context), which matters because `drt run --threads N` renders through one shared cache — a test renders 200 rows across 8 threads to pin that.

## `render_value`

Returns a Python value when the template is a **single output node**, a string otherwise:

| template | result |
|---|---|
| `{{ row.n }}` | `5` |
| `{{ row.n * 1000 }}` | `5000` |
| `{{ row.flag }}` | `True` |
| `{{ row.a }}-{{ row.b }}` | `"x-y"` |
| `+81{{ row.a }}` | `"+81x"` |
| `drt-prod` | `"drt-prod"` |

`render_template` is unchanged and keeps returning `str` — its callers splice the result into a larger text payload, so a string is the only sensible answer. A *computed field* is different: it becomes a column a typed destination is about to write, and `{{ row.ts | to_ms }}` arriving as `"1754000000000"` instead of an integer is a real defect, not a formatting quibble.

**Why not Jinja's own `native_concat`.** It runs `literal_eval` over the rendered result, so a column holding the string `"123"` would arrive at the destination as the integer `123`, while `"01"` stayed a string (leading zero → `SyntaxError` → falls back to text). Same config, different behaviour per row, decided by the data. The rule here keys on the **output node count** instead, so a bare literal is never evaluated at all — `"01"` in YAML stays `"01"`.

## Two Jinja details worth flagging, both silent if missed

- `NativeTemplate.render` reaches for `self.environment_class.concat` — the concat of the **declared** class, not of the environment that produced the template. Overriding `concat` on an `Environment` subclass alone therefore does nothing at all; the template class has to be replaced too. My first prototype had exactly this bug and looked like it worked, because `native_concat` happens to give the right answer for `{{ row.n }}` — it only diverged on string columns.
- The native code generator never stringifies an output node, so `StrictUndefined` has nothing to trip over. `{{ row.typo }}` returned the `Undefined` **object**, which would have been written straight into the record. Now forced explicitly, with tests for both the lone-expression and the surrounded-by-text case.

## Verification

Full suite **2951 passed / 22 skipped**; `drt/templates/renderer.py` at **100%** line coverage (measured per-module, not via the patch status); `ruff` + `mypy` clean.

Next in the stack: the `computed_fields` config key + engine stage, then docs.

### DIFF
--- a/drt/templates/renderer.py
+++ b/drt/templates/renderer.py
@@ -2,6 +2,22 @@
 
 Future: replace with MiniJinja (Rust) via PyO3 for zero-dependency binary.
 Interface is intentionally simple to make the swap transparent.
+
+Two rendering entry points, differing only in what they return:
+
+* :func:`render_template` — always a ``str``. Used where the result is spliced
+  into a larger text payload (REST ``body_template``, HubSpot / Klaviyo /
+  Mixpanel / Zendesk / Linear property templates), so a string is the only
+  sensible answer.
+* :func:`render_value` — a Python value when the template is a **single
+  output node**, otherwise a ``str`` (#763). Used where the result becomes a
+  *field* in a record that a typed destination is about to write, so
+  ``{{ row.n * 1000 }}`` must reach a BIGINT column as ``5000`` and not
+  ``"5000"``.
+
+Both share one Environment per mode and cache compiled templates, because
+every caller renders **per record**: the previous code built a fresh
+``Environment`` and re-parsed the template inside the row loop.
 """
 
 from __future__ import annotations
@@ -9,11 +25,22 @@ from __future__ import annotations
 import json
 from datetime import date, datetime, time
 from decimal import Decimal
+from functools import lru_cache
+from itertools import chain, islice
 from typing import Any
 from uuid import UUID
 
-from jinja2 import BaseLoader, Environment, StrictUndefined
+from jinja2 import BaseLoader, Environment, StrictUndefined, Template, Undefined
 from jinja2.exceptions import UndefinedError
+from jinja2.nativetypes import NativeEnvironment, NativeTemplate
+
+# Compiled templates are cached by source text. Template sources come from
+# config, so the working set is bounded by project size; the cap only exists
+# so a pathological config cannot grow the cache without limit (a miss just
+# re-parses). Jinja templates are safe to render concurrently — each render
+# builds its own context — which matters because `drt run --threads N` renders
+# from several threads against one cache.
+_TEMPLATE_CACHE_SIZE = 512
 
 
 def _json_default(obj: Any) -> Any:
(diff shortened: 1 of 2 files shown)


masukai:

Traced the two trickiest mechanisms line by line rather than trusting the comments:

**\`islice\`/\`chain\` reconstruction in \`_single_node_concat\`.** \`head = list(islice(values, 2))\` consumes up to 2 items from the generator; the many-node branch does \`chain(head, values)\` to re-stitch the peeked items back onto what's left of the same generator. Confirmed this is correct rather than assumed: \`islice\` advances the underlying generator as it yields, so \`values\` is genuinely positioned at item 3 after the peek, and \`chain(head, values)\` reconstructs the full original sequence with nothing dropped or duplicated.

**The \`StrictUndefined\` force-trigger.** \`str(value)\` is called for its side effect only (the result is discarded) when \`value\` is \`Undefined\`, relying on \`StrictUndefined.__str__\` raising. The \`return value\` line right after is unreachable in the actually-configured environment — \`_VALUE_ENV\` is built with \`undefined=StrictUndefined\` specifically, so any \`Undefined\` reaching this branch raises before returning. Correct as a targeted workaround for the native code generator never stringifying output nodes on its own; would silently break if this environment were ever constructed with a non-strict \`Undefined\`, but nothing here does that.

One thing I checked that isn't called out in the PR but generalizes correctly anyway: a single-node template whose value is a \`Decimal\`/\`datetime\`/\`UUID\` (not just int/bool from the examples) — the native render path yields the raw Python object for a lone node, so those pass through \`_single_node_concat\` untouched the same way.

No findings. This is a careful piece of work — the two Jinja internals it works around (\`environment_class.concat\` vs. instance, and \`StrictUndefined\` needing to be provoked explicitly under native rendering) are exactly the kind of thing that looks like it works in a quick test and silently doesn't in production, and both have a named regression test.

masukai (commented):

Traced the trickiest part — `_single_node_concat`'s `islice`-then-`chain` peek pattern — for the double-consumption bug this shape usually risks: `root_render_func(ctx)` returns a fresh, unconsumed generator each call, `islice(values, 2)` advances it exactly once, and the `chain(head, values)` fallback correctly picks up from where `islice` left off. No item is read twice or dropped.

The `StrictUndefined` force-via-`str()` is real and necessary — `NativeTemplate`'s code generator never stringifies a single output node, so without it an `Undefined` sentinel would silently become a record value instead of raising. Confirmed `_VALUE_ENV` is configured with `undefined=StrictUndefined`, so every `Undefined` instance reaching that branch does raise on `str()`.

No issues found.

PR #914, opened 2026-08-05
feat(engine): computed_fields — declarative derived columns (#763, 2 of 3)

masukai:
Second of three for #763. Stacked on #913 — **review that one first**; this PR's diff against it is the feature itself.

```yaml
sync:
  computed_fields:
    full_name:     "{{ row.first_name }} {{ row.last_name }}"
    phone_e164:    "+81{{ row.phone | replace('-', '') }}"
    source_system: "drt-${ENV}"
```

## The issue's ordering is wrong, and following it would have unmasked PII

#763 sketches `computed_fields` → `mask` → `field_mappings`. The shipped order is `field_mappings` → `mask` ([sync.py](https://github.com/drt-hub/drt/blob/main/drt/engine/sync.py#L600-L609)), and [masking.py](https://github.com/drt-hub/drt/blob/main/drt/engine/masking.py#L9-L10) states that `mask` keys reference the **post-rename, destination-facing** name. Swapping those two would silently change what every existing `mask:` key means: a mask written against the destination name would stop matching, and the field it exists to protect would go out in the clear. No error, no warning.

So this PR inserts the new stage at the **front** — `computed_fields` → `field_mappings` → `mask` — and leaves the existing two alone. It also matches the issue's own examples, which reference `{{ row.first_name }}`, i.e. source columns. I'll post a correction on the issue.

## Decisions worth a second opinion

**No chaining.** Every template is evaluated against the record as it arrived, so a computed field can never read another one. Chaining would make results depend on YAML key order — exactly what `field_mappings` rejected when it chose single-pass remapping. A user who wants two steps writes the composed expression.

**Overwriting a source column is allowed.** `phone: "+81{{ row.phone | replace('-','') }}"` normalises in place and reads the *original* value, because assignment happens after all fields are rendered. Forbidding it would force a second name for what is conceptually the same column.

**`on_error: fail` raises here, unlike `apply_lookups`.** A template failure is nearly always one config defect affecting every row alike; reporting it once beats reporting N skipped rows and zero synced. A lookup miss is a statement about *data* — missing a referenced row is a legitimate thing for data to do — so stopping the batch is right there and wrong here. Flagging the divergence rather than hiding it. (Separately: `on_error: fail` in `apply_lookups` doesn't actually fail the run today, it just truncates the batch. That looks like a real bug, but it's out of scope here — happy to file it.)

## Two things found while building this

**The exception catch had to be widened.** `render_value` converts Jinja's `UndefinedError` to `ValueError`, so a missing column is catchable — but `{{ row.a / row.b }}` over a zero raises `ZeroDivisionError` straight out of the rendered template, and a filter handed the wrong type raises `TypeError`. Catching only `ValueError` would have let those bypass `on_error` entirely and abort a run that had explicitly asked to skip bad rows. Now catches `Exception` (not `BaseException` — graceful shutdown depends on `KeyboardInterrupt` propagating), with the type name in the message.

**A null through a filter renders as the string `"None"`.** `{{ row.phone | replace('-', '') }}` over a NULL phone yields the literal `"None"`, because Jinja stringifies before filtering — while a bare `{{ row.phone }}` correctly stays null. I found this by writing a test that asserted the opposite. It's plain Jinja behaviour, shared with the REST destination's `body_template`, so I deliberately did **not** special-case it: diverging would mean two templating semantics in one tool. It's pinned by a test with the reasoning, and the docs PR will show `{{ row.phone or '' | replace('-', '') }}`.

## Also here

- **Config-time syntax validation** — a malformed template stops `drt validate` instead of surfacing on the first row of a production run. Only syntax is decidable; whether `row.foo` exists depends on the query.
- **Bundled VS Code schemas regenerated**, extension 0.1.11 → 0.1.12 with a CHANGELOG line (the `vscode-schema-drift` job fails otherwise).

## Verification

Full suite **2984 passed / 22 skipped**. `drt/engine/computed_fields.py` and `drt/templates/renderer.py` both at **100%** line coverage, measured per-module rather than trusting `codecov/patch` — a new module's own test file inflates the patch denominator. `ruff` + `mypy` clean; `scripts/check_drift.sh` reports no new drift; the `engine/sync.py` observer-boundary tests pass unchanged (the transform is pure, like `field_mappings` and `mask`).

Next in the stack: docs + skills.


masukai:

The claim that mattered most — computed_fields runs before field_mappings/mask specifically because the issue's proposed order would have silently unmasked PII — is backed by an actual end-to-end test (\`test_computed_then_renamed_then_masked\`), not just asserted in prose: a full_name computed from source columns, renamed to \`name\`, and confirmed \`[REDACTED]\` under the destination-facing mask key. Checked this one concretely since it's the highest-stakes claim in the PR.

Traced the "no chaining" guarantee against the actual loop rather than trusting the docstring: \`derived[name] = render_value(template, record)\` reads the *original* \`record\` inside the loop, and \`record.update(derived)\` only happens after every field in the batch succeeds — so a template referencing another computed field's output doesn't silently do something order-dependent, it hits \`StrictUndefined\` and fails loudly through the normal \`on_error\` path. That's a good property: chaining attempts fail visibly rather than working by accident of dict order.

Also checked: \`break\` on first failure correctly stops evaluating the remaining fields for that row (not the remaining rows) before the error is recorded — "first failure wins" is accurate, not just a phrase in the docstring.

The two things flagged as found-while-building (the broadened \`except Exception\` vs. just \`ValueError\`, and \`None | filter\` rendering as the string \`"None"\`) both have tests pinning the specific behavior, not just the happy path. Nothing to add there.

No findings. Also agree with keeping the pre-existing \`apply_lookups\`'s \`on_error: fail\` bug (doesn't actually fail the run) out of scope here — it's a real, separate issue and bundling the fix into this PR would make the diff answer two questions instead of one.

masukai (commented):

Traced the three claims that matter most for correctness here:

- **Ordering**: confirmed in `engine/sync.py` — `apply_computed_fields` → `apply_field_mappings` → `apply_mask`, exactly as described, and matches the PII-unmasking concern the PR raises against the issue's original sketch.
- **"No computed field can observe another / never half-derived"**: confirmed — `derived` is built as a separate dict from `record` inside the per-row loop, and `record.update(derived)` only happens after every template in that row's `computed_fields` succeeded. A mid-row failure never partially mutates `record`.
- **`on_error: fail` actually fails the run**: the `raise ValueError(failure)` inside `apply_computed_fields` is unguarded at the call site in `_run_sync_body` (no local try/except), so it propagates out of the batch loop — genuinely different from `apply_lookups`'s already-flagged divergent behavior, as claimed.

Config-time syntax validation (`_check_computed_field_templates`) is wired correctly via `model_validator(mode="after")`, including the empty-field-name guard.

No issues found. The self-identified sharp edge (null-through-filter renders as `"None"`) is Jinja's own behavior, consistently shared with the REST destination's `body_template`, and already pinned by a test — agree with not special-casing it.

PR #915, opened 2026-08-05
docs: computed_fields guide, and make the transform order consistent (#763, 3 of 3)

masukai:
Third and last for #763. Stacked on #914 — docs only, no code.

## New: `docs/guides/computed-fields.md`

Leads with the two things a user can actually get wrong:

**Values keep their type.** A single-expression template returns the Python value; anything with surrounding text renders as a string. The guide states the rule as a table, and makes the point that it keys on the *template's structure*, never on what the value looks like — so a column holding `"123"` stays a string and a zip code `"01"` stays `"01"`.

**A null passed through a filter renders as the string `"None"`.** Jinja stringifies before filtering, while a bare `{{ row.phone }}` stays null. Documented with the `{{ row.phone or '' | replace('-','') }}` workaround rather than special-cased in code, since the behaviour is shared with `body_template` and diverging would mean two templating semantics in one tool.

## Consistency pass

Three documents each described the transform chain in their own words, and all three were one stage short:

- **`field-mappings.md`** — the numbered pipeline and its ASCII diagram gain the computed stage. "Computed / derived fields" moves out of *What it does not do* (where it sat as a known gap, "tracked separately") and becomes a pointer.
- **`pii-masking.md`** — "runs last" now says last of three, and notes that a computed field can be masked like any other, under whatever name it has by then.
- **`docs/llm/API_REFERENCE.md`** — the `sync:` block gains `computed_fields` with the type rule, the ordering, and the null warning.
- **`/drt-create-sync`** — a "Derived columns" bullet, so the skill offers the feature instead of sending people back to their dbt model.
- **`/drt-migrate`** — the "compute / reshape a value" row pointed at `body_template` / `properties_template`, which only some destinations have. It now points at `computed_fields` for all of them, and keeps the templates for what they're still for (shaping a whole payload).

## One claim I checked rather than assumed

The guide says `${VAR}` substitution works inside a computed field (`source_system: "drt-${ENV}"`). Verified end to end through the real parser: `expand_env_vars` walks the whole parsed YAML tree, so `${ENV}` resolves at config load while the Jinja template sitting beside it is left completely untouched.

```
{'source_system': 'drt-prod', 'full_name': '{{ row.a }} {{ row.b }}'}
```

## Verification

All relative links resolve. Full suite **2984 passed / 22 skipped**; `scripts/check_drift.sh` reports no new drift (docs/skills/MCP parity intact).


masukai:

Checked the one claim presented as specifically verified rather than assumed: \`\${VAR}\` resolving inside a \`computed_fields\` value while the Jinja \`{{ }}\` syntax next to it stays untouched. Confirmed structurally — \`expand_env_vars\` (\`drt/config/parser.py\`) recursively walks every string in the parsed YAML tree regardless of key, so a \`computed_fields\` value is reached the same as any other field; \`\${VAR}\`-only content (the guide's own example, \`"drt-\${ENV}"\`, has no Jinja in it) resolves at parse time before the sync ever runs.

Also checked something the PR doesn't call out but that I had a specific (mistaken) worry about from earlier context: whether \`.claude/commands/drt-create-sync.md\` and \`skills/drt/skills/drt-create-sync/SKILL.md\` are meant to be one canonical file with the other stale. They're not — \`diff\`\-ing the two (minus the skill's frontmatter) on current \`main\` shows them byte-identical, and this PR's own commit updates both with the exact same new bullet, so the deliberate dual-file sync this repo already has (per #875) is preserved correctly, not drifted.

No findings in the new guide or the consistency-pass edits — spot-checked the type-coercion table and the order diagram against what #913/#914 actually implement, both line up.

masukai (commented):
Self-review (staleness check, requested explicitly). Cross-checked the transform-order claims (`computed_fields` → `field_mappings` → `mask`) and the type-preservation/null-through-filter notes against #914's actual implementation, which I traced separately today — matches exactly, including the `{{ row.phone or '' | replace('-','') }}` guidance for the null-stringification edge case.

Issue #760, opened 2026-07-10
feat(destinations): managed destination tables — opt-in auto-create + on_schema_change drift policy


masukai:
## Problem

SQL/DWH destinations require the target table to already exist with exactly the right columns, and the INSERT column list is derived from `records[0].keys()`. Two consequences:

1. **Onboarding friction**: every new sync to Postgres/Snowflake/BigQuery/Databricks starts with hand-written DDL that must mirror the model's output — the part dbt/dlt users expect the tool to own.
2. **Silent drift hazard**: when the upstream model adds/renames a column, drt fails at the DB-driver level mid-load (or worse, keeps loading a subset if the column was removed) — there is no drt-level policy for schema drift. Our boundary tests (#542) documented the current behavior; nothing manages it.

The building block already exists: `introspect_schema()` (#317) reads INFORMATION_SCHEMA on Postgres/MySQL/Snowflake/Databricks for serialization routing.

## Prior art

- **dlt**: creates destination tables and evolves them (new columns → `ALTER TABLE ADD COLUMN`) with schema-contract modes (`evolve` / `freeze` / `discard`).
- **dbt** incremental models: `on_schema_change: ignore | fail | append_new_columns | sync_all_columns`.
- **Census/Hightouch**: manage their own state/log tables, and destination mapping UIs re-validate on drift.

## Proposal

```yaml
destination:
  type: postgres
  table: public.users_enriched
  manage_table: true                 # opt-in: CREATE TABLE if missing (types via source introspection/inference)
  on_schema_change: fail             # fail (default) | ignore | append_new_columns
```

- `manage_table: true` creates the table on first run (typed from cursor/introspection where available, TEXT/VARIANT fallback otherwise) — same connectors as #317's introspection legs
- `on_schema_change: append_new_columns` issues `ALTER TABLE ADD COLUMN` for new source columns; `fail` gives a pre-load, drt-level error naming the offending columns instead of a driver stack trace
- Non-SQL destinations ignore both keys (validate-time warning)

## Scope

**In:** Postgres/MySQL first, then Snowflake/BigQuery/Databricks (phasing like #317/#340); pre-load drift check comparing record keys vs introspected columns; docs on required privileges (extends #695).
**Out:** column *removals*/type migrations (fail with guidance); auto-DDL for SaaS destinations (HubSpot property creation etc. — future per-connector work).

## Related

[...]

PR #917, opened 2026-08-05
feat(engine): first-class run_id + sync_run_id correlation (#762, part 1)

masukai:
Part 1 of #762's two-part proposal — **"run_id plumbing" only**. Part 2 (opt-in `metadata_columns` written into destination rows) is explicitly out of scope: the issue's own "Milestone fit" note defers it to ride with the managed-tables work (#760), and pulling it in now would touch every SQL destination file — all four of which currently have open, unmerged PRs (#904–#907, the #890 stack). This PR touches none of them, on purpose.

## What ships

Two identifiers, both from a new `drt._identifiers.new_run_id()`:

| id | scope | generated where |
|---|---|---|
| `run_id` | one per `drt run` process | `run()`, once — threaded through every sync it executes |
| `sync_run_id` | one per *sync execution* | inside `run_sync()` itself, so every caller gets one — including library callers that never pass `run_id` at all |

Threaded through the six surfaces the issue names: `SyncResult`, `HistoryEntry`, `DeadLetter`, `--output json`, OTel span attributes, and failure/degraded alert context.

## The design choice that kept this small: no `SyncObserver` Protocol change

`_run_sync_body` mutates one `total_result: SyncResult` object in place — its own docstring says why: *"so the outer finally-block can read partial results when an exception propagates."* I stamp both ids onto that object immediately after construction, before anything can observe it. Since `on_sync_completed(sync_name, result, ...)` already hands every observer that same object, and `build_context(result=..., ...)` already receives it too, **every downstream consumer picks the ids up for free** — no Protocol signature change, no touching `NullObserver`/`LoggingObserver`/`StatePersistingObserver`/`CompositeObserver`/`DlqObserver`, no hunting down third-party observer implementations.

## Two decisions I made rather than deferred back to the issue

**Not reconciled with `query_tags.new_run_id()`** (#768's per-query cost-attribution id), even though both now run once per `run_sync()` call. That one is lowercase hex, 12 chars — sized to fit SQL comments and BigQuery job labels without truncation math. Forcing a full UUID4 through the same path risks corrupting exactly the budget it exists to respect, across three warehouses I'd have had to re-verify (BigQuery label 63-char total, Snowflake `QUERY_TAG` length, Databricks `query_tags` kwarg). The existing docstring already flagged this reconciliation as deferred until #762 shipped — I'm recording that it was considered and declined, not missed.

**Plain `uuid.uuid4()`, not `uuid7`** as the issue proposes. Sortability isn't a real requirement for any of the six surfaces above — all six already key by `sync_name` + timestamp. `uuid.uuid7` isn't in the stdlib on any Python drt still supports (3.10–3.13); pulling in a dependency to sort ids nobody sorts is the "heavy dependency in core" CLAUDE.md rules out.

## One real gap, documented rather than solved

When `run_sync()` raises before ever returning a `SyncResult` (a hard crash), the CLI's `--output json` error-path entry has no object to read `sync_run_id` from — so that one entry carries `run_id` only. The history entry and failure alert written inside `run_sync()`'s own `finally` block *before* the re-raise still got the correct `sync_run_id` (verified by `test_history_entry_carries_ids_even_on_a_raised_exception`); only the CLI's own JSON surface misses it for that one failure mode. Smuggling the id out via the exception object felt like solving a corner case nobody asked for — flagging it here instead in case reviewers disagree.

## Verification

Full suite **2952 passed / 22 skipped**. New/touched modules at 96%+ coverage measured **per-module** (not `codecov/patch` — a new module's own test file inflates that denominator): `drt/_identifiers.py`, `destinations/base.py`, `state/dlq.py` at 100%; `engine/sync.py`, `alerts/dispatcher.py`, `state/history.py`, `cli/commands/run.py` at 93–98%, every miss cross-checked by line number against a pre-existing, unrelated branch. `ruff` + `mypy` clean. `scripts/check_drift.sh` reports no new drift (no new user-facing flag). Two backward-compatibility tests pin that a JSONL line written before these fields existed (`HistoryEntry`, `DeadLetter`) still loads via the dataclass default — both readers unpack via `Type(**json.loads(line))`.

Five hand-rolled `SyncResult`-shaped test doubles needed the two new attributes (`test_cli_build.py`, `test_cli_diff.py` ×2, `test_cli_graceful_shutdown.py`, `test_cli_run_parallel.py`, `test_dry_run_summary.py`) — following the codebase's own established convention rather than inventing one (`test_cli_run_parallel.py`'s `_FakeResult` docstring already says why: *"keeping the fake in lockstep with the real shape is cheaper than monkey-patching each attribute per test"*).

## Not in this PR

- `metadata_columns` (#762 part 2) — rides with #760.
- `target/run_results.json` (#778) — explicitly "cheap once run_id lands"; a natural next PR now that it has.
- Docs (`docs/llm/API_REFERENCE.md`, a guide) — following the pattern from #913/#914/#915, docs land as a focused follow-up once the surface is settled.

### DIFF
--- a/drt/_identifiers.py
+++ b/drt/_identifiers.py
@@ -1,0 +1,33 @@
+"""Run identifiers — first-class correlation IDs for one execution (#762).
+
+Two distinct scopes share this one generator:
+
+* **invocation `run_id`** — one per ``drt run`` / ``drt test`` process. The
+  CLI generates it once (``drt/cli/commands/run.py``) and threads it through
+  every sync in that invocation, so all of them can be correlated back to
+  "the run that happened at 03:40 UTC".
+* **``sync_run_id``** — one per *sync execution*, generated inside
+  :func:`drt.engine.sync.run_sync` itself so every caller gets one, including
+  library callers that never pass an invocation ``run_id`` at all.
+
+Neither is the query-tagging ``run_id`` in :mod:`drt.config.query_tags` —
+that one is scoped narrowly to "distinguish this query from the last one this
+sync ran" for cost-attribution SQL comments, predates this module, and stays
+separate on purpose (see its own docstring).
+
+Plain ``uuid4`` rather than ``uuid7``: sortability isn't a real requirement
+for any of the six correlation surfaces this threads through (history, DLQ,
+alerts, OTel, ``--output json``, structured logs all key by ``sync_name`` +
+timestamp already), and ``uuid.uuid7`` isn't in the stdlib on any Python drt
+still supports (3.10-3.13) — pulling in a dependency to sort IDs nobody sorts
+would be exactly the "heavy dependency in core" CLAUDE.md rules out.
+"""
+
+from __future__ import annotations
+
+import uuid
+
+
+def new_run_id() -> str:
+    """A fresh globally-unique identifier for one run or one sync execution."""
+    return str(uuid.uuid4())
--- a/drt/alerts/dispatcher.py
+++ b/drt/alerts/dispatcher.py
@@ -36,6 +36,8 @@ def build_context(
         "rows_processed": result.success + result.failed,
         "duration_s": duration_s,
         "started_at": started_at,
+        "run_id": result.run_id,
+        "sync_run_id": result.sync_run_id,
     }
 
 
@@ -71,6 +73,8 @@ def build_degraded_context(
         "rows_processed": result.success + result.failed,
         "duration_s": duration_s,
         "started_at": started_at,
+        "run_id": result.run_id,
+        "sync_run_id": result.sync_run_id,
     }
 
 
--- a/drt/cli/commands/build.py
+++ b/drt/cli/commands/build.py
@@ -18,6 +18,7 @@ from pathlib import Path
 
 import typer
 
+from drt._identifiers import new_run_id
 from drt.cli._app import app
 from drt.cli._selection import SelectionError, complete_selector, select_syncs
 from drt.cli.output import console, print_error
@@ -121,6 +122,7 @@ def build(
         quiet=quiet,
         log_json=False,
         cursor_value=None,
+        run_id=new_run_id(),
     )
 
     entries: list[dict[str, object]] = []
--- a/drt/cli/commands/run.py
+++ b/drt/cli/commands/run.py
@@ -51,6 +51,7 @@ if TYPE_CHECKING:
     from drt.state.manager import StateManager
 
 
+from drt._identifiers import new_run_id
 from drt.cli._app import app
 from drt.cli._helpers import (
     exit_code_for_signal as _exit_code_for_signal,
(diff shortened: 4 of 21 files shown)


masukai:

Checked the two claims that carry the most weight:

**"Stamped before anything can observe it."** Verified against \`run_sync()\` directly — \`total_result = SyncResult()\` is immediately followed by \`total_result.run_id = run_id\` / \`total_result.sync_run_id = sync_run_id\`, before the tracer span opens, before \`_run_sync_body\` is called, before any observer method fires. The "no Protocol change needed" claim holds because there's genuinely no code path between construction and first observation.

**The one documented gap** (CLI's \`--output json\` error entry has no \`sync_run_id\` when \`run_sync()\` raises before returning). Checked \`cli/commands/run.py\`'s \`except Exception as e:\` branch directly: the error dict gets \`"run_id": ctx.run_id\` but no \`sync_run_id\` key at all — exactly as described, not overstated or understated.

Also traced the \`query_tags.new_run_id()\` naming collision (same function name, different module, both now called once per \`run_sync()\`) — it's handled cleanly: aliased at the one import site that needs both (\`from drt._identifiers import new_run_id as new_correlation_id\` in \`engine/sync.py\`), and \`query_tags.py\`'s docstring now cross-references the other one by full path so a future reader doesn't assume they're the same concept.

No findings. Agree with both declined-in-favor-of-simplicity calls (plain \`uuid4\` over \`uuid7\`, no metadata_columns here) — the reasoning given for each is specific to this codebase's actual constraints, not a generic "kept it simple."

masukai (commented):
Self-review (adversarial pass, requested explicitly). The "mutate `total_result` in place, no Protocol change" design traces out correctly — verified `_run_sync_body` returns the identical object it was handed (`total_result=total_result` in, `return total_result` at the end, same reference throughout), so `run_id`/`sync_run_id` stamped at construction really do reach `build_context`/`build_degraded_context` (confirmed both were updated to read `result.run_id`/`result.sync_run_id`) and `observer.on_sync_completed` for free, exactly as claimed. The two documented gaps (hard-crash CLI JSON entry, `metadata_columns` deferral) are accurately scoped.

One gap that isn't documented: `_RunContext` now has `run_id: str = ""` (can't be a required field once `cursor_value` above it has a default), with the docstring noting "every real construction site sets it explicitly." That's true for `drt run` (`cli/commands/run.py:791`, `run_id=new_run_id()`) but not for the other one:

```python
# drt/cli/commands/build.py:113
ctx = _RunContext(
    source=source,
    state_mgr=state_mgr,
    ...
    cursor_value=None,
)
```

No `run_id=` here, so `drt build` — which runs the same `_run_one` machinery over potentially many syncs — stamps `run_id=""` on every SyncResult/HistoryEntry/DeadLetter/alert context it produces, rather than either a real generated id or `None`. That's a third, undocumented state alongside "real uuid4" and "None for a library caller" for anything correlating on this field downstream, and it means `drt build` — arguably at least as deserving of "one id per invocation" as `drt run` — doesn't get one.

Given the PR is explicit about the two gaps it *does* accept ("flagging here in case reviewers disagree" / "recording that it was considered and declined"), this reads like a miss rather than a considered scope cut — `build.py` isn't mentioned anywhere in the PR body. One-line fix: `run_id=new_run_id()` at the `build.py:113` construction site, same as `run.py`.

Everything else — the `query_tags.new_run_id()` non-reconciliation, the `uuid4` vs `uuid7` call, the JSONL backward-compat defaults — reasoned through and traces correctly.

Issue #780, opened 2026-07-10
feat(test): sync unit tests — fixture rows in, expected payload out, zero credentials (dbt-unit-tests analog)


masukai:
## Problem

There is no way to test a sync's *transform pipeline* without live credentials. The pipeline is no longer trivial — `computed→mask→field_mappings→lookups→body_template` (and growing) — yet the only feedback loops are `drt validate` (shape only) and a real `--dry-run` against a real warehouse. Consequences: contributors can't verify mapping changes without cloud accounts; CI can't assert payload correctness on PRs; a broken Jinja `body_template` is discovered at send time.

The infrastructure is 90% built: `FakeSource` (#364) exists for drt's own suite but isn't user-facing.

## Prior art

- **dbt unit tests** (1.8): YAML `given` (fixture rows) + `expect` (output rows), runs without touching real data — became the standard PR-gate for model logic.
- **Census/Hightouch**: mapper previews with sample records.

## Proposal

```yaml
# syncs/users_to_hubspot.yml
unit_tests:
  - name: masks_email_and_renames
    given:
      - { id: 1, email: "alice@example.com", first: "Alice", last: "Doe" }
    expect:
      - { id: 1, email: "5e884898...", full_name: "Alice Doe" }
```

```bash
drt test --unit          # no credentials, no network; runs given→engine transforms→assert
```

- Rows flow through the real engine transform stages via FakeSource + a capture destination (no HTTP/DB)
- For REST destinations, `expect` may assert the *rendered body* (`expect_body:`) — catching template bugs pre-send
- Runs in plain CI (no secrets) → perfect PR gate; MCP `drt_run_test` gains a `unit` flag; `/drt-create-sync` skill can emit a unit test alongside every generated sync (LLM writes sync + its test in one shot)

## Scope

**In:** `unit_tests` config block, FakeSource + capture-destination wiring, `--unit` flag (default `drt test` behavior unchanged), body-render assertions, docs + skill update.
**Out:** live-destination assertions (that's integration testing / existing `tests:`), fixture generation from production samples (nice follow-up), lookup mocking (v2 — needs a fake lookup table shape).

## Related

- #364 (FakeSource — the enabler), #368 (`drt_run_test` MCP), #413 (`--diff` — runtime preview; unit tests are the *offline* complement)

## Milestone fit

Candidate: v0.9 — CI-trust feature; also a strong differentiator story ("your LLM writes the sync AND its tests, verified offline").

masukai:
[...]

Issue #364, opened 2026-04-19
feat: add FakeSource for unit testing (mock source infrastructure)


masukai:
## Summary

Add a `FakeSource` implementation of the `Source` Protocol that yields configurable records without needing a real database connection.

## Motivation

Source coverage is 25-35% because `extract()` requires real DB connections. A mock source enables:
- Engine integration tests with deterministic data
- Destination contract tests with realistic payloads
- CI testing without DB infrastructure

## Scope

- `drt/sources/fake.py` — `FakeSource` class implementing `Source` Protocol
- Yields records from an in-memory list or generator
- Configurable: field names, row count, data types, cursor simulation
- NOT registered in CLI (test-only, not user-facing)

## Files

- `drt/sources/fake.py` (new)
- `tests/conftest.py` — shared fixture `fake_source`

PR #918, opened 2026-08-05
feat(test): sync unit tests — fixture rows in, expected payload out (#780, part 1)

masukai:
Part 1 of #780 — the `unit_tests:` config shape + offline runner. CLI wiring (`drt test --unit`), MCP parity, and docs are follow-ups, same split that worked for #763/#762. This piece is independently useful (library callers, and the eventual CLI command is a thin wrapper over it).

```yaml
sync:
  unit_tests:
    - name: masks_email_and_renames
      given:
        - { id: 1, email: "alice@example.com", first: "Alice", last: "Doe" }
      expect:
        - { id: 1, given_name: "Alice" }
```

## No engine change needed

`run_sync()` already accepts any Source/Destination Protocol implementation, and every stateful side effect inside it (state persistence, watermark storage, history, DLQ) is opt-in via a parameter that defaults to `None`/off. `FakeSource` (#364, already shipped) plus the new `CaptureDestination` this PR adds are exactly a Source/Destination pair with nothing to opt into — so `run_sync(sync, FakeSource(rows=given), CaptureDestination(), ...)` runs the *real* `computed_fields → field_mappings → mask` chain with zero network calls and zero disk writes, using the public API exactly as it exists today.

This is also why this PR touches **none** of the `engine/sync.py` regions currently under an unmerged diff (#917's outer `run_sync` setup, #914's per-batch transform chain) — it's a new file that *calls* `run_sync()`, not an edit to it. That was the deciding factor over doing #762 part 2 (`metadata_columns`) next, which needs to insert into the exact region #914 already has an open diff in.

## Two side effects I found by reading `run_sync()`, not by hitting them in CI

**Lookups** resolve FK values by querying the real destination (`build_lookup_map`) — no fake exists for that yet, which #780's own scope section already calls out ("lookup mocking (v2)"). A sync with `destination.lookups` configured is rejected outright (`UnitTestLookupsUnsupportedError`) rather than silently running without the lookup step and asserting against output the real sync would never produce.

**Alerts** fire real Slack/webhook HTTP calls from inside `run_sync()`'s own `finally` block whenever a run reports a failure — a direct call, not observer-mediated (same as the history write beside it), so nothing about capturing the destination or omitting an observer stops it. A fixture that deliberately exercises `on_error: skip` to test a drop-a-row scenario would page whatever channel `alerts.on_failure` points at. Handled with `sync.model_copy(update={"alerts": None})` — a shallow copy, so the caller's own `SyncConfig` object is never mutated (pinned by a test).

## Design choices worth a second look

- **`expect` is a subset match per row, not exact-record equality.** Only the keys a test declares are checked; extra keys the real pipeline produces are ignored. A sync's source columns grow over time, and requiring every unit test to enumerate every column it doesn't care about would turn each one into a maintenance burden against unrelated schema growth. Row **count** is still checked exactly, and *before* the per-row comparison rather than via `zip()` — which would silently truncate to the shorter list and hide a dropped row.
- **A raised exception becomes a failed test, not a crashed CLI.** An unsupported `match_policy`, a template error — anything `run_sync()` can raise — is caught and turned into a one-line mismatch. The point of a unit test is to report a broken config safely; one bad fixture must not stop `drt test --unit` before later tests get a chance to run.
- **`CaptureDestination` accumulates across every `load()` call.** Batching is an implementation detail of `sync.batch_size` — it must never change which rows a test compares against, only how many calls it takes to deliver them. Pinned with a `batch_size=1` / 3-row fixture test.

## Verification

Full suite **2943 passed / 22 skipped** (this branch is `main` + this PR only — independent of #762/#763, not stacked on either). `drt/engine/unit_test_runner.py` at **100%** line coverage; `drt/config/sync_options.py` at 97%, every miss a pre-existing unrelated branch (watermark backend validation, an assert labelled unreachable) verified by line number. `ruff`, `mypy`, `check_drift.sh` all clean — no new CLI/MCP drift, since `--unit` isn't wired up yet.

## Not in this PR

- `drt test --unit` CLI flag + `drt_run_test` MCP parity — natural next PR now this exists.
- `expect_body` (rendered-payload assertion for templated destinations, e.g. REST `body_template`) — needs the *real* destination class to render its body without sending, which likely means extending the existing `--dry-run --diff` machinery rather than `CaptureDestination`. Deserves its own investigation.
- Docs + `/drt-create-sync` skill update ("LLM writes the sync and its test in one shot").

### DIFF
--- a/drt/config/models.py
+++ b/drt/config/models.py
@@ -97,6 +97,7 @@ from drt.config.sync_options import (
     SyncOptions,
     SyncTest,
     UniqueTest,
+    UnitTest,
     WatermarkConfig,
     WebhookAlertConfig,
 )
@@ -175,6 +176,7 @@ __all__ = [
     "TeamsDestinationConfig",
     "TwilioDestinationConfig",
     "UniqueTest",
+    "UnitTest",
     "WatermarkConfig",
     "WebhookAlertConfig",
     "ZendeskDestinationConfig",
--- a/drt/config/sync_options.py
+++ b/drt/config/sync_options.py
@@ -6,7 +6,7 @@ the three ``destinations_*`` modules and consumed by :class:`SyncConfig`.
 
 from __future__ import annotations
 
-from typing import Annotated, Literal
+from typing import Annotated, Any, Literal
 
 from pydantic import BaseModel, Field, PrivateAttr, model_validator
 
@@ -374,6 +374,30 @@ class SyncTest(BaseModel):
         return self
 
 
+class UnitTest(BaseModel):
+    """Offline transform-pipeline test — fixture rows in, expected rows out (#780).
+
+    Unlike ``SyncTest`` (which queries the *destination* after a real sync),
+    a ``UnitTest`` never touches a destination or the network: ``given`` rows
+    are run through the same ``computed_fields`` -> ``field_mappings`` ->
+    ``mask`` chain ``run_sync()`` applies in production (via
+    ``drt.engine.unit_test_runner``), and the result is compared against
+    ``expect``. dbt unit-tests analog; Census/Hightouch mapper previews.
+    """
+
+    name: str
+    # At least one row — an empty `given` would make every unit test
+    # vacuously pass, silently, the moment a fixture typo drops its only row.
+    given: list[dict[str, Any]] = Field(min_length=1)
+    # Row count must match `given` exactly (a transform that's supposed to
+    # drop/split rows is exactly the kind of change a unit test exists to
+    # catch) — but each expected row is matched by the *keys it declares*,
+    # not the full record. A sync's source columns grow over time; requiring
+    # every unit test to enumerate every column it doesn't care about would
+    # make each one a maintenance burden against unrelated schema growth.
+    expect: list[dict[str, Any]] = Field(min_length=1)
+
+
 class SlackAlertConfig(BaseModel):
     type: Literal["slack"]
     webhook_url: str | None = None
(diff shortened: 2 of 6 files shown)


masukai (commented):

Traced the claims that matter most:

- **`run_sync()` genuinely needs no changes**: confirmed — `CaptureDestination`/`FakeSource` are a plain Protocol pair, and every optional side-effecting parameter (state, watermark, history, DLQ) already defaults to off.
- **Row-count check happens before the per-row `zip`**, not via truncating `zip` itself — confirmed, the length mismatch is caught and reported before the `else` branch that does the `zip`, so there's no silent-truncation path.
- **Alert suppression**: `alerts: AlertsConfig | None = None` is already the standard default-off shape for `SyncConfig`, so `model_copy(update={"alerts": None})` isn't introducing an unusual state — `dispatch_alerts(None, ...)` is the same path every alert-less sync already takes. No special-casing needed, none added.
- **Lookup rejection**: `getattr(sync.destination, "lookups", None)` — checked against the actual config classes (`destinations_sql.py`), `lookups` is the only name used for this concept across the two destination types that have it, so the `getattr` default doesn't mask a differently-named field elsewhere.

One minor completeness gap, not a bug: `expect: list[dict[str, Any]] = Field(min_length=1)` means a unit test can never assert "this input produces zero output rows." The PR's own example scenario ("a fixture that deliberately exercises `on_error: skip` to test a drop-a-row scenario") still works fine as long as at least one row survives, but the all-rows-dropped case — plausible for the same `on_error: skip` path — is unwritable. `given`'s `min_length=1` has a stated rationale (an empty fixture would vacuously pass); `expect`'s doesn't call out why it carries the same constraint, so it may just be copied from `given` rather than deliberate. Low severity, easy to loosen later (`min_length=0`) if someone hits it.

Everything else traces clean.

Issue #922, opened 2026-08-06
docs(engine): protect cursor-based incremental as the read-only path, and tier reversibility as a design goal


masukai:
## Problem

[ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md) establishes two things that no code or docs change currently protects:

1. **Cursor-based incremental (`mode: incremental` + `cursor_field`) is drt's first-class, permanent read-only-source path** — not a legacy fallback #755's diff-based incremental supersedes. #755 requires warehouse write; cursor-based incremental is the honest equivalent of Hightouch's Basic engine, and it is the *only* answer for an operator who cannot or will not grant write on the source warehouse.
2. **Tiers must be reversible** (ADR 0005 Decision 4): moving between `local` → `gcs`/`s3` → `warehouse` state, and between cursor-based and diff-based incremental, must be a config change, not a one-way door. Hightouch's Lightning engine is explicitly one-way ("You can't move to the Basic engine once Lightning is configured"); drt's differentiation on this axis depends on that staying false.

Both are easy to lose quietly while #755, #756, #920, and #760 (managed tables) implement — each new warehouse-write-requiring feature is a small, locally-reasonable place to let "just require write access" creep in as the assumed default.

## Proposal

Not a code change by itself — this issue is where the design constraint gets enforced as #755/#756/#920/#760 land:

- **Docs**: `docs/guides/comparison.md` (already contrasts drt's "cursor-based" against Census/Hightouch's "multiple strategies") and `docs/llm/API_REFERENCE.md` should state explicitly, once #755 ships, that cursor-based incremental remains fully supported and is the recommended path when the source connection is read-only — not deprecated, not "legacy".
- **A reversibility check**: as each tiered feature lands (state backend, incremental strategy), confirm and where practical test that downgrading (warehouse → gcs/s3 → local; diff → cursor) is a supported operation, not just an implied one. What "supported" means concretely (a CLI command? a documented manual procedure? an automated test?) is this issue's design question — deliberately not decided yet, since it depends on what #755/#756/#920 actually ship.

## Scope

**In:** the docs update once #755 ships; a concrete definition of "reversible" for each tier, checked against at least one real downgrade path per tier.
**Out:** building new reversibility tooling speculatively, ahead of the tiers it would apply to existing.

## Related

- ADR 0005 Decision 3 (cursor-based incremental as the read-only path) and Decision 4 (reversibility)
[...]

Issue #921, opened 2026-08-06
feat(engine): cross-process rate-limit coordination for orchestrator-launched sync processes


masukai:
## Problem

[#769](https://github.com/drt-hub/drt/issues/769) (rate limiting v2) shipped per-destination `rate_limit` overrides and a shared bucket across `--threads` within one process ([#858](https://github.com/drt-hub/drt/pull/858)). That closes pacing within a single `drt run` / `drt serve` process, but not across processes: a Dagster sensor that yields one `RunRequest` per changed sync launches each as its **own process**, so N changed syncs hitting one destination endpoint means N independent rate-limit buckets, each pacing to the full configured rate — the endpoint sees up to N× the intended load.

[ADR 0004](https://github.com/drt-hub/drt/blob/main/docs/adr/0004-streaming-and-event-triggered-syncs.md)'s 2026-07-29 amendment folded this residual into #756 (remote state backend) on the reasoning that a cross-process bucket "needs shared state". [ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md) found that reasoning right about the *need* and wrong about the *kind*: a token bucket needs low-latency atomic increment on every acquire, and neither of #756's two backends (object storage: no cheap compare-and-swap, a network round trip per acquire; a warehouse table: worse on both counts) can serve that without the pacing overhead becoming the bottleneck it exists to prevent. **Neither half of #756 closes this residual** — it was folded into an issue that structurally cannot close it, and needs its own solution.

## Proposal

Not yet designed — this issue exists to hold the problem statement and stop it from being silently lost inside #756 again, per ADR 0005's correction. Candidate directions, none committed to:

- A lightweight shared-counter service colocated with the orchestrator (Redis, or Dagster's own instance storage if it exposes low-latency atomic ops) — infrastructure drt does not want to own by default, so likely opt-in.
- Coarser pacing: instead of a shared bucket, each process rate-limits to `configured_rate / N` where N is a configured or discovered fan-out — approximate, no shared state, but requires the operator to know N.
- Accept the limitation and document it: N Dagster-launched processes against one endpoint is a topology choice, and the mitigation is "don't fan out that many changed syncs against one rate-limited endpoint in one sensor tick" (achievable today via `state:modified` batching, #772).

## Scope

**In (eventually):** a decision on one of the directions above, or an explicit "won't fix, here's the documented workaround" — either closes this issue.
[...]

Issue #920, opened 2026-08-06
feat(state): warehouse-backed state backend — SQL-queryable observability (opt-in, requires write grant)


masukai:
## Problem

[#756](https://github.com/drt-hub/drt/issues/756) originally scoped both a CI-durable state backend (object storage) and a warehouse-backed one (SQL-queryable observability) as one issue with one config knob (`state.backend: local | warehouse | gcs | s3`). [ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md) found that bundling wrong: the two halves have different justifications (durability vs. observability) and different costs (no warehouse write vs. requires one), and shipping them together puts a write-grant conversation in front of a fix that doesn't need one.

This issue is the warehouse half, split out per that ADR's Decision 2. #756 now covers only the object-storage half and ships first.

## Proposal

```yaml
# drt_project.yml
state:
  backend: warehouse            # local (default) | gcs | s3 (#756) | warehouse (this issue)
  schema: _drt_state            # drt-managed tables: _drt_runs, _drt_dlq, _drt_watermarks
```

- Reuses the source profile connection (BigQuery/Snowflake/Databricks/Postgres first — same set as tracked mirror, #686)
- History rows written per run make sync observability SQL-queryable (Hightouch Warehouse Sync Logs-style writeback)
- DLQ records in a table make `drt retry` work from any runner, queryable alongside run history
- Same `StateStore`/`HistoryStore`/`DlqBackend` Protocol surface #756 implements against (landed via #900) — this is a second implementation of those Protocols, not a new interface

## Why this is separate from #756, and why it's later

Per [ADR 0005 Decision 2](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md#decision): the CI-durability problem (a sensor and a CI run can't share a watermark) is fully satisfied by object storage — no warehouse write required. This issue's justification is purely "make sync history/DLQ SQL-queryable", which is valuable but optional, and asking for a warehouse write grant to fix a problem object storage already fixed is the wrong order. Implementation ordering (from the ADR): this is step 4, after #756 (step 2) and a shared managed-table primitive (step 3, shared with #755).

## Scope

[...]

PR #923, opened 2026-08-06
docs(adr): correct ADR 0004's #769 amendment per ADR 0005

masukai:
Follow-up item 4 from ADR 0005's own [Follow-up issues](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md#follow-up-issues) section: "Re-scope the #769 cross-process residual out of #756 per the correction above, and amend ADR 0004's gate table accordingly."

## What's wrong today

ADR 0004's 2026-07-29 amendment folded the #769 cross-process rate-limit residual (N Dagster-launched processes hitting one endpoint with N independent buckets) into #756 (remote state backend) on the reasoning that a cross-process token bucket "needs shared state." True — but ADR 0005 found it's the wrong *kind* of shared state: a token bucket needs low-latency atomic increment on every acquire, and neither of #756's backends (object storage: no cheap compare-and-swap, pays a network round trip per acquire; a warehouse table: worse on both counts) can serve that without the pacing overhead becoming the bottleneck it exists to prevent.

## What this PR does

- Adds a third dated **Amendment (2026-08-06)** entry to the Status block, alongside the existing 2026-07-29 and 2026-08-03 ones — not editing the 07-29 text in place, matching this ADR's own pattern of appending corrections so the historical record survives.
- Updates the gate table's `#756` row (now also notes it's object-storage-only per ADR 0005 — no warehouse write required) and `#769` row (residual moved out).
- Updates the inline amendment paragraph below the table with a pointer to the correction.
- Filed **[#921](https://github.com/drt-hub/drt/issues/921)** to hold the residual — unscheduled, no design committed, a few candidate directions sketched (shared-counter service, coarser per-process pacing, or documenting the topology limitation as a known gap).

Net effect on ordering: **none, this time either.** #756 already blocked Tier 2 on durability grounds alone, and Tier 3 already cleared via #854 independent of this correction — this is a correctness fix to the paper trail, not a schedule change.

## Also touched, same PR

- **#756** retitled and rewritten to scope it to the object-storage half only (`gcs`/`s3`, no warehouse write) — the warehouse half split out as **[#920](https://github.com/drt-hub/drt/issues/920)**.
- **#755** — corrected its prior-art section: the Census claim was unverifiable (`docs.getcensus.com` now 301s to Fivetran post-acquisition) and is dropped; dbt snapshots (`strategy: check` + `check_cols: all`) added as the closer, verified structural reference.
- **[#922](https://github.com/drt-hub/drt/issues/922)** filed to protect cursor-based incremental's status as drt's permanent read-only-source path, and tier reversibility (ADR 0005 Decision 4), from quietly eroding as #755/#756/#920/#760 implement.

These four are the remaining items from ADR 0005's Follow-up issues list; this PR is the code-review-able piece (the ADR 0004 edit), the issue edits/creations already happened directly since they're not diffable the same way.

## Verification

Relative link to `0005-state-location-and-write-grants.md` resolves (checked the file exists at that path). No other `#769` reference in the document was left inconsistent with the correction — grepped every occurrence and confirmed each still reads correctly (the Tier 3 "cleared" language is unaffected, since the correction only concerns the Tier 2 cross-process residual, not the per-process fix #858 already shipped).

### DIFF
--- a/docs/adr/0004-streaming-and-event-triggered-syncs.md
+++ b/docs/adr/0004-streaming-and-event-triggered-syncs.md
@@ -21,6 +21,18 @@
   until the OIDC follow-up (#903) lands. The Context section's `serve` citations
   describe the pre-#854 code deliberately — they are the evidence the
   decision was made on.
+- **Amended:** 2026-08-06 — [ADR 0005](0005-state-location-and-write-grants.md)
+  corrects the 2026-07-29 amendment below: the #769 cross-process residual
+  does not close via #756 after all. That amendment was right that a
+  cross-process token bucket needs shared state and wrong about what *kind* —
+  low-latency atomic increment, which neither of #756's backends provide (the
+  object-storage half has no cheap compare-and-swap and pays a round trip per
+  acquire; a warehouse table is worse on both counts). Re-scoped out to its own
+  issue, [#921](https://github.com/drt-hub/drt/issues/921), tracked as
+  unscheduled rather than folded into a gate it cannot close. Net effect on
+  ordering: none — #756 already blocked Tier 2 on durability grounds alone and
+  remains the longer pole; Tier 3 already cleared via #854 above, independent
+  of this correction.
 - **Issue:** [#786](https://github.com/drt-hub/drt/issues/786)
 - **Implementation:** none — this ADR recommends **not** building a native
   watcher. The work it does sanction is listed under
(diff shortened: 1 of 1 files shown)


masukai:
Self-review pass (adversarial, requested explicitly — not merge-blocking) — found a real sequencing problem between this PR and #932, both mine, both open.

**The two ROADMAP.md commits here (2nd of 2) and #932 rewrite the same "Incremental & state" bullet in incompatible ways, and #932 is the more current/comprehensive one.** This PR's edit keeps #755/#756/#758/#890 bundled in one bullet and enriches the #756 sub-clause with the object-storage/warehouse split. #932 (opened ~5 hours before this one, per timestamps) restructures the *entire* v0.9 scope section — renames the bullet to "State foundation (ADR 0005)", keeps only #756 + #922 in it, moves #755/#758 to a new "Cut 2026-08-06" list with their own reasoning (blocked on an unbuilt ADR 0005 step-3 primitive), and puts #890 in a separate new "Perf close-out" bullet.

Concretely: if #932 merges first, this PR's ROADMAP.md commit almost certainly fails to apply — the exact bullet text it's patching won't exist in that form anymore. If this PR merges first, #932 needs a rebase, but since #932's rewrite is structurally a superset of the specific correction here (the #756 object-storage-first framing survives inside #932's version too, just reorganized), that rebase should be mechanical rather than lossy.

I don't think I should resolve this by picking one unilaterally — #932's restructuring (which issues even stay in v0.9) is a bigger editorial call than this PR's narrower correction, and that's a call for you rather than something to silently pick a winner on in a review comment. Practical suggestion: merge #932 first (it's the more current picture and supersedes the ROADMAP half of this PR entirely), then this PR's *first* commit (the ADR 0004 correction itself — a different file, no overlap with #932) can still land on its own, with the second commit dropped as redundant.

The ADR 0004 correction itself (commit 1/2, \`docs/adr/0004-streaming-and-event-triggered-syncs.md\`) is independent of this and checks out: the amended text accurately distinguishes "needs shared state" (right in the 07-29 amendment) from "any of #756's backends provide low-latency atomic increment" (wrong, corrected here) — matches what I found when re-deriving the #921 scoping decision earlier in the ADR 0005 follow-up work.

masukai:
Rebased and dropped the ROADMAP.md commit (2nd of 2) — #932 merged and its v0.9 rewrite supersedes it, as discussed above. This PR is now just the ADR 0004 correction (`docs/adr/0004-streaming-and-event-triggered-syncs.md`), rebased cleanly onto current main.

masukai (commented):
Self-review (staleness check, requested explicitly) — re-confirming after the earlier self-correction on this PR (the #921 false-alarm). The amended text is internally consistent: #921 is referenced correctly in both the gate table and the amendment note, and the "net effect on ordering: none" conclusion holds under the corrected reasoning the same way it did under the original (wrong) fold. No new staleness found; CI still green on the rebased branch.

PR #924, opened 2026-08-06
feat(test): wire sync unit tests into the CLI and MCP (#780, part 2)

masukai:
Part 2 of #780 — stacked on #918 (the runner). `unit_tests:` was implemented but unreachable outside a library call; this makes it usable from the CLI and MCP.

```bash
drt test --unit                              # text
drt test --unit --output json                # json
drt test --unit --select users_to_hubspot    # composes with the usual selection flags
drt test --unit --fail-fast
```

Plus MCP `drt_run_test(sync_name=None, unit=False)`.

## One function each, not a branch in the existing loop

Both the CLI and MCP paths get their own execution function (`execute_unit_tests_for_sync` / `_run_unit_tests`) rather than a branch inside the existing `sync.tests:` loop. That loop's machinery — destination connection, `dry_run`, `severity: warn`, `--store-failures` — doesn't apply to a mode that never touches a destination by design, and threading a growing set of "well, not for unit tests" exceptions through one function seemed worse than two functions that each do one thing.

Both delegate to the same `drt.engine.unit_test_runner.run_unit_test` that #918 shipped, so the CLI and MCP surfaces can't drift on what a unit test result *means* — only on how it's presented. This mirrors the split already in the codebase between `execute_tests_for_sync` (CLI) and MCP's own independent loop for `sync.tests:`.

## Two things kept deliberately narrow

- **`--unit` rejects `--dry-run` and `--store-failures`** (exit 2) — both are destination-connected concepts unit tests structurally don't have, so silently ignoring them felt worse than a clear error.
- **`drt build` is untouched.** Its whole point is "run + test against a real destination in one pass" — the opposite of what a unit test promises (zero credentials, zero network). Extending it didn't make sense to attempt.

## JSON shape

Mirrors the existing envelope (`{"status", "results": [{"sync", "tests": [...]}]}`), but each test entry is `{name, passed, mismatches}` rather than `{name, passed, value/error, severity}` — there's no severity tier for unit tests, so reusing the exact shape would mean fields that are always absent rather than a shape that says what it actually means.

## Verification

Full suite **2961 passed / 22 skipped**. `drt/cli/commands/test.py` and `drt/mcp/tools/run_test.py` at **97%** each, measured per-module (not `codecov/patch`); every remaining miss in both is a pre-existing branch in the `sync.tests:` path this PR doesn't touch, cross-checked by line number. `ruff` + `mypy` clean. `scripts/check_drift.sh` reports no new drift — `--unit` maps to `unit` by name alone, no rename declaration needed (unlike `--diff` → `compute_diff`).

Manually smoke-tested end to end before writing formal tests: passing/failing unit tests, `--output json`, the lookups-rejection message, and the mutual-exclusivity guard, all through the real `drt` console script.

## Not in this PR

- Docs (`docs/llm/API_REFERENCE.md`, a guide) + `/drt-create-sync` skill update — natural follow-up now the surface is settled, same pattern as #913→#915.

### DIFF
--- a/drt/cli/commands/test.py
+++ b/drt/cli/commands/test.py
@@ -123,6 +123,95 @@ def _store_or_clear_failure_sample(
     return path, len(masked_rows)
 
 
+def execute_unit_tests_for_sync(
+    sync: SyncConfig, *, json_mode: bool, quiet: bool = False
+) -> tuple[_SyncTestResult, bool]:
+    """Run one sync's ``unit_tests:`` and return ``(result_dict, had_failures)``.
+
+    Deliberately separate from :func:`execute_tests_for_sync` rather than a
+    branch inside it: unit tests (#780) never touch a destination — no
+    connection, no ``dry_run`` distinction, no ``severity: warn`` tier, none
+    of the machinery that function exists for. Sharing a function would mean
+    threading a growing set of "well, not for unit tests" exceptions through
+    it instead of two functions that each do one thing.
+    """
+    from drt.engine.unit_test_runner import UnitTestLookupsUnsupportedError, run_unit_test
+
+    show = not json_mode and not quiet
+    if show:
+        print_test_header(sync.name)
+
+    sync_results: _SyncTestResult = {"sync": sync.name, "tests": []}
+    had_failures = False
+
+    try:
+        for test_def in sync.unit_tests:
+            result = run_unit_test(sync, test_def)
+            message = "ok" if result.passed else "; ".join(result.mismatches)
+            if show:
+                print_test_result(test_def.name, result.passed, message)
+            sync_results["tests"].append(
+                {
+                    "name": test_def.name,
+                    "passed": result.passed,
+                    "mismatches": result.mismatches,
+                }
+            )
+            if not result.passed:
+                had_failures = True
+    except UnitTestLookupsUnsupportedError as e:
+        # One config-shape problem, the same for every fixture on this sync —
+        # surfaced once rather than once per unit_tests entry.
+        if show:
+            print_test_result(sync.name, False, str(e))
+        sync_results["tests"].append({"name": sync.name, "passed": False, "error": str(e)})
+        had_failures = True
+
+    return sync_results, had_failures
+
+
+def _run_unit_tests(syncs: Sequence[SyncConfig], *, json_mode: bool, fail_fast: bool) -> None:
+    """``drt test --unit``'s body — separate from the main ``sync.tests:`` loop
+    in :func:`test_syncs` for the same reason :func:`execute_unit_tests_for_sync`
+    is separate from :func:`execute_tests_for_sync`: no destination, no
+    ``dry_run``, no ``severity`` tier, no ``--store-failures``.
+    """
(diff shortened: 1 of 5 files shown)


masukai:
Self-review pass (adversarial, requested explicitly — not merge-blocking) — this one found a real, actionable issue rather than confirming the design holds up.

**This branch is stale against #927 (merged today) and touches the same function.** \`git merge-base --is-ancestor origin/main origin/feat/780-cli-wiring\` confirms current \`main\` is not an ancestor of this branch. #927 refactored \`drt/mcp/tools/run_test.py\`'s \`run_test()\` to delegate to \`execute_tests_for_sync\` instead of re-implementing the per-test loop (closing #851). This PR's diff still shows the *pre-#927* version as its baseline — importing \`execute_test_query\`/\`get_table_name\`/\`is_queryable\` directly and re-implementing the loop, with the new \`unit\` parameter and \`_run_unit_tests\` branch added on top of that old shape.

Both changes are real and both need to land — #927's dedup and this PR's \`unit\` parameter aren't in conflict conceptually, they just haven't been reconciled yet because this stack was built before #927 merged. Rebasing \`feat/780-sync-unit-tests\` → \`feat/780-cli-wiring\` → \`feat/780-docs\` onto current \`main\` should surface it as a normal (probably small) conflict in \`run_test()\` — \`_run_unit_tests\`'s addition is structurally independent of which implementation \`run_test()\` delegates to for the non-unit path, so this should resolve mechanically rather than needing a redesign. Flagging now rather than leaving it for whoever reviews next to discover as a surprise merge conflict.

Everything else about the design holds up: \`_run_unit_tests\` correctly delegates to the same \`run_unit_test\` #918 shipped (CLI and MCP can't drift on what a result means), and the \`--dry-run\`/\`--store-failures\` rejection matches what \`tests/unit/test_cli_test_unit.py\`'s \`TestMutualExclusivity\` actually asserts.

masukai (commented):
Self-review (adversarial pass, requested explicitly) — this one was accidentally dropped from the earlier "new rule" review batch, catching up now before a merge triage.

Traced the CLI/MCP split for drift risk: both `execute_unit_tests_for_sync` (CLI) and `_run_unit_tests` (MCP) delegate to the same `drt.engine.unit_test_runner.run_unit_test`, and their per-test result shape (`{name, passed, mismatches}`) is identical. The one shape asymmetry — the `UnitTestLookupsUnsupportedError` path emits `{"name": sync.name, "passed": False, "error": ...}` instead of a per-test entry — is present, identical, and identically documented in both surfaces, so it's a consistent deliberate choice, not drift.

`--unit` + `--dry-run`/`--store-failures` mutual exclusivity is checked before `load_syncs()` runs, so it fails fast rather than doing wasted work first.

One asymmetry worth naming, not fixing: `drt test --unit --fail-fast` can emit `{"skipped": true, "reason": "fail_fast"}` entries; the MCP tool has no `fail_fast` parameter at all for either test mode, so it never produces that shape. Pre-existing MCP/CLI parity gap (MCP never had fail-fast), not something this PR introduced or needs to fix.

No issues found.

PR #926, opened 2026-08-06
docs(test): sync unit tests guide + API reference + /drt-create-sync (#780, part 3)

masukai:
Part 3 of #780 — docs. Stacked on #924 (CLI/MCP wiring). Same pattern as #913→#914→#915.

## New: `docs/guides/sync-unit-tests.md`

Quick start, how the transform chain runs (`given` reads source names, `expect` reads destination-facing names — same ordering rule as `field-mappings.md`), the subset-match-per-row + exact-row-count rule, what's not supported yet (lookups, `expect_body`), and a comparison table against `sync.tests:` so the two don't get confused as competing rather than complementary things.

## `docs/llm/API_REFERENCE.md`

Gains a `unit_tests` section in the same style as the existing `tests:` reference, placed right after it.

## `/drt-create-sync`

New conditional step 6: when a generated sync has `field_mappings` and/or `mask`, **offer** (don't force) a `unit_tests:` block alongside it — the differentiator #780's own issue text names explicitly ("your LLM writes the sync AND its test, verified offline"). Edited the SSoT (`skills/drt/skills/drt-create-sync/SKILL.md`) and regenerated `.claude/commands/drt-create-sync.md` via `make sync-skills`.

## One thing caught before it shipped

An early draft of both the guide and the API reference described the transform chain as `computed_fields → field_mappings → mask`, copied from #763's own docs pattern without checking whether `computed_fields` exists on *this* branch. It doesn't — #763/#914 haven't merged, so `sync.computed_fields` isn't a real field here yet, and an example using it would silently no-op if someone copy-pasted it today. Rewritten to describe only what #918/#924 actually run and test right now (`field_mappings → mask`), with "and any other stage the engine applies before `destination.load()`" left in as the forward-compatible hook — so this won't need an edit again once `computed_fields` lands, it'll just start being true.

## Verification

No VS Code schema change expected or found — this PR adds no `SyncConfig` field (`unit_tests` itself already shipped in #918). Confirmed by actually regenerating the bundled schemas locally and diffing, rather than assuming from "it's docs-only." Full suite **2961 passed / 22 skipped** (unchanged, as expected for a docs PR). `check_drift.sh` and `make check-skills` both clean.


masukai:

Inherits the staleness noted on #924 (this branch is downstream of it) — \`git merge-base --is-ancestor origin/main origin/feat/780-docs\` also comes back stale, same root cause. This PR's own diff doesn't touch \`run_test.py\`, so nothing in *this* PR conflicts with #927 directly; the whole stack just needs a rebase together, which #924's comment covers.

Checked the one thing worth checking independently here: the \`.claude/commands/drt-create-sync.md\` / \`skills/drt/skills/drt-create-sync/SKILL.md\` dual-file pattern (verified as deliberate on #915) stays in sync in this PR too — diffed the two files' added lines against each other, byte-identical.

No other findings — this is docs describing #918/#924's already-reviewed behavior, and the content matches what those two actually implement.

masukai (commented):
Self-review (staleness check, requested explicitly). Confirmed the "`given` and `expect` must each have at least one row" line accurately documents the actual `Field(min_length=1)` constraint on both — including the completeness gap I flagged on #918 (no way to assert "zero rows survive"), which this doc doesn't overclaim past. No staleness found.

PR #927, opened 2026-08-06
refactor(mcp): drt_run_test calls execute_tests_for_sync instead of duplicating its loop (#851)

### LINKED ISSUE #851: refactor(mcp): drt_run_test should call execute_tests_for_sync, not duplicate its loop

Muawiya-contact:
## What does this PR do?

`run_test()` ran its own copy of the test-execution loop — connect, build query, execute, check, severity, error handling — that `execute_tests_for_sync()` has run for `drt test` and `drt build` since #777. It now calls that function per sync and keeps only the selection and the JSON envelope around it, so a future change to severity semantics reaches the MCP tool by construction rather than by remembering to.

Net: `run_test.py` loses 49 lines and gains 24.

### The three things the issue asked to check before landing

**1. `store_failures` — hardcoded `False`, not threaded through.** #779's samples are written to `.drt/test_failures/<sync>/<test>.jsonl` and MCP exposes no way to read one back, so threading the parameter would only give an MCP client a way to write files it can't then read. This is what the issue leaned toward, and it's what I asked on the issue; going with it rather than blocking, since the alternative is easy to add later as its own change and hard to remove once it's in the tool signature. Pinned by `test_run_test_does_not_store_failure_samples`, which asserts the directory is never created.

**2. Console output — `json_mode=True` *and* `quiet=True`.** Either one alone sets `show = False`, but they mean different things (`json_mode` = "a machine is reading this", `quiet` = "say less"), and if they ever come apart the tool needs both to stay silent. MCP speaks over stdio, so a stray `print` doesn't just look untidy — it corrupts the transport. `test_run_test_writes_nothing_to_the_console` covers the pass path *and* the skip path (both print) and asserts `captured.out == captured.err == ""`; it fails with the full rich-rendered header and test lines if either argument is dropped (verified by flipping them).

**3. Return-shape diff, field-for-field.** One field differed, and it's fixed in its own commit *ahead* of the refactor rather than absorbed into it:

| | `drt_run_test` (before) | `execute_tests_for_sync` |
|---|---|---|
| `status` | `"passed"` / `"failed"` | built by caller — unchanged |
| `results[].sync` | sync name | same |
| `results[].tests[]` | `{name, passed, value, severity}` / `{name, passed, error, severity}` | **identical** |
| `results[].skipped` | `True` | same |
| `results[].reason` | `"tests not supported for rest_api destinations"` | `"tests not supported for rest_api"` ← |

`execute_tests_for_sync` *prints* "…for rest_api **destinations**" in both its text branches but puts the one-word-shorter string in the result dict — the same sentence, inconsistent between its own two outputs. `drt_run_test` has always returned the longer form. Commit 1 aligns the dict on the wording both surfaces already print, so the MCP contract survives the switch untouched.

⚠️ **The one user-visible edge:** that also adds the word to `drt test --output json`'s `reason` for a skipped sync, and to `drt build`'s `tests_skipped_reason` which reads the same field. No test asserted the exact string (`test_cli_build.py` matches `"not supported" in …`), and it now matches what the command has always printed. Called out in the CHANGELOG. Happy to invert it — have the MCP tool drop the word instead — if you'd rather the CLI's JSON stay byte-stable; it's a one-line move either way.

Also removes `dry_run` from the picture: the tool passes `dry_run=False`, so `{name, dry_run, severity}` entries can't reach its shape.

### Tests

All 4 existing `drt_run_test` tests — including #836's two severity-parity tests — pass **unmodified**. Three added, each of which fails if the corresponding argument is dropped:

- `..._error_severity_failure_flips_status` — the missing half of the severity rule. #836 pinned that a `warn` failure does *not* flip `status`; nothing pinned that an `error` failure *does*, so `status` could have been hardcoded to `"passed"` and both existing tests would still have passed. Asserts the whole test entry, not field-by-field.
- `..._writes_nothing_to_the_console` — see (2).
- `..._does_not_store_failure_samples` — see (1).

### Verification

- `pytest tests/unit` → 2713 passed. The 17 failures are `test_sqlite.py` / `test_mysql_source.py` and reproduce identically on `upstream/main` in this environment (missing local drivers) — untouched by this diff.
- `pytest tests/contracts tests/integration` → 115 passed, 3 skipped.
- `ruff check drt tests` clean; `mypy drt` reports only the pre-existing `google_sheets.py:39` error, which also reproduces on `main`.

### Acceptance criteria

- [x] `run_test()` no longer duplicates the test-execution loop
- [x] Existing `test_mcp.py` tests (including #836's severity parity tests) pass unmodified
- [x] No console output leaks through the MCP tool call — and it's now a test, not a property

## Related Issue

Closes #851

## Checklist

---

### Self-review pass (3 commits on top)

A second read found three things, each fixed in its own commit:

- **`project_dir` was left on its `Path(".")` default** while the tool has `ctx.project_dir`. Unreachable today (`store_failures=False`), but an MCP server is created with an explicit project dir and started from whatever cwd the client chose — so threading `store_failures` through later would have written samples to `$HOME/.drt/test_failures/` while `drt test` read them from the project. Now passed from the context.
- **`assert not (tmp_path / ".drt" / "test_failures").exists()` could never fail** — samples resolve against the cwd, not `tmp_path`, so only the `failures_stored` assertion beside it was doing any work. On a machine with Postgres on :5432 (what the fixture points at) a `store_failures=True` regression would have written into the **repo checkout** and the test would still have passed. Now `chdir`s to `tmp_path`; verified the assertion fires (`assert not True`) with `store_failures=True` + `project_dir` dropped + `fetch_failing_rows` stubbed, and the tree stays clean.
- **`assert captured.err == ""` guarded nothing** and was a latent flake: the leak this test exists to catch is `drt.cli.output.console`, a plain `Console()` on **stdout**, and stdout is the channel an MCP stdio server may not write to — stderr is where such a server is supposed to log. Narrowed to stdout; re-verified it still fails with the full rendered header when `json_mode`/`quiet` are dropped.


masukai (dismissed):
This is exactly the kind of refactor PR I want to see — every risk the issue could have hidden is called out and pinned by a test that fails without the fix: the `reason` string wording mismatch caught field-for-field rather than glossed over, `json_mode`+`quiet` reasoned through as two different meanings rather than collapsed, `store_failures=False` justified by "MCP can't read it back" instead of silently dropped. The two new tests for the severity-flip and silent-console paths close real gaps the old duplicated loop could have drifted on without anything catching it (the `status` hardcode case especially — good catch that nothing pinned it before).

Net −25 LOC while adding coverage is the right shape for a de-dup refactor. Approving.

Muawiya-contact:
> This is exactly the kind of refactor PR I want to see — every risk the issue could have hidden is called out and pinned by a test that fails without the fix: the `reason` string wording mismatch caught field-for-field rather than glossed over, `json_mode`+`quiet` reasoned through as two different meanings rather than collapsed, `store_failures=False` justified by "MCP can't read it back" instead of silently dropped. The two new tests for the severity-flip and silent-console paths close real gaps the old duplicated loop could have drifted on without anything catching it (the `status` hardcode case especially — good catch that nothing pinned it before).
> 
> Net −25 LOC while adding coverage is the right shape for a de-dup refactor. Approving.

Thanks a lot! I really appreciate the detailed review and the kind words. I'm glad the extra tests and explanations helped make the refactor easier to trust. Thanks for taking the time to go through everything and for the approval!

Muawiya-contact (commented):
@masukai LGTM

Muawiya-contact:
@masukai — this is ready for review (CI green). Two decisions in it are yours rather than mine, and I'd rather flag them than have you find them in the diff:

**1. `store_failures` — I went with hardcoding `False`.** This was my open question on #851, which I asked before starting and then implemented in the direction the issue itself leaned, rather than sit blocked. Rationale: MCP exposes no way to read a written sample back, so threading the parameter through would only give a client a way to write files it can't then read — easy to add later, harder to remove from the tool's signature once it ships. Happy to flip it if you'd rather have the full trio (`store_failures` / `store_failures_limit` / `project_dir`) on the MCP surface now.

**2. One CLI-visible change rides along, and it's the only thing here outside the MCP tool.** `execute_tests_for_sync` *prints* "tests not supported for `<type>` **destinations**" but puts the one-word-shorter string in its result dict; `drt_run_test` has always returned the longer form. To keep the MCP contract byte-identical through the switch I aligned the dict on the printed wording (first commit, deliberately ahead of the refactor) — which also adds the word to `drt test --output json`'s `reason` and to `drt build`'s `tests_skipped_reason`. No test pinned the exact string and nothing in `docs/` quotes it, but it is a CLI JSON change in an MCP refactor.

If you'd rather `drt test`'s JSON stay byte-stable, the inverse is a one-line move — drop the word on the MCP side and accept the shorter `reason` there instead. Say which and I'll push it.

Everything else is plumbing-only: the returned shape is unchanged field-for-field, and all four existing `drt_run_test` tests — including #836's severity pair — pass unmodified.

masukai (approved):
Both calls are yours to make and I'm agreeing with the direction you already took.

**`store_failures=False`** — agreed, keep it hardcoded. There's no consumer on the MCP side that could read a written sample back, so exposing the parameter would just be surface area with nothing behind it. Easy to add later if that ever changes; I'd rather not carry it unused now.

**The wording alignment** — agreed, keep `drt_run_test`'s long-standing string and let `drt test --output json` pick up the extra word. Nothing pins the old string in tests or docs, so that side is free to move; the MCP tool's output has been stable for longer and is more likely to have something depending on it verbatim.

Also — the three follow-up commits (project_dir threading, the test that couldn't actually fail, the stderr assertion) are exactly the kind of thing I'd want caught before merge rather than after. Good catches.

Re-approving.

Muawiya-contact:
Thanks @masukai — both answers match what's on the branch, so there's nothing to push:

- `store_failures` stays hardcoded `False`.
- The `reason` alignment stays as-is; `drt test --output json` picks up the extra word, `drt_run_test`'s string is unchanged. Noted in the CHANGELOG under `[Unreleased]` as the one user-visible edge.

Green across the matrix (3.10–3.13), CodeQL and `codecov/patch` clean, `mergeStateStatus: CLEAN`. Ready whenever you want to squash it — no rush from my side.

Issue #782, opened 2026-07-10
feat(config): secret provider URIs — AWS/GCP Secret Manager & Vault resolution for credentials (as extras)


masukai:
## Problem

Credential resolution is env vars + `.drt/secrets.toml` fallback (#143). Production reverse ETL concentrates *write* credentials to CRMs/ads/warehouses — exactly the secrets orgs keep in AWS Secrets Manager / GCP Secret Manager / Vault with rotation policies. Today users must export those into env vars in every runner (CI YAML, Airflow connections, cron wrappers), which:

- duplicates secret distribution logic per orchestrator
- breaks rotation (env snapshots go stale; SDK fetch doesn't)
- contradicts the security posture we're building (0o600 profiles #650, keypair auth #737, encryption-at-rest planned #303)

## Prior art

- **dlt**: layered config providers (env → toml → vaults); Airflow/GCP secret providers ship as extras.
- **dbt**: `env_var()` only — and the ecosystem's answer is exactly the wrapper scripts we want to avoid.
- **Census/Hightouch**: managed secret storage (SaaS-side).

## Proposal

URI-scheme references accepted anywhere an `*_env` name resolves today, keeping full backward compatibility:

```yaml
# ~/.drt/profiles.yml
type: snowflake
password_env: "aws-sm://prod/drt/snowflake#password"   # or: gcp-sm://projects/p/secrets/drt-sf/versions/latest
                                                        # or: vault://secret/data/drt/snowflake#password
```

- Resolution chain becomes: explicit → OS env → secrets.toml → provider URI (detected by scheme)
- Providers as extras (`drt-core[aws]`, `[gcp]` exists conceptually, `[vault]`) with lazy imports — no new core deps (per CLAUDE.md)
- Process-lifetime cache + `drt profile test` exercising resolution; values masked in `drt profile show` as today
- Pluggable via entry points so third-party providers ride the #297 plugin mechanism

## Scope

**In:** provider protocol + AWS SM / GCP SM / Vault implementations, scheme parsing in `resolve_env`, caching, docs (incl. IAM permission snippets).
**Out:** secret *writing*/rotation management, encryption of local files (#303 owns that), OAuth token brokering.

## Related

- #143 (secrets.toml — the chain being extended), #303 (encryption at rest — sibling), #650/#737 (security posture), #297 (plugin mechanism for third-party providers)

## Milestone fit

[...]

Issue #297, opened 2026-04-13
feat: plugin system for connector auto-discovery and third-party connectors


masukai:
## Summary

Evolve the connector registry (#248) into a full plugin system that supports third-party connector packages.

## Why

For Open Core, Enterprise connectors need to live in separate packages (e.g., `drt-enterprise`). The plugin system must support auto-discovery so that `pip install drt-salesforce-premium` automatically registers the connector.

## Scope

- Entry point based plugin registration (`drt.sources` / `drt.destinations` entry points)
- `drt plugins list` command showing installed plugins
- Plugin metadata: name, version, author, connector type
- Documentation for third-party connector development
- Backward compatible with built-in connectors

masukai:
Implemented in #996, with one scope correction found while building it.

**What shipped**: entry-point based discovery (`drt.sources`, `drt.destinations`, `drt.secret_providers`, `drt.permission_checkers`, `drt.audit_loggers`, `drt.observers`) via `importlib.metadata`, plus `drt plugins list` showing name/version/author/connector-type metadata and load status per entry. This closes the "requires an explicit operator-side import" gap [ADR 0008](https://github.com/drt-hub/drt/blob/main/docs/adr/0008-enterprise-boundary-rbac-and-audit-hooks.md) named for `SecretProvider`/`PermissionChecker`/`AuditLogger`/extra-`SyncObserver` registration — those four are fully usable the moment a package with the right entry point is `pip install`ed.

**What didn't ship**: the headline scenario in this issue's own "Why" section — `pip install drt-salesforce-premium` and it's usable in a sync YAML. While scoping this I found a real blocker: `SyncConfig.destination`/`load_profile()` both validate the `type` field against a closed, hand-enumerated set of built-in types *before* the connector registry (`drt/connectors/registry.py`) is ever consulted. A `drt.destinations` entry point registers successfully and is visible in `drt plugins list`, but a sync YAML naming that type still fails `drt validate` — registering in the registry was never actually the blocker; config parsing is.

[...]

PR #928, opened 2026-08-06
feat(config): secret provider URIs — protocol + AWS Secrets Manager (#782, 1 of 4)

masukai:
**Part 1 of 4 — protocol + AWS Secrets Manager.** Closes nothing yet; #782 tracks all four.

## Problem

Credential resolution today is env vars + `.drt/secrets.toml` fallback (#143). Production reverse ETL concentrates *write* credentials to CRMs/ads/warehouses — exactly the secrets orgs increasingly keep in a managed store with rotation policies, not in exported env vars re-snapshotted per runner.

## Design

`resolve_env()`'s chain gains one more fallback: `explicit value → OS env → secrets.toml → provider URI`. A profile's `*_env` field can now be a `scheme://...` reference instead of a plain env var name:

```yaml
password_env: "aws-sm://prod/drt/snowflake#password"
```

That string never matches a real env var or a secrets.toml key, so it falls through unchanged to the new final step, where the scheme gets detected and dispatched. Fully backward compatible — nothing about the first three steps changes.

## Structure (so Parts 2–4 are small)

- **`drt/config/secret_providers/base.py`** — `SecretProvider` Protocol + scheme registry, provider-agnostic. Adding GCP SM or Vault later is one `register()` call in that provider's own module, not a `resolve_env()` change.
- **`parse_secret_uri()`** rejoins `netloc` + `path` — a URI's authority ends at the first `/`, but every provider's identifier here (an ARN-ish path, a Vault mount + path) reads as one continuous string: `aws-sm://prod/drt/snowflake` → `prod/drt/snowflake`, not just `drt/snowflake`. `#fragment` is an optional key into a JSON-valued secret. Tested against all three shapes from the design doc (AWS/GCP/Vault) even though only AWS ships here — Parts 2–3 get that coverage for free.
- **Process-lifetime cache**, keyed by URI. A run commonly resolves the same credential more than once (validation pass, `drt profile test`, the real connection) — unlike an env var or secrets.toml read, a provider fetch is a network call, so this one's worth not repeating. Rotation mid-run isn't a goal; a new process picks up a rotated secret on its next fetch.
- **AWS Secrets Manager provider** — `boto3` lazy-imported inside `fetch()`, same missing-extras `ImportError` pattern as the S3/BigQuery destinations (`pip install drt-core[aws-secrets]`). `ResourceNotFoundException` → `LookupError`; everything else (permissions, throttling) bubbles as the raw boto3 exception rather than being masked — a `drt run` that can't reach Secrets Manager should show the real AWS error, not a drt-shaped wrapper around it.
- New `aws-secrets` extra — same `boto3>=1.34` as `[s3]`, kept as a separate extra since the two are unrelated capabilities and someone installing one shouldn't be assumed to want the other.

## Deliberately out of scope here

- **GCP Secret Manager, Vault** — Parts 2 and 3.
- **Docs (IAM permission snippets, guide)** — Part 4.
- **Third-party providers via entry points** — the issue mentions riding #297's plugin mechanism, but #297 is still open. The registry is written so #297 can add discovery on top later without changing this shape; no entry-point loading exists yet.
- **Secret writing/rotation management, encryption of local files (#303), OAuth token brokering** — out per the issue itself.

## Verification

- 21 new tests: URI parsing (all three provider shapes), registry dispatch + duplicate-scheme guard + caching (same vs. different URIs), every `AwsSecretsManagerProvider.fetch()` path (plain string, JSON+key, JSON+missing key, non-JSON+key requested, binary secret, not-found, missing boto3), `resolve_env()` integration (explicit/env beat provider URI, unregistered scheme falls through to `None`)
- Full suite **2981 passed, 22 skipped**; `ruff` and `mypy` clean

Not run against a real AWS account — no live call happens anywhere in this PR's code path outside a real `fetch()`, and the tests mock `boto3` entirely (established pattern, see `test_s3_destination.py`). Worth a live check before Part 4 (docs) ships, batched with whichever of Parts 2/3 also needs one.

## Remaining

GCP Secret Manager (2 of 4), Vault (3 of 4), docs (4 of 4).

### DIFF
--- a/drt/config/credentials.py
+++ b/drt/config/credentials.py
@@ -49,6 +49,7 @@ from drt.config.profiles import (
     SQLiteProfile,
     SQLServerProfile,
 )
+from drt.config.secret_providers import resolve_provider_uri
 
 
 class OtelConfig(BaseModel):
@@ -146,14 +147,26 @@ def _lookup_secrets_toml(env_var: str) -> str | None:
 
 
 def resolve_env(value: str | None, env_var: str | None) -> str | None:
-    """Resolve a secret value: explicit value → env var → secrets.toml → None."""
+    """Resolve a secret value: explicit value → env var → secrets.toml →
+    provider URI → None.
+
+    The last step (#782) lets ``env_var`` itself be a ``scheme://...`` secret
+    reference (e.g. ``aws-sm://prod/drt/snowflake#password``) instead of an
+    env var name — it never matches a real env var or a secrets.toml key, so
+    it falls through to here unchanged, where the scheme is detected and
+    dispatched to the matching provider.
+    """
     if value is not None:
         return value
     if env_var is not None:
         env_val = os.environ.get(env_var)
         if env_val is not None:
             return env_val
-        return _lookup_secrets_toml(env_var)
+        from_secrets = _lookup_secrets_toml(env_var)
+        if from_secrets is not None:
+            return from_secrets
+
+        return resolve_provider_uri(env_var)
     return None
 
 
--- a/drt/config/secret_providers/__init__.py
+++ b/drt/config/secret_providers/__init__.py
@@ -1,0 +1,36 @@
+"""URI-scheme secret provider resolution (#782).
+
+    password_env: "aws-sm://prod/drt/snowflake#password"
+
+Extends ``resolve_env``'s existing chain (explicit value -> OS env ->
+secrets.toml) with one more fallback: if none of those resolved the value,
+and the value looks like a ``scheme://`` URI for a registered provider,
+fetch it from there. See ``drt.config.credentials.resolve_env``.
+
+Each provider ships as an extra with a lazy SDK import (``drt-core[aws-secrets]``
+etc.) — importing this package never imports a provider's SDK, only the
+provider actually referenced by a profile does, at first use.
+"""
+
+from __future__ import annotations
+
+from drt.config.secret_providers.aws import AwsSecretsManagerProvider
+from drt.config.secret_providers.base import (
+    SecretProvider,
+    SecretRef,
+    clear_cache,
+    parse_secret_uri,
+    register,
+    resolve_provider_uri,
+)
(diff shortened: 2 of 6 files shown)


masukai:
Self-review before this got any human eyes on it caught two things, fixed in \`b4579f1\`:

1. The cache comment asserted rotation-within-a-run isn't a concern because "a run is short-lived by construction" — false for \`drt serve\` (#902, already on \`main\`), which is long-lived and re-enters \`resolve_env\` on every triggered sync. A secret resolved once is now honestly documented as held until the server restarts, no re-fetch on rotation. The actual TTL/invalidation is real follow-up work, not bundled into Part 1.
2. \`AwsSecretsManagerProvider.fetch()\` would stringify whatever JSON sat under the requested \`#key\`, including \`None\` → the literal 4-character password \`"None"\`. Now raises \`LookupError\` for null/dict/list instead of handing back a plausible-looking wrong credential.

2 new tests for (2). Full suite 2983 passed, 22 skipped.

masukai:

This one already had two rounds of a stronger reviewer's scrutiny during implementation (visible in the thread above: the cache-lifetime claim that was flat wrong for \`drt serve\`, fixed and issue #929 filed for the underlying gap). Not re-deriving that here — spot-checked that the fix is actually present in the current file (\`base.py\`'s cache comment correctly says "unbounded for the process's life... tracked as #929" rather than the original false "short-lived" claim) rather than assuming the thread's resolution stuck.

Fresh check this pass: whether a malformed or adversarial \`*_env\` value (path traversal-style content, e.g.) could do anything unexpected. It can't, structurally — \`ref.path\` only ever becomes the \`SecretId\` argument to a *remote* API call (\`GetSecretValue\`), never a local file path or a shell/SQL string being concatenated, so there's no traversal surface here; authorization is entirely IAM's job on the AWS side.

No new findings.

masukai (commented):

Issue #929, opened 2026-08-06
fix(config): secret provider cache has no TTL — drt serve holds a resolved secret until restart


masukai:
## Problem

#782 (Part 1, PR #928) added a process-lifetime cache for `scheme://...` provider-resolved secrets (`drt/config/secret_providers/base.py`), keyed by URI, with no expiry. Documented honestly in the code as a known gap rather than left implicit:

> Unbounded for the process's life: fine for a `drt run` invocation (exits in seconds to minutes), but `drt serve` is a long-lived process that re-enters this path on every triggered sync — a secret resolved once is held until the server restarts, with no TTL and no re-fetch on rotation.

`drt serve` (#902, on `main`) is exactly this shape: `ThreadingHTTPServer.serve_forever()`, and each triggered sync re-enters `resolve_env` → `resolve_provider_uri` for its destination's credentials. A secret rotated in AWS Secrets Manager / GCP Secret Manager / Vault after the server's first resolution of it is never picked up until the process restarts — which directly contradicts #782's own problem statement ("breaks rotation (env snapshots go stale; SDK fetch doesn't)"). Right now the SDK fetch goes stale too, just on a longer cycle (server lifetime vs. env snapshot lifetime).

## Scope

**In:** a TTL (or explicit invalidation hook) on `drt/config/secret_providers/base.py`'s `_value_cache`, so a long-lived `drt serve` process re-fetches periodically rather than never. Needs a design call on default TTL (balance: too short defeats the "don't hit the network every request" reason the cache exists; too long reintroduces the staleness this issue is about) and whether it should be configurable (env var? `drt serve` CLI flag? per-provider?).

**Out:** proactive invalidation (webhook/event-driven from the secret store) — polling-based TTL is enough for v0.9; push-based invalidation is a different feature.

## Related

#782 (parent), #902 (`drt serve`, the deployment mode this actually bites)

PR #930, opened 2026-08-06
feat(config): GCP Secret Manager provider (#782, 2 of 4)

masukai:
**Part 2 of 4 — GCP Secret Manager.** Stacked on #928 (protocol + AWS). Part of #782.

## Design

```yaml
password_env: "gcp-sm://projects/p/secrets/drt-sf/versions/latest#password"
```

One real dialect difference from AWS, worth calling out specifically: AWS's `GetSecretValue` accepts a bare secret id and implicitly resolves to its latest version, but GCP's `AccessSecretVersion` resource name always names a version explicitly — `ref.path` is the whole `projects/*/secrets/*/versions/*` string, with `latest` being a real, resolvable alias rather than an implicit default. Nothing in `parse_secret_uri` needed to change for this — the whole resource name already reads as one continuous path per how that function rejoins `netloc`+`path` (tested against exactly this shape back in #928).

Otherwise the shape matches the AWS leg on purpose: `boto3`/`google-cloud-secret-manager` both lazy-imported inside `fetch()`, same missing-extras `ImportError` pattern, `NotFound` → `LookupError`, everything else bubbles as the raw SDK exception.

## Refactor: `extract_key()` moves to `base.py`

AWS's `#key`-into-JSON-payload extraction (parse JSON, pull a field, reject `null`/`dict`/`list` rather than silently stringifying them — the null-becomes-`"None"` bug caught in #928's self-review) needed to be identical here. Two consumers is where copying ~20 lines stops being the right call, so it's now `base.extract_key(raw, ref, scheme)` — provider-agnostic, `scheme` only shapes the error message so a failure still names the provider a user actually configured. `aws.py` now calls it too; its existing tests pass unmodified since the observable behaviour didn't change, only where it lives. Vault (Part 3) gets this for free.

## Verification

- 11 new tests: 7 GCP `fetch()` paths mirroring the AWS coverage (plain value, JSON+key, not-found, missing extra, client-reuse), 8 direct `extract_key` unit tests (replacing what used to be AWS-only incidental coverage of that logic — now tested at the layer it actually lives at)
- Full suite **2996 passed, 22 skipped**; `ruff` and `mypy` clean

Not run against a real GCP project — same reasoning as #928: no live call in this PR's code path, `google.cloud.secretmanager` / `google.api_core.exceptions` mocked via `sys.modules` injection (established pattern, see `test_gcs_destination.py`). Worth a live check before Part 4 (docs) ships.

## Remaining

Vault (3 of 4), docs (4 of 4).

### DIFF
--- a/drt/config/secret_providers/__init__.py
+++ b/drt/config/secret_providers/__init__.py
@@ -23,8 +23,10 @@ from drt.config.secret_providers.base import (
     register,
     resolve_provider_uri,
 )
+from drt.config.secret_providers.gcp import GcpSecretManagerProvider
 
 register("aws-sm", AwsSecretsManagerProvider())
+register("gcp-sm", GcpSecretManagerProvider())
 
 __all__ = [
     "SecretProvider",
--- a/drt/config/secret_providers/aws.py
+++ b/drt/config/secret_providers/aws.py
@@ -13,10 +13,9 @@ Requires: pip install drt-core[aws-secrets]
 
 from __future__ import annotations
 
-import json
 from typing import Any
 
-from drt.config.secret_providers.base import SecretRef
+from drt.config.secret_providers.base import SecretRef, extract_key
 
 
 class AwsSecretsManagerProvider:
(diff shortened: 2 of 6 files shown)


masukai:
Retroactive note, done while verifying Vault's hvac API for #933: installed \`google-cloud-secret-manager\` in a throwaway venv and checked \`inspect.signature(SecretManagerServiceClient.access_secret_version)\` against the real package rather than resting on documentation search alone (which is all this PR had when it was opened).

Confirmed: \`request: Union[AccessSecretVersionRequest, dict, None]\` — a plain dict (\`{"name": ref.path}\`, as this PR's \`gcp.py\` passes) is in the documented type union, not just something that happened to not crash a \`MagicMock\`. No code change needed here.

masukai:

Same note as #928: already went through a stronger reviewer during implementation (the \`extract_key\`/\`select_field\` split, prompted by GCP needing the JSON-key logic without the string-parsing step). Confirmed the retroactive signature-verification note (the \`request: dict\`-is-valid-per-the-real-type-union check against an actually-installed \`google-cloud-secret-manager\`) is still accurate — didn't just trust the comment, re-ran the same throwaway-venv \`inspect.signature\` check locally... actually didn't re-run it this pass, but the original check was against the real installed package rather than docs, which is the stronger form of evidence already.

No new findings this pass.

masukai (commented):

Issue #919, opened 2026-08-05
fix(state): GCSWatermarkStorage.save()/delete() race — concurrent writers clobber each other


masukai:
## Problem

`GCSWatermarkStorage.save()` (`drt/state/watermark.py:111-117`) loads the whole watermarks blob, mutates one key in the resulting dict, and writes the whole blob back with `upload_from_string()` — no generation precondition:

```python
def save(self, sync_name: str, value: str) -> None:
    data = self._load()
    data[sync_name] = value
    self._blob().upload_from_string(
        json.dumps(data, indent=2),
        content_type="application/json",
    )
```

Two `save()` calls that race — two syncs finishing near-simultaneously, or two `--threads N>1` runs against different syncs sharing one `watermark: {storage: gcs}` backend — both load the same base object, both mutate their own key locally, and whichever `upload_from_string()` lands last wins. The other call's watermark update is silently discarded: on the next run, that sync re-reads from a stale cursor and re-processes rows it already synced (duplicate work, not data loss, since it's a watermark not a delete — but still a correctness bug for any destination that isn't naturally idempotent).

`GCSWatermarkStorage.delete()` has the same shape (load, mutate, write, no precondition).

`BigQueryWatermarkStorage.save()` does **not** have this bug — it uses a `MERGE` statement, which is atomic per row at the SQL engine level.

## Fix

GCS supports preconditioned writes via `if_generation_match` on `upload_from_string()`/`upload_from_file()`. The fix is a load → mutate → conditional-write loop that retries on a precondition-failure (`google.api_core.exceptions.PreconditionFailed` or the 412 it wraps), reloading and re-applying the mutation on conflict — standard optimistic-concurrency pattern, no design change to the class's public interface (`get`/`save`/`delete` keep their current signatures).

`_blob().generation` (or `0` for "must not exist yet", on the first write to a bucket with no object) is the precondition value; `blob.reload()` after a conflict refreshes it before retrying.

## Found by

Caught during review of [ADR 0005](https://github.com/drt-hub/drt/pull/899) (`docs/adr/0005-state-location-and-write-grants.md`'s falsification section) by @Pawansingh3889 — flagged as a bug in code shipped today, independent of that ADR's #756 design questions.

masukai:
[...]

Issue #908, opened 2026-08-05
bug(destinations): Databricks mirror DELETE fails on any composite upsert_key — DELTA_UNSUPPORTED_MULTI_COL_IN_PREDICATE


masukai:
## Problem

`sync.mode: mirror` on Databricks fails at the DELETE pass whenever `destination.upsert_key` has **more than one column**:

```
databricks.sql.exc.ServerOperationError: [DELTA_UNSUPPORTED_MULTI_COL_IN_PREDICATE]
Multi-column In predicates are not supported in the DELETE condition.
```

`_delete_via_staged_keys` (`drt/destinations/databricks.py:614`) builds

```sql
DELETE FROM {table} WHERE (parent_id, id) IN (SELECT parent_id, id FROM {staging})
```

and Delta rejects the multi-column `IN`. A single-column key produces `id IN (SELECT id FROM …)`, which is fine — which is why nothing caught this.

## Scope

**Both mirror paths, so every mirror strategy on Databricks with a composite key:**

| call site | path |
|---|---|
| `databricks.py:732` | plain / scoped mirror (`negate=True`) |
| `databricks.py:970` | tracked mirror (`negate=False`) |

`mirror.scope` + `strategy: tracked` is hit **unconditionally**, because `scope` must be a subset of `upsert_key` — a scope that is the *whole* key is degenerate, so any meaningful scoped configuration has a composite key. Tracked+scoped mirror on Databricks has therefore never worked.

Present since the staging anti-join landed in #707 (v0.7.11) and shipped in every release since, including v0.8.4.

## Evidence

Live run against the smoke workspace, [dwh-smoke #30966965525](https://github.com/drt-hub/drt/actions/runs/30966965525):

```
tests/integration/dwh/test_databricks_smoke.py::test_databricks_mirror_deletes_unobserved_keys PASSED   # upsert_key: ["id"]
tests/integration/dwh/test_databricks_smoke.py::test_databricks_tracked_scoped_mirror_and_scope_backfill FAILED  # upsert_key: ["parent_id", "id"]
```

```
drt/destinations/databricks.py:970 in _finalize_mirror_tracked
    self._delete_via_staged_keys(
drt/destinations/databricks.py:686 in _delete_via_staged_keys
    cur.execute(delete_sql)
→ DELTA_UNSUPPORTED_MULTI_COL_IN_PREDICATE
```

The same run passed Snowflake's equivalent test (10/10), so this is specific to Delta rather than to the tracked+scoped feature.

[...]

Issue #903, opened 2026-08-03
feat(cli): OIDC JWT verification for drt serve — the Pub/Sub push leg of #854


Pawansingh3889:
Split out of #854 per the discussion there (masukai okayed splitting auth if it ballooned the PR).

## Problem

#854 made `drt serve` auth pluggable (`--auth none|bearer|hmac`). HMAC covers body-signing senders (GitHub, Shopify, bare hex). But Pub/Sub push authenticates differently: an **OIDC JWT in the `Authorization` header** — verify the signature against Google's public keys, check `aud` / `iss` / `email`. No body signature at all.

So the ADR 0004 Tier 3 example — a Pub/Sub push subscription pointed at `drt serve` — still needs a verifying proxy in front. The `--auth` dispatch was built so this is an addition, not a rewrite: this issue adds `--auth oidc`.

## The dependency question (why this was split out)

JWT signature verification against rotating JWKS is not stdlib-able in any reasonable way. Options:

- **PyJWT** (+ `cryptography`) — small, standard, but `cryptography` is a heavy wheel
- **google-auth** — purpose-built for exactly this (`google.oauth2.id_token.verify_oauth2_token`), already an indirect dependency for anyone using the BigQuery extra
- hand-rolled JWKS fetch + RS256 — no; verification code you write yourself is verification code you get wrong

Core's rule is no heavy deps outside extras, so whichever library wins, this lands as an extra (`drt-core[serve-oidc]`, or folded into an existing google extra) with a clear error message when `--auth oidc` is requested without it.

## Scope

- `--auth oidc` + `--oidc-audience` (required; the URL Pub/Sub was configured with) and `--oidc-issuer` / `--oidc-email` checks
- Verification failures → 401, same as the other schemes
- Docs: the Pub/Sub push section of `docs/guides/using-webhook-trigger.md` loses its "needs a proxy" caveat
- Out of scope: any other IdP than Google's (the header/JWKS mechanics generalize, but only the Pub/Sub leg has a named consumer today)

## Related

- #854 / PR #902 (the auth dispatch this plugs into)
- ADR 0004 Tier 3 (the Pub/Sub push example this unblocks)

Issue #896, opened 2026-08-03
feat(config): dbt-manifest-derived column contracts (source: dbt for #761)


masukai:
## Problem

#761 lets a sync declare `columns:` with types and enforce them before load — but that declaration is hand-maintained in the sync YAML, separate from the dbt model it actually reads. When a dbt model's output schema changes (a column renamed, dropped, or retyped), #761's contract only catches it if someone remembered to update the drt-side `columns:` block too. In practice the two drift independently, and the failure mode is exactly the one #761 was written to prevent: a schema change reaches HubSpot properties / ad-platform conversions / a production Postgres table before anyone notices.

dbt already has the answer to "what does this model actually output" in `target/manifest.json` (declared contract, if the model uses one) and `target/catalog.json` (the warehouse's actual introspected columns, from `dbt docs generate`) — drt just doesn't read either.

## Prior art

- dbt's own model contracts (`contract: enforced` + `columns:` in the model's `schema.yml`) are the source of truth this issue proposes reading *from*, not duplicating.
- Great Expectations / Elementary both have a "diff against last known schema" checkpoint that dbt-adjacent tools commonly wire into CI.

## Proposal

When a sync's `model: ref(...)` points at a dbt model that has its own `columns:` contract declared, let `columns:` in the drt sync be **omitted** and derived from the dbt manifest instead:

```yaml
sync:
  model: ref('mart_users')
contract:
  enforced: true
  source: dbt   # new — derive expected columns/types from target/manifest.json instead of a hand-written columns: block
```

- Requires a `--manifest-path` (or project-relative default `target/manifest.json`) so drt can resolve `ref()` → model → its dbt-declared columns without invoking dbt itself.
- A dbt model with no `columns:` contract of its own has nothing to derive from — falls back to #761's hand-written block, or no contract at all. This is additive, not a replacement.
- `drt validate` gains a check: does every `source: dbt` contract actually resolve a manifest entry with columns declared? (fails fast on a typo'd `ref()` or a model that hasn't opted into dbt's own contract yet).

## Scope

**In:** manifest parsing (`ref()` → model → `columns:`), the `contract.source: dbt` config surface, `drt validate` resolution check.
**Out:** generating a dbt contract from drt's side (dbt owns the model, drt only reads); catalog.json-based introspection as a fallback for models without a declared dbt contract (worth a follow-up if this lands and people ask for it).

## Related

- #761 (column contracts) — this is a `source:` option on top of it, not a competing mechanism.
[...]

PR #932, opened 2026-08-06
docs(roadmap): re-scope v0.9 to what's shippable, fix v0.8.5/#854 staleness

masukai:
## What

- Adds the missing **v0.8.5** section to ROADMAP.md (patch for #908, shipped 2026-08-05 — was documented in CHANGELOG.md but absent here).
- Fixes v0.10's #854 mention — it described the `drt serve` concurrency contract as future scope; it shipped ahead of schedule via #902. Points to its direct follow-on (#903) instead.
- Moves #781 (the one open issue left in the v0.8 milestone) out, so **v0.8 can close** — it shipped through .5 with 0 open issues left otherwise.
- Re-scopes **v0.9 "Engine Foundation"** from 24 open issues down to **8** that are actually close to shippable for the September target: #756, #762, #763, #772, #780, #782, #890, #922.
- Updates CLAUDE.md's "Current Status" — it claimed nothing had landed on `main` since v0.8.4, which was stale (v0.8.5 + 7 more commits had shipped).

## Why the v0.9 cut

Most of what's being moved out depends on an **unbuilt ADR 0005 primitive** (implementation-ordering step 3, "warehouse managed-table primitive") and has zero PRs in flight, one month from target:

- **#755** diff-based incremental — the milestone's own headline feature, but it's step 5 of 5 in ADR 0005's ordering; step 3 isn't built yet.
- **#760/#761/#896** — the whole "Schema management" subsection, same step-3 dependency, moved as a unit (#896 depends on #761's contract shape).
- **#758/#764/#778** — each individually large with nothing banked; #764 is additionally blocked on #762's `run_id` plumbing.
- **#425/#426/#428/#649/#825** — the "older backlog carried forward" cluster; never cleanly fit an earlier milestone, not ADR-0005 work either. #428 needs re-litigating against ADR 0004's "no daemon" posture before it's scheduled anywhere.
- **#757** match_policy — scoped as "remaining destinations," so it can never reach zero and can't sit inside a milestone that's supposed to close. Converted to an open-ended tracking issue, no milestone (comment: #757).
- **#919** — a live data-loss bug (`GCSWatermarkStorage` race), not a v0.9 feature. Flagged as a v0.8.6 patch candidate instead, same pattern as #908 → v0.8.5 (comment: #919).
- **#469** → v0.10 (Protocol-stability theme; premise partly overtaken by #900's Protocol refactor already landing).
- **#781** → v0.10 (Ecosystem theme; only landed in v0.9 by deferral from v0.8, not thematic fit).

What's staying: 6 of the 8 kept issues already have PRs in flight (13 open PRs across #763/#780/#890/#762/#772/#782) — that's the strongest signal of what's actually close.

Milestones re-labeled on GitHub to match (v0.9: 24→8 open, v0.10: +3, v0.8: 0 open).

## Test plan

- [x] Docs-only change, no code touched
- [x] Cross-checked every moved/kept issue number against `gh issue list --milestone` output before and after
- [x] `gh release list` / `git tag` confirm v0.8.5 is the real latest tag
- [x] `gh issue view 854` confirms closed, contradicting the old v0.10 text


masukai:
Self-review, cross-referencing from #923 (also mine, also open) — flagging a sequencing issue between the two rather than duplicating the full analysis.

This PR's v0.9 rewrite (the "State foundation (ADR 0005)" restructuring, #755/#758 moved to Cut, #890 to its own "Perf close-out" bullet) and #923's second commit both touch the same "Incremental & state" bullet in ROADMAP.md, and this one is the more current, more comprehensive rewrite — #923's is a narrower correction (enriching the #756 sub-clause with the object-storage/warehouse split) written against the *pre-restructure* bullet shape.

No content conflict in the sense of disagreeing about facts — everything #923 says about #756 (object-storage ships first, no warehouse write) is consistent with how this PR already describes it. It's purely a "these two diffs touch the same lines differently" sequencing problem. Suggest merging this one first since it's the fuller picture, then #923's ROADMAP.md commit becomes redundant (its ADR 0004 correction, a separate file, is independent and still worth landing on its own). Left the actual call on #923 rather than resolving it here — full detail on that thread.

Everything else in this PR's own "Why the v0.9 cut" reasoning checks out against what I have direct context on: #921 isn't explicitly named in the diff I read, worth double-checking it's captured somewhere (I filed it during the ADR 0005 follow-up work, timing-adjacent to this PR) — but that's a completeness check, not a correctness one, and may just be a timing gap between when this was drafted and when #921 got filed.

masukai:
Correction to my own comment above: #921 *is* accounted for — it's in the v0.10 'Event-driven activation' section of this same diff (correctly explained as the sensor-fan-out case), just not in the v0.9 section I was looking at when I wrote that. False alarm on the completeness point; the rest of the comment (the sequencing overlap with #923) stands.

PR #933, opened 2026-08-06
feat(config): HashiCorp Vault provider (#782, 3 of 4)

masukai:
**Part 3 of 4 — HashiCorp Vault.** Stacked on #930 (GCP) → #928 (protocol + AWS). Part of #782.

## Design

```yaml
password_env: "vault://secret/data/drt/snowflake#password"
```

`ref.path` mirrors Vault's own raw KV v2 HTTP path — mount point, the literal `data` segment KV v2's API inserts to distinguish a data read from a metadata/delete operation on the same secret, then the logical secret path. That's the exact shape `vault kv get` / curl examples against Vault already use, so it's what an operator configuring this recognizes without translation. hvac's higher-level client wants the mount and the logical path passed separately (`read_secret_version(path=, mount_point=)`), so `_split_kv2_path()` splits it here rather than making every caller do that parsing.

**`#key` is required**, unlike AWS/GCP — a KV v2 secret's payload is always a field map (there's no "the secret is just one string" case), so there's no sensible value to return without naming which field. Fails fast with a clear message rather than guessing.

## Refactor: `extract_key` splits into `extract_key` + `select_field`

Vault's payload is already a parsed dict by the time it reaches this code (`response["data"]["data"]`) — it never needs the JSON-parsing step AWS/GCP's `extract_key` does. Rather than have Vault round-trip through `json.dumps` just to reuse a string-shaped function, `base.py` now has:
- `select_field(payload: dict, ref, scheme)` — the actual guard logic (key present? not `null`/`dict`/`list`?)
- `extract_key(raw: str, ref, scheme)` — parses `raw` as JSON, then calls `select_field`

AWS/GCP call `extract_key` (unchanged behavior — their own tests pass unmodified, confirming the split didn't move any observable behavior). Vault calls `select_field` directly.

## Verification done differently this time

Rather than resting on documentation search the way #928/#930 initially did, I installed both `hvac` and `google-cloud-secret-manager` in a throwaway venv and checked `inspect.signature()` against the real packages:

```
hvac read_secret_version: (path, version=None, mount_point='secret', raise_on_deleted_version=None)
gcp access_secret_version: (self, request: Union[AccessSecretVersionRequest, dict, None] = None, ...)
```

Both confirm what's already in the code (this PR's `path=`/`mount_point=`, and retroactively #930's `request={"name": ...}` — a plain `dict` is in the documented type union, not just something that happened to not crash a mock). `hvac.exceptions.InvalidPath` and `google.api_core.exceptions.NotFound` both exist with the expected import paths. This closes a real gap the mocked tests structurally can't: `assert_called_once_with(...)` can't catch a wrong keyword-argument name, since `MagicMock` accepts anything silently.

## ⚠️ The client is *not* cached — asymmetric with the AWS/GCP legs, deliberately

`boto3` and `google-auth` both refresh their own credentials internally, so caching those SDKs' clients for the process's lifetime (as `AwsSecretsManagerProvider`/`GcpSecretManagerProvider` do) is harmless. `hvac.Client()` captures `VAULT_TOKEN` once at construction and **never refreshes it**, and Vault service tokens are conventionally short-TTL. Caching the client here would mean a long-lived `drt serve` process eventually holds an expired token — every subsequent Vault fetch would then fail outright, which is worse than #929's "may serve a rotated-but-still-valid value" gap (that one degrades gracefully; this one wouldn't have). `hvac.Client()` construction is cheap (no network call, just wraps connection params), so a fresh client per `fetch()` call is the simple fix rather than adding token-renewal handling. Documented at length next to `_get_client()`.

## Verification

- 18 new tests: 5 `_split_kv2_path` (default/custom mount, nested paths, both malformed-path rejections), 5 `select_field` (mirroring what used to be `extract_key`-only coverage, now tested at the layer each behavior actually lives at), 8 `VaultProvider.fetch()` paths (success, custom mount, no-`#key` rejection, malformed-path rejection before any client call, not-found, missing extra, and the fresh-client-per-call behavior replacing what would otherwise have been a "cached and reused" assertion)
- Full suite **3013 passed, 22 skipped**; `ruff` and `mypy` clean

Not run against a real Vault server — no live call happens in this PR's code path; `hvac` is mocked via `sys.modules` injection (established pattern). The signature verification above closes the biggest gap that leaves, but a live check is still worth doing before Part 4 ships, batched with AWS/GCP.

## Remaining

Docs (4 of 4) — IAM/policy snippets for all three providers, plus the `sync-unit-tests`-style guide.

### DIFF
--- a/drt/config/secret_providers/__init__.py
+++ b/drt/config/secret_providers/__init__.py
@@ -24,9 +24,11 @@ from drt.config.secret_providers.base import (
     resolve_provider_uri,
 )
 from drt.config.secret_providers.gcp import GcpSecretManagerProvider
+from drt.config.secret_providers.vault import VaultProvider
 
 register("aws-sm", AwsSecretsManagerProvider())
 register("gcp-sm", GcpSecretManagerProvider())
+register("vault", VaultProvider())
 
 __all__ = [
     "SecretProvider",
--- a/drt/config/secret_providers/base.py
+++ b/drt/config/secret_providers/base.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 
 import json
 from dataclasses import dataclass
-from typing import Protocol
+from typing import Any, Protocol
 from urllib.parse import urlparse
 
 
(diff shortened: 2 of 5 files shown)


masukai:

This is the one that got the most scrutiny during implementation — the Vault client-caching asymmetry (\`hvac.Client()\` doesn't refresh \`VAULT_TOKEN\` the way boto3/google-auth self-refresh) was a real finding from a stronger reviewer, not something I'd have caught on my own. Confirmed it's actually implemented as described rather than just documented: \`VaultProvider\` has no \`__init__\`/\`self._client\` at all now, and \`_get_client()\` constructs fresh every call — checked the current file directly rather than trusting the PR description.

Fresh angle this pass: the \`_split_kv2_path\` string-splitting on \`/\` — could a mount or path name containing a literal \`/\` inside a single path segment (unusual but not impossible for a Vault mount name) confuse the split? \`parts = path.split("/")\` then \`parts[0]\` as mount, \`"/".join(parts[2:])\` as the rest — this only assumes \`parts[1] == "data"\`, and doesn't care how many \`/\`-separated segments follow, so a deeply-nested secret path is fine (already covered by \`test_deeply_nested_path\`). A mount name itself containing a \`/\` isn't something Vault allows (mount names are a single path segment by Vault's own convention), so not a real gap.

No new findings.

masukai (commented):

PR #934, opened 2026-08-06
docs(config): secret provider URIs guide + API reference (#782, 4 of 4)

### LINKED ISSUE #782: feat(config): secret provider URIs — AWS/GCP Secret Manager & Vault resolution for credentials (as extras)

masukai:
**Part 4 of 4 — guide + API reference.** Stacked on #933 (Vault) → #930 (GCP) → #928 (protocol + AWS). Closes #782.

## What's here

- New `docs/guides/secret-provider-uris.md`, following the existing `pii-masking.md` template (quick start → per-topic detail → composition/caveats → what it isn't → related):
  - Quick start showing a mixed profile (plain env vars alongside a provider URI)
  - The resolution chain, extending `.drt/secrets.toml`'s existing precedence list with the new final step
  - Per-provider section (URI shape, `#key` semantics, required extra, auth model) **plus an IAM/policy snippet each** — AWS policy JSON scoped to a secret prefix, a `gcloud` IAM binding command, a Vault HCL policy
  - A caching section that states the `drt serve` caveat plainly (secret held until restart, no TTL yet, links #929) and separately documents Vault's client-not-cached asymmetry from #933
  - "What this isn't" — no write/rotation, no encryption-at-rest (that's #303's scope), no OAuth brokering, no third-party plugin loading (that needs #297 first)
- `docs/llm/API_REFERENCE.md`: the resolution-order line now ends in `> provider URI`, plus a compact scheme/extra/example table pointing at the full guide

## Verified before writing

- Both relative links (`using-webhook-trigger.md` for the `drt serve` reference, `../llm/API_REFERENCE.md` for the secrets.toml cross-reference) resolve to files that exist on this branch — caught and fixed one broken link (`../guides/../llm/...`, a leftover from an earlier draft) before it shipped
- The "profile or destination config" claim in the intro is accurate, not just plausible-sounding: `resolve_env` is called from both `drt/sources/*.py` (5 files) and `drt/destinations/*.py`, confirmed by grep, not destinations only
- No CHANGELOG.md entry, matching how this same stack's earlier PRs (and #918/#924 before it) handled it — feature entries land at release time from the full commit log, not per-PR
- No markdown linter exists in this repo's toolchain (`make lint` is `ruff`+`mypy` over `drt`/`tests` only), so there's no CI signal for a docs-only PR beyond the standard test matrix — checked manually instead

## Verification

Docs-only change. Full suite still **3013 passed, 22 skipped** (reran to confirm rather than assume markdown can't affect it); `ruff` and `mypy` clean (unchanged, no `.py` touched).

### DIFF
--- a/docs/guides/secret-provider-uris.md
+++ b/docs/guides/secret-provider-uris.md
@@ -1,0 +1,184 @@
+# Secret Provider URIs — Resolve Credentials from a Managed Secret Store
+
+Production reverse ETL concentrates *write* credentials to CRMs, ad
+platforms, and warehouses — exactly the secrets an org keeps in AWS
+Secrets Manager, GCP Secret Manager, or Vault, with rotation policies
+attached. The default path (export into an env var in every runner — CI
+YAML, Airflow connections, cron wrappers) duplicates secret-distribution
+logic per orchestrator and breaks rotation the moment it's exported: the
+env snapshot goes stale even though the store itself rotated cleanly.
+
+Any `*_env` field in a profile or destination config can be a
+`scheme://...` URI instead of a plain env var name. drt resolves it from
+the matching provider at connection time — no wrapper script, no
+re-export step.
+
+## Quick Start
+
+```yaml
+# ~/.drt/profiles.yml
+prod:
+  type: snowflake
+  account_env: SNOWFLAKE_ACCOUNT
+  user_env: SNOWFLAKE_USER
+  password_env: "aws-sm://prod/drt/snowflake#password"
+```
+
+`account_env` and `user_env` above are still plain env var names — nothing
+about this feature requires every credential in a profile to go through a
+provider. Mix and match per field.
+
+## Resolution chain
+
+```
+explicit YAML value  >  OS environment variable  >  .drt/secrets.toml  >  provider URI
+```
+
+The four steps are tried in order and the first that resolves wins — see
+`docs/llm/API_REFERENCE.md`'s `.drt/secrets.toml` section for the first
+three. A provider URI only gets tried once the other three have all come
+back empty, which is automatic: `aws-sm://prod/drt/snowflake#password`
+doesn't collide with a real env var name or a `secrets.toml` key, so it
+falls through unchanged.
+
+## Providers
+
+### AWS Secrets Manager — `aws-sm://`
+
+```
+aws-sm://<secret-id-or-arn>#<key>
+```
+
+- `<secret-id-or-arn>` is whatever `GetSecretValue` accepts — a plain
+  secret name (`prod/drt/snowflake`) or a full ARN.
+- `#<key>` is optional. Omit it when the secret **is** the value (a
+  single password stored as a plain string). Include it to select one
+  field out of a secret stored as a JSON object holding several related
+  values under one id — `aws-sm://prod/drt/snowflake#password` alongside
+  a sibling `#username` in the same secret, say.
+- Requires: `pip install drt-core[aws-secrets]`
+- Auth: boto3's own default credential chain (env vars,
+  `~/.aws/credentials`, an attached IAM role) — nothing drt-specific to
+  configure.
+
+Minimal IAM policy for a secret (or a prefix of them) drt should be able
+to read:
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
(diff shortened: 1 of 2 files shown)


masukai:

Docs-only, and already covered the two things worth independently checking when I wrote it: both relative links resolve to real files, and the "profile or destination config" claim in the intro is grep-verified against \`drt/sources/*.py\` and \`drt/destinations/*.py\`, not just plausible-sounding.

Nothing new on a fresh pass — this PR's job is to describe #928/#930/#933 accurately, and having just re-checked those three, the guide's claims (resolution chain order, the Vault \`#key\`-required rule, the \`drt serve\` caching caveat pointing at #929) all still match what's actually in the code.

masukai (commented):
Self-review (staleness check, requested explicitly) — checking whether this PR's prose still matches what's actually shipping, since #938 (a later PR in this same stack) landed after this one was written.

Real gap: the "Caching, and what it means for `drt serve`" section describes the cache as having **no TTL**, tells operators to "restart the server after rotation... until that lands," and links #929 as the open tracking issue:

> There is currently no TTL or background refresh ([tracked in #929](...)). If your deployment rotates secrets on a schedule and runs `drt serve` continuously, restart the server after rotation... until that lands.

#938 (`fix(config): TTL on secret provider cache so drt serve picks up rotation`, closes #929) already ships a `DRT_SECRET_CACHE_TTL_SECONDS`-configurable TTL (default 300s) — I reviewed and confirmed it earlier today, including tracing that `drt serve` genuinely re-enters `resolve_provider_uri()` on every triggered sync, so the fix isn't cosmetic. #938 doesn't touch this doc file at all.

If #938 merges after this PR (it's 5th on the stack, based on this branch), the guide will ship describing a restart-workaround for a limitation the code no longer has, and #929 will read as "open" in a published doc after the issue is closed. Worth updating this section before or alongside #938 — the TTL/env var, its default, and that it's now bounded rather than unbounded.

Everything else — the IAM/policy snippets, the Vault client-not-cached explanation, the resolution-order description — checked against the corresponding provider code and matches.

PR #935, opened 2026-08-06
fix(state): guard GCS watermark writes with a generation precondition

### LINKED ISSUE #919: fix(state): GCSWatermarkStorage.save()/delete() race — concurrent writers clobber each other

Pawansingh3889:
GCSWatermarkStorage.save() read the whole watermarks blob, set one key on the result, and uploaded the whole blob back with no precondition. Two writes that overlapped both read the same base object, so the later upload won outright and dropped the other sync's watermark with no error raised anywhere. That sync then resumed from a stale cursor on its next run and re-processed rows it had already synced. delete() had the same shape.

Both now read the object's generation, apply the change, and upload with if_generation_match, so a write that lost its race gets a 412 and retries against freshly read state instead of overwriting it. The download is pinned to the generation just read, so the bytes and the precondition token cannot come from different versions. A first write to a bucket with no object asserts if_generation_match=0.

Retries are bounded at 5. Exhausting them raises WatermarkContentionError rather than returning quietly, because a write that gives up in silence is the failure this change exists to prevent.

get() is untouched, being a plain read with nothing to be conditional about. BigQueryWatermarkStorage was never affected: its save() is a MERGE.

The write path now needs google-api-core for the precondition exception types. It is already transitive via the [gcs] extra, and is resolved through the same lazy import as the GCS client so a core install is unaffected.

Closes #919

## Verification

The tests model GCS generation semantics rather than asserting on mock call args, so the
race is exercised rather than described: a competing write lands between the reader's read
and its upload, at exactly the point a real race occurs.

Reverting only the `if_generation_match` argument, leaving the retry loop intact:

```
FAILED test_concurrent_save_does_not_discard_the_other_writer
E       AssertionError: assert {'sync_a': 'new_a'} == {'sync_a': 'new_a', 'sync_b': 'b_value'}
E         Right contains 1 more item:
E         {'sync_b': 'b_value'}
```

`sync_b` vanishing is the reported bug. Restored, the suite is green.

Three existing tests needed the widened mock surface, since the write path now resolves the
precondition exception types.

## Relevance to #756

ADR 0005's write-amplification note describes this same read-modify-write shape, so the
generation-precondition design here is a usable precedent for the remote state backend
rather than a one-off.

## Checklist

- [x] Tests pass (`make test`): 2943 passed, 18 skipped
- [x] Linter passes (`make lint`): ruff clean, mypy clean across 175 source files


masukai (changes requested):
Nice fix — the concurrent-write test (`test_concurrent_save_does_not_discard_the_other_writer`) is exactly the right shape to catch the original bug, and pinning the download to the just-read generation is the correct instinct.

One gap in the retry coverage though: `_read_for_update`'s `download_as_text` call is only guarded against `not_found`, not `precondition_failed`:

```python
try:
    raw = blob.download_as_text(if_generation_match=generation)
except not_found:
    return {}, 0
```

But a generation mismatch on a `GET` with `ifGenerationMatch` — object still exists, just at a newer generation than the `reload()` a moment earlier saw — is a 412, i.e. `PreconditionFailed`, not 404. Your own fake proves it:

```python
def download_as_text(self, if_generation_match: int | None = None) -> str:
    if self._store.generation == 0:
        raise _NotFound("no such object")
    if if_generation_match is not None and if_generation_match != self._store.generation:
        raise _PreconditionFailed("generation mismatch on read")
```

So the window between `blob.reload()` and `blob.download_as_text(if_generation_match=generation)` is itself racy: if a competing writer's full read-modify-write completes in that gap, `download_as_text` raises `PreconditionFailed`, which `_read_for_update` doesn't catch, and `_write`'s `try/except precondition_failed` only wraps the `upload_from_string` call, not the call to `_read_for_update`. So instead of consuming one of the `_MAX_WRITE_ATTEMPTS` and retrying, this specific race propagates an unhandled `PreconditionFailed` straight out of `save()`/`delete()` — the one outcome `WatermarkContentionError` exists to avoid.

None of the current tests exercise this path — the `before_upload` hook lands the competing write right before the upload, not between `reload()` and `download_as_text()`.

Suggest folding the read into the same retry boundary as the write, e.g. catching `precondition_failed` around the `_read_for_update()` call in `_write` and `continue`-ing, same as the upload branch does. Happy to be wrong if I'm missing a reason GCS can't actually produce this ordering, but the fake object you wrote models it as reachable.

(Self-review note: I read this as a maintainer, not as the PR author — flagging before merge rather than after, per the same standard I've been holding my own PRs to this pass.)

Pawansingh3889:
You're right, and the fake did model it as reachable, which makes it worse: the ordering was already expressible in the harness I wrote and I still only guarded the upload.

Fixed in `730f781`. `_read_for_update()` now sits inside the same retry boundary as the upload rather than in front of it:

```python
for _ in range(_MAX_WRITE_ATTEMPTS):
    blob = self._blob()
    try:
        data, generation = self._read_for_update(blob)
    except precondition_failed:
        continue  # someone else wrote mid-read; re-read from scratch
```

So a writer landing between `reload()` and the pinned `download_as_text()` costs one attempt and re-reads, instead of escaping `save()`/`delete()` as an unhandled `PreconditionFailed`.

On the test gap, you were right that `before_upload` cannot reach this ordering. The fake now has a `before_download` hook that fires between `reload()` and the download, and two tests use it:

- `test_writer_landing_mid_read_is_retried_not_raised`: one competing writer lands mid-read, the save retries, and both watermarks survive rather than one being dropped.
- `test_sustained_mid_read_contention_fails_loudly`: every read loses its pin, and it raises `WatermarkContentionError` after `_MAX_WRITE_ATTEMPTS`, the same way losing every upload does.

I also extended the CHANGELOG entry so it describes the read sitting inside the boundary, not just the write.

Separately, codecov flagged the two error branches on that same read as uncovered, so `8fb1ac5` closes both:

- `test_object_deleted_mid_read_starts_from_absent`: the object is deleted between `reload()` and the pinned download. That is a 404 rather than a 412, so it is not contention and should not cost an attempt, and the write then proceeds under `if_generation_match=0`.
- `test_unparseable_object_is_replaced_at_its_own_generation`: unreadable contents are given up on, but the generation is not. Returning `0` there would pin the write to "only if still absent" against an object that plainly exists, turning a self-healing case into `WatermarkContentionError`.

That takes patch coverage from 83.33% to 89.58%. The 5 statements still uncovered are all of `_gcs_precondition_errors()`, which I have left deliberately: it is the lazy-import helper and every test patches it, the same reason `_gcs_client()` directly above it is uncovered today. Happy to exercise the real import instead if you would rather not carry that gap.

Pawansingh3889:
You're right, and the fake did model it as reachable, which makes it worse: the ordering was already expressible in the harness I wrote and I still only guarded the upload.

Fixed in `730f781`. `_read_for_update()` now sits inside the same retry boundary as the upload rather than in front of it:

```python
for _ in range(_MAX_WRITE_ATTEMPTS):
    blob = self._blob()
    try:
        data, generation = self._read_for_update(blob)
    except precondition_failed:
        continue  # someone else wrote mid-read; re-read from scratch
```

So a writer landing between `reload()` and the pinned `download_as_text()` costs one attempt and re-reads, instead of escaping `save()`/`delete()` as an unhandled `PreconditionFailed`.

On the test gap, you were right that `before_upload` cannot reach this ordering. The fake now has a `before_download` hook that fires between `reload()` and the download, and two tests use it:

- `test_writer_landing_mid_read_is_retried_not_raised`: one competing writer lands mid-read, the save retries, and both watermarks survive rather than one being dropped.
- `test_sustained_mid_read_contention_fails_loudly`: every read loses its pin, and it raises `WatermarkContentionError` after `_MAX_WRITE_ATTEMPTS`, the same way losing every upload does.

I also extended the CHANGELOG entry so it describes the read sitting inside the boundary, not just the write.

Separately, codecov flagged the two error branches on that same read as uncovered, so `8fb1ac5` closes both:

- `test_object_deleted_mid_read_starts_from_absent`: the object is deleted between `reload()` and the pinned download. That is a 404 rather than a 412, so it is not contention and should not cost an attempt, and the write then proceeds under `if_generation_match=0`.
- `test_unparseable_object_is_replaced_at_its_own_generation`: unreadable contents are given up on, but the generation is not. Returning `0` there would pin the write to "only if still absent" against an object that plainly exists, turning a self-healing case into `WatermarkContentionError`.

That takes patch coverage from 83.33% to 89.58%. The 5 statements still uncovered are all of `_gcs_precondition_errors()`, which I have left deliberately: it is the lazy-import helper and every test patches it, the same reason `_gcs_client()` directly above it is uncovered today. Happy to exercise the real import instead if you would rather not carry that gap.

masukai (approved):
This closes the gap cleanly. I traced through `_write` and `_read_for_update`
line by line: the read is now inside the same retry loop as the upload,
`precondition_failed` around `_read_for_update()` retries instead of
escaping, and the 404-vs-412 distinction is preserved correctly in both
directions — `test_object_deleted_mid_read_starts_from_absent` confirms a
genuine delete mid-read still returns generation `0` (fresh write), while
`test_unparseable_object_is_replaced_at_its_own_generation` confirms a
corrupt-but-present object keeps its real generation rather than falling
back to `0` and deadlocking against `WatermarkContentionError`. That second
case is the one I'd have gotten wrong if I'd written this fix myself —
good catch keeping those two failure modes distinct instead of collapsing
them into one "return 0" branch.

`test_writer_landing_mid_read_is_retried_not_raised` /
`test_sustained_mid_read_contention_fails_loudly` are exactly the race
window my comment described, and the `before_download` hook is the right
way to land a competing writer at that specific point rather than only at
upload time.

On the remaining 5 uncovered lines in `_gcs_precondition_errors()`: agreed,
leave it. It's the same shape as `_gcs_client()` right above it being
uncovered today for the same "every test patches it" reason — not a gap
this PR introduced, and not worth a special-case test here. If it's ever
worth closing, it'd be a small follow-up covering both lazy-import helpers
uniformly rather than something specific to this fix.

Approving.

masukai:
Merged — thanks for the careful follow-through here, this was a real bug (#919) and the fix + tests are in great shape now. Appreciate the thoroughness on the mid-read race in particular, that's an easy thing to miss.

PR #937, opened 2026-08-07
chore: point AGENTS.md at CLAUDE.md for Codex compatibility

masukai:
## Summary
- Add `AGENTS.md` as a symlink to `CLAUDE.md` so Codex CLI picks up the same agent context as Claude Code, with `CLAUDE.md` remaining the single source of truth.

## Test plan
- [x] Verified `AGENTS.md` resolves to `CLAUDE.md` content locally
- [x] Confirmed no existing `AGENTS.md` was overwritten

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ cd drt
 uv sync --extra dev --extra bigquery
 ```
 
-**Note for Windows contributors:** `AGENTS.md` is a symlink to `CLAUDE.md` (kept in sync for Codex CLI). If `git clone` checks it out as a plain text file containing just `CLAUDE.md` instead of the real file contents, your Git isn't creating real symlinks — run `git config core.symlinks true` (may also require enabling Windows Developer Mode) and re-clone.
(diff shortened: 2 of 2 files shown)


masukai:
## Review

Reviewed as a 1-line, non-breaking change. CI (test x4, CodeQL, Analyze, CLA, codecov) all green.

**Good:**
- Keeps `CLAUDE.md` as the single source of truth (symlink, not a copy) — no drift risk.
- Branched from `main`, no unrelated changes.

**Flagged and addressed:**
- No `.gitattributes`/config pins symlink checkout behavior — on a contributor environment without real symlink support (e.g. Windows without Developer Mode / `core.symlinks=false`), `AGENTS.md` checks out as a plain text file containing the literal string `CLAUDE.md` rather than the real content, so Codex CLI would silently get 9 bytes instead of the intended context. This is the first symlink in this repo, so there was no existing precedent/handling for it.
  - Added a note to `CONTRIBUTING.md` (68aa213) telling affected contributors how to fix it (`git config core.symlinks true` + re-clone). Not adding checkout-time tooling for this — disproportionate to a docs-pointer file, and the failure mode is silent-but-harmless (no context, not broken behavior).
- Did not independently verify Codex CLI follows the symlink at read time (only verified `readlink`/`cat` resolve correctly locally) — reasonable to trust OS-transparent file reads here.

No further changes requested. Approving.

masukai:
**Codex review:**

The symlink itself is valid, but the newly documented Windows recovery procedure loses its configuration when the user re-clones. This leaves the compatibility problem unresolved for the exact users the note targets.

Review comment:

- [P2] Persist the symlink setting before re-cloning — CONTRIBUTING.md:27
  On Windows clones where `AGENTS.md` is checked out as text, this command runs after `cd drt`, so `git config` writes `core.symlinks` only to that clone's `.git/config`; deleting and re-cloning discards the setting and can reproduce the same plain file. Use a persistent setting before cloning, or set it locally and re-check out `AGENTS.md`, so the documented recovery actually restores Codex context.

masukai:
Addressed in 6eb1812: `core.symlinks` set locally now points to `--global` (before cloning) or a local set + re-checkout instead of re-clone, so the recovery steps actually work.

PR #938, opened 2026-08-07
fix(config): TTL on secret provider cache so drt serve picks up rotation (#929)

### LINKED ISSUE #929: fix(config): secret provider cache has no TTL — drt serve holds a resolved secret until restart

masukai:
## Summary
- `_value_cache` in `drt/config/secret_providers/base.py` cached provider-resolved secrets (`aws-sm://`, `gcp-sm://`, `vault://`) for the process's full lifetime — fine for a short-lived `drt run`, but `drt serve` is a long-lived process that re-enters this path on every triggered sync, so a rotated secret was never picked up until restart. Contradicted #782's own "breaks rotation" problem statement.
- Cache entries now store `(value, fetched_at)`; a lookup past TTL refetches and overwrites the entry.
- Default TTL: 300s. Configurable via `DRT_SECRET_CACHE_TTL_SECONDS`, read lazily per lookup. Non-positive value disables caching entirely.
- Updated `vault.py`'s comment referencing the old "known gap, unbounded" framing to describe the now-bounded staleness window.

Closes #929.

**Stack:** 5th PR on the #782 stack, based on #934 (`feat/782-docs`). Merge order: #928 → #930 → #933 → #934 → this.

## Test plan
- [x] `pytest tests/unit/test_secret_providers.py -q` — 55 passed (added 2 new tests: post-TTL-expiry refetch, TTL=0 disables caching; extended the existing cache test to assert pre-expiry behavior)
- [x] `ruff check` — clean
- [x] `mypy drt/config/secret_providers/base.py` — clean

### DIFF
--- a/drt/config/secret_providers/base.py
+++ b/drt/config/secret_providers/base.py
@@ -11,6 +11,8 @@ level, matching the existing ``s3``/``bigquery`` destination pattern).
 from __future__ import annotations
 
 import json
+import os
+import time
 from dataclasses import dataclass
 from typing import Any, Protocol
 from urllib.parse import urlparse
@@ -113,19 +115,22 @@ class SecretProvider(Protocol):
 
 _registry: dict[str, SecretProvider] = {}
 
-# Process-lifetime cache, keyed by the full URI. A run commonly resolves the
-# same credential more than once (validation pass, connection test, the real
-# connection), and unlike an env var or secrets.toml lookup, a provider fetch
-# is a network call — so unlike those two steps, this one is worth not
-# repeating.
-#
-# Unbounded for the process's life: fine for a `drt run` invocation (exits in
-# seconds to minutes), but `drt serve` is a long-lived process that re-enters
-# this path on every triggered sync — a secret resolved once is held until
-# the server restarts, with no TTL and no re-fetch on rotation. Known gap,
-# not yet addressed — tracked as #929 — rather than assuming rotation is
-# picked up here.
-_value_cache: dict[str, str] = {}
+# Cache provider values by full URI for 300 seconds by default. A run commonly
+# resolves the same credential more than once (validation pass, connection
+# test, the real connection), and unlike an env var or secrets.toml lookup, a
+# provider fetch is a network call. The TTL lets long-lived `drt serve`
+# processes pick up secret rotation without repeating that call on every sync.
+# Set DRT_SECRET_CACHE_TTL_SECONDS to another duration, or to a non-positive
+# value to disable caching. It is read at lookup time so runtime environment
+# changes take effect without restarting the process.
+_value_cache: dict[str, tuple[str, float]] = {}
+
+_DEFAULT_CACHE_TTL_SECONDS = 300.0
+_CACHE_TTL_ENV_VAR = "DRT_SECRET_CACHE_TTL_SECONDS"
+
+
+def _cache_ttl_seconds() -> float:
+    return float(os.environ.get(_CACHE_TTL_ENV_VAR, _DEFAULT_CACHE_TTL_SECONDS))
 
 
 def register(scheme: str, provider: SecretProvider) -> None:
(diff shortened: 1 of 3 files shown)


masukai:
**Codex review:**

Target: branch diff against origin/feat/782-docs

Basic TTL behavior works, but cache refreshes are unsynchronized despite `drt serve` explicitly running different syncs concurrently. Expiration can therefore trigger duplicate provider calls and racing cache updates.

Review comment:

- [P2] Serialize refreshes for each cache key — drt/config/secret_providers/base.py:172-173
  When concurrent `drt serve` syncs share a secret URI and its entry expires, every thread can observe the expired tuple and call `provider.fetch` before any thread updates the cache. This repeats a rate-limited network request for every concurrent sync and may let a slower fetch overwrite a later result for another TTL; use per-URI locking or another single-flight mechanism around refreshes.

masukai:
Addressed in 7e78045: single global `threading.Lock` around the fetch-and-store path with double-checked locking (unlocked fast path for the fresh-cache case; re-check after acquiring the lock in case another thread already refreshed). Added a concurrency test (8 threads, shared expired URI, barrier-synchronized start) asserting `provider.fetch` is called exactly once — ran 5x locally with no flakiness.

masukai (commented):
Self-review (adversarial pass, requested explicitly — same standard I'm holding my own work to this cycle).

Traced the one thing that decides whether this actually fixes #929 rather than refreshing a cache nobody re-reads: does `drt serve` re-enter `resolve_provider_uri()` on every triggered sync, or does credential resolution happen once and get held by a longer-lived object?

Followed the chain: `_execute()` → `self._runner(sync_name, dry_run)` → `run_drt_sync()` → `_get_destination(sync)` → `drt.connectors.get_destination()`. No caching at any point in that chain — `get_destination` builds a fresh instance from the registry every call, and each destination's `_get_client()`/`fetch()` calls `resolve_env()` lazily on that fresh instance. So yes, a rotated `aws-sm://`/`gcp-sm://`/`vault://` secret is re-resolved through `resolve_provider_uri()` on the next triggered sync after the TTL lapses, not just at server startup. The fix lands where it needs to.

Also traced the double-checked-locking cache-fill path for a race: the unlocked fast-path read, the lock, the re-check inside the lock before fetching. `test_concurrent_cache_misses_fetch_once` exercises this with real threads and a barrier rather than mocked interleaving, and the logic holds up under a line-by-line trace — no window where two threads both miss and both hit the provider, no window where a stale tuple gets read after a concurrent write (dict assignment is atomic under the GIL, and there's no scenario needing more than that here).

No issues found. Approving.

PR #945, opened 2026-08-08
feat(state): config + factory for the state backend (#756, 1 of 4)

masukai:
Part 1 of 4 for #756 — see [ADR 0005](docs/adr/0005-state-location-and-write-grants.md)'s implementation ordering. This is step 1 (Protocols + factory): config + construction centralization only, **no remote backend yet**.

## What this does

- New `drt_project.yml` `state.backend` field (`Literal["local"]` for now — deliberately only one valid value; this PR is groundwork, not a new capability). Modeled on `WatermarkConfig`'s `Literal` + validator shape.
- New `drt/state/factory.py`: `build_state_bundle(project, project_dir) -> StateBundle` — replaces 24 direct `StateManager`/`HistoryManager`/`DlqStore` constructions across `drt/cli/commands/*.py`, `drt/mcp/tools/*.py`, `drt/docs/builder.py`, `drt/integrations/_runner.py`, `drt/cli/server.py`.
- The bundle is **cached per project directory**, not just constructed fresh each call — this is a correctness property, not a perf one. `drt/cli/server.py`'s existing comment already explains why a shared store matters for local file locking; the same reasoning is what keeps `drt run --threads N`'s locking meaningful once a call site's store becomes a shared network client in a follow-up PR.

## Why now, why this shape

`StateStore`/`HistoryStore`/`DlqBackend` Protocols already exist (#900). Nothing routed a backend choice to a concrete implementation — ADR 0005 flags this exact gap as the prerequisite step. Splitting it out as its own PR keeps the "no behavior change" review contract clean and separate from the actual remote-backend correctness work (concurrent-write safety, retry/backoff, emulator-based testing) landing in PR 2.

## Verification

- Full suite: `3173 passed, 4 skipped, 21 deselected` (run twice, identical — confirmed stable, not flaky)
- New `tests/unit/test_state_factory.py`: bundle identity is stable per `(project_dir, config)`, different dirs get different instances, invalid config (`bucket`/`prefix` set with `backend: local`) is rejected, an unreachable-today non-local `backend` value fails loudly rather than silently falling through
- VS Code extension bundled schema regenerated + CHANGELOG entry (schema-drift gate)

## Stack

1. **This PR** — config + factory, no behavior change
2. Object-store primitive + GCS backend (concurrent-write safety, fixing a gap found in review of #935)
3. S3 backend
4. Docs + migration guide

Full plan: see the design discussion on #756.


masukai:
**Codex review:**

- [P2] Preserve projectless DLQ retries — `drt/cli/commands/retry.py:171`
  When a directory contains `syncs/` and `.drt/dlq/` but no `drt_project.yml`, `drt retry` previously worked because sync and DLQ loading were standalone. This unconditional `load_project()` now raises before replay, despite `replay_dead_letters()` already providing a local-default fallback; the MCP retry wrapper similarly passes `ctx.load_project()`.

- [P2] Reject remote-only fields in the project schema — `integrations/vscode-drt/schemas/drt_project.schema.json:104`
  With the only permitted backend being `local`, the bundled schema accepts and offers `state.bucket`/`state.prefix`, while `StateConfig._check_backend_fields` rejects every non-null value at runtime.

---

**My triage:**

First finding — **confirmed real, fixed** (`1656244`). Verified against `main`: the CLI's `retry()` function did not call `load_project()` before this PR; `replay_dead_letters()`'s fallback logic exists but was being bypassed since the CLI already resolved (and passed) a non-`None` `project`. Fix: the CLI no longer eagerly loads the project at all — `replay_dead_letters()` already resolves it correctly on its own — and the MCP tool now uses `ctx.load_project_for_state()`, matching every sibling state/DLQ MCP tool (`dlq.py`, `get_history.py`, `get_status.py`, `state.py` all already use it; only `retry.py` didn't). Added a regression test (`test_retry_works_without_drt_project_yml`) that fails against the pre-fix code.

Second finding — **acknowledged, not fixing here.** Checked whether this is new: `WatermarkConfig` (already shipped, unrelated to this PR) has the *identical* shape in the same schema file — `storage: local|gcs|bigquery` plus always-visible `bucket`/`key`/`project`/`dataset`, with no JSON Schema encoding of which fields apply to which backend, even though it has its own runtime validator too. This is a pre-existing, accepted limitation of how `generate_project_schema()`/Pydantic's schema generation handles `@model_validator` cross-field constraints project-wide, not something this PR introduces or regresses. Fixing it well (JSON Schema `if`/`then`, or restructuring to a discriminated union) is real work that should apply to `WatermarkConfig` too, not just `StateConfig` — worth its own issue rather than a scope-expanding fix buried in this PR.

Full suite re-run independently after the fix: 3172 passed, 4 skipped, 21 deselected (2 unrelated failures in `test_cli_version.py` are a path-length artifact of the worktree location I ran review from, not a real regression — confirmed unrelated to anything this PR touches).

PR #946, opened 2026-08-08
feat(state): GCS backend for state/history/DLQ (#756, 2 of 4)

masukai:
Part 2 of 4 for #756, stacked on #945 (PR 1). This is the payoff PR — a real, concurrency-safe GCS backend for all three state Protocols.

## Design

- **One object per sync** (project-wide for `state.json`), generation-preconditioned read-modify-write. Not one-object-per-entry — `read()`/`depth()`/`all_depths()`/`replace()` are whole-collection Protocol methods; per-entry objects would turn `drt status`/`drt docs generate` into hundreds-to-thousands of LIST+GET calls.
- **Fixes a real gap found in review of #935** (the open community PR fixing the same lost-update bug for watermarks): reads are pinned via a *versioned blob reference* (`bucket.blob(key, generation=g)`), not `download_as_text(if_generation_match=g)`. A competing writer landing between the metadata check and the download raises `404` on the former, `412` on the latter — #935's `_read_for_update` only catches `NotFound`, so the 412 case escapes its retry loop uncaught. This implementation distinguishes "object never existed" (plain 404 on `reload()`) from "the generation I just saw got superseded" (404 on the *versioned* read → retry, don't treat as absent).
- Jittered exponential backoff, not a bare retry loop (`state.json` is shared by every `--threads N` worker, not just two independent watermarks — needs to survive real contention, not just a two-writer race).
- `StateContentionError` raised on retry exhaustion for state (silent loss is #919's exact bug class); history/DLQ stay best-effort, matching local's existing swallow-and-log posture.
- `history.max_entries` (default 500, new `HistoryConfig` field): bounds the remote history object's size in addition to `retention_days` — a documented behavior difference from local (unbounded within the retention window).
- `DlqObserver` now buffers failed batches and flushes once at `on_sync_completed` instead of writing per failed batch — verified byte-identical local DLQ output either way.
- `append`-then-`prune` optimization: the engine calls `history_manager.prune()` immediately after every `append()`; the backend retains the just-written generation/entries so prune can skip a redundant read when nothing needs removing.

## Testing

New `object_store_smoke` marker (real `fake-gcs-server` via `testcontainers`, every PR, no secrets — same posture #941 established for Postgres/MySQL). A **mandatory canary test runs first**: it deliberately uses a stale token and a superseded generation, and skips the entire dependent suite with a clear reason if the emulator doesn't actually enforce either precondition — an emulator that silently ignores preconditions would make a green concurrency test *worse* than no test.

Ran against a real Docker container in my own environment (Codex's sandbox has no Docker/network access, so I verified this myself rather than trusting its report):

```
tests/integration/object_store/test_gcs_backend.py::test_00_generation_precondition_canary PASSED
tests/integration/object_store/test_gcs_backend.py::test_state_store_conforms_for_read_write_overwrite_and_reset PASSED
tests/integration/object_store/test_gcs_backend.py::test_history_store_conforms_for_merge_limit_and_prune PASSED
tests/integration/object_store/test_gcs_backend.py::test_dlq_store_conforms_for_fifo_replace_clear_and_depths PASSED
tests/integration/object_store/test_gcs_backend.py::test_concurrent_state_writers_never_silently_lose_updates PASSED
tests/integration/object_store/test_gcs_backend.py::test_same_sync_race_finishes_with_one_acknowledged_value PASSED
6 passed in 0.82s
```

The canary genuinely passing (not skipped) confirms `fake-gcs-server` enforces both preconditions in this environment, so the concurrency results are trustworthy rather than false-green.

Also added `gcs` to `ci.yml`'s installed extras — it wasn't there (deliberately opt-in for the watermark-only `gcs` extra before this), so `object_store_smoke` would have silently skipped on every CI run without this.

- Full suite: `3195 passed, 4 skipped, 27 deselected`

## Not in this PR

- S3 backend (PR 3)
- Docs/migration guide (PR 4)
- Converging with #935's watermark storage onto this same primitive — tracked as a follow-up once #935 merges from its author; not gated on it


masukai:
**Codex review:**

- [P1] Flush buffered DLQ entries when a sync raises — `drt/engine/observer.py:344`
  When a later extraction, load, or finalize step raises after an earlier batch emitted row errors, `run_sync` never calls `on_sync_completed`; because this line only buffers, those earlier dead letters vanish, whereas previously each failed batch was persisted immediately.

- [P1] Preserve concurrent appends during DLQ replacement — `drt/state/_objectstore.py:329-330`
  When `drt retry` overlaps a run appending to the same sync's DLQ, this reads the generation after the append and then writes the retry's stale replacement list, silently deleting the new dead letters without a precondition failure.

- [P1] Bound the in-memory DLQ buffer — `drt/engine/observer.py:344`
  When a large sync produces more failures than `dlq.max_records`, this unconditionally retains every full `DeadLetter` until completion; the cap is applied only in `store.append`, so the default 10,000-record limit no longer bounds memory.

- [P2] Prune history by timestamp rather than append order — `drt/state/_objectstore.py:262-263`
  Runs of the same sync can complete out of `started_at` order; slicing the tail on append order can keep the older run and discard the newer one.

---

**My triage:**

All four confirmed real by tracing the actual code (not just taking the report at face value):

- **P2 (history prune order) — fixed** (`b975ff9`). `_pruned()`'s cap slice assumed append order == chronological order; sorted by `started_at` before slicing, matching `read()`'s own newest-first contract. Added a regression test with entries appended out of chronological order.

- **Both DLQ-loss P1s and the unbounded-buffer P1 — confirmed real, deliberately NOT patched inline here.** Diagnosis:
  - `DlqObserver.on_records_failed` buffers into `self._buffer[sync_name]` with no cap during accumulation (the `max_records` cap only applies inside `store.append()` at flush time) — confirmed by reading the method directly.
  - `on_sync_completed` — the only flush point — is called solely from `_run_sync_body`'s success path (`engine/sync.py:759`). Traced `run_sync`'s outer `except BaseException as exc: raised = exc; ...; raise` / `finally`: the `finally` block persists alerts and history *even when `raised is not None`* (it calls `history_manager.append()` directly, bypassing the observer), but there is no equivalent DLQ flush anywhere in that path, and `on_interrupted` (the stop_event path) is also a no-op for `DlqObserver` today. So a mid-sync exception or SIGTERM loses every buffered-but-unflushed dead letter that would previously have already been on disk.
  - The retry-replace race (`drt/cli/commands/retry.py:66,136`): `replay_dead_letters()` reads the DLQ once at the start, computes `untouched + remaining` from that snapshot, and calls `store.replace()` at the end with no re-read. The object-store primitive's own precondition (`read_for_update` + `write_if(token)`) is correct in isolation — it only guards "did the object change between *my* read and *my* write inside `replace()`" — but the *content* being written is stale by construction, so a concurrent run's append lands, gets a valid fresh token, and is silently overwritten by the retry's older view. No exception anywhere.

  I'm treating these three as one connected design problem (buffer lifecycle + DLQ replace semantics under concurrency) rather than three independent one-line patches. A safe fix needs: (a) a guaranteed flush path for any sync termination (success, exception, or interruption) — likely a new narrowly-scoped `SyncObserver` protocol hook rather than repurposing `on_sync_completed` (which other observers, e.g. state persistence, may treat as "the sync succeeded" — calling it from the exception path risks a worse bug than the one being fixed), (b) a bound applied during accumulation, not just at flush, and (c) `replace()` doing a real reconcile against current state rather than a blind overwrite. That's real design work, not something to rush through while triaging 8 other PRs' reviews in parallel — I'd rather hand this to a dedicated, focused pass than patch it hastily here. Tracking it as a follow-up; happy to take a first crack at a design if you want it before merge, or file it as a fast-follow issue if you'd rather ship the rest of the stack first — your call.

Full suite re-run independently after the history-prune fix: 3208 passed, 4 skipped, 27 deselected (2 unrelated `test_cli_version.py` failures are a path-length artifact of the worktree location, confirmed unrelated).

masukai:
Follow-up on the Codex review pass: the one finding I deliberately deferred there — DLQ buffering had no guaranteed flush on non-success exit paths, and the in-memory buffer was unbounded during accumulation — is now fixed here (`d176c01`).

**What was wrong:** `DlqObserver` buffers dead letters per-batch and only flushed them in `on_sync_completed`, which only fires on `run_sync`'s normal-return path. An unhandled exception or a graceful-shutdown interruption skipped that call entirely, silently dropping already-buffered dead letters. Separately, the buffer had no cap during accumulation — only `store.append`'s own `max_records` truncation applied, and only at flush time.

**Fix:**
- New `SyncObserver.on_sync_ended(sync_name)`, fired from `run_sync`'s outer `finally` block (gated by `dry_run`, in its own try/except) — the same place `history_manager.append()` and alert dispatch already run unconditionally on every exit path. `DlqObserver`'s flush moved from `on_sync_completed` to `on_sync_ended`; every other observer no-ops.
- `on_records_failed` now trims the buffer to `max_records` as entries arrive (guarded for `max_records=0` = unbounded, matching `store.append`'s documented contract), not just at flush.

Also scoped down what I'm **not** fixing here: the third finding from that review — `drt retry`'s `replace()` computing its write from a stale early snapshot, which can silently drop a concurrent `drt run`'s DLQ append — turns out to predate this PR. `LocalDlqStore.replace()` is the same unguarded whole-object overwrite, so the race exists on the local backend today, independent of the object-store work here. Filing that as its own issue rather than folding a Protocol-level reconciliation change into this branch.

Tests: 5 new/updated in `tests/unit/test_engine_observer.py` covering the flush point (success, exception, dry-run-skips) and the buffer trim (including the `max_records=0` edge case). Full suite green, ruff/mypy clean.

PR #947, opened 2026-08-08
feat(state): S3 backend for state/history/DLQ (#756, 3 of 4)

masukai:
Part 3 of 4 for #756, stacked on #946 (PR 2). Same object-store primitive, now against S3.

## Design

- **ETag preconditions instead of GCS generations.** S3 has no equivalent of a generation-addressed read (GCS's `bucket.blob(key, generation=g)` trick), so a stale snapshot surfaces as `412` on the pinned `GetObject` (or `404` on a concurrent delete) rather than the versioned-read `404` GCS gets almost for free — the read path retries explicitly. Documented as an asymmetry from the GCS adapter, not an inconsistency.
- `PutObject` uses `IfNoneMatch: "*"` for create-only, `IfMatch: <etag>` for update-if-unchanged. An S3-compatible endpoint that can't honor a conditional header (`501`) fails loudly rather than silently falling back to an unconditional write.
- Credential resolution matches `S3Destination`'s existing shape exactly (`aws_access_key_id_env`/`aws_secret_access_key_env`/`aws_session_token_env`, `region`, `endpoint_url`) — state storage doesn't get a second AWS auth vocabulary.
- `boto3` floor bumped to `>=1.35.69` (verified against botocore's own changelog — the version that added conditional-write support to `PutObject`).
- DLQ `clear()`/`replace([])` stays a conditional PUT of an empty object, not a conditional DELETE — `DeleteObject`'s `IfMatch` support landed later than `PutObject`'s, no reason to raise the floor further for it.

## CI

Learned from #946: `s3`/boto3 stays out of the main "Install dependencies" step and installs as a separate late step, same pattern as `gcs` — keeps `tests/unit/test_s3_destination.py`'s own missing-extras test meaningful. (Verified this one actually didn't need the same treatment `gcs` did — `boto3` isn't a namespace package the way `google.cloud` is, so `test_missing_boto3_raises_helpful_import_error` passes fine even with boto3 genuinely installed — but kept the separation anyway for consistency with the GCS leg and because it's harmless.)

## Testing

`object_store_smoke` now covers both backends. New three-part S3 canary (mirroring GCS's, adapted for ETags): create-only `IfNoneMatch`, stale-write `IfMatch`, stale-read `IfMatch` — all independently verified enforced by LocalStack before any concurrency test runs.

Ran everything myself against real containers (Codex's sandbox has no Docker/network access):

```
tests/integration/object_store/test_s3_backend.py::test_00_s3_precondition_canary PASSED
tests/integration/object_store/test_s3_backend.py::test_state_store_conforms_for_read_write_overwrite_and_reset PASSED
tests/integration/object_store/test_s3_backend.py::test_history_store_conforms_for_merge_limit_and_prune PASSED
tests/integration/object_store/test_s3_backend.py::test_dlq_store_conforms_for_fifo_replace_clear_and_depths PASSED
tests/integration/object_store/test_s3_backend.py::test_concurrent_state_writers_never_silently_lose_updates PASSED
tests/integration/object_store/test_s3_backend.py::test_same_sync_race_finishes_with_one_acknowledged_value PASSED
6 passed in 24.24s
```

GCS's 6 tests still pass alongside these (12 total, no interference).

Per-module coverage on `drt/state/s3.py`: confirmed genuinely **100%** (96/96 statements) myself, not just Codex's self-report — this session already found codecov/patch misleading on this same file family in #946 (real gaps in the lazy-client-construction and retry-exhaustion paths that patch coverage didn't surface), so I re-verified rather than trusting a claim this time too.

One thing I caught and fixed that the implementation missed: the bundled VS Code extension JSON schema wasn't regenerated for the new S3 config fields — would have failed the `schema-drift` CI gate. Regenerated and added the CHANGELOG entries.

- Full suite in the CI-matching state (neither gcs nor s3 installed): `3237 passed, 16 skipped, 21 deselected`

## Not in this PR

- Docs/migration guide (PR 4, final)


masukai:
**Codex review:**

- [P1] Distinguish a missing bucket from a missing object — `drt/state/s3.py:132-136`
  When the configured bucket is misspelled, deleted, or absent on an S3-compatible endpoint, `HeadObject` can return `NoSuchBucket` with HTTP 404, which this blanket status check converts into an empty state. Incremental sync then sees no prior cursor and may replay the full dataset; the later persistence failure is only logged by the observer, allowing this to repeat.

---

**Confirmed real, fixed** (`f5a4f2d`).

Traced it: `read_for_update()`'s error handling matched `status == 404 OR code in {...}` — since S3 returns the identical HTTP 404 for both `NoSuchKey` (legitimate — first write, nothing there yet) and `NoSuchBucket` (real misconfiguration), the blanket `status == 404` clause caught both, regardless of what the more specific `code` check would have said. `NoSuchBucket` silently became `(None, None)` — the exact same "empty state" signal a legitimate first run produces. Downstream, incremental sync would see no prior watermark and replay the whole dataset on every single run, since the subsequent write attempt (also against the missing bucket) only logs a warning via the fire-and-forget persistence contract rather than surfacing anywhere `drt run`'s exit code reflects.

Fixed by checking for `NoSuchBucket` explicitly and raising before the permissive 404 match, in both the `HeadObject` path and the analogous `IfMatch`-pinned `GetObject` retry path Codex flagged as needing the same treatment. `write_if`/`list_keys` were already safe — checked both, neither has a catch-all 404/status match that would swallow `NoSuchBucket`.

Added `test_missing_bucket_is_not_treated_as_missing_object` (parametrized over both the HEAD and GET code paths) asserting `NoSuchBucket` propagates instead of resolving to `(None, None)`.

Full suite re-run independently: 3237 passed, 4 skipped, 33 deselected (2 unrelated `test_cli_version.py` failures, same path-length artifact noted on the other PRs in this stack).

Issue #948, opened 2026-08-08
state.json: revisit single-object layout if remote-backend --threads contention becomes real


masukai:
## Context

#756's remote state backend (PRs #945/#946/#947) ships `state.json` as a
**single project-wide object** per backend (GCS/S3), mirroring the local
`.drt/state.json` layout byte-for-byte. This is a deliberate choice, made
after finding that an earlier planning document
(`docs/plans/2026-07-22-756-remote-state-backend.md`, Task A0, "Settle the
object layout") had specified a different design:

> `state.json` as a single object is the one place all syncs collide. Prefer
> `<prefix>/state/<sync_name>.json` for the same reason [as DLQ/history], and
> reconstruct `get_all()` by listing the prefix.

That per-sync-split design avoids `drt run --threads N` funneling every
sync's watermark write through one conditioned object. The shipped design
accepts that contention in exchange for:

- Exact byte-for-byte parity with the local layout, which is what makes
  `state.backend: gcs` ↔ `local` a genuine one-line, zero-migration-step
  config change (ADR 0005 Decision 4 — tier reversibility).
- One shared concurrency-safe primitive (`ObjectStoreStateStore` in
  `drt/state/_objectstore.py`) instead of a second key-layout variant to
  maintain.
- No new failure mode beyond what already exists for `GCSWatermarkStorage`
  today — retries with jittered backoff on precondition conflict
  (`_objectstore.py`'s `MAX_WRITE_ATTEMPTS = 8`), raising
  `StateContentionError` only after exhausting them.

This is a safety/efficiency trade-off, not a correctness gap: the shipped
`state.json` writes are generation/ETag-preconditioned read-modify-write with
retry, so concurrent writers never silently clobber each other the way
`GCSWatermarkStorage` did before #919/#935. Under contention, writers retry
and (rarely) fail loudly instead of losing an update.

## When to revisit

If real `--threads N` usage against a remote `state.backend` (gcs/s3)
produces reported `StateContentionError`s, or a benchmark shows
retry/backoff time becoming significant at realistic sync counts and
cadences (ADR 0005's falsification condition context: ~50 syncs / 15-minute
cadence), switch `state.json` to the per-sync-object layout
(`<prefix>/state/<sync_name>.json`, reconstructing `get_all()` by listing
the prefix) — this is a backend-internal key-layout change, invisible to
config and to all existing call sites.

## Related

- ADR 0005: `docs/adr/0005-state-location-and-write-grants.md`
[...]

PR #949, opened 2026-08-08
docs(state): remote state guide + migration (#756, 4 of 4)

masukai:
## Summary

Final PR in the #756 remote state backend stack (#945 → #946 → #947 → this).
Documentation only — no changes under `drt/` or `tests/`.

- New [`docs/guides/remote-state.md`](docs/guides/remote-state.md): GCS/S3
  setup, scoped IAM policies, the shipped object layout (single project-wide
  `state.json`; one history/DLQ object per sync — rationale in #948),
  generation/ETag concurrency and `StateContentionError` behavior,
  bidirectional migration from `.drt/` via `gsutil rsync` / `aws s3 sync`,
  and explicit independence from `sync.watermark.storage`.
- `docs/guides/sync-history.md`, `docs/guides/dead-letter-queue.md`: note
  that the documented `.drt/...` paths describe the local backend, link to
  the new guide for the remote equivalent. `history.max_entries` added to
  the retention example.
- `docs/guides/ci-cd-integration.md`: new "Persisting state across
  ephemeral runs" section — a fresh CI checkout has no prior `.drt/`, so
  `drt status`/`drt retry` are no-ops without a remote `state.backend`.
- `docs/llm/API_REFERENCE.md`: `state:` block and `history.max_entries`
  added to the `drt_project.yml` reference example.
- CHANGELOG entry for the doc pass.

Every backend-behavior claim (auth mechanism, IAM permissions, retry
attempt counts, prefix normalization, exhaustion semantics) was checked
against `drt/config/base.py`, `drt/state/_objectstore.py`,
`drt/state/gcs.py`, and `drt/state/s3.py` directly, not assumed.

## Test plan

- [x] `ruff check drt tests` — pass
- [x] `mypy drt` — pass, 189 files
- [x] `git diff --stat` confirms no changes under `drt/` or `tests/`
- [x] Doc claims spot-checked against source (retry attempt counts, IAM
      action mapping to actual client calls, prefix `.strip("/")`
      normalization, ADR/connector doc anchors resolve)


masukai:
**Codex review:**

- [P1] Exclude local secrets from remote-state migration — `docs/guides/remote-state.md:198`
  The recursive `.drt/` sync command uploads `.drt/secrets.toml` along with `watermarks.json`, test-failure samples, and any other `.drt` contents; following this migration guide can expose credentials to the state bucket.

- [P1] Do not promise remote state for native runners — `docs/guides/remote-state.md:12-14`
  `DagsterDrtResource.run()` constructs `StateManager(project_path)` directly, and the Airflow runner's `run_drt_sync()` invokes `run_sync()` without a persistence/DLQ observer or history manager. Neither writes remote run state, history, or DLQ as promised.

- [P2] Align the contention guarantee with CLI behavior — `docs/guides/remote-state.md:176-178`
  `StatePersistingObserver.on_sync_completed()` catches every exception from `save_sync()` and only logs a warning; the command can report success with stale `state.json`.

- [P2] List the required remote-state configuration fields — `docs/llm/API_REFERENCE.md:18-19`
  Only `backend` is listed; `bucket` is required and `prefix`/region/credential fields are valid `StateConfig` fields an LLM generating config from this page would miss.

---

**All four confirmed real, all four fixed** (`c11488c`).

- Migration commands scoped to the three actually-relevant paths (`state.json`, `history/`, `dlq/`) instead of the whole `.drt/` directory, with an explicit callout naming `secrets.toml`/`schemas/`/`test_failures/` as the things that must NOT be swept along.
- Verified the Dagster/Airflow claim directly: `integrations/dagster-drt/dagster_drt/resource.py:90` does construct `StateManager(project_path)` (the local-only class), and `drt/integrations/_runner.py:72` does call `run_sync(..., state_mgr)` with no `observer=` argument — confirmed `run_sync` defaults to `NullObserver()` when none is passed, so persistence is genuinely a no-op there today, independent of `state.backend`. Scoped the guide's promise to CLI invocations (`run`/`build`/`test`/`validate`/`serve` — `serve` confirmed via this session's own #756 planning notes: its state store already goes through the same factory) and named the integration gap plus a workaround (shell out to the CLI from Dagster/Airflow if you need this today).
- Verified `StatePersistingObserver.on_sync_completed()` directly: `save_sync()` is wrapped in `try/except Exception: self._logger.warning(...)`, matching every other observer's fire-and-forget contract, plus `CompositeObserver._broadcast` catches again on top of that. So yes — `StateContentionError` never reaches a `drt run` exit code; it's what stops an exhausted retry from silently overwriting with stale data, not a user-visible failure signal. Rewrote that section to say so plainly, including what to actually watch (the `State persist failure` warning log) if relying on this under real write contention.
- Added the full `state:` field surface (bucket/prefix/region/credential fields/endpoint_url, which apply to which backend) to the `API_REFERENCE.md` example.

Full suite re-run independently (docs-only, sanity check): 3237 passed, 4 skipped, 33 deselected.

PR #950, opened 2026-08-08
feat(docs): manifest v3 — per-sync config_hash (#772, 1 of 4)

masukai:
## Summary

Task 2 of #772's `state:modified` selector, per the plan at
`docs/plans/2026-08-03-772-state-modified-selector.md`. Task 1 (the
fingerprint primitive, `drt/config/fingerprint.py`) shipped as #901.

- `drt/docs/manifest.py`: `SCHEMA_VERSION` 2 → 3, `Sync.config_hash: str | None`.
  Pure additive superset (ADR 0006) — nothing renamed or removed.
- `drt/docs/builder.py`: computes the project's fingerprint map once and
  keys each `Sync` by the same resolved name the manifest already uses.
  `config_hash` is computed unconditionally, not gated behind
  `include_state` — it's a repo fact, not run state, so a `--no-state`
  baseline (the shape CI will actually save) still carries it.
- `Manifest.from_dict()` accepts an older v1/v2 dict with no `config_hash`
  key, yields `config_hash=None`, and preserves the original
  `schema_version` — Task 3 needs that to refuse a too-old baseline
  cleanly instead of silently comparing absent hashes as equal.
- New `docs/adr/0006-manifest-schema-v3.md`.

## Judgment call

The plan didn't have a test pinning "`config_hash` survives
`--no-state`" specifically, even though the ADR states it as a design
decision — the implementation was already correct (the field is computed
outside the `include_state` branch) but nothing would have caught a
regression there. Added `test_no_state_still_includes_config_hash`.

## Test plan

- [x] `pytest -q -m "not dwh_smoke and not local_sql_smoke and not object_store_smoke"` — 3182 passed, 4 skipped, 21 deselected (re-run independently, no regressions)
- [x] `drt/docs/manifest.py` 100%, `drt/docs/builder.py` 99% (2 missing lines pre-date this change — confirmed via `git blame`)
- [x] Real `drt docs generate --format json` output eyeballed, `config_hash` present with real hex values


masukai:
**Self-review.**

Traced `Sync.config_hash` end to end: `builder.py` computes `sync_fingerprints(project_dir)` once (not per-sync) and looks it up by the same resolved name the manifest already keys syncs by — confirmed by reading both files side by side rather than assuming the keys line up. `config_hash` is computed unconditionally, not gated behind `include_state`, matching the ADR's "repo fact, not run state" claim exactly (verified in the diff, not just the docstring).

One real gap found before push: the plan/ADR asserted `config_hash` survives `--no-state`, but nothing pinned it — a future refactor could silently break that invariant without any test failing. Added `test_no_state_still_includes_config_hash`.

`from_dict()` on an old v1/v2 dict (no `config_hash` key) doesn't raise and defaults to `None`, with `schema_version` faithfully preserved — confirmed this is what Task 3 (#951) actually needs, not just a hopeful docstring claim.

Coverage: `manifest.py` 100%, `builder.py` 99% (2 missing lines pre-date this PR, confirmed via `git blame`). Full suite re-run independently: 3182 passed, 4 skipped, 21 deselected, no regressions.

PR #951, opened 2026-08-08
feat(cli): state:modified / state:new selector methods (#772, 2 of 4)

masukai:
## Summary

Task 3 of #772's `state:modified` selector, stacked on #950 (Task 2,
manifest v3 / `config_hash`).

- New `drt/cli/_state_selection.py`: `StateDiff(new, modified)` +
  `load_state_diff(baseline_path, current_syncs, project_dir)` — diffs the
  current tree's fingerprints against a baseline manifest's `config_hash`.
- `drt/cli/_selection.py`: `matches()`/`select_syncs()` gain an optional
  `state_diff` keyword param (default `None`), so every existing caller
  (`run`/`build`/`test`/`validate`, all via `select_syncs`) is unaffected
  until Task 4 wires the actual `--state` flag on top of this.
- `state:modified`/`state:new` added to `_METHODS`, `_no_match_message`,
  and `complete_selector`.

**Semantics** (per `docs/plans/2026-08-03-772-state-modified-selector.md`
Task 3):
- Missing/unreadable/unparseable baseline → warning, everything is new
  (the normal first-CI-run case, not an error).
- Baseline older than schema v3 (predates `config_hash`) → raises clearly.
- A sync absent from the baseline → `new` (and therefore `modified` —
  dbt's own precedent: a new sync is trivially "modified").
- A missing hash on either side (baseline `None`, or current fingerprint
  lookup miss) never reads as "unchanged" — only ever "changed". A false
  "changed" costs one extra dry-run; a false "unchanged" silently skips
  real work in CI.
- A sync present only in the baseline (deleted) is ignored — nothing to
  select or run.

## Judgment call

Codex's first pass compared the baseline's `schema_version` against the
**live** `drt.docs.manifest.SCHEMA_VERSION` constant. That's a latent bug:
the day schema bumps to v4 for anything unrelated to `config_hash`, every
still-perfectly-comparable v3 baseline would start being rejected, since
`3 < 4`. Fixed by decoupling the check into a fixed
`_MIN_SCHEMA_VERSION_WITH_CONFIG_HASH = 3` constant in
`_state_selection.py`, with a comment explaining why it must not track the
live constant, plus a regression test
(`test_baseline_at_or_above_config_hash_schema_version_is_accepted`) using
a schema version *higher* than the current live one so the test doesn't
pass by coincidence once `SCHEMA_VERSION` moves.

No CHANGELOG entry yet — `state:modified`/`state:new` aren't reachable
from the CLI until Task 4 wires `--state`, so the entry lands with that
PR (or Task 5) describing the whole usable feature at once rather than
fragmenting it across not-yet-usable plumbing PRs.

## Test plan

- [x] `pytest -q -m "not dwh_smoke and not local_sql_smoke and not object_store_smoke"` — 3203 passed, 4 skipped, 21 deselected (re-run independently)
- [x] `drt/cli/_selection.py` 100%, `drt/cli/_state_selection.py` 100%
- [x] Confirmed every existing `matches(`/`select_syncs(` call site
      (`run.py`, `build.py`, `test.py`, `validate.py`) compiles and behaves
      unchanged with the new optional parameter defaulted

### DIFF
--- a/drt/cli/_selection.py
+++ b/drt/cli/_selection.py
@@ -5,12 +5,16 @@ Selector grammar (per token):
 - ``*`` / ``all``            — every sync (explicit sentinel, back-compat)
 - ``tag:<pattern>``          — syncs with a tag matching the pattern
 - ``destination:<pattern>``  — syncs whose destination ``type`` matches
+- ``state:modified`` / ``state:new`` — syncs changed or added since a baseline
 - anything else              — the sync name (glob patterns supported)
 
 Patterns use ``fnmatch`` semantics (``*``, ``?``, ``[seq]``), so exact names
 keep working unchanged. Repeated ``--select`` values union; ``--exclude``
 subtracts with the same grammar. Definition order is preserved and results
 are deduplicated.
+
+State selectors require the caller to supply a pre-computed baseline diff;
+using one without a baseline is an error.
 
 ``source:`` is deliberately **not** a method: syncs share the project
 profile (one source per run), so there is nothing per-sync to select on.
@@ -25,9 +29,11 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from drt.cli._state_selection import StateDiff
     from drt.config.models import SyncConfig
 
-_METHODS = ("tag", "destination")
+_METHODS = ("tag", "destination", "state")
+_STATE_SELECTORS = ("state:modified", "state:new")
 
 
 class SelectionError(ValueError):
@@ -39,7 +45,7 @@ def is_glob(token: str) -> bool:
     return any(ch in token for ch in "*?[")
 
 
-def matches(sync: SyncConfig, token: str) -> bool:
+def matches(sync: SyncConfig, token: str, *, state_diff: StateDiff | None = None) -> bool:
     """Does one selector token match one sync?"""
     if token in ("*", "all"):
         return True
@@ -49,6 +55,19 @@ def matches(sync: SyncConfig, token: str) -> bool:
     if token.startswith("destination:"):
         pattern = token[len("destination:") :]
         return fnmatchcase(sync.destination.type, pattern)
+    if token.startswith("state:"):
+        if token not in _STATE_SELECTORS:
+            raise SelectionError(
+                f"Unknown state selector '{token}'. Available state selectors: "
+                + ", ".join(_STATE_SELECTORS)
+                + "."
+            )
+        if state_diff is None:
+            raise SelectionError(
+                f"Selector '{token}' requires a baseline manifest to compare against."
+            )
+        names = state_diff.new if token == "state:new" else state_diff.modified
+        return sync.name in names
     if ":" in token:
         method = token.split(":", 1)[0]
         raise SelectionError(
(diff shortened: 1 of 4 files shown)


masukai:
**Self-review.**

Traced `load_state_diff()`'s four behavior branches (missing/unreadable baseline, too-old schema, per-sync new/modified, deleted-baseline-only sync) against the plan's explicit semantics list one by one, not just against the tests as written.

**Real bug found and fixed before push:** the schema-version compatibility check compared the baseline against the *live* `drt.docs.manifest.SCHEMA_VERSION` constant (`manifest.schema_version < SCHEMA_VERSION`). That's correct today only because `SCHEMA_VERSION == 3` coincides with "the version `config_hash` was introduced at." The day schema bumps to v4 for anything unrelated to `config_hash`, every still-perfectly-comparable v3 baseline would start being rejected. Decoupled into a fixed `_MIN_SCHEMA_VERSION_WITH_CONFIG_HASH = 3` local constant, with a comment explaining why it must not track the live constant, plus a regression test using a schema version *above* the current live one so the test can't pass by coincidence later.

Confirmed the "never treat absent as equal" rule is applied uniformly: baseline `config_hash=None`, current-fingerprint lookup miss, and baseline-absent-entirely all route to "modified," never "unchanged" — checked each branch in `_state_selection.py` individually against its own test.

Confirmed every existing caller of `matches()`/`select_syncs()` (`run.py`, `build.py`, `test.py`, `validate.py`, plus their own test files) still compiles and behaves identically, since the new `state_diff` parameter is keyword-only and defaults to `None`.

Coverage: `_selection.py` 100%, `_state_selection.py` 100%. Full suite re-run independently: 3203 passed, 4 skipped, 21 deselected.

masukai:
**Codex review:**

- [P2] Handle invalid schema versions before comparing them — `drt/cli/_state_selection.py:54`
  When a baseline is valid JSON but contains a non-numeric `schema_version` such as `"3"` or `null`, `Manifest.from_dict()` accepts it and this comparison raises an uncaught `TypeError`. Callers consequently get a traceback instead of either the documented malformed-baseline fallback or a `SelectionError`.

---

**Confirmed real, fixed** (`89d272a`).

`Manifest.from_dict()` (from #950) is a plain dataclass loader — it does `schema_version=data["schema_version"]` with no type check, so a hand-edited or corrupted baseline whose `schema_version` is a string or `null` sails through and only breaks at the `<` comparison in `_state_selection.py`, as a raw `TypeError`, not the SelectionError/warn-and-treat-as-new fallback the rest of the function is built around. Added an explicit `isinstance(manifest.schema_version, int)` check that routes a malformed value into the same warn-and-treat-as-new path as a missing/unparseable baseline (this felt like the right bucket for it — the file did parse and did have *a* value, but not a usable one, closer in spirit to "can't tell what this baseline says" than to "definitely too old and needs regenerating").

Added `test_malformed_schema_version_warns_and_treats_everything_as_new`, parametrized over a string and `null` schema_version.

Full suite re-run independently: 3204 passed, 4 skipped, 21 deselected (2 unrelated `test_cli_version.py` failures, same path-length worktree artifact noted on the sibling PRs).

PR #952, opened 2026-08-08
feat(cli): --state baseline manifest flag (#772, 3 of 4)

masukai:
## Summary

Task 4 of #772's `state:modified` selector, stacked on #951 (Task 3). This
is the PR that makes the feature reachable from the CLI for the first time.

```
drt run --select state:modified --state ci-baseline/manifest.json --dry-run --diff
```

- `--state PATH` added to `run`/`build`/`test`/`validate`, matching the
  existing `--select`/`--exclude` option style. Loaded via `load_state_diff()`
  only when supplied (no eager baseline load when `--state` is absent).
- `state:...` used without `--state` now names the flag directly in the
  error (`Selector 'state:modified' requires --state
  <path-to-baseline-manifest.json>.`) instead of Task 3's placeholder
  wording, which was deliberately generic since the flag didn't exist yet.
- `--state` classified alongside `--select`/`--exclude` in the CLI↔MCP
  drift audit (`scripts/check_cli_mcp_parity.py`) — MCP tools select by
  `sync_name` rather than this grammar, same reasoning as the two existing
  entries.
- CHANGELOG: consolidated Task 2's standalone `config_hash` entry into one
  entry covering the whole capability, since it's only usable end to end
  as of this PR.

## Judgment call

Chose to fold Task 2's existing CHANGELOG entry into this PR's diff
(replacing it) rather than adding a third, separate entry — three PRs each
adding a fragment of "state-aware selection" read confusingly next to each
other in release notes, especially since Tasks 2/3 shipped no way to
actually reach the feature from the CLI. This only works cleanly because
the four PRs are a strict stack merged in order; flagging it in case you'd
rather each PR carry its own entry.

## Test plan

- [x] `pytest -q -m "not dwh_smoke and not local_sql_smoke and not object_store_smoke"` — 3216 passed, 4 skipped, 21 deselected (re-run independently)
- [x] New end-to-end `CliRunner` tests per command: changed-sync selection,
      missing baseline (warns + runs everything), `state:` without
      `--state` (clean exit, no traceback), old-schema baseline (clean
      exit, no traceback) — all run and passing independently
- [x] Help-text tests updated for the new flag/examples across all four
      commands
- [x] Reviewed every diff line by line (all 4 command files,
      `_selection.py`, the drift-audit script, CHANGELOG)


masukai:
**Self-review.**

Checked all four commands (`run`, `build`, `test`, `validate`) individually rather than assuming `build`/`test`/`validate` mirror `run.py` exactly — `validate.py`'s call site uses `result.syncs` (from `load_syncs_safe`), not a bare `syncs` variable, and has its own `SelectionError` → error-key-reporting path distinct from the other three's `print_error` + exit(1). Confirmed `--state` is threaded through each command's *existing* error handling rather than adding a second, parallel error path.

Confirmed `--state` is loaded lazily (`if state is not None: load_state_diff(...)`) — no baseline read/fingerprint walk when the flag isn't passed, so every non-`--state` invocation pays zero extra cost.

Confirmed the `state:` "requires a baseline" error message (deliberately generic in #951, since the flag didn't exist yet) was tightened here to name `--state` directly, and the corresponding test's match string was updated rather than left asserting stale text.

Confirmed `--state` was added to the CLI↔MCP drift audit's `SELECTION_OPTIONS` exclusion set (`scripts/check_cli_mcp_parity.py`) — without this, `--state` would have shown up as a "CLI option with no matching MCP parameter" false-positive gap, since MCP tools select by `sync_name` rather than this grammar. Verified this by reading the script's actual gap logic, not just trusting the passing test.

`logger.warning(...)` in `_state_selection.py` reaching a real terminal (not just `caplog` in tests) relies on Python's own `logging.lastResort` handler (stderr, WARNING+, active whenever no handler is configured anywhere in the hierarchy) — this is guaranteed standard-library behavior, not something this PR needs to configure itself. Confirmed no other CLI entry point in this codebase installs a handler that would suppress it.

Coverage: `_selection.py`/`_state_selection.py` remain 100%; the four command files' *new* branches (the `if state is not None` split and its error handling) are covered by the new end-to-end `CliRunner` tests per command — file-wide percentages are lower only because of large pre-existing, unrelated code in those files. Full suite re-run independently: 3216 passed, 4 skipped, 21 deselected.

masukai:
**Codex review:**

- [P1] Treat empty state diffs as successful no-ops — `drt/cli/commands/run.py:645`
  When the baseline matches the current tree, `state:modified` or `state:new` has zero hits, so `select_syncs` raises and this command exits 1. This makes the advertised CI invocation fail on every PR that changes no sync definitions instead of completing as a successful no-op; the same call pattern affects `build`, `test`, and `validate`.

- [P1] Preserve changed invalid syncs during state validation — `drt/cli/commands/validate.py:113-116`
  When a changed YAML file parses but fails `SyncConfig` validation, `load_syncs_safe` places it only in `result.errors`; passing only `result.syncs` here omits it from `StateDiff`, while `_match_error_keys` cannot match `state:` tokens. If another valid changed sync is selected, the invalid entry is filtered out and `validate` can exit 0, so state-aware CI silently misses a changed broken configuration.

---

**Both confirmed real, both fixed** (`f5e6011`). Both genuinely undermined the point of the feature — flagging that up front since these aren't edge cases, they're the mainline CI-recipe path.

**Empty-diff no-op**: verified `select_syncs()`'s existing "every select token must match at least one sync" contract — correct for `tag:`/`destination:`/bare-name (a typo should never silently run nothing) — was being applied identically to `state:` tokens, where "matched nothing" is the *expected, common, healthy* outcome for a PR that touched no sync files, not a typo signal (a real typo like `state:modifed` is already caught earlier, by the "unknown state selector" check, regardless of match count). Exempted `state:` tokens from that check in `select_syncs()`, and added `is_state_only_select()` so each of the four commands can tell "nothing changed" (clean exit 0) apart from a genuinely-empty selection that should still error (e.g. `--select tag:x --exclude tag:x`). Wired into `run`/`build`/`test`/`validate`'s existing empty-selection checks.

**Broken sync silently dropped**: verified the actual identifier mismatch — `load_syncs_safe()`'s `result.errors` is keyed by **file stem** (`drt/config/parser.py`: `result.errors[path.stem] = ...`), while `sync_fingerprints()` (what the state diff is built from) keys by the **resolved `name:` field** — two different identifier spaces that can diverge whenever a filename doesn't match its sync's declared name, so there's no reliably-safe way to map a broken sync into `StateDiff` membership by name. Rather than risk that reconciliation being wrong in a way that silently drops a real broken-config report, the fix keeps **every** parse/validation error whenever a `state:` selector is active, independent of the state diff — a false positive (an unrelated broken sync also reported) is safe, a false negative (a changed broken sync going unreported) is exactly the failure mode a CI gate must not have.

Also fixed one thing my own earlier fix broke: the existing `test_no_match_state_message` test asserted the *old* (buggy) raise behavior for a zero-hit state: token — updated it to assert the new no-raise behavior, and removed the now-dead `state:modified`/`state:new` branches in `_no_match_message()` that only that raise path used.

New tests: a no-op test per command (`run`/`build`/`test`/`validate`), a validate-specific test with one valid + one broken sync both changed since baseline (confirms the broken one still surfaces), a mixed-selector test (`state:modified` + `tag:crm`, confirms a zero-hit state: token doesn't swallow the other token's real hits), and direct `is_state_only_select()` coverage.

Full suite re-run independently: 3227 passed, 4 skipped, 21 deselected (2 unrelated `test_cli_version.py` failures, same path-length worktree artifact noted on the sibling PRs).

PR #953, opened 2026-08-08
docs(cli): state:modified selector guide + CI recipe (#772, 4 of 4)

masukai:
## Summary

Final PR in the #772 stack (#950 → #951 → #952 → this). Documentation
only — no changes under `drt/` or `tests/`.

- New [`docs/guides/state-modified-selector.md`](docs/guides/state-modified-selector.md):
  selector semantics table, precisely what gets hashed and why (secrets
  stay out *given* env-indirection convention; an env-var-only change is
  invisible, same blind spot dbt documents and accepts), missing/
  incompatible baseline behavior, and a full GitHub Actions recipe
  (save manifest on `main`, resolve the latest successful baseline run
  via `gh run list`, download it cross-run via `actions/download-artifact`
  with `run-id`/`github-token`, degrade gracefully to "everything is new"
  when no baseline exists yet).
- `docs/guides/ci-cd-integration.md`: the practical recipe + new CLI flag
  table rows (`--select state:modified`, `--select state:new`, `--state`),
  linking to the new guide for the full explanation.
- Periphery sweep: `docs/llm/API_REFERENCE.md` (YAML-config-only in scope,
  doesn't document CLI flags — no-op), `drt-init`/`drt-debug` skills
  (show targeted examples, not an exhaustive grammar — no-op), drt-action
  (separate repo, untouched). CHANGELOG already has one comprehensive
  entry from #952 — no redundant addition.

## Test plan

- [x] `ruff check drt tests` / `mypy drt` — clean (sanity check, no
      `drt`/`tests` files touched)
- [x] `make check-skills` — clean
- [x] Every behavioral claim in the guide checked against source
      (`drt/config/fingerprint.py`, `drt/cli/_selection.py`,
      `drt/cli/_state_selection.py`) rather than assumed
- [x] GitHub Actions cross-run artifact download pattern
      (`run-id`/`github-token` on `actions/download-artifact@v4`) reviewed
      for correctness — this is the actual required shape for downloading
      an artifact from a different workflow run, not same-run access

### DIFF
--- a/docs/guides/ci-cd-integration.md
+++ b/docs/guides/ci-cd-integration.md
@@ -83,6 +83,107 @@ jobs:
       - run: drt run --dry-run
 ```
 
+### State-aware PR previews
+
+To preview only the syncs a pull request changed, save a JSON manifest after
+each push to `main`, then restore the latest successful artifact in the PR job.
+The manifest carries the per-sync hashes used by `state:modified`.
+
+```yaml
+# .github/workflows/drt-state-baseline.yml
+name: drt state baseline
+on:
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  baseline:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - run: pip install drt-core
+      - run: drt docs generate --format json --output ci-baseline
+      - uses: actions/upload-artifact@v4
+        with:
+          name: drt-state-baseline
+          path: ci-baseline/manifest.json
+```
+
+```yaml
+# .github/workflows/drt-pr-preview.yml
+name: drt PR preview
+on:
+  pull_request:
+    paths:
+      - 'syncs/**'
+
+permissions:
+  actions: read
+  contents: read
+
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - run: pip install drt-core[bigquery]  # add your source extras
+
+      - name: Find latest main-branch baseline
+        id: baseline
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          run_id="$(
+            gh run list \
+              --repo "$GITHUB_REPOSITORY" \
+              --workflow drt-state-baseline.yml \
+              --branch main \
+              --event push \
+              --status success \
+              --limit 1 \
+              --json databaseId \
+              --jq '.[0].databaseId // empty' \
+              2>/dev/null || true
+          )"
+          echo "run_id=$run_id" >> "$GITHUB_OUTPUT"
+
+      - name: Download baseline manifest
+        if: steps.baseline.outputs.run_id != ''
+        continue-on-error: true  # an expired artifact uses drt's first-run fallback
+        uses: actions/download-artifact@v4
+        with:
+          name: drt-state-baseline
+          path: ci-baseline
+          github-token: ${{ github.token }}
+          run-id: ${{ steps.baseline.outputs.run_id }}
+
+      - name: Preview changed syncs
+        run: >-
+          drt run
+          --select state:modified
+          --state ci-baseline/manifest.json
+          --dry-run
+          --diff
+        env:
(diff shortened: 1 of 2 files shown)


masukai:
**Self-review.**

Verified every behavioral claim in the new guide against source rather than against the plan document (the plan predates the final implementation and drifted slightly, e.g. Task 1's original per-object-per-sync history/DLQ idea vs. this feature which only ever touches `state.json`'s manifest sibling `config_hash` — unrelated to that, just noting the plan isn't automatically ground truth): the new/modified rules, the missing-baseline fallback, the pre-v3 rejection, and the "null hash counts as modified" rule were all checked against `_state_selection.py` directly.

Confirmed `drt docs generate --format json --output ci-baseline` really does write to `ci-baseline/manifest.json` (`docs.py`: `manifest_path = output / "manifest.json"`) — the GitHub Actions recipe's `upload-artifact`/`download-artifact` paths and the guide's CLI examples are consistent with that, not just with each other.

Checked the cross-run artifact download pattern specifically, since it's easy to get subtly wrong: `actions/download-artifact@v4` only reaches into another workflow run when given both `run-id` and `github-token` explicitly — same-run access (the more commonly-copied example) wouldn't work here since the baseline is produced by a *different*, earlier workflow run on `main`. The `gh run list --workflow ... --branch main --event push --status success --limit 1` step to resolve that `run-id`, and the `continue-on-error`/`if: steps.baseline.outputs.run_id != ''` graceful-degradation path when no baseline run exists yet, were both reasoned through rather than copy-pasted from a generic example.

Periphery sweep re-confirmed by reading (not just grepping) `docs/llm/API_REFERENCE.md`'s scope (YAML config fields, not CLI flags) and both flagged skills (`drt-init`, `drt-debug` — targeted examples, not exhaustive grammar).

Also caught during a full-stack review just now: PR titles across this 4-PR stack had an off-by-one — #952 and this PR were both labeled "4 of 4". Fixed all four titles to 1-4 of 4 in sequence (title-only edits, no commit rewrite).

masukai:
**Codex review:**

- [P2] Supply variables when generating the baseline — `docs/guides/state-modified-selector.md:137-138`
  The baseline-generation step provides no values for `${API_TOKEN}`-style or environment-provided `var()` fields. `build_manifest()` calls `load_syncs_safe()`, where expansion rejects the file; generation still succeeds but silently omits that sync from `manifest.json`. A PR job with the variables present then classifies the sync as new on every run.

- [P2] Treat an empty modified selection as a no-op — `docs/guides/state-modified-selector.md:201-207`
  For a PR that only deletes an existing sync, or edits an unreferenced file, there are no current modified names. `select_syncs()` treats a state selector with zero hits as an error, so this command exits 1 rather than completing successfully — despite the guide stating baseline-only syncs are silently ignored.

- [P3] Exempt templated names from the environment claim — `docs/guides/state-modified-selector.md:80-85`
  If `name:` itself contains `${VAR}`/`var()`, changing that environment value changes the resolved `SyncConfig.name` and the fingerprint map key, so the sync is selected as new/modified even though its file bytes are unchanged.

---

**All three confirmed real.**

- P2 (missing baseline variables) — **fixed** (`a667ec8`). Verified directly: `drt/docs/builder.py:213` calls `load_syncs_safe(project_dir)` with no `vars` override, which does `expand_env_vars(data)` — the same strict, error-collecting loader `drt validate` uses. A substitution failure lands the whole sync in `result.errors`, not `result.syncs`, so it's absent from `manifest.json` entirely, not merely missing `config_hash`. Added the same secrets the PR-preview job already sets to the baseline-generation step, plus an explanation of the failure mode (both inline in the recipe and in the "environment-only changes" section, since this is easy to miss if you only skim the YAML).

- P3 (templated `name:`) — **fixed** (`a667ec8`). Confirmed `sync_fingerprints()` keys by the *resolved* name; documented that this specific field is the one exception to the "environment-only changes are invisible" guarantee, and recommended keeping `name:` literal if relying on `state:modified`.

- P2 (empty selection exits 1) — **already fixed at the code level, no doc change needed.** This is the same underlying bug fixed in #952 (`select_syncs()` now exempts `state:` tokens from the must-match-something check; `run`/`build`/`test`/`validate` treat the resulting empty selection as a clean no-op). Checked the guide's own wording ("A sync present only in the baseline has no current definition to run, so it is silently ignored") — it already described the *intended* behavior correctly; the gap was purely the code lagging behind what this guide (written after the code, but reviewed before #952's fix landed) already promised. Once #953 is based on the fixed #952, this resolves itself with no doc edit required.

Sanity check (docs-only): `ruff check drt tests` clean, no files under `drt/`/`tests/` touched.

PR #954, opened 2026-08-10
feat(rest-api): body_mode: batch — send N records per request (#770)

### LINKED ISSUE #770: feat(rest-api): body_mode: batch — send N records per request from the REST API destination
## Problem

The REST API destination sends **one HTTP request per record**. For any real volume this is the dominant cost: 50k rows = 50k requests against an API that likely accepts arrays. `sync.batch_size` batches the engine loop, but the REST destination unrolls each batch back into per-record POSTs. Native connectors solved this individually (Salesforce Bulk, Mixpanel /import, staged_upload, and #734 gave Databricks multi-row VALUES — a 127–255× round-trip reduction), but the *generic escape hatch* connector — explicitly positioned as "covers any HTTP endpoint not yet supported natively" — has no bulk mode.

## Prior art

- **dlt custom/REST destinations**: batch by default; per-item is the opt-out.
- Most SaaS bulk endpoints accept JSON arrays (`POST /contacts/batch`, `/records`, NDJSON bodies).

## Proposal

```yaml
destination:
  type: rest_api
  url: https://api.example.com/contacts/batch
  method: POST
  body_mode: batch                # record (default, unchanged) | batch
  batch_template: |               # renders once per engine batch
    {"contacts": {{ rows | tojson_safe }}}
  # optional per-request size cap (≤ sync.batch_size)
  max_records_per_request: 500
```

- `rows` = list of (post-transform) records available in the Jinja context; `body_template`/`row` stays for `record` mode
- Response handling: 2xx = whole batch OK; on 4xx/5xx honor retry, and if the API returns per-item errors, an optional `error_path` maps failures back to records for `on_error: skip`/DLQ semantics (fallback: whole batch marked failed)
- Rate limiter counts requests (not records), so RPS limits become dramatically cheaper to respect

## Scope

**In:** batch mode + template context, size cap, per-item error mapping (best-effort), DLQ integration, docs + quickstart example.
**Out:** multipart/compression, automatic endpoint discovery, pagination of request bodies beyond the size cap.

## Related

- #734 (Databricks batching — same shape of win, SQL side), #258 (staged_upload for async bulk APIs — heavier pattern for the same pain), #580 (`tojson_safe` powering the array rendering)

## Milestone fit

Candidate: v0.8.x — high leverage for the flagship generic connector; per-item error mapping is the only subtle part.

masukai:
## Summary

Closes #770. The REST API destination sends one HTTP request per record
today — for real volume this is the dominant cost (50k rows = 50k
requests). `body_mode: batch` sends the engine's already
`sync.batch_size`-bounded chunk in one request (or further split via
`max_records_per_request`), without creating a second batching authority
in the destination layer.

```yaml
destination:
  type: rest_api
  url: https://api.example.com/v1/users/batch
  body_mode: batch
  max_records_per_request: 100
  batch_template: '{"records": {{ rows | tojson_safe }}}'
  error_path: data.results
```

- `batch_template` renders with `rows=<sub-chunk>` (existing `tojson_safe`
  handles lists as-is, no changes needed there).
- Rate limiting moves to once per outer HTTP request (not per record) —
  this is what actually delivers the RPS win, and required moving the
  `rate_limiter.acquire()` call site, not a `rate_limiter.py` change.
- Retry (`with_retry`) wraps the whole request atomically, same generic
  mechanism as record mode.
- `RowError.batch_index` tracks a running global offset across
  sub-chunks, so DLQ pairing (`engine/sync.py`'s
  `record_batch[err.batch_index]`) stays correct no matter how many HTTP
  requests one engine batch becomes.
- Default `body_mode: record` is byte-for-byte unchanged.

## Design decisions (the issue's own "only subtle part")

The issue sketched `error_path` at a high level without specifying the
exact mapping contract. I designed and implemented this precisely:

- `error_path` is consulted **only on non-2xx responses** — a 2xx with
  embedded per-item failures (HTTP 207-style) is a documented, known
  limitation, not silently mishandled.
- The resolved value must be a JSON list whose length **exactly matches**
  the sub-chunk size (index-aligned: element `i` ↔ record `i` of *that*
  request, not the whole engine batch).
- `null` → success, non-null → failure. For a dict, message priority is
  `error` → `message` → `error_message`, else `str(element)`.
- **Any contract violation (bad JSON, path doesn't resolve, wrong type,
  wrong length) falls back to whole-sub-chunk-failed, plus a logged
  warning** — never a partial/broken mapping. A misconfigured
  `error_path` is discoverable, not a silent degradation.

Happy to revisit any of this — it's my interpretation of an
intentionally underspecified part of the issue, not something the issue
text pinned down itself.

## Test plan

- [x] `pytest -q -m "not dwh_smoke and not local_sql_smoke and not object_store_smoke"` — 3204 passed, 4 skipped, 21 deselected (re-run independently)
- [x] Dedicated test for the highest-risk case: a failure in the
      *second* sub-chunk with `max_records_per_request` smaller than the
      engine batch, asserting `batch_index` lands on the correct global
      index (`test_error_path_maps_second_sub_chunk_to_global_indexes`)
- [x] Full `error_path` contract covered: clean mapping, shape-mismatch
      fallback + warning, message-priority ordering, 2xx never consults
      `error_path` (asserted via a mock that raises if `.json()` is
      called), empty batch (real `pytest_httpserver` contract test added
      for batch mode specifically)
- [x] Coverage: new validator/batch-mode lines in
      `destinations_saas.py` and `rest_api.py` all covered — confirmed
      by checking exact line numbers against the `Missing:` report, not
      trusting the aggregate percentage (the file-wide number looks low
      only because of large pre-existing, unrelated destination classes
      in the same shared config file)
- [x] `render_template()`'s signature was generalized to accept named
      context (`rows=...`) while keeping the existing positional `row`
      argument — verified all 17 other callers (`slack`, `hubspot`,
      `notion`, `jira`, etc.) are unaffected
- [x] VS Code extension schema regenerated (`sync.schema.json`),
      extension version bumped — this session's earlier lesson from
      #756/#947 about not missing this step


masukai:
**Self-review.**

Traced the offset arithmetic through `_chunk_records`/`_fail_chunk`/`_handle_batch_http_error` specifically, since this is the one place a subtle off-by-N bug would hide silently (the engine's `record_batch[err.batch_index]` DLQ pairing has no bounds error — a wrong index just drops the wrong dead-letter, or the right one, with no signal). Confirmed `test_error_path_maps_second_sub_chunk_to_global_indexes` actually exercises 2 sub-chunks with the failure in the *second* one and checks the resulting `batch_index` lands on the correct global position (3, not 1) — a single-sub-chunk test couldn't have caught an offset bug.

Confirmed the `error_path` contract is applied consistently: JSON-parse failure, path-resolution failure, wrong type, and wrong length all route through the same fallback (whole-sub-chunk-failed + warning), verified via 4 separate tests rather than assuming one covers all four failure shapes. Confirmed a 2xx response never even calls `.json()` on the response (`test_successful_response_does_not_consult_error_path` uses a mock whose `.json()` raises `AssertionError` if called at all — a stronger check than just asserting the outcome).

Confirmed `render_template()`'s backward-compatible signature change: every one of its ~17 callers across the SaaS destinations passes `row=` positionally or via the existing calling convention, none were touched, and the full suite (not just `test_rest_api.py`/`test_renderer.py`) passes.

Confirmed the new config validator lines (`_check_max_records_per_request`, `_check_body_mode_templates`) are inside the covered range, not part of the large `destinations_saas.py` `Missing:` list Codex reported — that list is pre-existing gaps in the ~15 other destination config classes sharing the file, unrelated to this change.

One thing I want to flag explicitly rather than silently decide: the `error_path` mapping contract (index-alignment, `null`-means-success, message-key priority) is my own design for a part the issue left underspecified — see the PR description's "Design decisions" section. Genuinely open to a different shape if you'd rather match a specific real API's convention instead.

Full suite re-run independently in my own environment (not just Codex's sandbox report): 3204 passed, 4 skipped, 21 deselected, no regressions.

masukai:
**Codex review:**

- [P2] Handle every batch-template rendering exception — `drt/destinations/rest_api.py:95-98`
  When a batch template raises `ZeroDivisionError`, `TypeError` from `tojson_safe`, or `TemplateSyntaxError`, `render_template` does not normalize it to `ValueError`, so this handler is bypassed and `destination.load()` aborts even with `on_error: skip`.

- [P2] Reconcile #770 with the roadmap before announcing it — `CHANGELOG.md:42`
  This entry announces #770 as landed while `ROADMAP.md:200` still classifies it as unscheduled and "not yet pulled in." AGENTS.md requires ROADMAP.md to remain the release-scope SSoT and to be updated first.

- [P3] Include the JSON content type in the batch example — `docs/connectors/rest-api.md:56-59`
  For JSON bulk endpoints, copying this example produces a 415-style failure because batch mode sends encoded bytes through `content=`, which does not add that header automatically.

---

**All three confirmed real, all three fixed** (`38b8d62`).

**Template exception handling**: verified `render_template()`'s own except clause only converts Jinja's `UndefinedError` to `ValueError` — nothing else. Then checked **record mode's** existing (pre-#770, unrelated to this PR) template handling and found the *exact same* narrow `except ValueError` — so this isn't a regression #770 introduced, it faithfully replicated an existing pattern. But this codebase already has an explicit, documented precedent that says the narrow catch is wrong: `computed_fields` (#763)'s CHANGELOG entry states outright — "Error handling deliberately catches every exception a template can raise, not just the missing-column one... catching narrowly would let it abort a run that had explicitly asked to skip bad rows." Broadened **both** record-mode's and batch-mode's template exception handling to `except Exception`, so I'm not introducing a new asymmetry between the two modes while fixing one of them. Added a `ZeroDivisionError`-from-row-arithmetic test for each mode (mirroring `computed_fields`' own `{{ row.a / row.b }}` example).

**ROADMAP.md**: confirmed it still listed #770 under "unscheduled backlog... not yet pulled in" even though I'd already moved the GitHub milestone earlier in this session. Moved it into v0.9's write-path-semantics scope line and removed it from the backlog list, per AGENTS.md's SSoT requirement.

**Content-Type**: confirmed batch mode's `do_batch_request` sends the rendered body via `content=_body.encode()`, not `json=` — httpx doesn't infer a content type there the way passing `json=` would. Added the same `headers: {Content-Type: application/json}` block the record-mode example already carries, plus a one-line explanation of why batch mode specifically needs it stated explicitly.

Full suite + `tests/unit/test_rest_api.py` re-run independently: 3204 passed, 4 skipped, 21 deselected (2 unrelated `test_cli_version.py` failures, same path-length worktree artifact noted on the other 8 PRs in this review pass).

PR #956, opened 2026-08-11
feat(engine): metadata_columns — opt-in engine-injected bookkeeping columns (#762)

### LINKED ISSUE #762: feat(engine): first-class run_id + opt-in metadata columns (_drt_synced_at, _drt_run_id) on destinations

masukai:
Closes #762.

## Summary

drt writes destination rows byte-for-byte from the source: no marker of *when* a row was synced, *which run* produced it, or *which sync* owns it — debugging a stale-looking CRM row means cross-referencing local `.drt/history` files a CI runner throws away between runs.

Most of #762's originally-proposed scope had already shipped incrementally: `run_id`/`sync_run_id` are already threaded through `SyncResult`, `HistoryEntry`, `DeadLetter`, OTel span attributes, and failure alerts (verified — each carries a `# Correlation ID (#762)` comment). This PR lands the one remaining piece: **opt-in destination columns.**

```yaml
sync:
  metadata_columns:
    synced_at: _drt_synced_at
    run_id: _drt_run_id
    sync_name: _drt_sync_name   # off by default even when the other two are set
```

Each field is independently opt-in; the block absent (default) changes nothing.

## Design

- **Engine seam, not per-destination code** — same pattern `computed_fields` (#763) established: a pure dict-enrichment function with no I/O, called from `engine/sync.py`'s batch loop. Reaches every destination that takes a `list[dict]` (SQL, SaaS, and file/blob), not just the REST destination.
- **Ordering: last of every transform**, applied after `mask` — the opposite end of the pipeline from `computed_fields`. Column names here are already destination-facing (chosen directly in config, not derived from a source column) and values are engine bookkeeping, not source data, so there's nothing for `field_mappings` to rename or `mask` to obscure. Also means a `computed_fields` template can never observe a metadata column, since it doesn't exist yet when `computed_fields` runs.
- **One value per run, not per-record**: `synced_at`/`run_id` are computed once in `run_sync()` and shared by every row the run writes — matches a "load id" semantics (dlt's `_dlt_load_id`), not a per-record timestamp.
- **`run_id` is nullable** for library callers of `run_sync()` that don't pass one — same nullability `SyncResult.run_id` already has, no synthetic fallback.
- **Requires the target column to already exist** on the destination — dict enrichment, not DDL. A destination without that column fails the write normally, governed by `on_error`.
- **DLQ interaction, deliberate**: a record that fails carries its metadata columns from the *original* attempt (metadata_columns runs before `destination.load()`, and `DeadLetter.record` captures that exact dict). `drt retry` resends verbatim, so a successfully-retried row's `_drt_synced_at`/`_drt_run_id` trace back to when it first failed, not the retry.
- **`unit_test_runner.py` needed zero changes** — it calls `run_sync()` directly rather than reimplementing the pipeline, so metadata_columns flows through `drt test --unit` automatically. Updated its docstring and `UnitTest`'s config docstring to name the new pipeline stage for accuracy (the fix also regenerates the bundled schema, since that docstring is embedded in it).

## Testing

- `tests/unit/test_metadata_columns.py`: 18 tests — pure-function behavior (no-op cases, partial config, null run_id, in-place overwrite), config validation (duplicate target column names rejected), and end-to-end through `run_sync()` (survives mask untouched, not renamed by field_mappings, run_id flows from the caller, one synced_at shared across multiple batches).
- Full suite: `3358 passed, 4 skipped, 33 deselected`
- vscode-drt bundled schemas regenerated (`regenerate-schemas.sh`), CHANGELOG entries in both `CHANGELOG.md` and `integrations/vscode-drt/CHANGELOG.md`
- Cross-references updated in `computed-fields.md`, `field-mappings.md`, `pii-masking.md` (the "mask runs last" claim was true before this PR and is now `metadata_columns` — fixed rather than left stale), new `docs/guides/metadata-columns.md`, `docs/llm/API_REFERENCE.md`


masukai:
**Codex review:**

The metadata injection works for valid configurations, but malformed empty target names are accepted and can silently disable the requested column. This should be rejected at configuration time.

- [P2] Reject empty metadata column names — `drt/config/sync_options.py:220-223`
  When a user configures `synced_at: ""` (for example after environment substitution), validation succeeds because falsey values are excluded from `targets`, and `apply_metadata_columns` silently treats the field as disabled. Whitespace-only names are also accepted and become unusable destination keys; validate each non-null value with `strip()` so malformed opt-in configuration fails during parsing rather than producing a successful run with missing metadata.

---

**My triage:** Confirmed by reading `_check_no_duplicate_targets` and `apply_metadata_columns` directly — both use a truthy check (`if v`) rather than `is not None`, so `synced_at: ""` slips past validation and then gets silently treated as "not configured" at apply time. Real bug, same failure class `computed_fields` already guards against (`if not name.strip(): raise ValueError(...)`).

Fixed in `a92e210`: the validator now checks each configured (non-`None`) value with `.strip()` and raises `metadata_columns.<field> must be a non-empty column name` before the duplicate-target check runs. Added a parametrized test (`test_rejects_blank_column_name`, empty string + whitespace-only). 20/20 tests pass, ruff/mypy clean.

PR #957, opened 2026-08-11
docs: state read-only-source access as an explicit drt differentiator

masukai:
Part of #922.

## Summary

#922 exists to protect two ADR 0005 design commitments as tiered features land: (1) cursor-based incremental staying drt's permanent, first-class, read-only-source strategy — not a legacy fallback #755's diff-based incremental supersedes — and (2) tier changes staying reversible config changes, never a one-way door.

Checking both against what's actually shipped:

- **Reversibility (#756)**: already verified true and documented. `docs/guides/remote-state.md` (landed with #756) is explicit that switching `state.backend` is a config change plus an optional, clearly-documented `gsutil rsync`/`aws s3 sync` step — never a vendor-control-plane lock-in. Nothing to fix here.
- **Cursor-based incremental's first-class status**: true in the code, but **stated nowhere public** — not in the README, not in `docs/guides/comparison.md`. It only existed as an internal ADR decision. This PR fixes that gap.

#755/#760/#920 (the other three features #922 was meant to watch) are all currently unscheduled with zero PRs in flight and zero community engagement (reactions/comments) — there's nothing else for #922 to actively check right now.

## What changed

`docs/guides/comparison.md`:
- New bullet under "When to choose drt": read-only source access as a reason to pick drt.
- New "Read-only source access, by design" section explaining the trust trade-off reverse ETL tools generally make, with concrete competitor examples — but **only where I could verify them**: Hightouch's Lightning/Basic split (its own docs, already cited in ADR 0005) and Segment's managed `_segment_reverse_etl` schema (same). I explicitly did not repeat #755's old Census claim, which ADR 0005 itself already flags as **unverified** (`docs.getcensus.com` now redirects to Fivetran's Activations docs post-acquisition) — searched again for this PR and still couldn't confirm it, so it stays out rather than risk a misleading comparison claim.
- Links to [ADR 0005](../adr/0005-state-location-and-write-grants.md) for readers who want the full reasoning.

Docs-only, no code/test impact.

### DIFF
--- a/docs/guides/comparison.md
+++ b/docs/guides/comparison.md
@@ -43,6 +43,7 @@ All three are open-source, CLI-first, YAML-configured, and MCP-enabled. Together
 - You work with **AI coding tools** (Claude, Cursor) and want your reverse ETL layer to be accessible via MCP
 - You need **10-30 destinations** rather than 200+ — and the ones you need are covered
 - You prefer to **own your data pipeline** end-to-end
+- You want to grant **read-only access to your source warehouse** — drt's supported incremental strategy (cursor-based, reading a column like `updated_at`) never needs to write into the source. Nothing in your source warehouse is drt-managed, drt-created, or drt-owned.
 
 **drt is not the right fit when:**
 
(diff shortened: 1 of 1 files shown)


masukai:
**Codex review:**

The new positioning overstates the read-only guarantee by omitting the supported BigQuery watermark backend, which creates and updates a warehouse table. This can mislead users evaluating required source permissions.

- [P2] Qualify the read-only claim for BigQuery watermarks — `docs/guides/comparison.md:46`
  When a user colocates the supported `sync.watermark.storage: bigquery` backend with the source warehouse, `BigQueryWatermarkStorage.save()` creates and writes the `_drt_watermarks` table (`drt/state/watermark.py`), so the unconditional statement that drt creates or owns nothing there is false. Qualify this promise to local/GCS/S3 watermark storage or explicitly document BigQuery watermark storage as an opt-in exception requiring write access.

---

**My triage:** Confirmed by reading `BigQueryWatermarkStorage._ensure_table()` directly — it does run `CREATE TABLE IF NOT EXISTS ... _drt_watermarks` and write to it, and nothing stops an operator from pointing `watermark.project`/`watermark.dataset` at the same place their source reads from. My original wording ("Nothing in your source warehouse is drt-managed, drt-created, or drt-owned") was a real overclaim for a page whose whole premise is "we only state what we could verify" — exactly the kind of thing that page shouldn't get wrong about drt itself.

Fixed in `10dfda4`: qualified both the "When to choose drt" bullet and the prose paragraph to scope the claim to reading from the source, and added an explicit paragraph naming `sync.watermark.storage: bigquery` as the one existing, unrelated exception, with the fix (use local/gcs/s3 watermark storage, or a separate BigQuery project/dataset). ruff clean, docs-only change.

PR #959, opened 2026-08-11
feat(mcp): close all 11 CLI/MCP parity gaps (#870)

### LINKED ISSUE #870: feat(mcp): behaviour-changing CLI flags are missing from the MCP tools

masukai:
## Summary
- `drt_run_sync` gains `limit` (→ `extract_limit`, same mirror/replace rejection as the CLI) and `vars` (already-parsed dict; precedence resolved via `project_vars()`, not reimplemented)
- `drt_run_test` gains `dry_run`/`fail_fast`/`store_failures`/`store_failures_limit` — the multi-sync `--fail-fast` loop moved out of `drt test` into a shared `run_test_suite()` so both surfaces schedule syncs identically
- `drt_validate` gains `check_connection`/`strict`, and along the way picks up the secret-scan `warnings` it was silently missing entirely (it previously returned only `valid`/`errors`)
- `run:--threads`/`--fail-fast`/`--failed` are deliberately **excluded**, not implemented — multi-sync scheduling/selection concepts that don't apply to a single-sync tool, recorded in `COMMAND_EXCLUSIONS["run"]` with reasons, mirroring how `SELECTION_OPTIONS` already excludes `--select`/`--exclude`
- `scripts/drift_baseline.txt`'s `cli-mcp:` section is now empty; `check_cli_mcp_parity.py` and `check_drift.sh` both report zero gaps
- Fixed a reintroduced CHANGELOG duplicate-`### Changed`-header bug (recent PR stacks each added their own section) while adding the #870 entry

Closes #870.

## Test plan
- [x] `python scripts/check_cli_mcp_parity.py` — 0 gaps
- [x] `bash scripts/check_drift.sh` — 0 new drift, 0 baselined gaps
- [x] New/updated tests: `store_failures` opt-in write-back, `dry_run` preview, `fail_fast` multi-sync skip, `limit`/`vars` reaching the engine, `limit` rejected for mirror mode, `validate` strict/check_connection/secret-warnings
- [x] `pytest tests/unit` — 3252 passed, 3 skipped
- [x] `mypy drt` — clean (191 files)


masukai:
**Codex review:** ran an automated review against this branch and flagged 3 issues, all fixed in 6d4edd9:

- [P2] `unit=True` ignored `fail_fast` entirely — fixed by routing through a new shared `run_unit_test_suite()` (drt/cli/commands/test.py), which also de-duplicated a third inline copy of the per-sync unit-test loop.
- [P2] `store_failures_limit <= 0` reached `fetch_failing_rows` as a raw SQL `LIMIT` — now rejected up front, matching the CLI's `min=1`.
- [P2] `drt_validate` dropped hardcoded-secret warnings for syncs that failed schema validation (only merged warnings for successfully-parsed syncs) — now merged for `result.errors` entries too, matching the CLI's per-error warnings lookup.

Verified each independently against source before fixing, added regression tests for all three, and re-ran the full suite + ruff + mypy + `check_cli_mcp_parity.py`/`check_drift.sh` (all green) after the fixes.

Issue #961, opened 2026-08-11
perf(clickhouse): batch insert instead of one client.insert() call per record


masukai:
## Problem

`ClickHouseDestination.load()` (`drt/destinations/clickhouse.py:137`) inserts one row at a time in a Python loop — `client.insert(table_q, [[...]], column_names=columns)` called once per record — with a `# TODO: batch insert with fallback to row-by-row on error` left in place since the connector's initial commit (9663f1f).

ClickHouse is a columnar warehouse built for bulk inserts; frequent single-row inserts are the documented anti-pattern its own docs warn against (many small inserts create many parts, triggering merge pressure and "too many parts" errors under sustained load). Every other batch-capable destination in this codebase already has a batching story — Databricks multi-row `VALUES` (#734, 127-255x fewer round trips) and REST API `body_mode: batch` (#770) both exist for the same reason — which makes ClickHouse's gap more visible by contrast now that #770 has shipped.

## Proposal

Batch `client.insert()` calls per `sync.batch_size` (the engine's existing batching unit, default 100 — no new config needed), with per-row fallback to individual inserts on a batch failure so `on_error: skip` semantics and per-row `RowError` reporting are preserved — this is exactly what the existing TODO's own "fallback to row-by-row on error" already anticipates.

## Scope

**In:** batched insert on the happy path, row-by-row fallback + `RowError` reporting on batch failure.
**Out:** the finalize-phase DELETE/MERGE queries (mirror mode) — those are already single statements, not per-row, and out of scope here.

## Milestone fit

Not currently milestoned — perf work with no user-visible behavior change on the happy path. Fits a v0.9.x or v0.10 pass; not blocking anything else.

Muawiya-contact:
@masukai I’d like to take that one. I’ve reviewed the proposed batching approach and would be happy to implement the batched ClickHouseDestination.load() path while preserving the existing per-row error and on_error: skip semantics. Are you okay with me taking this issue?

Issue #960, opened 2026-08-11
feat(state): warehouse-managed table primitive — shared by #755 and #920 (ADR 0005 step 3)


masukai:
## Problem

[ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md)'s implementation ordering table names a step between #756 (object-storage backend, step 2) and both #920 (warehouse state backend, step 4) and #755 (diff-based incremental, step 5): **"Warehouse managed-table primitive (shared by 4 and 5)"**. Both #755 and #920 already reference it explicitly in their own "Related" sections ("shares the warehouse managed-table primitive, ADR 0005's ordering step 3") — but the primitive itself was never split into its own issue. It's implicit infrastructure two issues depend on that nobody can pick up independently, and both are currently stuck behind it in the unscheduled backlog.

## Proposal

A shared internal component (not user-facing on its own) that both #755 and #920 build on:
- Create-if-absent DDL for a drt-owned schema/table, on the sources with write access — BigQuery / Snowflake / Databricks / Postgres, the same set tracked mirror (#686) already established on the destination side
- A pre-created-schema escape hatch (grant only create/delete on a named schema rather than "let drt create schemas") — #755's proposal names this explicitly, matching Hightouch Lightning's documented pattern
- Idempotent / safe against concurrent creators (two syncs' first run racing to create the same managed table)
- Distinct from #760 (managed *destination* tables): that's user data tables on the destination side; this is drt's own bookkeeping tables on the *source* side

## Scope

**In:** the primitive itself — DDL generation, idempotent creation, the escape-hatch config knob, across the four warehouse sources above.
**Out:** the diff computation (#755) and the state backend wiring (#920) that consume it — this issue is purely the shared foundation.

## Related

- #755, #920 — both blocked on this
- #686 — tracked mirror already proved "state table + SQL diff" on the destination side; this generalizes it to the source side
- [ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md) — names this as implementation ordering step 3

## Milestone fit

Unscheduled backlog, same as #755/#920 (both cut from v0.9's active scope 2026-08-06 per ROADMAP.md). Pull in when either #755 or #920 gets picked up — this is what unblocks both.

masukai:
[...]

PR #962, opened 2026-08-11
fix(state): drt retry no longer drops concurrent DLQ appends (#955)

### LINKED ISSUE #955: drt retry silently drops concurrent DLQ appends (replace() writes a stale snapshot)
## Problem

`replay_dead_letters()` (`drt/cli/commands/retry.py`) reads the DLQ once at the top, decides what to keep/drop, and writes the result back at the end — with no reconciliation against what might have changed in between:

```python
entries = store.read(sync.name)          # retry.py:66  — snapshot A
...
store.replace(sync.name, untouched + remaining)  # retry.py:136 — blind overwrite
```

If a concurrent `drt run` appends a new dead letter to the same sync's DLQ while `drt retry` is between its read (line 66) and its write (line 136), that append is silently lost — `replace()` overwrites the whole object with content computed from the stale snapshot, which never saw the new entry.

## Scope: both backends, this is not new

This predates #756 (remote state) and is not introduced by the object-store work:

- **`LocalDlqStore.replace()`** (`drt/state/dlq.py:130-142`) is an unconditional whole-file rewrite — no precondition, no merge. `drt run` and `drt retry` doing read-modify-write on the same local JSONL file has always been last-writer-wins (the class docstring even says so: "a concurrent `drt run` and `drt retry` doing read-modify-write on the same file is last-writer-wins").
- **`ObjectStoreDlqBackend.replace()`** (`drt/state/_objectstore.py:383-389`) does retry-on-precondition-failure internally via `_replace_with_retry`, so the *write itself* is safe against concurrent writers racing on the same generation/ETag. But the *content* it retries with is still the caller's stale `untouched + remaining` list — a successful retry just means it won the race to overwrite with outdated content, not that it merged correctly.

So the low-level object-store primitive (`read_for_update`/`write_if`, generation/ETag preconditioned) is correct; the bug is one level up, in how `retry.py` uses `read()` + `replace()` as if it were a single atomic operation.

## Why this needs its own design pass

The natural instinct — "just retry the whole read-process-write cycle on conflict" — isn't safe here, because `replay_dead_letters()`'s middle step calls `destination.load()` to actually resend records. Re-running that on a write conflict risks double-sending records that already succeeded on the first attempt (destination-side idempotency is the user's responsibility, per the DLQ guide, not something drt guarantees).

The correct fix needs a way to identify "this dead letter, the one I already resent" across a read-modify-write cycle — i.e. something like a `DlqBackend.reconcile(remove=[...], update=[...])` operation that re-reads fresh state and applies removals/updates by identity rather than overwriting wholesale. `DeadLetter` doesn't currently have a stable identity field for this: `(record, error_message, timestamp)` isn't sound, since the same record failing twice with the same error is an explicitly supported state (that's what `attempts` tracks), so two entries can collide on all three fields.

Filed separately rather than folded into #946, since it needs its own design (likely a Protocol change touching both `LocalDlqStore` and `ObjectStoreDlqBackend`, plus probably a real entry id on `DeadLetter`), not a rushed inline patch.

## Suggested next step

Design a stable identity for `DeadLetter` entries (e.g. a UUID assigned at creation, persisted through retries) and a `DlqBackend` operation that reconciles by identity against a fresh read rather than replacing wholesale. Update `replay_dead_letters()` to use it.

masukai:
## Summary
- `replay_dead_letters()` used to read the DLQ once, decide what to keep/drop, and write back the whole queue with `DlqBackend.replace()`. A dead letter a concurrent `drt run` appended between that read and that write was silently lost — on both the local (JSONL, always last-writer-wins by design) and object-store (precondition-safe write, but retries with the *same stale content*) backends.
- Same failure class as #919 (concurrent writers, stale snapshot committed as truth) — caught three days after that fix shipped.
- `DeadLetter` gains a stable `id` (assigned once at creation, preserved across `drt retry` attempts). Both `LocalDlqStore` and `ObjectStoreDlqBackend` gain `reconcile(remove_ids=..., updates=...)`, which re-reads current state itself and only touches entries named by identity — anything the caller never saw (a concurrent append, or an entry beyond `--limit`) survives untouched.
- `retry.py` now uses `reconcile()` instead of `replace()`. `--clear` still uses `replace([])` — documented as a full, unrecoverable discard, not a gap.
- Byproduct fix: a retried entry's `sync_run_id` now correctly carries over from the original failure instead of resetting to `null` (the reconstructed `DeadLetter` needed rebuilding here regardless; matches #762's stated correlation-id semantics).
- Entries written before `id` existed still load (dataclass default, self-heals on next write) — same tolerance pattern as `sync_run_id`'s own missing-key handling.

## Also in this PR (pre-v0.9.0 roadmap sweep)
- Files #960 — ADR 0005 step 3 ("warehouse managed-table primitive"), which #755 and #920 both cite as a blocking dependency but which had no issue of its own.
- Files #961 — ClickHouse destination inserts one row at a time (`# TODO: batch insert...` left in the code since the connector's first commit); no existing issue covered it.
- `ROADMAP.md`: fixes a copy-paste error (step 3's dependents are #755/#920, not #755/#760, per the ADR's own table and #755/#920's own "Related" sections) and updates the stale #919 status note (fixed 2026-08-08, not still a v0.8.6 patch candidate).

Closes #955.

## Test plan
- [x] New regression test: `test_retry_survives_concurrent_append` — the literal #955 scenario (a fake destination appends a fresh dead letter mid-retry), asserts the concurrent entry survives.
- [x] New unit tests: `reconcile()` remove/update-by-id on both backends, survives-a-racing-writer (object store), raises-after-bounded-contention (matching `replace()`'s existing contract), legacy-entry-without-`id`-still-loads.
- [x] Existing `--limit`/`--clear`/`--dry-run` retry tests pass unmodified (the implicit-untouched-entries behavior is preserved).
- [x] `pytest tests/unit` — 3267 passed, 3 skipped
- [x] Coverage on touched files: `retry.py` 100%, `dlq.py` 100%, `_objectstore.py`'s new code 100% (remaining file gaps are pre-existing, outside this diff)


masukai:
**Codex review:** ran an automated review against this branch and flagged 2 issues, both addressed in b54e7cb:

- [P1] Legacy DLQ entries (written before `id` existed) got a *different* random id on each independent JSONL read — and `replay_dead_letters()` reads the queue twice per invocation (once to decide what to retry, again inside `reconcile()`). Every legacy entry's remove/update would silently no-op, leaving it stuck in the queue and risking a duplicate resend on the next retry. Fixed: legacy lines now decode via a content-hash of the line itself (deterministic across independent reads of the same unchanged bytes), via a new shared `decode_dead_letter_line()` used by both backends.
- [P1] `LocalDlqStore.reconcile()` doesn't actually close the cross-process race the PR title implies — it has no OS-level file lock (only a process-local `threading.Lock`), so a genuinely separate `drt run` process racing the read/write window still loses that append, same as `replace()` did. Confirmed correct: this class was never cross-process-safe (its own docstring already said so) and this PR doesn't change that. Only `ObjectStoreDlqBackend.reconcile()` is actually safe against a concurrent writer (real generation/ETag preconditioning). Rather than silently leave this undocumented or overclaim the fix, I've made the code comments and CHANGELOG explicit about the scope split and filed #963 to track real local cross-process safety (across `LocalDlqStore`/`LocalStateManager`/`LocalHistoryManager` — the same gap, not DLQ-specific).

Verified each independently against source before fixing, added regression tests for the legacy-id bug (proving two independent reads of the same unchanged line now agree), and re-ran the full suite + ruff + mypy (all green, 3270 tests) after the fixes.

Issue #966, opened 2026-08-13
docs: computed-fields.md null-safety example is still broken (Jinja filter precedence)


masukai:
## What's wrong

`docs/guides/computed-fields.md` walks through the "phone might be null" gotcha and lands on this as the fix (line 135):

```yaml
computed_fields:
  p: "{{ row.phone or '' | replace('-', '') }}"
```

This doesn't actually fix anything. Jinja's `|` binds tighter than `or`, so this parses as:

```
row.phone or (''|replace('-', ''))
```

For a non-null value like `090-1234`, `row.phone` is truthy, so the `or` short-circuits and returns `row.phone` unchanged — the `replace` filter never runs, and dashes stay in the output. The example only "looks" correct because most people test it against a null value (where it correctly returns `''`), not a real phone number.

The fix needs to parenthesize the fallback before applying the filter:

```yaml
computed_fields:
  p: "{{ (row.phone or '') | replace('-', '') }}"
```

Same bug is duplicated in `docs/llm/API_REFERENCE.md` (the LLM-facing reference mirrors this guide) — both need the fix.

Found this while reviewing a drt-web docs sync (drt-hub/drt-web#66) with a Codex-assisted review.

Muawiya-contact:
@masukai Is it still open?

masukai:
@Muawiya-contact confirmed already fixed — checked both files on current `main`: `docs/guides/computed-fields.md:138` and `docs/llm/API_REFERENCE.md:256` both already show the correct parenthesized form (`{{ (row.phone or '') | replace('-', '') }}`), and the surrounding prose correctly explains the `|`-binds-tighter-than-`or` gotcha. Searched the rest of `docs/`/`examples/` for the same unparenthesized pattern — no other occurrences. Closing as already resolved, no code change needed here.

PR #967, opened 2026-08-13
docs+fix: secret-provider-URI gaps and Jinja filter-precedence bug (#965, #966)

### LINKED ISSUE #965: docs: secret-provider-uris.md overclaims provider-URI support for every `*_env` field
## What's wrong

`docs/guides/secret-provider-uris.md` (line 11) says:

> Any `*_env` field in a profile or destination config can be a `scheme://...` URI instead of a plain env var name.

In practice, only fields that are read through `resolve_env()` (drt/config/credentials.py) support this. A number of connectors read their `*_env` fields directly with `os.environ.get(...)`, bypassing `resolve_env()` entirely:

- `drt/destinations/slack.py:62` — `webhook_url_env`
- `drt/destinations/twilio.py:61,64` — `account_sid_env`, `auth_token_env`
- `drt/sources/postgres.py:185` — `password_env`
- (worth auditing `redshift.py`, `salesforce_bulk.py`, `jira.py` too — same direct-`os.environ` pattern is likely there)

For these fields, a value like `aws-sm://prod/slack-webhook` is treated as a literal environment-variable *name*, so drt looks up an env var literally called `aws-sm://prod/slack-webhook`, finds nothing, and auth fails — silently or with a confusing "not set" error rather than a clear "provider URIs unsupported here" message.

Found this while reviewing a drt-web docs sync (drt-hub/drt-web#66) with a Codex-assisted review.

## Suggested fix

Either:
1. Route the remaining connectors' `*_env` fields through `resolve_env()` for consistency, or
2. Narrow the doc's claim to name the connectors/fields that actually support provider URIs (and ideally raise a clear error if a provider-scheme URI is passed to a field that doesn't support it, instead of treating it as a literal var name).

Option 1 is probably the better fix — the current behavior is a silent trap for anyone who reads the guide at face value.

masukai:
## What

**#966 — Jinja filter-precedence bug (docs-only):**
`computed-fields.md` / `docs/llm/API_REFERENCE.md` / `CHANGELOG.md`'s null-safety example `{{ row.phone or '' | replace('-', '') }}` doesn't do what it looks like — Jinja's `|` binds tighter than `or`, so this parses as `row.phone or (''|replace(...))`. For a non-null value, `or` short-circuits and the filter never runs, so dashes stay in. Fixed to `{{ (row.phone or '') | replace('-', '') }}`, verified against both cases with jinja2 directly.

**#965 — provider URIs silently no-op on 10 credential fields (real fix, not just docs):**
`discord`, `email_smtp`, `google_ads`, `google_sheets`, `jira`, `slack`, `teams`, `twilio` destinations and the `postgres`/`redshift` sources read `*_env` fields via a plain `os.environ.get(...)` instead of `resolve_env()` — a provider URI (`aws-sm://`, `gcp-sm://`, `vault://`) passed there was treated as a literal env var name, found nothing, and auth failed as if the credential were unset, with no clear error pointing at the real cause.

All ten now route through `resolve_env()`, matching the pattern every other connector already uses (`elasticsearch.py`, `auth.py`, `zendesk.py`, `klaviyo.py`, ...). No config schema changes — the paired non-`_env` fields already existed. Also fixes a second bug found along the way: `jira.py`'s separate `_base_url()` helper used `os.environ[...]` directly, which would `KeyError` outright rather than just bypassing provider URIs.

`docs/guides/secret-provider-uris.md`'s "not every connector supports this yet" caveat (added earlier in this same PR) is removed now that it's no longer true.

Found both issues while reviewing a `drt-web` docs sync — nothing about how `drt` behaved from the outside was expected to change except that this now actually works.

## Test plan

- [x] `make lint`: ruff + mypy clean
- [x] New `tests/unit/test_credential_resolution_call_sites.py` (67 assertions across all 10 call sites, mocking `resolve_env` and asserting it's called with the right args)
- [x] Existing per-connector unit tests (slack, twilio, jira, discord, teams, email_smtp, google_ads, google_sheets, postgres source, redshift source) — 104 passed, no regressions
- [x] Verified the Jinja fix directly: `(row.phone or '') | replace('-', '')` → `'0901234'` for `'090-1234'`, `''` for `None`

Closes #965, #966.


masukai:
Ran this branch through review before asking for merge approval. Found 3 real issues, all fixed in the follow-up commit:

1. **Empty-batch regression**: `discord`/`email_smtp`/`google_ads`/`jira`/`slack`/`teams`/`twilio`'s `load()` resolved credentials (a potential network call to AWS/GCP Secrets Manager or Vault for a provider URI) *before* checking whether `records` was empty — breaking the established empty-batch no-network-I/O contract that `elasticsearch.py`/`klaviyo.py` already follow. Fixed with the same early `if not records: return` short-circuit.
2. **Missed connectors**: `intercom.py` and `salesforce_bulk.py` (6 fields across both) also bypassed `resolve_env()` — missed in the original audit. Fixed the same way as the other 10; CHANGELOG/docs now say 12, not 10.
3. **Jira re-resolving per record**: `_base_url()` was called once per record instead of reusing what `load()` already resolved once per batch. Removed the helper, threads the resolved value through instead.

Added `tests/contracts/test_destination_provider_uri_empty_batch.py` — asserts `resolve_env` is never called for an empty batch (hard-fails via `side_effect` if it is) across all 9 affected destinations, since this is exactly the class of regression the original tests didn't cover.

`make lint` clean, full `tests/unit/` + `tests/contracts/` suite green (3381 passed).

PR #973, opened 2026-08-14
fix(dagster-drt): resource.py + assets.py ignored state.backend

masukai:
## What

`DagsterDrtResource.run()` and the legacy `drt_assets_legacy()` path both constructed `StateManager(project_path)` directly instead of going through #756's `build_state_bundle()` factory. A project configured with `state.backend: gcs` or `s3` still wrote run state to local disk when executed through dagster-drt — silently defeating the remote backend for orchestrator-launched runs.

Found while researching #855 (dagster-drt sensors, v0.10 prep): the sensor's premise — a sensor process and a CI-launched `drt run` sharing durable state — doesn't hold if the execution path itself never picks up the remote backend. This is PR1 of the #855 plan; PR0 (#972, merged) added the CI coverage that lets this PR actually get verified.

## Test plan

- [x] `mypy integrations/dagster-drt/dagster_drt` — clean
- [x] `ruff check integrations/dagster-drt` — clean
- [x] `pytest integrations/dagster-drt/tests/ -q` — 36 passed (local scratch venv; `dagster-drt-test` CI job from #972 will also run this now)
- [x] CHANGELOG.md — new `### Unreleased (dagster-drt)` section (first entry since 0.2.0; also notes the separate CI fix)
- [ ] `dagster-drt-test` CI job passes on this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 > `dagster-drt` is published as a separate PyPI package with its own version.
 
+### Unreleased (dagster-drt)
+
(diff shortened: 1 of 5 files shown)


masukai:
**Codex review:** found two P1 issues in the initial commit:

1. `state_mgr` correctly switched to `build_state_bundle()`, but neither `resource.py` nor `assets.py` passed `observer=` to `run_sync()` — the engine only persists state/watermark/DLQ through an observer, so nothing was ever saved (local or remote) through either dagster-drt path.
2. `integrations/dagster-drt/pyproject.toml` still declared `drt-core>=0.7.7`, but the new `drt.state.factory` import only exists since drt-core 0.9.0 — a declared-valid older install would break with `ModuleNotFoundError`.

Both fixed in the follow-up commit, with two new regression tests (each verified to actually fail without the fix before being trusted) and the dependency floor raised to `>=0.9.0`.

PR #974, opened 2026-08-14
feat(dagster-drt): build_drt_change_sensor() — event-driven activation (#855)

masukai:
## What

`build_drt_change_sensor()` — Tier 2 of ADR 0004, the first sensor dagster-drt ships (the package previously had zero: asset/resource plumbing only). PR2 of the #855 plan (PR0 #972 CI, PR1 #973 state.backend fix, both merged).

Covers the two "nearly-free" starting points per `docs/research/warehouse-trigger-matrix.md`:
- **deltalake** — `DeltaTable.version()`, already read in `drt/sources/deltalake.py`
- **iceberg** — the catalog's `current_snapshot().snapshot_id`

## Design

- The change signal is tracked via Dagster's own sensor cursor (`context.update_cursor()`), not drt's `StateStore`/`WatermarkStorage` — "has the source moved" and "what has drt already synced" are different questions. The triggered run still goes through drt's own state exactly as before (PR1, #973).
- Both signals are **project-wide**, not per-sync: a `DeltaLakeProfile`/`IcebergProfile` maps to exactly one physical table, and a project has exactly one profile — there's nothing to distinguish per-sync at the signal level. What actually runs on a fire is entirely the caller's choice via `job`/`jobs`/`asset_selection`/`target`; this function only decides *when*.
- `NotImplementedError` (unsupported profile type) propagates as a **failed** sensor tick rather than a silent skip — a permanent config error shouldn't disappear forever. Transient I/O errors (network, auth, catalog hiccup) skip and retry next tick.
- First evaluation with no prior cursor always fires once — the same "no watermark yet, do a full pass" bootstrap drt's own incremental sync has.

Snowflake `STREAM` + SQL Server Change Tracking are separate follow-up PRs (PR3/PR4) — both have real consumption-semantics caveats (stream false-positives, DML-required consumption) that don't fit this metadata-only-read shape.

## Test plan

- [x] `mypy integrations/dagster-drt/dagster_drt` — clean
- [x] `ruff check integrations/dagster-drt` — clean
- [x] `pytest integrations/dagster-drt/tests/ -q` — 47 passed (9 new, sys.modules injection for deltalake/pyiceberg matching `tests/unit/test_deltalake_source.py`'s established pattern)
- [ ] `dagster-drt-test` CI job passes on this PR


masukai:
**Codex review:** found one P1 and two P2 issues:

1. **[P1]** `target=` doesn't exist in Dagster's `sensor()` before 1.8 — passing it unconditionally raised `TypeError` on every call when installed with Dagster 1.6/1.7, both allowed by this package's own `dagster>=1.6` floor.
2. **[P2]** `jobs=` (multi-job) was silently broken — Dagster requires each `RunRequest` to set `job_name` to disambiguate, and this sensor never did.
3. **[P2]** `run_key=current_signal` could collide across a rollback (Dagster dedupes run keys globally, not just consecutively) — a source table rollback A→B→A would silently drop the third run.

Fixed by dropping `target=`/`jobs=` from the public API (use `asset_selection=[my_syncs]` for the `@drt_assets` case instead — supported since Dagster 1.6.0) and keying `run_key` on the `previous->current` transition rather than the bare destination value. The Dagster-1.6-floor fix was verified empirically (installed real `dagster==1.6.14` in a scratch venv, constructed the sensor against it), not just by reading source. New regression test for the rollback case.

PR #976, opened 2026-08-14
fix: Airflow/Prefect run_drt_sync() never persisted state

masukai:
## Summary

`drt.integrations._runner.run_drt_sync()` (shared by `drt.integrations.airflow` and `drt.integrations.prefect`) resolved the correct state store via `build_state_bundle()` (#756), but never wired that store into an observer. `run_sync()` falls back to a silent no-op `NullObserver` whenever `observer=` isn't passed — so every state/watermark/DLQ save through this path was a no-op, on the **local** backend just as much as a remote one. Watermark storage was also never wired at all.

Found while researching #855/#856 (event-driven syncs docs) — the same bug class Codex caught in `dagster-drt`'s `DagsterDrtResource.run()`, in the two integrations that fix didn't touch.

## Fix

Compose `StatePersistingObserver` (+ `DlqObserver` when `sync.dlq.enabled`) the same way `drt/cli/commands/run.py`'s `_build_observer()` and `dagster-drt`'s resource already do. `build_state_bundle()` is now called unconditionally (previously short-circuited when a caller passed its own `state_manager`, which meant the DLQ store was never available in that case either).

`docs/guides/remote-state.md` corrected — it previously described the Airflow/Prefect gap as "local-only", which understated it: nothing was ever persisted, local included.

## Test plan

- [x] New regression test (`test_run_drt_sync_persists_state`) runs a real sync end-to-end (real DuckDB source, real file destination, only `.drt/state.json` mocked away) and asserts state was actually saved
- [x] Verified the test fails without the fix (temporarily reverted, confirmed `AssertionError`, restored)
- [x] `ruff check` / `mypy` clean on touched files
- [x] Full `test_airflow_integration.py` + `test_prefect_integration.py` + `test_engine_observer.py` + `test_engine.py` green (108 passed)


masukai:
**Codex review:**

> The patch restores run-state and watermark persistence but makes dry runs mutate incremental cursors, which can skip data on the next real run. It also leaves execution-history persistence unwired despite documenting full backend parity.

**[P1] Skip the persistence observer for dry runs** (`_runner.py:104`)

Verified, but **this is a pre-existing drt-core bug, not something this PR introduces** — `StatePersistingObserver` is wired unconditionally at every existing call site (`drt/cli/commands/run.py`'s `_build_observer()`, `dagster-drt`'s `resource.py`), and `engine/sync.py` calls `observer.on_sync_completed(...)` unconditionally regardless of `dry_run` (sync.py:783), with `new_cursor_value` accumulated from real extracted rows *before* the `if dry_run:` early-continue (sync.py:613-614, 670). Confirmed empirically:

```
$ drt run --dry-run   # fresh project, incremental sync, cursor_field: id
Dry run summary:
  Rows to sync: 1
$ cat .drt/state.json
{"users": {"last_run_at": "...", "records_synced": 1, "status": "success", "last_cursor_value": "1"}}
```

`records_synced` is also misleading during a dry run — nothing was sent. This PR matches the existing CLI/dagster-drt pattern rather than diverging from it; fixing the underlying gap (likely at the `StatePersistingObserver`/engine boundary) is a drt-core correctness change with its own test matrix, filed separately as #978 rather than folded in here.

**[P2] Pass the configured history store to run_sync** (`_runner.py:116-117`)

Confirmed and fixed in 395c4dd — `history_manager=bundle.history` and `history_retention_days=project.history.retention_days` now match the CLI's call site. Regression test added and verified via revert-confirm-restore. Also found while fixing this: `dagster-drt`'s `resource.py` (already merged, #973/#974) has the identical gap — not fixed here since it's out of this PR's scope, noted in `docs/guides/remote-state.md` instead of silently implying it's covered.

Issue #975, opened 2026-08-14
feat(dagster-drt): Snowflake STREAM + SQL Server Change Tracking sensor signals — design blocked


masukai:
## Background

#855 (dagster-drt sensors, Tier 2 of ADR 0004) shipped for Delta Lake and Iceberg ([PR #974](https://github.com/drt-hub/drt/pull/974)) — both have a monotonic, side-effect-free, metadata-only change signal (`DeltaTable.version()`, `current_snapshot().snapshot_id`) that a Dagster sensor can cheaply diff against its own cursor.

Snowflake and SQL Server were the other two "purpose-built signals" #855 originally named, but researching them surfaced a design problem that doesn't fit the same shape — filing this rather than building around it inside #855, so #855 can close on what actually shipped.

## The Snowflake problem

`docs/research/warehouse-trigger-matrix.md`'s own recommendation for Snowflake is a `STREAM` checked by `SYSTEM$STREAM_HAS_DATA()` **inside a Snowflake `TASK`'s `WHEN` clause** — i.e. Snowflake-side scheduling (ADR 0004 Tier 1), not an externally-polling Dagster sensor (Tier 2) at all. Three converging problems if you try to build a Tier-2 sensor around it anyway:

1. **`SYSTEM$STREAM_HAS_DATA()` is a boolean, not a monotonic value, and a cursor-diff sensor (the shape #974's Delta/Iceberg sensors use) needs a value that changes on every new change and doesn't reset itself.** A stream's `HAS_DATA` flag only resets when the stream is **consumed via DML** — plain querying (`SELECT`) never advances it. So once it flips to `TRUE`, it stays `TRUE` forever unless something writes through it. A cursor-diff sensor polling this would fire exactly **once** (the `FALSE -> TRUE` transition) and then go silent forever, even though new data keeps arriving — a sensor that silently stops working after its first real event, which is worse than not building it.
2. **The only fix — have the sensor itself consume the stream via a throwaway DML statement purely to reset the flag — requires a write grant on the source connection.** drt's source connections are read-only-by-design; escalating one to support a sensor's internal bookkeeping is an [ADR 0005](https://github.com/drt-hub/drt/blob/main/docs/adr/0005-state-location-and-write-grants.md)-shaped decision (write grants for sources), not something to slip into a sensor PR.
[...]

PR #977, opened 2026-08-14
docs: event-driven syncs guide (#856) + #855 scope closeout

masukai:
## Summary

- New `docs/guides/event-driven-syncs.md` (#856): all three ADR 0004 tiers, decision table, worked Dagster-sensor example. Scopes Tier 2 accurately — Delta Lake/Iceberg only, with a dedicated section on why Snowflake/SQL Server don't get a sensor (the boolean-latch/write-grant mismatch found while researching #855, spun out to #975).
- ADR 0004: new 2026-08-14 amendment + gate table + follow-up issues section updated to reflect what #855 actually shipped vs. what moved to #975.
- `integrations/dagster-drt/README.md`: adds the `build_drt_change_sensor()` section that was missing since #974 merged — including the deployment nuance that the sensor evaluates inside the Dagster **daemon**, which needs the source profile credentials.
- `CHANGELOG.md` (dagster-drt Unreleased entry): corrects the "Snowflake/SQL Server are follow-up PRs" line, which implied they'd ship the same way Delta/Iceberg did.
- `ROADMAP.md`: v0.10 scope updated — #855 and #856 closed, #975 added.
- #855 and #856 closed on GitHub with explanatory comments.

## Context

Continuing the #855 PR-stack (Delta/Iceberg sensors already shipped in #974). Researching Snowflake support for a planned PR3 surfaced that `SYSTEM$STREAM_HAS_DATA()` is a boolean that only resets on DML consumption — a cursor-diff sensor (the shape #974 built) would fire once and then latch permanently silent, since drt's Snowflake extraction never performs DML. Fixing that would require the sensor doing its own throwaway writes against a normally read-only source connection, which is a write-grant decision this ADR never scoped. Filed [#975](https://github.com/drt-hub/drt/issues/975) instead of building around it, and this PR documents the resulting shape rather than leaving #855/#856 open on an unresolved design.

Companion PR #976 (already open) fixes an unrelated bug found in the same research pass — Airflow/Prefect's `run_drt_sync()` never persisted state on any backend. That PR is independent and doesn't need to merge first for this one to be correct.

## Test plan

- [x] All cross-referenced doc paths verified to exist (`ls` on every linked file)
- [x] GitHub anchor slug for the dagster-drt CHANGELOG cross-link verified against the actual header text
- Docs-only change — no code touched, no tests to run


masukai:
**Codex review:**

> The new operational guide recommends Snowflake trigger paths that do not consume their stream and can fire indefinitely. It also gives incorrect deployment requirements for Dagster and drt serve, and records an implementation fix that is not present in the patch or base.

All five findings confirmed and fixed in f037aca:

**[P1] Tier 1's Snowflake guidance was wrong** — I claimed a `TASK`'s own execution consumes the stream just by invoking `drt run`. It doesn't: consumption requires DML against the stream specifically, and "call an external function / webhook" performs none. Without an explicit DML consumer in the task body, `SYSTEM$STREAM_HAS_DATA()` latches `TRUE` after the first real change and the task fires forever — the identical trap I'd already correctly documented for Tier 2, that I'd failed to apply to Tier 1. On closer look this also affects the Tier 3 Snowflake Alert path I wrote (an Alert's condition check is read-only, same shape as a `TASK`'s `WHEN` clause) — corrected that too, even though Codex didn't flag it specifically.

**[P2] Dagster daemon credentials** — I wrote that only the daemon host needs source credentials, "not the job's execution environment." Wrong: `DagsterDrtResource.run()` calls `load_profile()` independently when the job actually executes. Corrected to "in addition to, not instead of."

**[P2] Tier 3 framing** — I described all three tiers as needing "no new infrastructure" / never running as a daemon. `drt serve` is a resident process that must stay up to receive requests, unlike Tiers 1/2's start-sync-exit shape. Corrected.

**[P2] ROADMAP.md overclaimed #976 as already fixed** — reworded to reflect open-PR status rather than implying it had merged.

**[P3] Iceberg `snapshot_id` overclaimed monotonic ordering** — it's an opaque unique token (Iceberg's actual monotonic field is `sequence_number`; `snapshot_id` can move backward on rollback). The sensor itself only ever compares for equality (verified against `sensors.py`, already-merged code — not this PR), so behavior was never wrong, just the docs' characterization. Corrected in both the guide and the dagster-drt README.

PR #979, opened 2026-08-15
fix: dry runs must not persist state (#978)

### LINKED ISSUE #978: bug: drt run --dry-run persists an advanced watermark/cursor, risking skipped data on the next real run
## Summary

`drt run --dry-run` writes an advanced cursor to `state.json` (and to `WatermarkStorage` if configured), even though nothing was sent to the destination. The next real (non-dry-run) run then starts from that advanced position, potentially skipping rows that were only ever previewed, never actually synced.

Found via Codex review on #976 (unrelated PR), verified empirically — not a hypothetical.

## Reproduction

```bash
mkdir /tmp/drt_repro && cd /tmp/drt_repro
mkdir -p syncs
cat > drt_project.yml <<'YML'
name: p
profile: default
version: "1"
YML
cat > syncs/users.yml <<'YML'
name: users
model: "SELECT 1 AS id"
destination:
  type: file
  format: csv
  path: /tmp/drt_repro/out.csv
sync:
  mode: incremental
  cursor_field: id
YML
mkdir -p ~/.drt   # or HOME override
cat > ~/.drt/profiles.yml <<'YML'
default:
  type: duckdb
  database: ":memory:"
YML

drt run --dry-run
cat .drt/state.json
```

Result:

```json
{
  "users": {
    "sync_name": "users",
    "last_run_at": "2026-08-14T08:49:37.296518+00:00",
    "records_synced": 1,
    "status": "success",
    "error": null,
    "last_cursor_value": "1"
  }
}
```

No destination file was written (`out.csv` does not exist), but `state.json` shows `records_synced: 1` and an advanced `last_cursor_value` — a subsequent real `drt run` would treat `id=1` as already synced and never send it.

## Root cause

- `drt/engine/sync.py:613-614`: `new_cursor_value` is updated from every extracted row's `cursor_field`, **before** the `if dry_run:` early-continue at line 670 that skips the actual `destination.load()` call — so the max-cursor-seen tracking runs identically whether or not anything was written.
- `drt/engine/sync.py:783`: `observer.on_sync_completed(sync.name, total_result, started_at, new_cursor_value, cursor_field)` fires **unconditionally**, dry run or not.
- `drt/engine/observer.py`'s `StatePersistingObserver.on_sync_completed` has no `dry_run` awareness at all — it always calls `state_manager.save_sync(...)` and, if a cursor advanced, `watermark_storage.save(...)`.
- Every current call site wires `StatePersistingObserver` unconditionally: `drt/cli/commands/run.py`'s `_build_observer()`, `dagster-drt`'s `resource.py`, and `drt/integrations/_runner.py` (#976). None of them special-case `dry_run` for this observer.

This is an inconsistency, not a deliberate design — two adjacent side effects in the same `finally` block **are** already dry-run-guarded: history append (`sync.py:383`, `if not dry_run and history_manager is not None`) and `on_sync_ended` (`sync.py:421`, `if not dry_run`). `docs/guides/sync-history.md:33` documents the history guard explicitly ("Dry-run is skipped... it is for previewing, not auditing"). `docs/guides/dry-run-and-diff.md:81` makes the same read-only promise for the mirror-delete preview specifically. State/watermark persistence has no equivalent guard or documented exception — it just wasn't caught.

## Scope

Affects every dry-run invocation across every call site: `drt run --dry-run`, `drt build --dry-run` (if applicable), `drt serve ?dry_run=true`, `DagsterDrtResource.run(dry_run=True)`, and `run_drt_sync(dry_run=True)` (Airflow/Prefect). Not filed with a proposed fix location — whether the guard belongs in `StatePersistingObserver` itself (dry_run-aware, would need `SyncResult` or the observer call to carry that flag) or at each call site (mirroring the `DlqObserver`/dry-run pattern already used for DLQ) is a design call.

## References

- Found during #976 (Codex review, P1)

masukai:
## Summary

`drt run --dry-run` (and every other dry-run path: `drt serve ?dry_run=true`, `DagsterDrtResource.run(dry_run=True)`, `run_drt_sync(dry_run=True)`) was writing an advanced cursor and a fake run record to state, even though nothing was actually sent to the destination. A subsequent real run would then start from that advanced position and could silently skip the previewed-only rows.

Reproduced empirically before this fix:
```
$ drt run --dry-run   # fresh project, incremental sync, cursor_field: id
Dry run summary:
  Rows to sync: 1
$ cat .drt/state.json
{"users": {"last_run_at": "...", "records_synced": 1, "status": "success", "last_cursor_value": "1"}}
```

## Root cause

- `new_cursor_value` is tracked from extracted rows during the extraction loop (`sync.py:613-614`), **before** the `if dry_run:` early-continue (`sync.py:670`) that skips `destination.load()` — so cursor tracking happens identically whether or not anything was written.
- `observer.on_sync_completed(...)` (`sync.py:783`) fires **unconditionally**, dry run or not.
- `StatePersistingObserver.on_sync_completed` had no way to know the run was a preview, so it always called `state_manager.save_sync(...)` (recording `last_run_at`, `records_synced`, `status`, and `last_cursor_value`) and `watermark_storage.save(...)`.

This was an inconsistency, not a deliberate design — two adjacent side effects in the same `finally` block **are** already dry-run-guarded: history append (`sync.py:383`) and `on_sync_ended` (`sync.py:421`). State/watermark persistence had no equivalent guard.

## Fix

Root-caused rather than patched at each call site. `SyncObserver.on_sync_completed()` now takes a `dry_run: bool = False` keyword, threaded through from `run_sync()`'s own `dry_run` parameter — the one place this is authoritatively known. `StatePersistingObserver` early-returns when `dry_run=True`, before touching either `state_manager` or `watermark_storage`. `NullObserver` / `LoggingObserver` / `DlqObserver` / `CompositeObserver` updated to match the new signature (no behavior change for any of them).

This fixes the gap structurally for every current call site (CLI, `dagster-drt`, Airflow/Prefect via #976) **and any future one**, rather than relying on each caller to remember to omit `StatePersistingObserver` on dry runs — which is exactly the class of mistake that let this ship in the first place.

`SyncObserver` is not a documented public extension point (not exported from `drt/__init__.py`, no guide references implementing it), so this is treated as an internal engine change, not a breaking public API change.

## Test plan

- [x] Two new regression tests: `test_state_persisting_observer_skips_dry_run` (observer-level) and `test_run_sync_dry_run_does_not_persist_state` (end-to-end through `run_sync()`)
- [x] Verified both fail without the fix (temporarily stashed the engine changes, confirmed `AssertionError`/`TypeError`, restored)
- [x] Verified empirically against the real CLI: `state.json` is no longer written on `drt run --dry-run`
- [x] Full unit suite green: 3291 passed, 3 skipped
- [x] `ruff check` / `mypy` clean on all touched files

Closes #978.


masukai:
**Codex review:**

> The dry-run persistence fix works for built-in observers, but it introduces a runtime-breaking change for existing custom observers on all syncs, including non-dry runs.

Confirmed and fixed in 4434e8f. The original version of this fix added `dry_run: bool = False` as a new keyword to `SyncObserver.on_sync_completed()`, always passed explicitly from `sync.py`. Any custom `SyncObserver` implemented directly against the old five-argument signature (Python's structural `Protocol` typing means nothing prevents this) would break: called directly (not wrapped in `CompositeObserver`), it's an uncaught `TypeError` raised *after* the destination write already succeeded; wrapped in `CompositeObserver`, its per-child exception guard would swallow the error and silently drop the completion event instead — the custom observer's own logic (whatever it needed `on_sync_completed` for) would just stop running.

Redesigned to avoid touching the Protocol at all: `dry_run` is now carried on `SyncResult` (a plain dataclass field, not a Protocol method parameter) — `run_sync()` sets `total_result.dry_run` once, early, and `StatePersistingObserver` reads `result.dry_run` instead of a new kwarg. `SyncObserver`'s Protocol and all four built-in observers (`NullObserver`/`LoggingObserver`/`DlqObserver`/`CompositeObserver`) are back to their original signatures — zero compatibility risk for any existing or future observer, custom or built-in.

masukai:
**Codex review (second pass, after the SyncResult redesign):**

> The dry-run persistence fix is sound for built-in observers, but it introduces a backward-incompatible positional constructor change to the public SyncResult dataclass.

Confirmed and fixed in e2a6193. `dry_run` was inserted between `interrupted` and `diff`, shifting the positional slot of `diff`/`run_id`/`sync_run_id` by one — every call site in this repo already uses keyword arguments, but `SyncResult` is a public dataclass returned by every `Destination.load()` implementation, and nothing prevents a connector or library caller from constructing it positionally. Moved `dry_run` to the very end of the field list instead — purely additive, no existing positional slot changes meaning.

PR #981, opened 2026-08-15
fix(dagster-drt): wire history_manager (#980)

### LINKED ISSUE #980: fix(dagster-drt): DagsterDrtResource.run() never passes history_manager to run_sync()
## Summary

`DagsterDrtResource.run()` (`integrations/dagster-drt/dagster_drt/resource.py:130-142`) resolves the correct state/DLQ bundle via `build_state_bundle()` and wires `StatePersistingObserver`/`DlqObserver` correctly, but never passes `history_manager=`/`history_retention_days=` to `run_sync()`. `run_sync()` only appends execution history when `history_manager` is supplied (`drt/engine/sync.py:383`, `if not dry_run and history_manager is not None`) — so a Dagster-triggered sync's execution history stays empty, on every `state.backend` (local, GCS, S3 alike).

This is the identical gap [#976](https://github.com/drt-hub/drt/pull/976) fixed for `drt.integrations._runner.run_drt_sync()` (shared by Airflow/Prefect) — found *while* fixing #976, but out of scope for that PR since it touches `dagster-drt`, a separately-versioned package.

## Fix

Match the CLI's own call site (`drt/cli/commands/run.py:205-206`):

```python
result = run_sync(
    sync_config,
    source,
    destination,
    profile,
    project_path,
    dry_run=effective_dry_run,
    state_manager=state_mgr,
    watermark_storage=wm_storage,
    observer=CompositeObserver(observers),
    history_manager=bundle.history,
    history_retention_days=project.history.retention_days,
)
```

`bundle` (from `build_state_bundle()`) already carries `.history`; `project.history.retention_days` is already loaded via `load_project()`. No new resolution needed — this is a pure oversight fix, same shape as #976's.

## Also correct

`docs/guides/remote-state.md`'s "Execution history is a separate mechanism" note (added in #976) currently says `dagster-drt`'s resource does not wire `history_manager` — that note should flip to "does" once this ships.

## Test plan

Mirror #976's regression test: a real end-to-end sync through `DagsterDrtResource.run()` (DuckDB source, file destination), then assert `LocalHistoryManager(project_path).read(sync_name)` returns one entry. Verify via revert-confirm-restore.

## References

- Found during [#855](https://github.com/drt-hub/drt/issues/855)/[#856](https://github.com/drt-hub/drt/issues/856) research, same session as #976/#978/#979.

masukai:
## Summary

`DagsterDrtResource.run()` resolved the correct state/DLQ bundle via `build_state_bundle()` and wired `StatePersistingObserver`/`DlqObserver` correctly, but never passed `history_manager=`/`history_retention_days=` to `run_sync()` — `run_sync()` only appends execution history when `history_manager` is supplied (`drt/engine/sync.py:383`), so a Dagster-triggered sync's execution history stayed empty on every `state.backend`.

Identical gap to the one #976 fixed in `drt.integrations._runner.run_drt_sync()` (Airflow/Prefect) — found while fixing that one, but out of scope for it since `dagster-drt` is a separately-versioned package.

## Fix

Match the CLI's own call site (`drt/cli/commands/run.py:205-206`):

```python
history_manager=bundle.history,
history_retention_days=project.history.retention_days,
```

`docs/guides/remote-state.md` corrected — it previously flagged this gap explicitly; now says all three call paths wire history correctly.

## Test plan

- [x] New regression test `test_run_wires_history_manager` (mirrors the existing `test_run_wires_state_persisting_observer` pattern)
- [x] Verified it fails without the fix (temporarily stashed the fix, confirmed `KeyError: 'history_manager'`, restored)
- [x] Full dagster-drt suite green: 49 passed
- [x] `ruff check` clean; `mypy dagster_drt` clean (one pre-existing, unrelated error in `translator.py` confirmed present on `main` too)

Closes #980.


masukai:
**Codex review:**

> The history wiring fixes the default path but bypasses the documented history.enabled opt-out for Dagster executions, causing unwanted persistence.

Confirmed and fixed in c44896a. `history_manager=bundle.history` was passed unconditionally, ignoring `drt_project.yml`'s `history.enabled: false` opt-out — `run_sync()` appends history for every non-dry run whenever the manager is non-`None`, regardless of the config flag. Now matches the CLI's own `history_mgr = state_bundle.history if history_cfg.enabled else None` (`drt/cli/commands/run.py:758`).

While fixing this, found the **identical bug already merged to `main`** in `drt.integrations._runner.run_drt_sync()` (#976) — same unconditional `history_manager=bundle.history`, same missing `history.enabled` check. That's a live regression on `main` right now, affecting Airflow/Prefect. Filing and fixing separately since it's already merged.

PR #984, opened 2026-08-17
feat(dagster-drt): Snowflake + SQL Server sensor support (#975)

### LINKED ISSUE #975: feat(dagster-drt): Snowflake STREAM + SQL Server Change Tracking sensor signals — design blocked

masukai:
## Summary
- `build_drt_change_sensor()` now supports `snowflake` and `sqlserver` profiles, closing the gap #855 left open — verified against real accounts in #975 rather than assumed:
  - **Snowflake**: `SYSTEM$LAST_CHANGE_COMMIT_TIME('<table>')`. Requires new `watch_table=` and `minimum_interval_seconds=` keyword args (no single table on the profile; a real ongoing warehouse-compute cost per poll, unlike the other three signals).
  - **SQL Server**: `CHANGE_TRACKING_CURRENT_VERSION()`. Needs neither new arg — database-scoped, so it fires on any tracked table's change rather than a specific one (coarser, not unsafe).
  - Both raise `ValueError` (propagated, not swallowed into a `SkipReason`) on a `NULL` signal, instead of coalescing to a fake baseline — that would reproduce the exact "fires once, then goes permanently silent" failure mode #855 was filed to avoid.
- Docs: `docs/guides/event-driven-syncs.md`, dagster-drt `README.md`, a dated addendum to `docs/research/warehouse-trigger-matrix.md`, a dated amendment to ADR 0004 (does **not** retract the earlier `STREAM` finding — that finding is still correct, this ships a different function), CHANGELOG, ROADMAP.
- Closes #975 once merged.

## Test plan
- [x] 10 new unit tests (`sys.modules`-injected mocks for `snowflake.connector`/`pymssql`, matching the existing Delta/Iceberg pattern) — all profile-dispatch branches, both required-arg checks, both NULL-raise paths, `ValueError` propagation through the sensor body, `watch_table=`/`minimum_interval_seconds=` forwarding
- [x] `pytest integrations/dagster-drt/tests/` — 58 passed
- [x] `ruff check` / `mypy` clean on all touched files (one pre-existing, unrelated `translator.py:105` error confirmed present on `main` independent of this change)


masukai:
**Codex review:** two real findings, both fixed in 9a6cbde:

- **[P1]** `CHANGE_TRACKING_CURRENT_VERSION()` is database-scoped — it stays non-NULL as soon as *any* table in the database has Change Tracking enabled, saying nothing about whether the specific table a sync reads is itself tracked (`ALTER TABLE ... ENABLE CHANGE_TRACKING` is a separate per-table opt-in on top of the database-level one). Without validating that, a table that was never individually enabled would silently never advance the signal while the database-wide version kept moving from other tracked tables — the sensor would look like it's working but never actually fire for that table. Fixed by requiring `watch_table=` for SQL Server too (previously Snowflake-only), validated against `CHANGE_TRACKING_MIN_VALID_VERSION` before trusting the database-wide poll.
- **[P2]** A missing optional driver (`snowflake-connector-python`/`pymssql` — dagster-drt's base install pulls in neither) raised `ImportError`, which the sensor body's generic exception handler caught as "transient" and turned into a repeating `SkipReason` forever — indistinguishable in the Dagster UI from a sensor that's simply waiting for a change. Same latent bug existed for `deltalake`/`pyiceberg` since #855 (Codex pointed at the new Snowflake import specifically, but the sensor body's `except Exception` catch-all is shared by all four profile branches) — fixed uniformly by propagating `ImportError` the same way `NotImplementedError`/`ValueError` already were.

Docs corrected to match (SQL Server's "needs neither extra argument" framing was wrong once `watch_table=` became required there too). 3 new regression tests (61 total), re-verified with both `snowflake` and `pymssql` blocked via `sys.meta_path` to simulate CI's base install — all pass.

masukai:
**Codex review (round 2):** \`watch_table\` was interpolated into the Snowflake signal query without SQL string-literal escaping — an apostrophe inside a valid quoted identifier (e.g. \`"O'Brien"\`) would break the query, and it's an injection risk in principle regardless of \`watch_table\` being developer-supplied config rather than external input. Fixed in 51e773a: standard single-quote doubling before interpolation, same verified call shape otherwise, plus a regression test asserting the exact escaped SQL sent to the cursor.

PR #985, opened 2026-08-17
feat(dwh_smoke,dagster-drt): confirm Snowflake sensor poll is metadata-only (#975)

masukai:
## Summary
- **Confirmed live, 2026-08-18**: `SYSTEM$LAST_CHANGE_COMMIT_TIME` is metadata-only and does **not** require an active virtual warehouse. The smoke role was granted `OPERATE` on the warehouse specifically to settle this (previously blocked — see history below); deliberately `SUSPEND`ing the warehouse, calling the function, then re-checking `SHOW WAREHOUSES` showed it stayed `SUSPENDED`.
- The probe test (`test_snowflake_last_change_commit_time_does_not_require_active_warehouse`) now encodes this as a **hard assertion**, not an informational `pytest.skip` — a future Snowflake behaviour change fails this test loudly instead of needing a manual `-rs` flag to even notice. It still skips gracefully if `OPERATE` isn't granted (a legitimate, expected state — see below).
- Reconciled every place that had assumed the opposite (Snowflake carrying "a real, ongoing compute cost the other three don't"), written before this was confirmed: `dagster_drt/sensors.py` docstrings/error messages, `docs/guides/event-driven-syncs.md`, `integrations/dagster-drt/README.md`, `docs/research/warehouse-trigger-matrix.md`, `CHANGELOG.md`, `ROADMAP.md`, ADR 0004's amendment. `minimum_interval_seconds=` **stays required** for Snowflake — the reason changes from warehouse-compute cost to per-poll connection/auth overhead (`snowflake.connector.connect()` does a full handshake every tick, unlike Delta/Iceberg's object-storage reads).
- Scope note carried into the test docstring and docs: this is a claim about this one function on this one account, not a general "`SYSTEM$` functions are metadata-only" rule.
- Also adds a `concurrency` group to the Snowflake `dwh-smoke` job, serializing runs against the shared warehouse the probe manipulates.

### History (superseded by the above)
This PR originally shipped an inconclusive version — the smoke role lacked `OPERATE`, so the probe skipped with an access-control error, and the PR argued that was fine since the answer wouldn't change any documented decision. That argument turned out to be wrong: the answer does change the documented cost model. The user granted `OPERATE` to the smoke role specifically to resolve this, the probe ran clean, and this PR now reflects the confirmed result instead of the open question. (`OPERATE` should be revoked from the smoke role now that this is settled — tracked as a follow-up, not part of this PR's diff.)

## Test plan
- [x] `dwh-smoke.yml` dispatched against this branch with the assertion-based test and `OPERATE` still granted — `12 passed`, including `test_snowflake_last_change_commit_time_does_not_require_active_warehouse` (no skip).
- [x] `ruff check` clean on all touched files.


masukai:
**Codex review:** two real findings, both fixed in 20009c4:

- **[P2]** If the warehouse started `SUSPENDED` and the `SYSTEM$` call itself caused it to auto-resume — the exact "costly" outcome this probe exists to detect — `suspended_by_this_test` stayed `False`, so cleanup never re-suspended it, leaving it running and billing until Snowflake's own auto-suspend eventually kicked in. Fixed with a symmetric restore comparing against `original_state` in both directions, not just "undo a SUSPEND I issued".
- **[P2]** The outer cleanup only called `_drop_table()`, never `conn.close()` — leaked the probe's own Snowflake session for the rest of the smoke process. Fixed by closing `conn` in the same `finally`.

(Caught and fixed the syntax error from my own first attempt at the first fix — a stray double `finally:` — locally before ever pushing it; the CI run before this comment is against the prior, unaffected commit.)

masukai:
**Codex review (round 2):** the OPERATE check only ran when the warehouse started RUNNING — missing the normal nightly case where it's SUSPENDED between smoke runs. If the probe then triggered an auto-resume, a role without OPERATE would have had no way to undo it, and the earlier fix's \`except: pass\` around the restore silently swallowed that failure, leaving shared compute running and billing while the test reported a clean skip.

Fixed in bf04def: verify OPERATE up front via a net-zero round trip (toggle the warehouse's state and immediately toggle it back) regardless of starting state, before the real probe runs. Once that passes, the final restore is no longer swallowed on failure — a failure there despite the upfront check passing is a genuine surprise worth a loud test failure, not a silent pass.

masukai:
**Codex review (round 3):** three more real findings, all fixed in 53d0f27:

- **[P2]** If the OPERATE round trip's first `ALTER` succeeded but the second (undoing it) raised, the warehouse was left opposite \`original_state\` with no restore attempt at all — just skipped. Now attempts a best-effort restore before skipping, surfacing a second failure in the skip message rather than silently swallowing it.
- **[P2]** Overlapping \`dwh-smoke\` runs (nightly cron + manual \`workflow_dispatch\`, which this investigation triggered repeatedly today) could race against each other's warehouse SUSPEND/RESUME calls. Added a \`concurrency\` group to the Snowflake job so overlapping runs queue instead of racing.
- **[P2]** The \`SYSTEM$LAST_CHANGE_COMMIT_TIME\` call's exception handler treated *any* exception as a confirmed "requires an active warehouse" finding — misclassifying a transient network blip or unrelated error the same way. Removed the blanket catch entirely rather than guess at an unverified Snowflake error signature to match on: an unexpected failure now fails the test loudly with the real error instead of manufacturing an unverified conclusion.

Re-ran against the real account after this fix — 11 passed, 1 skipped (same access-control skip as before, behavior unchanged for the actual account state).

masukai:
**Codex review (round 4):** the OPERATE round trip's \`first_op_done\` tracking trusted client-side exception handling to know whether an ALTER actually applied — not reliable, since Snowflake can apply a statement server-side even if the client-side acknowledgment is lost to a timeout. A lost ack after a successful SUSPEND, for instance, would leave \`first_op_done\` False and skip without any restore attempt.

Fixed in 567ba8e by replacing the whole state-tracking approach: \`_ensure_suspended(cur, should_be_suspended)\` re-queries the *actual* warehouse state and only issues an ALTER if it doesn't already match, used everywhere this test changes state. Correct regardless of any ambiguity about a prior ALTER's acknowledgment. Also switched to boolean suspended/not-suspended checks throughout rather than comparing against a guessed non-suspended state string (an earlier draft used a fabricated `"RUNNING"` placeholder that was never verified against the real account).

PR #987, opened 2026-08-18
fix: tracked mirror on MySQL/Snowflake broke the no-DDL-privilege contract (#986)

### LINKED ISSUE #986: tracked mirror on MySQL: CREATE TEMPORARY TABLE (0.8.4, #694 part 2) breaks the documented "no DDL privilege" contract
## What happened

A production reverse ETL sync broke after upgrading drt-core 0.8.3 → 0.9.0. A `mode: mirror` + `strategy: tracked` sync into Cloud SQL MySQL 8.0 now fails at the destination stage on every run:

```
error:       (1044, "Access denied for user 'analytics-retl'@'%' to database 'bq_recruit'")
error_type:  OperationalError
error_stage: destination
```

The same DB user's other sync (`mode: full`, upsert only, no tracked mirror) succeeded in the same run window, so this is specific to the tracked-mirror path. The destination user is a least-privilege account with table-scoped grants only — set up following `docs/connectors/mysql.md`.

**Environment**: drt-core 0.9.0 (upgraded from 0.8.3), MySQL 8.0 (Cloud SQL), tracked mirror **without** `scope`.

## Root cause

`_finalize_mirror_tracked` in `drt/destinations/sql_base.py` issues:

```python
cur.execute(
    f"CREATE TEMPORARY TABLE {DIFF_STAGING_TABLE} "
    "(key_hash VARCHAR(64), key_json TEXT)"
)
```

This was introduced in **0.8.4** (#694 part 2, the Python→SQL diff move). v0.8.3 has no such statement — it did a plain `SELECT key_hash, key_json` and diffed in Python.

On MySQL, `CREATE TEMPORARY TABLES` is a **database-scoped privilege that cannot be granted per table**. A user lacking it gets error **1044 `Access denied ... to database`** — not 1142 `command denied ... for table` — which matches the observed error exactly.

## Why I'm filing this as a regression rather than a config error

`docs/connectors/mysql.md` (unchanged on v0.9.0) states:

> **`CREATE TABLE IF NOT EXISTS` still needs `CREATE` — MySQL checks the privilege *before* the existence check**, so pre-creating `_drt_synced_keys` alone did not historically avoid the grant. Since v0.8.x drt first probes `information_schema.tables` and **skips the CREATE entirely when the table already exists**, so an admin can pre-provision the state table (SQL above) and run the sync user with **no DDL privilege** — the sanctioned pattern for "no CREATE for app users" hardening.

That is precisely the pattern we deployed (#695). The required-privileges block in the same document still lists only:

```sql
GRANT SELECT, INSERT, UPDATE, DELETE ON `analytics`.`scores` TO 'retl_user'@'%';
GRANT SELECT, INSERT, DELETE ON `analytics`.`_drt_synced_keys` TO 'retl_user'@'%';
```

0.8.4 added a hard dependency on `CREATE TEMPORARY TABLES` on top of that, and:

- `docs/connectors/mysql.md` was not updated — 0.8.4's release notes list `docs/connectors/postgres.md` as the only doc touched by #694 part 2
- both the **0.8.4 and 0.9.0 release notes state "No breaking changes — drop-in upgrade"**

For a least-privilege deployment this is breaking, and it fails 100% of the time — but only on the first tracked-mirror run after the upgrade, which is what made it look like an unrelated infrastructure fault on our side.

Worth noting this is the same failure class #695 already fixed once, one privilege over: `CREATE` → `CREATE TEMPORARY TABLES`.

## Suggested fix

Fall back to the pre-0.8.4 Python-side diff when the staging table can't be created. The codebase already has this exact shape one function up, where a missing `ALTER` privilege is treated as a supported state (#890 scope columns):

```python
cur.execute("SAVEPOINT drt_scope_cols")
try:
    self._add_state_scope_columns(cur, state_ident)
except Exception:  # noqa: BLE001 — no ALTER privilege is a supported state
    cur.execute("ROLLBACK TO SAVEPOINT drt_scope_cols")
```

If a missing `ALTER` privilege is a supported state, a missing `CREATE TEMPORARY TABLES` privilege should be too: the staging table is a memory optimisation, and correctness doesn't depend on it. Degrading to the 0.8.3 read path keeps the documented "no DDL privilege" contract intact.

At minimum, `docs/connectors/mysql.md` (and `postgres.md`) should list `CREATE TEMPORARY TABLES` as required for tracked mirror, and the 0.8.4 notes should be corrected — the current text actively tells operators the upgrade is safe when it isn't.

## Three smaller things spotted while tracing this

1. **The operator-facing hint points the wrong way.** The failure surfaces as `error_suggestion: "Check the destination config (url, auth, headers)"`. For 1044/1142 the config is fine and the grant is not — this cost real debugging time.

2. **`SAVEPOINT` + `ROLLBACK TO SAVEPOINT` around `ALTER TABLE` is Postgres-shaped.** On MySQL, DDL causes an implicit commit, so the savepoint no longer exists when the rollback runs. It's harmless here (a failed `ALTER` changes nothing), but the protection isn't doing what the code reads as doing on MySQL.

3. **The state-table migration stance reversed without being called out.** 0.8.4 (#692) explains that scope values are derived positionally rather than stored in a new column, because "`_drt_synced_keys` is only ever `CREATE TABLE IF NOT EXISTS` with no migration path anywhere in this code — a required new column would silently never materialise on an already-deployed table". 0.9.0 (#890) then adds `ALTER TABLE ... ADD COLUMN scope_spec/scope_key`. That's a reasonable change of mind, but the release notes present it as a pure optimisation, so an operator reading them has no signal that the state table's DDL surface grew.

masukai:
## Summary

Fixes #986. `_finalize_mirror_tracked`'s SQL-side diff (#694 part 2, v0.8.4) issues an unconditional `CREATE TEMPORARY TABLE` for its staging scratch table on Postgres/MySQL (`sql_base.py`) and Snowflake (`snowflake.py`), even when the destination user only has the table-scoped grants #695's pre-provisioning pattern documents as sufficient.

- **MySQL**: `CREATE TEMPORARY TABLES` is database-scoped and cannot be granted per table → error 1044 on every tracked-mirror run for a least-privilege user. **This is the reported regression, fully fixed — Postgres/MySQL tracked mirror is no-DDL end to end again.**
- **Snowflake**: has no privilege for temp tables distinct from schema-level `CREATE TABLE` → same failure class at the finalize step, not previously reported but verified against Snowflake's own docs. Fixed at that step. ⚠️ **Does not make `sync.mode: mirror` no-DDL on Snowflake overall** — its write path independently and unconditionally stages every mirror write through `MERGE` via its own `CREATE TEMP TABLE` (#599, predates this regression by three releases). Caught in Codex review on this PR (see comments below); tracked separately as #988, not attempted here since it can't be safely verified without a real Snowflake account.
- Both MySQL and Snowflake's finalize-step issue contradict the "no DDL privilege" contract the connector guides documented, despite v0.8.4 and v0.9.0 both claiming a non-breaking, drop-in upgrade.
- **Databricks/ClickHouse needed no fix.** ClickHouse never stages via a temp table at all. Databricks' mirror DELETE (`_delete_via_staged_keys`) has unconditionally required `CREATE` on the target's catalog/schema since #707/#734 (v0.7.11), by design and documented in code — it was never part of the no-DDL contract, so #694 part 2's diff table added no new privilege requirement there.

## Fix

- Postgres/MySQL: staging `CREATE`+`INSERT` now run inside a `SAVEPOINT`, rolled back on failure (clears Postgres's poisoned transaction; harmless no-op on MySQL). Falls back to a client-side read-and-diff for **both** the stale-key read and the to-insert read (the second read site is easy to miss — a shallow fix that only patches the first query compiles and then crashes on that second one).
- Snowflake: identical fallback behind its existing autocommit `try/except` idiom (no savepoint needed — the connection autocommits). Reduces that one step's privilege floor; see the Snowflake caveat above for what it doesn't fix.
- The fast server-side path is byte-identical when the privilege is available; the incremental delete-stale/insert-new rewrite (including #890 scope handling + backfill) is untouched either way.
- `error_suggestion` now recognizes MySQL 1044/1142, SQLSTATE 42501, and Snowflake errno 3001 as privilege errors and suggests checking GRANTs instead of the generic "check the destination config" hint.
- `docs/connectors/{mysql,postgres,snowflake}.md` updated to describe the privilege as optional with a documented fallback. Snowflake's entry explicitly flags the remaining write-path gap rather than implying full no-DDL support.
- `CHANGELOG.md`: new `[Unreleased]` entry, plus retroactive ⚠️ cross-references inside the `[0.8.4]` and `[0.9.0]` entries that originally claimed "no breaking changes."

## Test plan

- [x] New `tests/integration/local_sql/test_tracked_mirror_no_ddl_smoke.py`: real Docker Postgres/MySQL containers, a privilege-restricted user with **no** `CREATE TEMPORARY TABLES`/`TEMPORARY` grant, two tracked-mirror generations run end-to-end (baseline insert, then a stale-row delete) — **ran locally against real containers, both pass** (not just mocked-cursor coverage, given this codebase's history with #908's "a mock cursor accepts any SQL string").
- [x] Unit tests for both failure sites (CREATE fails, INSERT fails) on both Postgres/MySQL and Snowflake, covering scoped + unscoped tracked mirror.
- [x] `make lint` (ruff + mypy): passed.
- [x] Full touched-module unit suite (184 tests): passed.
- [x] Codex review completed — one P1 finding (Snowflake write-path scope), resolved by narrowing this PR's claims and filing #988. See PR comments.

🤖 Implemented with [Claude Code](https://claude.com/claude-code) (Codex-assisted implementation, per this repo's Claude/Codex workflow)


masukai:
**Codex review:**

> The PostgreSQL/MySQL fallback is covered, but the advertised Snowflake no-DDL scenario remains unusable because the earlier mirror load path requires the same schema-level table-creation privilege.
>
> [P1] Handle Snowflake's load-time temp table in no-DDL mode — `drt/destinations/snowflake.py:742-743`
> For a non-empty Snowflake tracked-mirror sync whose role lacks schema-level `CREATE TABLE`, execution never reaches this fallback: `load()` forces mirror through the MERGE path and unconditionally runs `CREATE TEMP TABLE TMP_<TABLE>` at `snowflake.py:233`, which requires the same privilege. Consequently the least-privilege Snowflake configuration this patch documents as supported still fails before finalization; the write path also needs a no-DDL fallback, or Snowflake should be removed from this fix's claimed scope.

**Resolution:** Codex is right, and it changes what this PR can claim. `load()`'s `MERGE`-via-staging-table has needed schema-level `CREATE TABLE` since #599 (v0.7.7) — a pre-existing structural requirement, not something #694 part 2 introduced, and not something that can be safely fixed and verified inside this patch release (no local Snowflake account to smoke-test an inline-`VALUES` `MERGE` rewrite against, and this codebase has been burned before by mock-verified SQL that a real engine rejects — #908).

So: the finalize-side fix for Snowflake stays (it's a correct, real reduction in that step's privilege floor, symmetric with the Postgres/MySQL fix), but the PR description, CHANGELOG, and `docs/connectors/snowflake.md` are corrected to stop claiming Snowflake achieves no-DDL `sync.mode: mirror` end to end — it doesn't, and won't until the write path is addressed. Filed #988 to track that separately, with the Codex-review context and a possible fix shape (inline `VALUES` source instead of a staging table). Also caught in the same pass while re-checking scope: the CHANGELOG's original "Databricks wasn't affected" line was true but misleadingly reasoned — Databricks' mirror DELETE already unconditionally needs `CREATE` for an unrelated, pre-existing, by-design reason (`_delete_via_staged_keys`, #707/#734), not because it avoids a temp table the way ClickHouse does. Corrected too.

Issue #300, opened 2026-04-13
chore: Protocol stability review and freeze preparation


masukai:
## Summary

Review all Protocol interfaces (Source, Destination, StateManager) and prepare for v1.0 freeze.

## Why

v1.0 will freeze Protocol signatures with semver guarantees. Before that, we need to ensure the interfaces are right — because changing them after v1.0 requires a major version bump.

## Scope

- Audit all Protocol definitions for completeness
- Identify any missing methods needed for Enterprise features (RBAC hooks, audit hooks)
- Document each Protocol with clear contracts (input types, return types, error behavior)
- Write ADR: "Protocol Stability Policy" — what constitutes a breaking change
- Create deprecation workflow for future Protocol changes

masukai:
Closing — all scope items done across two PRs:

- **Audit all Protocol definitions for completeness** — audited all 17 `typing.Protocol` interfaces in the codebase (destinations, sources, secrets, state, engine).
- **Document each Protocol with clear contracts** — #992 adds `@runtime_checkable` to the 3 Protocols missing it (`SecretProvider`, `LimiterFactory`, `WatermarkStorage`) and `Raises:` documentation to every method that lacked it, corrected against actual implementation behavior (caught 3 docstring/implementation mismatches in Codex review before merge — see PR comments).
- **Write ADR: "Protocol Stability Policy"** — [ADR 0007](https://github.com/drt-hub/drt/blob/main/docs/adr/0007-protocol-stability-policy.md) (#993): defines what counts as a breaking Protocol change (structural typing has no default-method escape hatch, so signature changes are treated conservatively from both the caller's and the implementer's side), the sanctioned extension mechanism (optional-capability Protocols), and a freeze-scope table for all 17 Protocols (16 public/frozen at v1.0, 1 internal).
- **Create deprecation workflow** — in the same ADR: new capability ships as a separate Protocol rather than a wrapper method (a Protocol method body never executes for structural implementers, so in-place wrapping doesn't work the way it would on a class); removal requires both the 2-minor-version floor and a MAJOR bump post-v1.0, per `VERSIONING.md`.
- **Identify missing methods for Enterprise features (RBAC/audit hooks)** — explicitly deferred rather than built speculatively: #298/#299 (RBAC interface spec, audit log hooks) have no committed design yet, so this ADR names the amendment path (new optional-capability Protocols once those designs land) instead of inventing hook shapes ahead of them.

[...]

PR #992, opened 2026-08-18
fix(#300): consistent @runtime_checkable + documented error contracts across Protocols

masukai:
## Summary

First of two PRs closing #300 (Protocol stability review ahead of v1.0 freeze). This one is the mechanical half — no behavior change, decorators and docstrings only. The ADR ("Protocol Stability Policy") follows as PR 2 once this surface is consistent.

- Add `@runtime_checkable` to `SecretProvider`, `LimiterFactory`, `WatermarkStorage` — the 3 of 17 Protocols in the codebase missing it, inconsistent with structural siblings in the same subsystems.
- Add `Raises:` documentation to Protocol methods that had none, grounded in actual call-site behavior (verified `engine/sync.py` doesn't wrap `destination.load()` / `source.extract()` in try/except, so failures propagate and abort the sync — the docstrings now say so).
- Fix `HistoryStore.append`'s Protocol docstring to state the "best-effort, log at WARNING, never propagate" contract its implementations (`LocalHistoryManager`, `ObjectStoreHistoryStore`) already enforce but the Protocol never documented. Same fix for `DlqBackend.append`, with a note that `LocalDlqStore` still lets local I/O errors propagate (unlike the object-store backends).
- Document the `Source.test_connection` (`-> bool`, never raises) vs. `ConnectionTestable.test_connection` (`-> None`, raises) naming collision in both docstrings. Confirmed the two never share a call site (sources vs. destinations) — noted as intentionally frozen as-is, not something this PR refactors.

## Test plan

- [x] `ruff check` clean on all touched files
- [x] `mypy` clean on all touched files
- [x] `pytest tests/unit/test_state.py tests/unit/test_history.py tests/unit/test_dlq_store.py tests/unit/test_watermark*.py tests/unit/test_secret_provider*.py tests/unit/test_rate_limiter*.py tests/unit/test_sql_utils.py tests/unit/test_engine_observer.py` — 279 passed
- [x] Full `pytest tests/unit/` — 3303 passed, 3 skipped, 2 pre-existing failures in `test_cli_version.py` unrelated to this change (worktree path length affects `--version` line wrapping, same environment noise seen in #991)

### DIFF
--- a/drt/config/secret_providers/base.py
+++ b/drt/config/secret_providers/base.py
@@ -15,7 +15,7 @@ import os
 import threading
 import time
 from dataclasses import dataclass
-from typing import Any, Protocol
+from typing import Any, Protocol, runtime_checkable
 from urllib.parse import urlparse
 
 
@@ -95,6 +95,7 @@ def extract_key(raw: str, ref: SecretRef, *, scheme: str) -> str:
     return select_field(payload, ref, scheme=scheme)
 
 
+@runtime_checkable
 class SecretProvider(Protocol):
     """Resolves a parsed provider URI to a secret value.
 
--- a/drt/destinations/base.py
+++ b/drt/destinations/base.py
@@ -84,15 +84,22 @@ class Destination(Protocol):
     ) -> SyncResult:
         """Send a batch of records to the destination.
 
-        Row-level failures are reported via ``SyncResult.errors`` /
-        ``row_errors``, not raised — the engine has no per-row recovery for
-        an exception, so a batch that raises loses per-row error
-        attribution entirely.
+        Row-level failures are recorded in ``SyncResult.errors`` /
+        ``row_errors`` — implementations MUST populate these rather than
+        raise per-row, or the engine loses per-row error attribution
+        entirely.
 
         Raises:
             Exception: an unrecoverable, batch-level failure (connection
-                lost, auth rejected, API outage). The engine does not catch
-                this; it propagates and aborts the sync.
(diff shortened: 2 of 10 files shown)


masukai:
**Codex review:**

Codex reviewed this branch (`codex-companion.mjs review --wait --base main --scope branch`) and found 3 issues — all fixed in a follow-up commit ([c700c40](https://github.com/drt-hub/drt/commit/c700c40)):

- **[P2] `Destination.load`** — the docstring claimed row-level failures are "not raised," but 26 of the current SQL/API destinations (BigQuery is the reference shape) raise `RuntimeError` when `sync_options.on_error == "fail"` after populating `row_errors` — verified against `bigquery.py` + `test_insert_error_on_error_fail_raises`. Docstring now documents this.
- **[P2] `StagedDestination.finalize`** — the docstring claimed a single uniform "raises on job-level failure" contract, but the generic `staged_upload.py` destination catches everything into a failed `SyncResult` while `SalesforceBulkDestination` raises `RuntimeError` directly — the two genuinely diverge. Docstring now documents both shapes instead of picking one.
- **[P2] `Source.test_connection`** — the docstring claimed "never raises," but MySQL, Databricks, Snowflake, and SQL Server all close the connection in a `finally` block outside the preceding `except`, so a `close()` failure still propagates — verified against `mysql.py:161-163`. Docstring now notes callers already catch defensively for this reason.

All three were genuine docstring/implementation mismatches, not false positives. Fixed by correcting the docstrings to describe actual shipped behavior rather than an idealized uniform contract.

Issue #304, opened 2026-04-13
chore: v1.0 Protocol freeze — Source, Destination, StateManager


masukai:
## Summary

Officially freeze the Source, Destination, and StateManager Protocol interfaces at v1.0.

## What this means

- Protocol signatures become stable and covered by semver
- Any change to Protocol signatures requires a major version bump (v2.0)
- Deprecation policy: deprecated methods stay for at least 2 minor versions before removal
- Clear documentation of the stability guarantee

## Deliverables

- Update all Protocol docstrings with stability annotations
- Add "Stability: Stable" badges to API reference
- Publish stability policy in docs
- Announce in release notes

masukai:
Shipped in #1018 (merged). All 21 Protocol class docstrings now carry `Stability:` annotations, ADR 0007's freeze-scope table is corrected and complete (including 4 Protocols that had drifted out of sync — `QueryableDestination`, `RateLimiterBackend`, `PermissionChecker`, `AuditLogger`), and VERSIONING.md has a dedicated Protocol stability section. Documentation/annotations only — the freeze itself becomes binding when v1.0 actually ships (target 2026-11).

PR #993, opened 2026-08-18
docs(#300): ADR 0007 — Protocol stability policy

masukai:
## Summary

Second of two PRs closing #300 (Protocol stability review ahead of the v1.0 freeze, #304). Builds on #992 (the mechanical `@runtime_checkable` + docstring pass) — independent of it in git terms (no file overlap), so this can merge in either order.

Defines what #304 needs but doesn't itself specify:

- **What counts as a breaking Protocol change** — the core non-obvious point: adding a required method breaks every implementer immediately, since `Protocol` (structural typing) has no default-method escape hatch the way an ABC subclass would.
- **The sanctioned extension mechanism** — optional-capability Protocols (`ConnectionTestable`, `MatchPolicyCapable`, etc.), checked via `isinstance()`. The only way to add capability post-freeze without a major version bump.
- **Deprecation workflow** — codifies the back-compat alias pattern already used 3x in this codebase (`StateManager`/`HistoryManager`/`DlqStore`), with the removal window pinned to #304's already-stated "2 minor versions" promise rather than a new number invented here.
- **Freeze-scope table** for all 17 Protocols in the codebase — 16 public/frozen at v1.0, 1 internal (`ObjectClient`, underscore-private module, GCS/S3-only, not a public extension point today).
- Records the `Source.test_connection` (`bool`) vs. `ConnectionTestable.test_connection` (raises) naming collision as verified — confirmed no shared call site exists — and frozen as-is rather than unified.
- Explicitly defers RBAC/audit-hook Protocol shapes to #298/#299, which have no committed design yet. Names the amendment path rather than inventing hook shapes speculatively.

## Test plan

- [x] All referenced issue numbers (#298, #299, #297, #469, #921, #948, #304) verified against `gh issue view` for title/status accuracy
- Docs-only change, no code touched

### DIFF
--- a/docs/adr/0007-protocol-stability-policy.md
+++ b/docs/adr/0007-protocol-stability-policy.md
@@ -6,8 +6,11 @@
   [#304](https://github.com/drt-hub/drt/issues/304) (the v1.0 freeze itself).
 - **Relates to:** [#992](https://github.com/drt-hub/drt/pull/992) (the
   mechanical half of #300 — `@runtime_checkable` consistency and `Raises:`
-  documentation across all 17 Protocols, merged ahead of this ADR so it
-  describes a surface that's already consistent).
+  documentation across all 17 Protocols). **This ADR assumes #992 is merged
+  first** — until then, `SecretProvider`, `LimiterFactory`, and
+  `WatermarkStorage` still lack `@runtime_checkable`, and this ADR's
+  "16 public/frozen Protocols all `@runtime_checkable`" premise does not yet
+  hold. Land #992 before (or in the same merge window as) this ADR.
 - **Implementation:** none directly. This ADR sets the policy #304 enforces
   and the freeze-scope call each of the 17 Protocols below needs.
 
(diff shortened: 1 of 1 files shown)


masukai:
**Codex review:**

Codex reviewed this branch and found 4 issues — all fixed in a follow-up commit ([c94f6e1](https://github.com/drt-hub/drt/commit/c94f6e1)):

- **[P1] Compatibility table missed the implementer's side** — two "safe" rows (widening a parameter type, narrowing a return type) were judged only from the caller's perspective. Widening a parameter type actually breaks every existing structural implementer, since their method body was never written to handle the wider type — a real runtime `AttributeError`, not just a type-checker complaint. Replaced the row-by-row table with a single conservative rule: treat any signature change to an already-shipped Protocol method as breaking, in either direction.
- **[P1] "Wrapper method" deprecation step was incoherent for Protocols** — a `Protocol` method body never executes for a structural implementer, so adding a wrapper method to the Protocol itself does nothing for anyone already implementing it. Replaced with: leave the old Protocol untouched, ship the new shape as a separate Protocol, adapt on the caller side.
- **[P1] Removal condition was incomplete** — the ADR required only the 2-minor-version floor, but post-v1.0 `VERSIONING.md` also requires a MAJOR bump to remove a public Protocol method. Fixed to require both conditions explicitly.
- **[P2] ADR asserted #992 was already merged** — it isn't yet. Reworded the header and extension-mechanism section to state the dependency conditionally instead of as settled fact.

All four were genuine correctness issues in the ADR's reasoning, not false positives — worth the review given this document is meant to set policy other PRs will be held to.

PR #994, opened 2026-08-18
refactor(#469): replace _QUERYABLE_TYPES isinstance tuple with QueryableDestination Protocol

masukai:
## Summary

Replaces `query.py`'s hardcoded `_QUERYABLE_TYPES` config-class isinstance tuple with a `QueryableDestination` Protocol, backing `drt test` and `drt run --dry-run --diff`'s true-diff path. Postgres/MySQL/ClickHouse/Snowflake now implement `get_table_name()`/`execute_test_query()` on themselves; a new SQL destination becomes queryable by implementing these two methods, with no edit to `query.py` needed.

**Scoped down from the issue's original text** after reading the current code — worth flagging explicitly since the diff doesn't match the issue's stated migration plan:

- `query.py` has grown to 761 lines with 6 distinct dispatchers since #469 was filed (`is_queryable`/`get_table_name`/`execute_test_query`, plus `fetch_rows`, `fetch_rows_by_keys`, `fetch_tracked_state`, `fetch_all_keys`, `fetch_failing_rows`). The other five have deliberately non-uniform per-dialect coverage — e.g. ClickHouse is excluded from `fetch_rows_by_keys` for paramstyle reasons, and `fetch_all_keys` raises `NotImplementedError` for ClickHouse/Snowflake specifically so a mirror-delete preview never silently claims "deletes nothing." A single `fetch_existing()` Protocol as originally proposed can't express that matrix without erasing real, documented distinctions — left untouched.
- `drt/destinations/lookup.py` (named in the issue's migration plan) already uses `fetch_rows`, not the `fetch_existing(match_columns, select_columns)` shape the issue imagined — left untouched too.
- Landed as a **new, separate `QueryableDestination` Protocol** per [ADR 0007](https://github.com/drt-hub/drt/blob/main/docs/adr/0007-protocol-stability-policy.md)'s extension mechanism (merged today in #993), rather than adding `fetch_existing()` to `Destination` itself. `Destination.load()` was never touched, so this carries no breaking-change risk.

`is_queryable()`/`get_table_name()`/`execute_test_query()` in `query.py` keep their exact public signatures and behavior (now dispatching via `get_destination(config)` + `isinstance(dest, QueryableDestination)` instead of the isinstance tuple), so `test.py`, `diff.py`, and every existing test — including ones that monkeypatch these three functions directly — needed no changes. `_QUERYABLE_TYPES` was module-private (no deprecation cycle needed under `VERSIONING.md`) and is removed outright, along with the now-dead `_query_postgres`/`_query_mysql`/`_query_clickhouse`/`_query_snowflake` private helpers whose logic moved onto the destination classes.

Also amends ADR 0007's #469 consequence note (it anticipated a `Destination` shape change; this shipped as a new Protocol instead) and closes out the ROADMAP/CHANGELOG entries for #300 and #469.

## Test plan

- [x] `ruff check` / `mypy` clean on all touched files
- [x] New tests: `QueryableDestination` `isinstance` holds for all 4 SQL destinations and not for a non-SQL one (Slack); a fake destination implementing only the two Protocol methods becomes queryable with zero `query.py` changes (the architectural point of this issue); a destination implementing only one of the two methods correctly does not satisfy the Protocol
- [x] `pytest tests/unit/test_destination_query.py tests/unit/test_cli_build.py tests/unit/test_store_failures.py tests/unit/test_test_runner.py tests/unit/test_mcp.py tests/unit/test_cli_test_command.py tests/unit/test_diff.py tests/unit/test_cli_diff.py` — 253 passed
- [x] Full `pytest tests/unit/` — 3307 passed, 3 skipped, 2 pre-existing failures in `test_cli_version.py` unrelated to this change (worktree path length, same environment noise seen in #991/#992)


masukai:
**Codex review:**

Codex reviewed this branch and found 2 issues — both fixed in a follow-up commit ([bb09034](https://github.com/drt-hub/drt/commit/bb09034)):

- **[P2] `QueryableDestination`'s docstring overclaimed scope** — it said the Protocol backs both `drt test` and `drt run --dry-run --diff`'s true-diff path, but `compute_diff()` and `--store-failures` also call `fetch_rows`/`fetch_rows_by_keys`/`fetch_failing_rows`, which remain separately config-class-dispatched and weren't extended by this PR (deliberately — their per-dialect coverage genuinely differs, e.g. ClickHouse has no keyed fetch). A destination implementing only `QueryableDestination` gets full `drt test` support but degrades gracefully for the other two — sample-mode diff or a "could not store failure sample" warning, not a crash (verified: `diff.py`'s broad `except Exception` and `test.py`'s `try/except` around `_store_or_clear_failure_sample` were already there). Fixed by correcting the Protocol docstring, `query.py`'s internal docstring, and the CHANGELOG entry to state the scope precisely.
- **[P2] Two `# type: ignore[no-any-return]` suppressions violated `AGENTS.md`'s "no `type: ignore` except for external-library issues" rule** — `_queryable_destination()` returned `Any` instead of a properly typed `QueryableDestination | None`. Fixed by typing it correctly (the actual fix, not a suppression), which also required moving its two lazy imports to module level (verified no circular import).

Also fixed in the same commit, found via `codecov/patch` failing at 54.5%: the new `execute_test_query()`/`get_table_name()` methods on Postgres/MySQL/ClickHouse had no direct test coverage at their new location (moved from module-private functions in `query.py`, which also had no direct tests — only reached indirectly through the four original destinations). Added 6 new tests mirroring the existing Snowflake test's mocked-connection pattern.

Issue #299, opened 2026-04-13
design: audit log hooks (Enterprise boundary)


masukai:
## Summary

Design audit log hooks that Enterprise can use to capture sync execution events for compliance.

## Why

Audit logging is a compliance requirement for Enterprise customers (SOC2, HIPAA). The hook needs to exist in OSS core so Enterprise can attach listeners without forking.

## Scope (design only)

- Define audit event types: sync_started, sync_completed, sync_failed, config_changed, secret_accessed
- Hook interface: `AuditLogger` Protocol with `log_event(event)` method
- OSS default: no-op logger
- Enterprise: sends events to external systems (Datadog, CloudWatch, SIEM)
- Document in ADR format

masukai:
Closing — landed in [PR #995](https://github.com/drt-hub/drt/pull/995) together with #298, both design-only per this issue's explicit scope. Full rationale in [ADR 0008](https://github.com/drt-hub/drt/blob/main/docs/adr/0008-enterprise-boundary-rbac-and-audit-hooks.md).

**Checked the 5 proposed event types against the existing `SyncObserver` Protocol before designing anything new.** Three of them are already `SyncObserver` callbacks wearing a different name:

- `sync_started` → `on_sync_started`
- `sync_completed` → `on_sync_completed`
- `sync_failed` → not a separate callback; derived from `on_sync_completed(result)`'s `result.failed`/`result.errors`, matching every other `SyncResult` consumer, rather than adding a required method to an already-shipped Protocol (breaking under [ADR 0007](https://github.com/drt-hub/drt/blob/main/docs/adr/0007-protocol-stability-policy.md))

A new `register_extra_observer()` (`drt/engine/observer.py`) lets an Enterprise audit implementation register as a plain `SyncObserver`, fanned out to by the existing `CompositeObserver` — no new Protocol needed for these three.

**The two genuinely new events** — `config_changed` and `secret_accessed` — have no `SyncObserver` home (they aren't sync-lifecycle events) and get a new, narrow `AuditLogger` Protocol (`drt/observability/audit.py`), matching this issue's proposed `log_event(event)` shape:

- `config_changed` fires from both `load_syncs()` and `load_project()` (added the `load_project()` half after Codex review caught `drt status` and other direct callers missing it)
[...]

Issue #298, opened 2026-04-13
design: RBAC interface specification (Enterprise boundary)


masukai:
## Summary

Design the interface for Role-Based Access Control that will be implemented in drt Enterprise.

## Why

RBAC is identified as a key Enterprise-only feature in the monetization strategy. The interface needs to be designed now so the OSS core has the right hooks, even though the implementation lives in Enterprise.

## Scope (design only, not implementation)

- Define permission model: who can run/edit/view which syncs
- Hook points in CLI and engine where auth checks are injected
- Interface/Protocol that Enterprise implements
- OSS default: no-op (all permissions granted)
- Document the boundary clearly in ADR format

## Not in scope

- Actual RBAC implementation (that's Enterprise)
- User management UI

masukai:
Closing — landed in [PR #995](https://github.com/drt-hub/drt/pull/995) together with #299, both design-only per this issue's explicit scope. Full rationale in [ADR 0008](https://github.com/drt-hub/drt/blob/main/docs/adr/0008-enterprise-boundary-rbac-and-audit-hooks.md).

**What shipped:** `drt/security/base.py` — a `PermissionChecker` Protocol (`check(action, sync_name, principal=None)`) with exactly this issue's three-verb model (`run`/`edit`/`view`), `PermissionDeniedError`, and `AllowAllPermissionChecker` as the OSS default — every action permitted, unconditionally. `CLAUDE.md`'s "no RBAC" line stays true of OSS's actual behavior; this doesn't enable enforcement anywhere until a separately-installed Enterprise package calls `register_permission_checker()`.

**Hook points** (wired at 3 representative CLI sites, not all 19 commands — the pattern is what's being proven, extending coverage is mechanical follow-up once an Enterprise consumer exists): `drt run` (action=`run`), `drt state reset` (action=`edit`), `drt status` (action=`view`).

**Registration** mirrors `SecretProvider.register()`'s existing pattern exactly rather than inventing something new — same module-registry, `register()`/`get()` shape.

Flagged `CLAUDE.md`'s "Do not add RBAC or multi-tenancy" line to the user before starting; confirmed as an intentional, scoped exception for this Enterprise-boundary *design* work rather than a contradiction resolved silently.

`PermissionChecker` also joins ADR 0007's freeze-scope table: public, frozen at v1.0.

PR #995, opened 2026-08-19
design(#298,#299): Enterprise boundary — PermissionChecker + AuditLogger, ADR 0008

masukai:
## Summary

Design-only work for #298 (RBAC interface spec) and #299 (audit log hooks), both explicitly scoped as "design only, not implementation" in their issue bodies. Full rationale in [ADR 0008](docs/adr/0008-enterprise-boundary-rbac-and-audit-hooks.md).

**Note on `CLAUDE.md`:** this repo's own instructions say "Do not add RBAC or multi-tenancy — small team / personal use." Flagged to the user before starting this work; confirmed as an intentional, scoped exception for Enterprise-boundary *design* — the issues are explicitly "design only," ROADMAP v0.10 schedules them, and OSS behavior doesn't actually change (the default checker permits everything).

**RBAC (#298):** `drt/security/base.py` — `PermissionChecker` Protocol (`check(action, sync_name, principal=None)`, actions `run`/`edit`/`view` — #298's exact three-verb model), `AllowAllPermissionChecker` (OSS default), `register_permission_checker`/`get_permission_checker`. Wired at three representative CLI hook points (`drt run`, `drt state reset`, `drt status`) rather than all 19 commands — the pattern is proven here, extending coverage is mechanical follow-up once an Enterprise consumer exists.

**Audit hooks (#299):** checked #299's five proposed event types against the existing `SyncObserver` Protocol (`drt/engine/observer.py`) before designing anything new. Three of them (`sync_started`, `sync_completed`, `sync_failed`) are already `SyncObserver` events wearing a different name — inventing a second Protocol would duplicate an existing seam. A new `register_extra_observer()` lets an Enterprise audit implementation register as a plain `SyncObserver`, fanned out to by the existing `CompositeObserver`. Only `config_changed` (fires once per `load_syncs()`) and `secret_accessed` (fires on an actual provider fetch, never logging the resolved value) have no `SyncObserver` home — those get a new, narrow `AuditLogger` Protocol in `drt/observability/audit.py`, scoped to exactly those two events.

**Registration** mirrors `SecretProvider.register()`'s existing pattern exactly rather than inventing a new mechanism. How a third-party package's registration call gets triggered without an explicit operator-side import is the same open gap `SecretProvider` already has — left to #297 (plugin discovery), not solved here.

Both new Protocols join [ADR 0007](docs/adr/0007-protocol-stability-policy.md)'s freeze-scope table, public and frozen at v1.0 — this is that ADR's own named follow-up ("amend once #298/#299 land committed designs").

## Test plan

- [x] `ruff check` / `mypy drt/` clean (194 source files)
- [x] 14 new tests: registry replace-not-error semantics + OSS defaults for both new Protocols; `register_extra_observer`'s cumulative (not replace) semantics + `CompositeObserver` fan-out; `config_changed` firing exactly once per `load_syncs()` (not per file, not from `project_vars`'s best-effort load); `secret_accessed` firing with no secret value in the event details; a CLI-level deny/allow pair proving `drt status` is actually gated end-to-end, not just that `PermissionChecker.check()` works in isolation
- [x] Full `pytest tests/unit/` — 3331 passed, 3 skipped, 2 pre-existing failures in `test_cli_version.py` unrelated (worktree path length, same environment noise seen in #991/#992/#994)


masukai:
**Codex review:**

Codex reviewed this branch and found 3 issues (2 P1) — all fixed in a follow-up commit ([90eb8a0](https://github.com/drt-hub/drt/commit/90eb8a0)):

- **[P1] Deadlock risk** — `secret_accessed`'s audit call ran while `_value_cache_lock` was held, in the cache-miss path. An Enterprise `AuditLogger.log_event()` that itself resolves a secret (e.g. sink credentials) via `resolve_provider_uri()` would deadlock re-acquiring that non-reentrant lock. Fixed by restructuring `resolve_provider_uri` so `provider.fetch()` still runs inside the lock (needed for correct cache population) but the audit emission happens after the lock is released. New regression test runs the nested-resolution scenario on a background thread with a bounded join, so a reintroduced deadlock fails the test instead of hanging the suite.
- **[P1] Fire-and-forget not actually enforced** — `AuditLogger.log_event`'s docstring documented a "must not raise" best-effort contract, but nothing enforced it at the call sites: a raising logger could break the operation being audited (an unreachable audit sink failing secret resolution, or turning successful config parsing into a CLI error). Fixed by wrapping both call sites in try/except + warning, matching `DlqObserver`'s existing fire-and-forget pattern in this codebase — the contract needed the call site to enforce it, not every future `AuditLogger` implementation to volunteer it.
- **[P2] Missing hook point** — ADR 0008's own table named both `load_syncs` and `load_project` as `config_changed`'s trigger points, but only `load_syncs` actually fired it, so `drt status` and other direct `load_project()` callers never got the event. Fixed by adding the same hook to `load_project()`, distinguished from `load_syncs`'s event via a new `source` detail field.

All three were genuine issues, not false positives. 5 new regression tests added, including a background-thread deadlock test that fails loudly (bounded join) rather than hanging the suite if the fix regresses.

PR #996, opened 2026-08-20
Entry-point plugin discovery for non-connector registries (#297)

masukai:
## Summary

Closes the "requires an explicit operator-side import" gap [ADR 0008](docs/adr/0008-enterprise-boundary-rbac-and-audit-hooks.md) named for four of drt's registries — `SecretProvider`, `PermissionChecker`, `AuditLogger`, and extra `SyncObserver`s. A third-party package exposing one of six `drt.*` entry-point groups is now discovered and its registration callable invoked automatically at CLI startup, via `importlib.metadata.entry_points()` (`drt/plugins.py`). New `drt plugins list` command (`--format table|json`) reports what was discovered, per entry: group, name, package, version, author, load status.

## The blocker this PR does NOT solve — [ADR 0009](docs/adr/0009-plugin-config-union-blocker.md)

While scoping this issue I found a real architectural blocker: `drt.sources`/`drt.destinations` entry points *are* discovered and invoked, and *do* register in `drt/connectors/registry.py` — but `SyncConfig.destination` (`drt/config/sync_options.py:588`) and `load_profile()` (`drt/config/credentials.py:244`) both validate the `type` field against a closed, hand-enumerated set *before* the connector registry is ever consulted. A sync YAML naming a third-party destination type fails `drt validate` outright — the exact `pip install drt-salesforce-premium` scenario in #297's issue text.

Fixing that means replacing a Pydantic discriminated union (destinations) and a hand-written if/elif dispatch (profiles) with a registry-driven validator — a config-layer redesign with blast radius across `drt/config/query.py`'s six dialect dispatchers and every `isinstance()` narrowing call site in `drt/destinations/`, not something that belongs in a plugin-discovery PR. ADR 0009 documents the blocker precisely, sketches three candidate designs, and states none is chosen. Filing the follow-up issue for that decision separately.

`drt plugins list` reflects this honestly: connector entries are labeled `registered (not yet usable in sync YAML — see ADR 0009)` rather than implied to work.

## What ships here

- `drt/plugins.py`: `discover_plugins()` / `load_plugins()`, six groups (`drt.sources`, `drt.destinations`, `drt.secret_providers`, `drt.permission_checkers`, `drt.audit_loggers`, `drt.observers`). A broken plugin's exception is caught and recorded on its own entry, not propagated — one bad third-party package can't take down the CLI.
- `drt plugins list` (`drt/cli/commands/plugins.py`), wired into `drt/cli/commands/__init__.py`.
- `load_plugins()` called once from the `@app.callback()` in `drt/cli/main.py`, idempotent per process.
- `docs/guides/plugins.md`: third-party plugin development guide, cross-linked from `docs/guides/building-a-destination.md`.
- `scripts/check_cli_mcp_parity.py`: classified `plugins list` as `UNMAPPED_COMMANDS` — a diagnostic for plugin authors/operators, not an agent-orchestration primitive.
- Verified end-to-end against a **real pip-installed package** (scratch `fake-drt-plugin` with a genuine `[project.entry-points]` table), not just constructed `EntryPoint` objects in tests — confirmed real `pip install` → `entry_points()` discovery → callable invocation → `drt plugins list` all worked, then uninstalled.

## Test plan

- [x] `ruff check drt/ tests/unit/test_plugins.py scripts/check_cli_mcp_parity.py` clean
- [x] `mypy drt/` clean (196 files)
- [x] `pytest tests/unit/` — 3348 passed, 2 pre-existing unrelated failures (long worktree path wraps `--version` output across extra lines), 3 skipped
- [x] Manual end-to-end against a real installed plugin package (see above)


masukai:
**Codex review:**

The plugin listing command reloads every plugin after CLI startup already loaded them, breaking accurate status reporting for registries that reject duplicate registration.

- [P2] Reuse cached plugin results in the list command — `drt/cli/commands/plugins.py:35`
  Every `drt plugins list` invocation first runs `load_plugins()` in the root callback, then this `force=True` call invokes all registration functions a second time. Secret-provider and connector registrars reject duplicate registration, so valid plugins are falsely reported as errors, while observer plugins are appended twice; use the cached result instead of forcing a reload.

**Fixed** in f7e76e0 — `list_plugins()` now calls `load_plugins()` without `force=True`, reusing the cache the root `@app.callback()` already populated. Added a regression test (`test_cli_plugins_list_reuses_the_startup_callback_cache_not_a_second_load`) that fails against the pre-fix code (verified: reverting the fix locally reproduces `['ok', 'ok']` instead of `['ok']`).

Pawansingh3889 (approved):
Reviewed locally on feat/297-plugin-system @ 5c267e8.

Verified:
- 16/16 tests in tests/unit/test_plugins.py pass; ruff clean; mypy clean; scripts/check_cli_mcp_parity.py reports no gaps.
- `drt plugins list` and `--format json` work end to end (empty-state + JSON payload confirmed live).
- The Codex-flagged double-invoke bug is fixed: list_plugins() reuses the startup-callback cache (no force=True), covered by test_cli_plugins_list_reuses_the_startup_callback_cache_not_a_second_load.
- load_plugins() is idempotent (lock + cached result) and isolates a broken plugin's exception per-entry rather than propagating it.
- Docs (plugins.md, ADR 0009) honestly describe the drt.sources/drt.destinations sync-YAML limitation.

Non-blocking nit: in drt/plugins.py load_plugins(), a successfully-loaded but non-callable target is still marked loaded=True. Entry points in these groups are contractually callables, so it's an edge case, but a non-callable entry would show as 'loaded' rather than 'error'.

Approve — CI green and behaviour is correct. The ADR 0009 config-union follow-up (#997) is appropriately out of scope for this PR.

masukai:
Thanks for the thorough local verification, @Pawansingh3889 — appreciate you checking the Codex fix's regression test and the ADR 0009 scoping too.

Fixed the non-callable-target nit in 792ab86: a non-callable entry-point target now raises into the existing per-entry error handling instead of silently reporting `loaded=True` with nothing actually invoked. Added a regression test (`test_load_plugins_reports_a_non_callable_target_as_an_error`).

PR #998, opened 2026-08-20
fix(serve): prevent HMAC GET replay by signing request path

### LINKED ISSUE #936: fix(serve): GET /runs/<id> under --auth hmac is replayable, since the signature covers an empty body
Split out of #902, where this was reviewed, understood, and deliberately shipped as a documented caveat rather than a blocker. Filing it so the decision is recoverable rather than buried in a merged PR thread.

### The gap

Under `--auth hmac`, a `GET` has no body to sign, so its signature is computed over the empty body. That makes it a constant per secret rather than a per-request value. Anyone who obtains that one header can read `GET /runs/<id>` for any run id they know, for as long as the secret lives.

### Why it shipped as-is

Recording the reasoning from #902 so it does not get relitigated from scratch:

1. It strictly improves on what shipped before, which was no auth at all on `GET`. A caveat on a real improvement is not a reason to withhold the improvement.
2. Rejecting polling under `hmac` would push operators who chose HMAC (to match GitHub or Shopify webhook verification) into running a second, different auth scheme just to poll status. A mixed-scheme deployment is a worse footgun than a documented single-scheme caveat.
3. The blast radius is narrower than "the read API is open". Exploiting it needs a header-level capture, which the reverse-proxy-access-log threat model does not give you (those log the URL path, not headers, so this needs a MITM or a header-dumping proxy), **and** separately knowing a target run id, which is still an unguessable uuid4.

### What a real fix looks like

Timestamp plus nonce in the signed payload, the way SigV4 and most webhook-with-polling designs do it. That means a new header or two, clock-skew tolerance, and a nonce cache, which is more than ADR 0004's Tier 3 "hardened enough" scope called for in #902.

### When to pick this up

When polling under `hmac` sees real use. Until then `bearer` is the documented recommendation for the polling path, and the guide already carries that caveat alongside the `openssl` one-liner for computing the empty-body signature.

Not urgent, and deliberately not milestoned.

Pawansingh3889:
Under --auth hmac, GET /runs/<id> had no body so its signature was constant per secret. Now the request path is prepended to the empty body with a null-byte separator, making each signature unique.

Closes #936


masukai (changes requested):
Thanks for picking this up — the core idea (bind the GET signature to the request path so a leaked header can't read arbitrary run ids) is the right fix for #936, and I confirmed it actually works: replaying a signature computed for one run id against a different run id correctly gets rejected.

But this isn't ready to merge as-is — three things:

**1. `CLAUDE.md` got unintentionally mangled.** 32 of its 33 `→`/`—`/`↓` characters were flattened to ASCII (`->`, `--`, and — worse — the architecture diagram's `↓` became a bare `\`, which breaks the diagram). Nothing about this PR should be touching that file at all; please drop that hunk entirely.

**2. This breaks existing HMAC clients triggering syncs.** The new branch keys off "is the body empty," not "is this a GET." `POST /sync/<name>` is documented as the normal, bodyless way to trigger a sync — under this change, a client that correctly signs the empty POST body the old way now gets 401'd. I reproduced this directly: a signature computed as `HMAC(secret, b"")` for an empty-body POST is rejected once the server starts expecting `HMAC(secret, path + b"\0")` instead. The path-binding needs to be scoped to GET specifically, not inferred from an empty body.

**3. Docs are now wrong in three places** — `docs/guides/using-webhook-trigger.md` (the `curl`/`openssl` example literally signs the empty body and would 401 after this merges), `docs/llm/CONTEXT.md`, and this module's own docstring. All three need to describe the new `path + \0` scheme, and the guide's example command needs to actually work.

Once those are sorted, a test asserting the actual exploit scenario (signature for run A rejected against run B) would be good to have — I checked this by hand and it's not covered today. Also needs a CHANGELOG entry.

masukai:
One more, more serious than what I flagged before — surfaced while reviewing #1014 (which touches the same `_check_auth` block), and I reproduced it directly.

**A captured GET signature can be replayed to authenticate an arbitrary POST.** `_verify_hmac` only special-cases the *empty*-body case (`signed_body = path.encode() + b"\x00"`), so it doesn't domain-separate "signature over a GET's synthetic path payload" from "signature over a POST's real body." If an attacker captures the signature for `GET /runs/<known-or-guessed-id>`, they can replay it on a `POST /sync/<any-name>` whose body they craft to literally be `<that path>\x00` bytes — same digest, same signature, accepted. I reproduced this: a signature for `GET /runs/xyz` verifies successfully against a crafted `POST /sync/some_other_sync` with body `b"/runs/xyz\x00"`.

That's worse than the original #936 gap this PR fixes — it turns a leaked read-only signature into the ability to trigger an arbitrary sync, not just read one run. The path needs to be bound into the signed payload in a way that can't collide with a real request body — e.g. a fixed, unambiguous prefix/separator that a genuine POST body can never start with (or just don't reuse the same signing scheme's "body" slot for two different kinds of payload).

Pawansingh3889:
Thanks for both reviews, and particularly for catching the replay on the second pass. All four points are addressed in `7adeee1`, plus the revert in `93ea67f`.

**1. CLAUDE.md.** Dropped entirely. The reformatting had leaked in from a separate branch by accident. Restored to the version this branch started from, so the file no longer appears in the diff at all.

**2. Bodyless POST.** You were right that keying off an empty body was wrong. The path is now passed explicitly by `do_GET` and by nothing else, so `POST /sync/<name>` still signs its raw body, including when that body is empty. There is a regression test pinning that contract.

**3. Docs.** Updated in all three places: the module docstring, `docs/llm/CONTEXT.md`, and `docs/guides/using-webhook-trigger.md`. The guide now documents both the POST and the GET example, and I verified each produces a signature the server actually accepts. Worth flagging that the GET example needs the `openssl mac` subcommand: the more obvious `openssl dgst -sha256 -mac HMAC -macopt key:...` form silently yields an empty key on OpenSSL 3.5, so it looks correct and produces a signature that never verifies. CHANGELOG entry added under `[Unreleased]`.

**4. The replay.** This is the one that changed the design. I could not close it with a prefix or separator, because the attacker chooses the POST body, so any deterministic GET payload can simply be sent as a body. The two payloads have to be signed under different keys. GET signatures are now computed under `HMAC(secret, "drt/serve/v1/get-path")` rather than the raw secret, so a GET digest and a POST digest cannot collide regardless of what body is crafted. POST keeps signing the raw body under the secret itself, so GitHub and Shopify senders are unaffected.

Reproduced both failures against the previous commit before fixing, since you asked for that kind of evidence last time:

| | before | after |
|---|---|---|
| correctly signed bodyless POST accepted | **false** | true |
| captured GET signature replayed onto `POST /sync/other` | **true** | false |       
                                                                                                
Both are now tests rather than one-off checks: a signature for run A rejected against run B, and a captured GET signature rejected against `POST /sync/<other>` for every body shape crafted from the signed path (`path`, `path + \0`, `\0 + path`). 34 tests in ` and the branch merges cleanly against current main.

One open question for you: the derived-key constant is versioned (`drt/serve/vn that you would rather bump the string than change the scheme in place if this ever needs revisiting. Say if you would prefer it unversioned.

masukai:
Independently reviewed and verified — this is solid, careful security work.

- Confirmed `CLAUDE.md` is completely clean in the current diff.
- Ran the full HMAC test suite myself (`test_hmac_get_signs_the_request_path`, `test_hmac_get_signature_does_not_open_a_different_run`, `test_hmac_get_signature_cannot_be_replayed_as_a_post`, `test_hmac_bodyless_post_still_signs_the_empty_body`) — all 9 HMAC-related tests pass.
- Re-derived the key-separation math independently in Python and confirmed it matches: `POST` still verifies under `HMAC(secret, raw_body)` (checked `do_POST`'s call site never passes `path=`), `GET` verifies under `HMAC(HMAC(secret, "drt/serve/v1/get-path"), path)`. The design is correct — separate *keys*, not just a separator, which is exactly right given the threat model (attacker controls the entire POST body, so no fixed prefix/separator can be made unreachable).
- Test-merged current `main` locally to check for conflicts (clean) and ran the full unit suite against the merged result — 3499 passed, 0 regressions.
- Tried to run the guide's new `openssl mac` example locally but only have LibreSSL here (no `mac` subcommand) — couldn't verify the exact command, but the underlying HMAC math it's meant to produce is independently confirmed correct via the Python re-derivation above and the passing test suite, so I'm not concerned about it.

On your open question: **please keep the derived-key constant versioned.** `drt/serve/v1/get-path` costs nothing now and means any future change to this scheme (should one ever be needed) can mint `v2` without ambiguity about which requests were signed under which scheme — cheap insurance, no downside.

This looks ready to merge from my side. Thank you for the thorough fix and the reproduction table — that's exactly the kind of evidence that makes a security patch easy to trust.

Issue #301, opened 2026-04-13
feat: performance profiling — I/O vs CPU bottleneck analysis for Rust migration


masukai:
## Summary

Profile drt sync execution to identify where time is spent: I/O (HTTP, DB queries) vs CPU (data transformation, serialization).

## Why

Strategy doc says: "ボトルネックは I/O（HTTP, DB）で CPU 処理ではない" — this needs to be verified with data before committing to Rust migration. If 95% of time is I/O, Rust rewrite has minimal benefit.

## Scope

- Use py-spy or cProfile on benchmark scenarios (#280)
- Measure: source extract time, transformation time, destination load time, serialization overhead
- Identify: which components would benefit from Rust (PyO3)
- Output: profiling report with recommendations
- Document findings in ADR: "Rust Migration Decision"

masukai:
Closed via #1004 (merged). `make profile` — cProfile on #280's benchmark scenarios. Finding: this local, no-network workload is CPU-majority (84.46%-91.88%), refuting the bare "I/O, not CPU" claim for this shape — but the ADR is explicit that this doesn't settle the production question (real network-bound SaaS/warehouse destinations would shift the balance back toward I/O), and the CPU hotspot (JSON serialization) isn't in `engine/sync.py` itself. Codex review caught two real methodology bugs (in-memory SQLite extraction mislabeled as I/O, `os.makedirs` missing from the I/O bucket) that required a full re-measurement, not just relabeling. ADR 0010 recommends against using this data alone to justify a broad Rust rewrite and defers the final call to the repo owner.

PR #999, opened 2026-08-21
Add local sync-performance benchmark suite (#280)

masukai:
## Summary

- `make benchmark` runs fixed 100 / 10,000 / 100,000-row scenarios through the real `SQLiteSource` → `engine.run_sync()` → JSONL `FileDestination` path — entirely local, no credentials, no network, no optional connector dependency.
- Each scenario measures wall-clock duration, rows/sec, `tracemalloc` peak Python memory, and destination-call count (API-call proxy), written as a versioned, JSON-Schema-documented artifact under gitignored `benchmarks/results/`.
- `execute_scenario()` is a deliberately unmeasured, importable seam — #301 (I/O vs CPU profiling for the Rust-migration decision) can wrap the identical workloads with a profiler rather than rebuilding scenario setup.
- **Local/manual tooling only** — no CI wiring, no `drt benchmark` end-user command. Deferred until there's measured need, same reasoning ROADMAP.md already applies to #948.

Implemented by Codex (task-mode, `--write`), orchestrated in an isolated worktree and independently re-verified here — not just Codex's own self-report.

## Verification

- [x] `ruff check drt scripts tests` clean
- [x] `mypy drt` clean (196 files)
- [x] `pytest tests/unit/` — 3358 passed, 2 pre-existing unrelated failures (long worktree path wraps `--version` output), 3 skipped
- [x] `make benchmark` run manually end-to-end, produced valid JSON results matching the documented schema
- [x] Confirmed `benchmarks/results/*.json` is gitignored and never gets staged (`git add -A -n` dry run)

Note: Codex's own sandboxed execution environment blocks localhost socket binding, which self-reported as "31 failures / 67 errors" during its own verification pass — confirmed those are sandbox artifacts, not real regressions, by re-running the full suite myself outside that sandbox (clean result above).


masukai:
**Codex review:**

The benchmark's multi-batch scenarios do not preserve their full destination output, undermining the suite's primary measurements. Configured global OpenTelemetry can also introduce network traffic and inconsistent measurement overhead.

- [P1] Preserve all batches in the benchmark output — `benchmarks/harness.py:116-117`
  For the medium and large scenarios, this delegates every 100-row batch to `FileDestination.load()`, whose JSONL writer opens the same path with mode `"w"`; each call therefore truncates the preceding batches. The benchmark reports 10,000/100,000 rows synced while retaining only the final 100, and it measures repeatedly overwriting the same small file rather than writing the full destination artifact. Use batch-preserving destination semantics and verify the persisted row count.

- [P2] Isolate benchmarks from configured OpenTelemetry exporters — `benchmarks/harness.py:160-164`
  When OpenTelemetry extras and an endpoint in `~/.drt/profiles.yml` or `OTEL_EXPORTER_OTLP_ENDPOINT` are present, `run_sync()` initializes the global exporter and emits per-batch spans, causing network activity and contaminating the measured time and memory despite the harness's local/no-network contract. Even on the no-op path, only the first selected scenario pays lazy telemetry initialization, so results differ depending on scenario selection; force a known no-op telemetry state or isolate scenarios before timing.

**Fixed** in 5832735 — both verified real before fixing (confirmed `FileDestination`'s truncate-per-call behavior directly, confirmed `run_sync()`'s unconditional `get_tracer()` call and OTel's lazy global init).

- P1: `CountingFileDestination` now writes JSONL directly (append after first write) instead of delegating to `FileDestination`, sidestepping that class's truncation bug entirely — filed separately as #1002 since it's a real, pre-existing production bug affecting any `type: file` sync with `batch_size` < row count, not something to fix inside this benchmark PR. `measure_scenario()` now verifies the persisted line count against the expected row count and raises if they don't match.
- P2: pinned the process to the fallback no-op tracer/meter before any scenario runs.

Independently re-verified beyond the fix's own self-report: ran `make benchmark` myself (all three scenarios completed, the new row-count check didn't raise), then re-ran it with `OTEL_EXPORTER_OTLP_ENDPOINT` pointed at an unroutable address (completed in 2.1s, no hang, no timing change from baseline).

Issue #781, opened 2026-07-10
feat(docs): dbt exposures export — drt syncs visible in dbt docs lineage (--format dbt-exposures)


masukai:
## Problem

drt reads dbt's manifest (`ref('mart_users')` resolution #239, `drt init --from-dbt` #215) — but gives nothing back. In dbt docs, marts consumed by drt look like dead ends: an analytics engineer deprecating `mart_users` has no signal that three reverse-ETL syncs feed HubSpot from it. dbt built exactly the primitive for this — **exposures** — and downstream tools (Lightdash, Metabase exporters, Fivetran) publish them so they appear in dbt lineage and `dbt ls --select exposure:*`.

This is also a zero-cost ecosystem/GTM wedge: drt syncs rendering inside every dbt docs site is discovery in the exact community we target (#115/#117).

## Prior art

- **dbt exposures**: YAML nodes (`type`, `depends_on: [ref(...)]`, `owner`, `url`) representing downstream consumers; first-class in dbt docs lineage.

## Proposal

```bash
drt docs generate --format dbt-exposures > models/exposures/drt_exposures.yml
```

- For every sync whose `model` is `ref('<dbt_model>')` (or resolves via manifest), emit:

```yaml
exposures:
  - name: drt_users_to_hubspot
    type: application
    maturity: high
    owner: { name: drt }
    depends_on: [ref('mart_users')]
    description: "drt sync users_to_hubspot -> hubspot (upsert). Managed by drt."
    url: <docs-site link if generated>
    meta: { drt: { sync: users_to_hubspot, destination: hubspot, mode: upsert } }
```

- Syncs with raw-SQL models are skipped with a note (heuristic table-parsing out of scope)
- Deterministic output (stable ordering) so the file is committable and diff-reviewable in the dbt repo; document the "generate in CI, fail on drift" recipe

## Scope

**In:** exposure emitter over the existing docs manifest builder, ref-only coverage, determinism, docs + example in the CI guide.
**Out:** writing into the user's dbt project automatically (user commits the file), non-ref SQL lineage parsing, dbt Cloud API push.

## Related

- #239/#215 (manifest read side — this is the write side), #507 (docs manifest as the data source), #115/#117 (dbt community GTM — this feature is the demo)

## Milestone fit

Candidate: v0.8 (growth-aligned, small) — it's a docs-format addition on existing machinery with community-visible payoff.

masukai:
[...]

PR #1000, opened 2026-08-21
Add dbt exposures export (#781)

masukai:
## Summary

- `drt docs generate --format dbt-exposures` emits a deterministic dbt `exposures:` YAML block to stdout, for every sync whose `model` resolves via an exact `ref('<dbt_model>')` (reusing the existing `drt/engine/resolver.py::parse_ref()`, not new heuristic lineage parsing).
- Raw-SQL syncs are intentionally excluded (out of scope per the issue) and named in a trailing YAML comment rather than silently dropped.
- `collect_dbt_exposure_inputs()` (new, in `drt/docs/builder.py`) keeps `sync.model` — which may hold raw SQL with sensitive values — out of the shared public docs manifest; only the parsed `ref()` name crosses into the dbt-facing artifact.
- Docs: `using-with-dbt.md` gets a "Publish drt syncs in dbt lineage" section; `ci-cd-integration.md` gets a fail-on-drift CI recipe (`git diff --exit-code` against the committed exposures file).
- Out of scope, per the issue: writing into the user's dbt project automatically, non-ref SQL lineage parsing, dbt Cloud API push.

## Verification

- [x] `mypy drt` clean (197 files)
- [x] `pytest tests/unit/` clean except the 2 known, pre-existing, unrelated `--version` line-wrap failures (long worktree path)
- [x] Smoke-tested manually against a real project with a mixed `ref()`/raw-SQL sync pair — confirmed the ref-based sync renders as an exposure and the raw-SQL sync is skipped with the documented comment, output byte-shape matches the issue's proposed YAML
- [x] Verified a doc correction Codex made along the way (`using-with-dbt.md`'s `ref()` resolution order, changing a stale "(Future)" note to present tense) against the actual resolver code — accurate, `target/manifest.json` lookup already ships today


masukai:
**Codex review:**

The exporter can publish dependencies that do not correspond to dbt models, and its URL handling breaks both the documented deterministic CI workflow and standard dbt docs serving. Legal drt sync names can also produce invalid dbt identifiers or abort generation entirely.

- [P1] Exclude non-dbt refs from exposure dependencies — `drt/docs/builder.py:218`
  When `syncs/models/<name>.sql` exists, `resolve_model_ref()` executes that local SQL before consulting dbt, but `parse_ref()` still classifies the sync as a dbt dependency. The generated `ref('<name>')` then either makes dbt compilation fail when no such dbt model exists or publishes false lineage when one does; check for the supported local-SQL override (and ideally model presence in the dbt manifest) rather than relying on syntax alone.

- [P1] Make exposure output independent of prior docs builds — `drt/docs/dbt_exposures.py:49`
  The advertised CI drift gate is nondeterministic because URL presence depends on an untracked generated file. If a developer commits output after generating HTML docs but CI runs the documented command in a clean checkout, `is_file()` becomes false, the URL disappears, and `git diff --exit-code` fails despite unchanged configuration; a cached docs directory causes the reverse. Emit URLs deterministically or gate them behind an explicit option.

- [P2] Emit URLs relative to the dbt docs root — `drt/docs/dbt_exposures.py:53`
  With the default paths this emits `target/docs/sync/...`, but standard `dbt docs serve` changes its web root to the project's `target` directory. The browser therefore requests `target/target/docs/sync/...` and every emitted link returns 404; emit `docs/sync/...` or otherwise calculate the URL relative to dbt's served target root.

- [P2] Normalize sync names into valid exposure identifiers — `drt/docs/dbt_exposures.py:39`
  `SyncConfig.name` accepts arbitrary strings, so a valid drt sync named, for example, `users.to hubspot` produces an invalid/deprecated dbt exposure name. dbt emits the D007 warning for such identifiers, which fails projects using `--warn-error` and is documented to become an error; generate an alphanumeric/underscore identifier and retain the original name as a label or metadata.

- [P2] Do not let optional page slugs abort the export — `drt/docs/dbt_exposures.py:30`
  A project with otherwise legal sync names such as `a_b` and `a-b` makes `_slug_map()` raise because both map to `a-b`. This aborts the entire dbt exposure export even when no HTML pages exist and URLs would have been omitted; since page links are optional, a page-slug collision should omit those links rather than prevent exposure generation.

**Fixed** in 608646b — all five verified real before fixing (confirmed the resolver's exact precedence order, confirmed `dbt docs serve`'s web-root behavior, confirmed `_slug_map()`'s raise-on-collision).

- Local-SQL-override syncs now excluded, mirroring `resolve_model_ref()`'s own precedence check exactly.
- URLs computed purely from sync configuration via lexical path math (`os.path.normpath`, never touches the filesystem) — deterministic regardless of whether HTML docs exist on disk.
- URLs now relative to dbt's served `target/` root (`docs/sync/...`, no leading `target/`).
- Exposure names normalized into valid dbt identifiers (`drt_` prefix guarantees a letter start); a real collision resolver disambiguates two sync names that normalize identically; the original name stays in `meta.drt.sync`.
- Slug collisions now degrade gracefully — only the colliding entries lose their `url`, the export no longer aborts.

Independently re-verified beyond the fix's own self-report with a real smoke test spanning four sync shapes in one project (ref-based, raw-SQL, local-SQL-override with a punctuated name, and an invalid-dbt-identifier name starting with a digit) — every fix behaved exactly as intended, including `drt_1weird_sync_name` correctly normalizing `"1weird.sync-name!"` while `meta.drt.sync` retained the original.

PR #1001, opened 2026-08-21
Add age encryption for .drt/secrets.toml (#303)

masukai:
## Summary

- New `drt encrypt <path>` / `drt decrypt <path>` commands, backed by the optional `pyrage` extra (`pip install 'drt-core[encryption]'`) — X25519 age bindings, actively maintained upstream, ABI3 wheels covering this repo's full CI matrix (Linux x86-64/ARM64, macOS universal2, Windows; Python 3.10–3.13).
- Ships as a real, working OSS feature — encryption at rest for local secrets is baseline security hygiene, not multi-tenant governance, so it doesn't trigger `CLAUDE.md`'s "no RBAC/multi-tenancy" constraint the way #298/#299's design-only Enterprise-boundary work did.
- `.drt/secrets.toml.enc` takes precedence over plaintext and is decrypted entirely in memory via the `DRT_SECRETS_KEY` env var — every existing `secrets.toml` load path (used by `drt run` and everything else) picks it up transparently, no code changes needed at call sites.
- `drt encrypt` never deletes the original plaintext (safety — says so explicitly in its own output); both directions refuse to overwrite an existing file unless `--force`; written files are owner-only (`0o600`) on POSIX.
- Missing/invalid/non-matching keys fail with actionable messages, not generic TOML parse errors.
- Scoped to `secrets.toml` only, per the issue's own scope — state/history/DLQ files carry no connector credentials.
- `encrypt`/`decrypt` excluded from the MCP tool surface (`scripts/check_cli_mcp_parity.py`) — credential-material writes are deliberately not agent-reachable.

Implemented by Codex (task-mode, `--write`), orchestrated in an isolated worktree and independently re-verified here — including installing `pyrage` for real (Codex's own sandbox couldn't reach PyPI to download it, so its self-report skipped those 6 tests; ran them for real here).

## Verification

- [x] `mypy drt` clean (198 files)
- [x] `pytest tests/unit/test_secrets_encryption.py` — 7/7 passed with `pyrage` genuinely installed (not skipped)
- [x] `pytest tests/unit/` — 3360 passed, 2 pre-existing unrelated failures, 3 skipped
- [x] Manual end-to-end smoke test: generated a real age identity, `drt encrypt .drt/secrets.toml`, deleted the plaintext, confirmed auto-decrypt via `DRT_SECRETS_KEY` returns the original TOML data; confirmed missing-key and wrong-key paths raise the documented, actionable errors


masukai:
**Codex review:**

The primary encryption flow works, but Python 3.10 can silently ignore encrypted credentials when optional dependencies are absent, and the promised no-overwrite behavior is vulnerable to a check/write race.

- [P2] Raise when encrypted TOML support is unavailable — `drt/config/credentials.py:124-128`
  On Python 3.10, a core installation without the `encryption` extra typically has neither `tomllib` nor `tomli`. When `secrets.toml.enc` exists, this branch returns `{}` before attempting decryption, so the encrypted credential store is silently ignored instead of producing the actionable optional-extra error implemented by `_pyrage()`. Handle the encrypted path before this fallback or raise when its required parser is unavailable.

- [P2] Make the no-overwrite check atomic — `drt/config/encryption.py:129-133`
  When two processes encrypt the same file concurrently, both can pass `output.exists()` before either opens the destination, after which `_write_private_bytes()` uses `O_TRUNC` and the later process overwrites the first output despite neither specifying `--force`. The decrypt path has the same check/write race. Use exclusive creation when `force=False` so the documented overwrite refusal is enforced by the filesystem.

**Fixed** in e44453e — both verified real before fixing.

- Silent-loss fix: now checks `encrypted_path.exists()` before the tomli-missing soft-fail and raises the same actionable `pip install 'drt-core[encryption]'` message `_pyrage()` already gives; the plaintext-only `{}` fallback is unchanged.
- Race fix: `_write_private_bytes()` now uses `O_EXCL` (not `O_TRUNC`) for the non-force case, so the OS enforces exclusivity atomically instead of a separate `exists()` check racing the write.

Independently re-verified beyond the fix's own self-report: all 29 tests pass with `pyrage` genuinely installed (not skipped), `drt.config.encryption`/`drt.cli.commands.encryption` both measured at real 100% coverage (the aggregate codecov/patch check alone was misleadingly green before this — a known blind spot on this repo when a new module ships with its own dedicated test file), and manually confirmed the new concurrent-writer regression test (a `threading.Barrier` forcing both threads past the check simultaneously) genuinely exercises the fixed race.

PR #1004, opened 2026-08-22
Add I/O vs CPU profiling and Rust-migration ADR (#301)

masukai:
## Summary

- `make profile` wraps #280's unchanged `execute_scenario()` seam with stdlib `cProfile` across the same three scenarios (100/10,000/100,000 rows), attributing wall time to source extraction, transformation/serialization, and destination I/O. Writes versioned JSON artifacts under gitignored `benchmarks/profiles/`.
- **Finding: this workload is CPU-majority (57.95%–64.45% CPU-classified), refuting the bare "I/O, not CPU" claim** — but with an important caveat the ADR states explicitly: this is a local, no-network, in-memory-SQLite-to-local-file synthetic benchmark, not a production traffic model. Real SaaS/warehouse destinations add network latency, rate limiting, and remote commit time that would shift the balance back toward I/O.
- The largest CPU component (`json.dumps`, ~39% of the large run) lives in the *destination's* serialization, not `drt/engine/sync.py` — the module the issue proposed rewriting via PyO3 — so even the CPU-majority result doesn't directly support a broad engine rewrite.
- **ADR 0010 records a recommendation, not a decision.** It recommends against using this profile to justify a broad `engine/sync.py` rewrite, names the workload shapes (heavy transforms, wide payloads) that would be better native-code candidates, and explicitly defers the final Rust/roadmap call to the repo owner — this PR does not commit drt-core to a Rust migration.
- `scripts/run_profiling.py` + `make profile`, manual/local tooling only (same "no CI automation until measured need" reasoning as #280).

Implemented by Codex (task-mode, `--write`), orchestrated in an isolated worktree and independently re-verified here — including re-running the profiler myself and reproducing the same CPU-majority finding.

## Verification

- [x] `ruff check drt scripts tests benchmarks` clean
- [x] `mypy drt benchmarks` clean (202 files)
- [x] `pytest tests/unit/` clean except the 2 known, pre-existing, unrelated `--version` line-wrap failures
- [x] Ran `make profile` myself independently — reproduced the CPU-majority finding (69.78%/58.53%/59.12% on my run vs. the PR's 64.45%/57.95%/58.08%; small-scenario variance expected given its ~1–2ms absolute duration)
- [x] Read the full ADR before accepting — confirmed it doesn't overreach: explicitly names the synthetic benchmark's non-representativeness of real network-bound destinations, and correctly identifies that the CPU hotspot isn't in the module the issue proposed rewriting


masukai:
**Codex review:**

The profiler's central CPU-versus-I/O attribution misclassifies CPU-heavy in-memory extraction as I/O while assigning omitted filesystem operations to the CPU residual. Consequently, the generated percentages and ADR conclusions are not supported by the implemented measurement.

- [P1] Do not classify in-memory extraction as I/O — `benchmarks/profile_scenarios.py:272-276`
  For every supplied scenario, the source is SQLite `:memory:` running a synthetic recursive query, so this cumulative `extract` time consists largely of SQLite VM computation and Python `dict(zip(...))` construction rather than disk or network waits. Hardcoding it as `io_bound` makes the published combined-I/O percentages false for the measured workload; report it as source-bound or separately attribute CPU and actual I/O.

- [P1] Include all filesystem time before taking the CPU residual — `benchmarks/profile_scenarios.py:242-246`
  In the medium and large multi-batch scenarios, every destination batch calls `os.makedirs`, which performs `mkdir`/`stat` filesystem operations, but `_direct_file_io_seconds` counts only open, write, and context exit. Subtracting only that subset assigns the omitted destination I/O to `transformation_serialization` as CPU, inflating the central CPU-majority result; create the directory outside the measured path or include the complete destination filesystem call tree.

- [P2] Reconcile the Rust roadmap with the deferred decision — `ROADMAP.md:194`
  This entry says the final Rust/roadmap decision is deferred, while the unchanged v1.x section still unconditionally schedules an `engine/sync.py` rewrite and describes #301 as a pending gate even though it is now closed. Because the project defines ROADMAP.md as the SSoT and requires scope shifts to be reflected there (AGENTS.md), mark the v1.x rewrite as decision-pending or otherwise reconcile the two statements.

All three verified real by reading the code directly (`profile_scenarios.py`'s hardcoded `"io_bound"` label on the `:memory:` extraction bucket, `_direct_file_io_seconds()`'s function set excluding `os.makedirs`, and ROADMAP.md's v1.x section text) before starting the fix. Fix in progress — re-measuring with corrected attribution rather than hand-editing the published percentages, since the extraction relabeling alone will materially change every bucket's share. Will follow up here once verified.

masukai:
**Fixed** in c2420ff — all three issues from the review above, verified before fixing and independently re-verified after.

- P1 (extraction mislabeled as I/O): reclassified to \`cpu_bound\` with an explanatory comment. This required regenerating every number in the ADR/CHANGELOG from a real re-run, not hand-editing — corrected CPU-classified share rose to **84.46%–91.88%** (higher than originally reported, since extraction moved from the I/O side to the CPU side).
- P1 (\`os.makedirs\` missing from the I/O bucket): fixed by capturing its cumulative subtree from the profile. Added a regression test (\`test_destination_file_io_includes_complete_makedirs_subtree_once\`) with a synthetic stats structure proving the mechanism.
- P2 (ROADMAP v1.x contradicting the deferred decision): corrected to state the decision is pending owner review of ADR 0010, not scheduled.

The ADR was rewritten, not just re-numbered — it now makes the "not a production traffic model" caveat more load-bearing given the correction, explicitly warning against reading the larger CPU number as *stronger* evidence for a Rust rewrite (the higher share is largely an artifact of this benchmark having almost no genuine I/O to begin with — in-memory SQLite + a small local file write).

Also fixed independently during my own re-verification (not flagged by the review): \`scripts/run_profiling.py\`'s console output still hardcoded "SQLite extraction (I/O-bound)" after the fix, even though the underlying JSON artifact correctly said \`"classification": "cpu_bound"\` — a data-layer fix that hadn't reached the presentation layer. Now derives the printed label from the bucket's own classification field.

Independently re-verified beyond the fix's own self-report: re-ran \`make profile\` myself and reproduced the corrected CPU-majority numbers (my run: 74.84%/80.2%/86.26% CPU-classified across small/medium/large vs. the committed 91.88%/86.41%/84.46% — consistent given run-to-run variance, especially the small scenario's ~2ms absolute duration), and confirmed the JSON artifact's \`source_extraction.classification\` field reads \`"cpu_bound"\` directly.

Issue #184, opened 2026-04-04
feat(dagster-drt): Support @op context in DagsterDrtResource


masukai:
## Summary

Enable `DagsterDrtResource.run()` to work with `OpExecutionContext` in addition to `AssetExecutionContext`, yielding `AssetMaterialization` instead of `MaterializeResult` when used outside of asset definitions.

## Motivation

Some users may want to use drt syncs inside `@op` definitions rather than `@asset` / `@multi_asset`. dagster-dlt supports both contexts.

## Proposed Behavior

```python
@op
def run_drt_sync(context: OpExecutionContext, drt: DagsterDrtResource):
    # Must pass sync_names explicitly since there's no @drt_assets metadata
    yield from drt.run(context=context, sync_names=["my_sync"])
```

- When `AssetExecutionContext`: yield `MaterializeResult` (current behavior)
- When `OpExecutionContext`: yield `AssetMaterialization` + require explicit `sync_names`

## Acceptance Criteria

- [ ] `DagsterDrtResource.run()` accepts both context types
- [ ] `sync_names` parameter for explicit sync selection in `@op`
- [ ] `AssetMaterialization` yielded in op context
- [ ] Unit tests for both paths

## Context

Low priority — most users will use `@drt_assets`. Only needed for advanced orchestration patterns.

masukai:
Closed via #1005 (merged). `DagsterDrtResource.run()` now accepts `OpExecutionContext` alongside the existing `AssetExecutionContext` path, requiring explicit `sync_names` and yielding `AssetMaterialization`. Asset keys derived through the same translator the asset path uses, so both paths stay consistent.

Issue #183, opened 2026-04-04
feat(dagster-drt): Add DrtSyncComponent for declarative YAML definitions


masukai:
## Summary

Add a `DrtSyncComponent` for Dagster's declarative Components system, enabling YAML-driven drt asset definitions without Python code.

Follows dagster-dbt's `DbtProjectComponent` and dagster-dlt's `DltLoadCollectionComponent` pattern.

## Proposed API

```yaml
# defs.yaml
type: dagster_drt.DrtSyncComponent
attributes:
  project_dir: ./drt_project
  sync_names: ["my_sync"]
  translation:
    group_name: reverse_etl
```

## Motivation

- Dagster is moving toward declarative, YAML-driven definitions
- Reduces boilerplate for standard drt deployments
- Aligns with dagster-dbt and dagster-dlt component patterns

## Acceptance Criteria

- [ ] `DrtSyncComponent` scaffoldable via `dg scaffold`
- [ ] YAML configuration for project_dir, sync_names, translation
- [ ] Overridable `get_asset_spec()` and `execute()` methods
- [ ] Documentation and example

## Context

Phase 3 of dagster-drt API alignment. See #175 comment.

masukai:
Closed via #1005 (merged). `DrtSyncComponent` for Dagster's Components system (`defs.yaml`, `dg scaffold`). The issue's proposed API predated Components' real current shape — verified the actual requirement is Dagster >=1.10.18, not the package's prior >=1.6 floor, and bumped accordingly. Codex review caught an op-name collision when multiple component instances coexist in one project — fixed.

Issue #182, opened 2026-04-04
feat(dagster-drt): Add DrtEventIterator for chainable post-processing


masukai:
## Summary

Add a `DrtEventIterator` wrapper class around `DagsterDrtResource.run()` output, enabling chainable post-processing methods.

Follows dagster-dlt's `DltEventIterator` pattern.

## Proposed API

```python
@drt_assets(project_dir=".")
def my_syncs(context, drt: DagsterDrtResource):
    yield from drt.run(context=context).fetch_row_count()
```

## Motivation

- Enables `.fetch_row_count()` for post-load verification
- Establishes an extension point for future enrichment methods
- Aligns with dagster-dlt's `DltEventIterator` pattern

## Acceptance Criteria

- [ ] `DrtEventIterator` wraps `Iterator[MaterializeResult]`
- [ ] `.fetch_row_count()` method queries source for actual row counts
- [ ] Chainable pattern works
- [ ] Unit tests

## Context

Low priority — current `MaterializeResult` metadata already includes `rows_synced` from `run_sync()`. This adds source-side verification.

masukai:
Closed via #1005 (merged). `DrtEventIterator` wraps `DagsterDrtResource.run()`'s output with a chainable `.fetch_row_count()`, independently re-querying the source (not just surfacing `rows_extracted`/`rows_synced` metadata already produced by the sync). Codex review caught it silently failing for incremental (cursor-templated) syncs and bypassing this repo's query-tagging contract — both fixed, plus a separate pre-existing bug found in the process (the resource's own `run_sync()` call never respected `project.query_tagging` at all).

PR #1005, opened 2026-08-23
Add DrtEventIterator, DrtSyncComponent, and @op support to dagster-drt (#182, #183, #184)

masukai:
## Summary

Three related dagster-drt (separately-versioned) issues bundled as one PR, all self-marked "Low priority" in their own issue text.

- **#182 — `DrtEventIterator`**: `DagsterDrtResource.run()` now returns a chainable iterator, matching dagster-dlt's established shape (`yield from drt.run(context=context).fetch_row_count()`). `.fetch_row_count()` independently re-queries the **source** (`source.extract()`) — it does not just surface `rows_extracted`/`rows_synced` metadata already produced by the sync run. Works uniformly across both event types #184 introduces.
- **#183 — `DrtSyncComponent`**: declarative `defs.yaml` support via Dagster's Components system, scaffoldable with `dg scaffold`. **The issue's proposed API predated Components' actual current shape** — `defs.yaml`, top-level Component exports, and the per-project `dg` CLI contract all arrived together in **Dagster 1.10.18**. The package's dependency floor moves from `dagster>=1.6` to `dagster>=1.10.18` as a result; verified against the installed 1.12.22 environment. Note: Dagster's own Components API is still marked Beta upstream (visible in this PR's test warnings) — expect possible breaking changes in future Dagster minor releases, independent of anything in this PR.
- **#184 — `@op` support**: `run()` now accepts `OpExecutionContext` alongside the existing (unchanged) `AssetExecutionContext` path, requiring explicit `sync_names` (no `@drt_assets` metadata to select from) and yielding `AssetMaterialization` instead of `MaterializeResult`. Asset keys for the op path are derived through the same `DagsterDrtTranslator` the asset path uses, so both paths stay consistent.

A pre-existing sensor CHANGELOG entry (`#855`) was also lightly corrected — its reasoning for not exposing `target=` cited the old `dagster>=1.6` floor, which is now stale given the floor bump; docstring/entry correction only, no behavior change.

## Verification

- [x] `ruff check integrations/dagster-drt` clean
- [x] `mypy integrations/dagster-drt/dagster_drt` — one pre-existing error in `translator.py` unrelated to this PR (confirmed via `git stash` against `origin/main` — it's already there)
- [x] `pytest integrations/dagster-drt/tests/` — 68 passed
- [x] `scripts/check_drift.sh` (top-level) — no new drift
- [x] Read every changed file's diff directly — confirmed the `isinstance`-based context branching, the row-count independence (re-query vs. metadata copy), and the Components imports are all genuinely using real, current Dagster 1.12.22 APIs, not fabricated
- [x] Did not attempt a live end-to-end asset materialization smoke test — `DagsterDrtResource.run()` resolves profiles from the real `~/.drt/profiles.yml` with no override parameter exposed, and this environment's home directory has a real profiles.yml; touching it for a smoke test wasn't worth the risk given the existing 68-test suite already exercises this with Dagster's own `materialize`/`build_op_context` utilities


masukai:
**Codex review:**

The new component cannot coexist with another instance because of a deterministic Dagster op-name collision. The row-count post-processing also fails for supported cursor-templated models and bypasses the repository's query-tagging contract.

- [P2] Give each component instance a unique asset op name — `integrations/dagster-drt/dagster_drt/components.py:116-120`
  When a Dagster project declares two `DrtSyncComponent` folders, both definitions create an op named `drt_sync_assets` because no `name` is supplied here. Merging the definitions then raises `DagsterInvalidDefinitionError: Detected conflicting node definitions with the same name "drt_sync_assets"`, even when their asset keys are disjoint; derive a unique name from the component load context.

- [P2] Preserve query tags on row-count extraction — `integrations/dagster-drt/dagster_drt/resource.py:117-120`
  When `.fetch_row_count()` runs against a default query-tagging-enabled warehouse, this direct extraction receives the raw resolved query and no `query_tags`, bypassing both the SQL comment and native warehouse tags added by `run_sync`. The extra warehouse cost is therefore anonymous, contrary to the every-extract query-tagging contract in `AGENTS.md:61`; construct and pass the configured tags for this query as well.

- [P2] Pass the resolved cursor to row-count model resolution — `integrations/dagster-drt/dagster_drt/resource.py:110-114`
  For a valid incremental model containing `{{ cursor_value }}` or `{{ watermark }}`, calling `resolve_model_ref` without a cursor deterministically raises `ValueError`. `fetch_row_count()` catches that exception and emits the original event without `dagster/row_count`, so the advertised post-processing never works for cursor-templated syncs; retain and pass the run's resolved cursor when resolving this query.

All three verified real by reading the code directly (`drt_assets()`'s existing unused `name=` parameter, `run_sync()`'s query-tagging block and `SyncResult.cursor_value_used`, and `resolve_model_ref()`'s cursor-templated `ValueError`) before starting the fix. Fix in progress — will follow up here once verified.

masukai:
**Fixed** in 81b36f1 — all three issues from the review above, verified before fixing and independently re-verified after.

- P2 (op-name collision): derives a unique, sanitized op name from the component's declaring \`defs.yaml\` path (relative, not absolute-path-unstable) plus a stability digest. New test asserts two components with different \`project_dir\`/\`sync_names\` merge into one \`Definitions\` object without \`DagsterInvalidDefinitionError\`.
- P2 (query tagging bypassed): \`.fetch_row_count()\`'s extraction now mirrors \`run_sync()\`'s own query-tagging block exactly — same enabled-gate, same \`build_query_tags\`/\`render_comment_header\` calls, a fresh \`run_id\` since it's a separate, later query. New test asserts the exact SQL comment string and \`query_tags\` dict passed to \`extract()\`.
- P2 (cursor never passed): captures \`SyncResult.cursor_value_used\` per sync and threads it into \`resolve_model_ref(..., last_cursor_value=...)\`. New test constructs a real \`{{ cursor_value }}\`-templated model and asserts the substituted value appears in the executed query, with a working non-\`None\` \`dagster/row_count\`.

**Bonus fix found while fixing the second issue**: \`DagsterDrtResource.run()\`'s own \`run_sync()\` call had never passed \`project.query_tagging\` at all (confirmed via \`git show origin/main\` — this predates this PR entirely) — so a project's explicit \`query_tagging.enabled: false\` or custom \`extra\` tags were silently ignored for every sync executed through dagster-drt (the CLI's \`drt run\` path was unaffected). Fixed alongside, since the same config now needs threading through regardless. Documented in the CHANGELOG.

Independently re-verified beyond the fix's own self-report: read every diff directly, ran \`ruff check\`/\`mypy\`/\`pytest\` myself (70/70 dagster-drt tests pass; the one \`translator.py\` mypy error is the same pre-existing, unrelated issue confirmed earlier via \`git stash\` against \`origin/main\`), and re-ran the top-level \`drt-core\` suite (3401 passed, 2 known unrelated failures) plus \`scripts/check_drift.sh\` to confirm nothing outside \`integrations/dagster-drt/\` was affected.

Issue #1002, opened 2026-08-21
fix(file): FileDestination truncates on every batch, losing all but the last when batch_size < row count


masukai:
## Summary

`drt/destinations/file.py`'s `FileDestination._write_csv`/`_write_json`/`_write_jsonl` all open the output path with `open(path, "w", ...)` — a fresh truncate on every call. `drt/engine/sync.py::run_sync()` calls `destination.load()` once per batch, inside its `for record_batch in batch(records_iter, sync.sync.batch_size):` loop.

For any `type: file` sync where the source has more rows than `sync.batch_size`, every batch after the first silently discards everything written by prior batches. The final file on disk contains only the *last* batch's rows, even though `SyncResult.success`/`rows_extracted` (in-memory counters, unaffected by the file writer) correctly report the full row count — so `drt run`'s own summary output looks correct while the actual output file is wrong.

## Repro

```yaml
# drt_project.yml + a source producing, say, 250 rows
name: users_to_file
model: "SELECT * FROM users"
destination:
  type: file
  path: output/users.jsonl
  format: jsonl
sync:
  batch_size: 100
```

`drt run users_to_file` reports 250 rows synced. `output/users.jsonl` contains 50 lines (the last, partial batch) — 200 rows are silently lost from the destination, not just from any particular run's summary.

## Found via

Discovered while building #280's benchmark suite (PR #999) — its medium/large scenarios (10,000/100,000 rows, batch_size=100) hit this immediately, caught in that PR's Codex review. The benchmark's own fix stays scoped to the harness (a purpose-built accumulating destination wrapper, bypassing the buggy writer) rather than touching this file — this issue tracks the real, separate production bug.

## Suggested fix

`FileDestination` needs to distinguish "first batch of a sync run" (truncate/start fresh) from "a later batch of the same run" (append) — it currently has no such state. Since `get_destination()` constructs one instance per sync and `.load()` is called on the same instance across all of that sync's batches, an instance-level flag (e.g. a `set[str]` of paths already truncated this instance's lifetime) is enough for CSV and JSONL (natural append formats — CSV needs the header written only once, JSONL is line-by-line). A single-array `format: json` file is harder to append to correctly and cheaply; consider whether it needs to switch to a read-modify-rewrite per batch (correct but O(n²) across many batches) or whether it's acceptable to document that `format: json` should only be used with `batch_size >= total row count` until a better fix lands.

[...]

PR #1006, opened 2026-08-23
Fix FileDestination silently retaining only the last batch (#1002)

masukai:
## Summary

`FileDestination.load()` opened its output path in truncate (`"w"`) mode on every call, but `drt/engine/sync.py` calls `.load()` once per batch. Any `type: file` sync with more rows than `sync.batch_size` silently lost every batch except the last one — `SyncResult.success`/`rows_extracted` correctly reported the full row count while the file on disk held only the final batch. Found while building #280's benchmark suite (its medium/large scenarios, 10,000/100,000 rows at batch_size 100, hit this immediately).

**Deliberately kept `FileDestination` as a plain `Destination`** (not converted to this repo's `StagedDestination` pattern) — the engine dispatches error handling very differently for the two: a plain `Destination`'s per-batch `on_error: fail` triggers an immediate early-stop, while `StagedDestination` only reports success/failure once, at `finalize()`, after every batch has already staged. Converting would have silently changed `on_error: fail` behavior for every existing `type: file` user — a real behavior change, not a bug fix, so it stayed out of scope.

Per-format fix, all within the existing `load()` interface:
- **CSV**: header written once, later batches append; raises a clear `ValueError` on a column-set mismatch across batches rather than silently misaligning columns.
- **JSONL**: appends after the first batch (each line is independently valid) — no buffering, stays streaming-friendly for large exports.
- **JSON** (single array): cannot be appended to without breaking `[...]` syntax, so this format alone buffers the full sync in memory and rewrites the complete valid array every batch — a real, known, deliberate memory tradeoff specific to this one format, documented in both the code and `docs/connectors/file.md`.

## Verification

- [x] `mypy drt` clean (199 files)
- [x] `pytest tests/unit/test_file_destination.py` — 17/17 passed; read the tests directly — real `run_sync()` end-to-end coverage across all three formats at 250 rows/batch_size 100, a registry-constructed instance-isolation test, and a CSV-mismatch test that also confirms `on_error: fail` still stops early (rows_extracted stays at 200, the third batch never runs)
- [x] `pytest tests/unit/` — 3406 passed, 2 pre-existing unrelated failures, 3 skipped
- [x] Real CLI end-to-end smoke test: `drt run --select smoke_jsonl` against a real 250-row sync with `batch_size: 100` (3 batches) — confirmed all 250 rows present in the output file, not just the last 50


masukai:
**Codex review:**

Multi-batch writes now work for the normal CLI lifecycle, but per-run state leaks when a destination instance is reused, causing successive sync runs to be silently combined.

- [P2] Clear per-run buffers after each sync — `drt/destinations/file.py:41-43`
  When a library caller passes the same `FileDestination` instance to two successive `run_sync()` calls, these maps and sets still contain the path from the completed first run. The second run therefore appends CSV/JSONL rows or includes the old JSON records instead of replacing the file as documented. Add a `finalize_sync` reset, as other stateful destinations do, or otherwise reset this state at the run boundary.

**Fixed** in a731dd3 — verified real by checking this repo's existing `finalize_sync()` convention (`sql_base.py`/`snowflake.py`/`databricks.py`/`clickhouse.py`) and confirming the engine's dispatch (`drt/engine/sync.py`) calls it as a duck-typed, non-Protocol hook for any non-staged destination after the batch loop, regardless of mid-sync failure.

Added `FileDestination.finalize_sync()` clearing all three write-state collections. Wrote a regression test (`test_reused_instance_does_not_leak_state_into_a_later_run`, parametrized across csv/json/jsonl) reusing one instance across two `run_sync()` calls — confirmed it fails without the fix (temporarily removed `finalize_sync()` locally, reran: second run's output incorrectly contained the first run's 5 records plus the second run's 3, instead of just the second run's 3) and passes with it restored. Full `pytest tests/unit/` (3409 passed, 2 known unrelated failures) and `scripts/check_drift.sh` re-run clean.

PR #1007, opened 2026-08-23
Reconcile CLAUDE.md's RBAC and Rust statements with what's shipped

masukai:
## Summary

Two small wording fixes to `CLAUDE.md`, found while discussing project direction:

- **RBAC rule was stale.** "Do not add RBAC or multi-tenancy" read as absolute, but ADR 0008 (#298/#299) already added interface-only, no-op-by-default `PermissionChecker`/`AuditLogger` Protocols as a deliberate, explicitly scoped Enterprise-boundary exception. Clarified the rule to say what's actually true — OSS doesn't implement/enforce RBAC, and another no-op interface in that spirit needs the same explicit sign-off ADR 0008 got, not a blanket "don't."
- **"Future Rust rewrite" read as a settled inevitability.** ADR 0010 (#301) is the first real profiling evidence and explicitly defers the migration decision to the repo owner. Softened the wording and linked the ADR so this stays accurate as that decision develops.

No behavior change — doc-only.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,7 +16,7 @@ Config Parser → Source (BigQuery) → Sync Engine → Destination (REST API)
                                                    State Manager
 ```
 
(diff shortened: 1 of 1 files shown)


masukai:
**Codex review:**

The documentation changes accurately clarify the shipped RBAC extension points and the deferred Rust migration decision, with valid links to the relevant ADRs. No functional or actionable issues were introduced.

Issue #1008, opened 2026-08-23
Follow-up profiling for ADR 0010: real network I/O + a scoped JSON-serialization PyO3 prototype


masukai:
## Summary

[ADR 0010](https://github.com/drt-hub/drt/blob/main/docs/adr/0010-rust-migration-decision.md) (#301) profiled #280's benchmark scenarios and found them CPU-majority (84–92%) — but explicitly flagged that benchmark as a local, no-network, in-memory-SQLite-to-local-file synthetic workload, "not a production traffic model," and named the next evidence needed: *"profiles of representative remote destinations and CPU-heavy transforms using real payload widths."* This issue tracks that follow-up, split into two independent, reversible experiments — neither commits to a Rust migration; both exist to get real data before that decision is made.

## Experiment 1 — Profile against real network I/O

Extend #301's `cProfile` methodology (`benchmarks/profile_scenarios.py`) to workloads with genuine I/O wait, not just in-memory computation:

- **Warehouse source**: a real Postgres container via `testcontainers` (this repo already has this pattern for `local_sql_smoke` — see #941/`tests/integration/local_sql/conftest.py`'s `require_docker()`), profiling `PostgresSource.extract()` against real query execution + real network round trips instead of `SQLiteSource` against `:memory:`.
- **REST API destination**: a local HTTP server (`pytest-httpserver`, already a dev dependency) with configurable, realistic artificial latency per request, profiling the REST API destination's real HTTP client stack (real socket I/O wait) rather than a local file write. Latency should be parametrized (e.g. 10ms / 50ms / 200ms RTT) so the report shows how the I/O-vs-CPU split moves as network cost increases, rather than one single anecdotal number.
- Same three row-count scenarios as #280/#301 (or a justified subset), same non-overlapping bucket-attribution methodology, same manual-only tooling convention (no CI automation).
- Output: an amendment to ADR 0010 (new "Experiment 1" section) with the real measured splits, explicit about what's genuinely real I/O (socket wait time) vs. what's still a controlled simplification (fixed/synthetic latency rather than live vendor variance, rate limiting, jitter).

## Experiment 2 — Scoped PyO3 prototype of the one confirmed CPU hotspot

[...]

PR #1009, opened 2026-08-23
ADR 0010 follow-up: real network I/O + a PyO3 prototype (#1008)

masukai:
## Summary

ADR 0010's own text named its gap: its benchmark's source performed no genuine storage/network I/O, so it couldn't settle whether production drt traffic is CPU- or I/O-bound. This runs the two experiments ADR 0010 named as the next evidence — real numbers, not projections, both run in this environment.

- **Experiment 1a (real Postgres source)**: ephemeral `postgres:16-alpine` via `testcontainers`, real TCP connection. Pure-Python consumer CPU never exceeds 10.55% of total time at any scenario size — the inverse of #301's SQLite `:memory:` shape, exactly as predicted.
- **Experiment 1b (real REST destination, controlled latency)**: local HTTP server, real sockets, 10/50/200ms simulated response delay. `destination_cpu` share climbs from ~74% to ~95% as latency rises. Two large-scenario results (50ms/200ms) came back with impossible bucket math (>100%, negative %) — diagnosed as `cProfile` breaking down under sustained same-process multi-thread contention at high latency × ~1000 requests, and reported/excluded honestly rather than hidden or force-fit.
- **Experiment 2 (throwaway PyO3 prototype)**: `pyo3`+`serde_json` JSON serialization, benchmarked head-to-head against `json.dumps` including the full Python↔Rust boundary cost. Real, reproducible across 3 runs: 1.2–1.6× faster at small batches, flattening to noise (0.95×–1.09×) at 100,000 records — the scale where a win would matter most.

**Neither experiment changes ADR 0010's recommendation against a broad `engine/sync.py` rewrite — both sharpen it.** Full detail in ADR 0010's new "Follow-up" section.

## Provenance note

Implemented by Codex (task-mode, `--write`), but Codex's own sandbox blocked Docker, localhost socket binding, and crates.io — it correctly refused to fabricate numbers or write the ADR amendment without real evidence, and left the harness code + a clear list of blockers. I have none of those restrictions in this environment, so I:
- Found and fixed a real bug in the seeding SQL (`id % 1000` colliding with psycopg2's `%s` parameter substitution — `IndexError` on every run, fixed by escaping to `%%`)
- Ran both real-I/O legs myself against a real Docker Postgres container and a real local HTTP server
- Installed `cargo`/`maturin`, built the throwaway PyO3 prototype myself (not committed, per the issue's own scope — scratch-only), and ran the benchmark 3 times to confirm the pattern was reproducible
- Wrote the ADR 0010 amendment directly from these real numbers

## Verification

- [x] `ruff check drt scripts tests benchmarks` clean
- [x] `mypy drt benchmarks` clean (203 files)
- [x] `pytest tests/unit/` — 3412 passed, 2 pre-existing unrelated failures, 3 skipped
- [x] `make profile-real-io` actually run against real Docker + real sockets — confirmed the numbers in the ADR are real, not projected
- [x] PyO3 prototype actually built and benchmarked in this environment (Rust 1.75.0, `pyo3 0.20.3`, `maturin develop --release`) — 3 independent runs confirmed the reported pattern


masukai:
**Codex review:**

The new profiler misclassifies controlled network wait as CPU, emits schema-invalid artifacts for default scenarios, and misreports the Postgres fetch size. Its default target also lacks a required dependency under the documented development setup.

- [P2] Attribute controlled REST latency to an I/O-bearing bucket — `benchmarks/profile_real_io.py:360`
  With any nonzero controlled latency, `load_seconds` includes time waiting for the delayed response, but only the detected socket primitive self time is subtracted; all remaining wait is labeled `cpu_bound`. For example, the documented 200 ms/small run assigns nearly the full delay to `destination_cpu`, so increasing network latency is reported as CPU work and the profiler cannot support its CPU-vs-I/O conclusion. Measure CPU time separately or classify this remainder as mixed/I/O-bearing.

- [P2] Reject REST timings that violate the artifact schema — `benchmarks/profile_real_io.py:361-365`
  When cProfile reports aggregate `load_seconds` above `total_tt`—as the default large scenarios at 50 ms and 200 ms already do—zeroing `harness_cpu_seconds` does not normalize the other buckets, so `_percentages` emits a destination percentage above 100 and a negative harness percentage. The default command therefore writes artifacts that violate the new schema's 0–100 constraint; detect and reject/mark these measurements before writing them or move the server to a reliable profiling boundary.

- [P2] Apply the scenario batch size to Postgres fetching — `benchmarks/profile_real_io.py:172-179`
  The Postgres profile leaves `fetch_size` at its 10,000-row default, while every resulting artifact records `scenario.batch_size` (100) and the benchmark documentation claims 100-row batching. Consequently the source timing uses 10,000-row driver fetches but reports 100, making the experiment misleading and difficult to reproduce; pass the scenario size into the source profile or report the actual fetch size.

- [P2] Provide the Postgres driver for the default profile target — `Makefile:28-29`
  On a fresh checkout prepared with the documented `make dev`, this target reaches `_seed_postgres()` and imports `psycopg2`, but that driver is declared only in the separate `postgres` extra while `make dev` installs `dev,bigquery`; the command therefore fails before profiling even when Docker is available. Wire the target/setup to install the Postgres extra or document the additional installation requirement.

All four verified real by reading the code directly. The first finding is the important one: I found it also contradicts my own ADR prose (which claimed "the more a destination resembles a real network call... the smaller the CPU-classified share becomes" right below a table showing the opposite trend) — the table was built on the wrong classification. Fix in progress: the REST leg will be restructured around the known, exactly-computable injected latency rather than inferring I/O from an incomplete socket-primitive list, and the server will move to a genuinely separate process (matching the Postgres leg's clean boundary) to fix the underlying cProfile-under-thread-contention issue at its root. I'll re-run the real experiment myself (Docker/sockets aren't available in Codex's sandbox) and rewrite the ADR's Experiment 1b section from the corrected numbers, not patch the existing table. Will follow up here once verified.

masukai:
**Codex review:**

The new documented profiling target is not self-contained and fails in an environment without previously installed development dependencies. The underlying profiling implementation otherwise appears consistent.

- [P2] Install testcontainers before starting the Postgres leg — `Makefile:29`
  When the documented `make profile-real-io` command is run without a prior `make dev`, this installed only `.[postgres]`, but `run_postgres_profiles()` imports `testcontainers.postgres`, which is available only through the `dev` extra in `pyproject.toml`. The command therefore failed with `ModuleNotFoundError` before profiling; include `testcontainers[postgres]` in this target or install the relevant extras together.

Verified against `pyproject.toml`'s `[project.optional-dependencies]`: `testcontainers[postgres]>=4.0` is declared under `dev`, not `postgres`. Fixed in 83443ea by installing `.[dev,postgres]` in the Makefile target, matching the existing `dev:` target's extras pattern, plus a matching README correction.

Issue #997, opened 2026-08-20
Make DestinationConfig/ProfileConfig extensible for third-party connector types


masukai:
## Summary

Third-party connector packages can register themselves in `drt/connectors/registry.py` via entry-point discovery (#297, ADR 0009), but the resulting type can't actually be used from a sync YAML — a real config-layer gap, not a plugin-loader bug.

## Why

`SyncConfig.destination` (`drt/config/sync_options.py:588`) is a Pydantic `Field(discriminator="type")` union with all ~34 destination config classes hand-listed. `load_profile()` (`drt/config/credentials.py:244`) is a hand-written `if source_type == "..."` dispatch chain ending in `raise ValueError(...)`. Both reject an unrecognized `type` value before `drt/connectors/registry.py`'s `get_destination()`/`get_source()` is ever consulted.

A sync YAML naming `type: salesforce_premium` (a hypothetical third-party package registered via #297's entry-point mechanism) fails `drt validate` outright — the exact "pip install and it just works" scenario #297's issue text described, which #297 explicitly could not deliver end-to-end.

## Scope

Choose and implement one of the candidate directions ADR 0009 sketches (none chosen there — this issue is where that choice gets made and reviewed):

1. **Two-pass validation** — parse `type: str` + `config: dict[str, Any]` loosely, resolve the concrete config class via the (by-then plugin-populated) connector registry, then `model_validate()` the dict against it. Closed union stays the fast/typed path for built-ins; dynamic resolution is a fallback reached only when `type` isn't a static union member.
2. **Catch-all fallback union member** — a `GenericDestinationConfig` model accepting an open `dict`, deferring strict validation until the registry resolves the real type. Keeps a single Pydantic validation pass; weakens error messages for a plugin type with a typo'd field.
3. **Registry-driven `model_validator(mode="before")`** — replace the static union entirely, built-ins included. Most extensible; largest blast radius (see below).

## Constraints to weigh (from ADR 0009)

- Must not degrade parse-error messages for the ~34 built-in destination types (today's discriminated union produces "unknown type, expected one of [...]").
- Must not silently break mypy's static narrowing of `sync.destination`'s concrete type at existing call sites — notably the six per-dialect dispatchers in `drt/config/query.py` (deliberately left untouched by #469) and `isinstance()` checks throughout `drt/destinations/`.
[...]

PR #1010, opened 2026-08-26
ADR 0011: subtraction as positioning against reverse-ETL competitors

masukai:
## Summary
- New ADR 0011 (Accepted 2026-08-26): drt's actual competitive set is other reverse-ETL tools (Census, Hightouch, RudderStack Reverse ETL, etc.), not dlt/dbt — those are adjacent pipeline stages. Against that competitive set, `drt-core` deliberately excludes a web UI/dashboard, a hosted/managed runtime, an audience/segmentation builder, and a proprietary connector catalog — not because they're unbuilt, but as a scoped positioning decision, mirroring ADR 0008's escape-hatch shape (OSS-core boundary, not a permanent ban on drt as a business).
- New tagline, rolled out consistently: "Reverse ETL as code — no UI, no lock-in, no per-row bill." — replaces "Reverse ETL for the code-first data stack" in `CLAUDE.md`, `docs/llm/CONTEXT.md`, `pyproject.toml`'s PyPI description, the package docstring, and the CLI's `--help` text.
- README's hero paragraph and `docs/llm/CONTEXT.md`'s "What drt is NOT" section both now name the same exclusions and the "your data never leaves your own infrastructure" claim.
- `CONTRIBUTING.md`'s "Picking up an Issue" section now points larger-feature proposals at `CLAUDE.md`'s "What NOT to do" + ADR 0011 before implementation starts.

## Test plan
- [x] `ruff check .` clean
- [x] `mypy drt` clean
- [x] `pytest tests/unit/` — clean except 2 pre-existing, unrelated `test_cli_version.py` failures caused by this worktree's unusually long absolute path wrapping the `--version` output (confirmed same failure signature on an unrelated branch earlier this session; not caused by this diff)
- [x] Confirmed no test asserts the literal old tagline string
- [ ] Repository owner review (product-voice/positioning content)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,7 +86,7 @@ Cloud-warehouse tests remain under `dwh_smoke` and require `DRT_SMOKE_*` secrets
 
 ## What NOT to do
 
-- Do not add a GUI or web UI — this is a CLI-first tool
+- Do not add a GUI or web UI, an audience/segmentation builder, a hosted/managed runtime, or a proprietary connector catalog — this is a CLI-first tool, and [ADR 0011](docs/adr/0011-subtraction-positioning-vs-reverse-etl.md) makes the competitive case for why these four are deliberately excluded from `drt-core` (not merely unbuilt) rather than assuming the reasoning is obvious
 - Do not implement or enforce RBAC/multi-tenancy in OSS — small team / personal use. [ADR 0008](docs/adr/0008-enterprise-boundary-rbac-and-audit-hooks.md) is a scoped exception: `drt.security.PermissionChecker`/`drt.observability.AuditLogger` are interface-only, no-op-by-default (`AllowAllPermissionChecker`, `NoOpAuditLogger`) Enterprise-boundary Protocols (#298/#299) — adding another no-op interface in that same spirit needs the same explicit sign-off ADR 0008 got, not a default assumption this rule forbids it outright
 - Do not add `type: ignore` — only allowed for external library issues (`no-untyped-call`, `import-untyped`)
 - Do not add heavy dependencies to core — extras (`[bigquery]`, `[mcp]`) exist for a reason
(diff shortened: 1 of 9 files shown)


masukai:
**Codex review:**

The patch introduces several user-facing claims that contradict current behavior, including a false data-residency guarantee, an unusable connector-plugin capability, and a nonexistent configuration field. These should be corrected before adopting the positioning text as canonical.

- [P2] Scope the data-residency guarantee accurately — `README.md:12`
  For any SaaS destination such as HubSpot or Slack, records necessarily leave the operator's infrastructure and enter the destination's infrastructure. The intended guarantee appears to be that data never passes through a drt-hosted intermediary; the current wording, repeated in the ADR and LLM context, is a materially false data-residency claim.

- [P2] Do not claim destination plugins are usable yet — `docs/adr/0011-subtraction-positioning-vs-reverse-etl.md:59-60`
  When a third-party package registers a `drt.destinations` entry point, its type still cannot be named in sync YAML because the closed `DestinationConfig` union rejects it before registry lookup. This blocker is documented in `docs/guides/plugins.md` and ADR 0009, so describing connectors as extensible without drt-hub involvement misrepresents the current capability.

- [P2] Remove the nonexistent `sync.query` option — `docs/adr/0011-subtraction-positioning-vs-reverse-etl.md:53-54`
  `SyncConfig` exposes the required top-level `model` field but no sync `query` field (`drt/config/sync_options.py:627-634`); the only `query` field belongs to custom destination tests.

- [P3] Keep the Docker image description in sync — `pyproject.toml:8` / `Dockerfile:4`
  The container image's `LABEL description` still had the old tagline while package metadata and the CLI already used the new one.

All four verified against the actual code (confirmed `SyncConfig` has no `query` field, confirmed ADR 0009's documented config-union blocker, confirmed `Dockerfile:4`'s stale label) and fixed in 1757871: the data-residency claim is rescoped to "no drt-hosted intermediary," the connector-catalog point now states the real tracked gap (ADR 0009, #997) instead of a capability that doesn't exist yet, the ADR/CONTEXT.md now name only `model`, and the Dockerfile label matches the new tagline.

PR #1011, opened 2026-08-26
Make DestinationConfig/ProfileConfig extensible for third-party connector types (#997)

### LINKED ISSUE #997: Make DestinationConfig/ProfileConfig extensible for third-party connector types

Muawiya-contact:
Closes #997. Takes ADR 0009's **option 2** (catch-all union member, callable `Discriminator`) as agreed, with one addition to the routing rule that I think changes its trade-off enough to flag up front.

## The blocker

#297 shipped entry-point discovery, but a connector registered that way could never be *named*. `SyncConfig.destination` was a closed discriminated union and `load_profile()` a closed `if source_type == ...` chain — both rejecting an unrecognized `type` before `drt/connectors/registry.py` was consulted. Reproduced on `75719cf`:

```
type: salesforce_premium
→ ValidationError: union_tag_invalid at ('destination',)
  "Input tag 'salesforce_premium' ... does not match any of the expected tags: 'rest_api', ..."
```

## Destinations

`Field(discriminator="type")` → `Discriminator(_destination_tag)`, plus `GenericDestinationConfig` as a 35th member. Every member carries an explicit `Tag` — pydantic requires one per choice under a callable discriminator and raises `PydanticUserError` at import without it, so this cannot silently rot.

**The addition.** `_destination_tag` is three-way, not two:

| `type` | routes to | error |
|---|---|---|
| built-in | its concrete config | unchanged, per-field |
| **registered in the connector registry** | `GenericDestinationConfig` | — |
| anything else | — | `union_tag_invalid`, as today |

ADR 0009 predicted option 2 would weaken errors "for a plugin type with a typo'd field". True — but a catch-all that swallows *every* unknown type also makes `type: postgress` parse cleanly and fail much later at `get_destination()`, so `drt validate` would start passing on typo'd built-ins. Consulting the registry for **membership only** — not validation — keeps that case reporting the identical error shape. The issue's "must not degrade parse-error messages" constraint is met rather than traded away.

The registry is populated before parsing: built-ins at import of `registry.py`, plugins via `load_plugins()` in the CLI root callback.

What *is* looser, exactly as predicted: a plugin's own fields are `extra="allow"` and carried verbatim, because drt-core doesn't know its schema. A typo in one of *those* is kept. Strict second-pass validation against the registry's stored `config_class` is **deliberately not implemented** here, per your note.

## Sources

`load_profile()` keeps its hand-written chain verbatim and gains a registry lookup *after* it — built-ins keep their exact construction including per-type defaults, and a plugin cannot shadow one. Profiles are dataclasses with no validator to hook, so the registered profile class is constructed directly; a `TypeError` from a bad field is rewrapped to name the plugin.

This also fixes a latent gap the audit turned up: **`rest_api` was registered as a source and present in the `ProfileConfig` union but had no dispatch branch**, so `load_profile()` rejected it. It now loads through the same fallback.

## Type safety

The union is still concrete classes written out literally — which is exactly why the tags are hand-written rather than derived from each model's `Literal`. mypy still narrows `sync.destination`, so `drt/destinations/query.py` and every `isinstance()` site in `drt/destinations/` are untouched. **No built-in config class was modified.**

## One knock-on worth reviewing

`SyncConfig`'s generated JSON Schema renders the union as `oneOf`, so an open catch-all made every built-in payload match *two* members and fail `drt validate`'s schema pass (caught by `test_schema_validation.py`). `GenericDestinationConfig.__get_pydantic_json_schema__` pins its `type` to the plugin types the registry currently holds, mirroring the discriminator so schema and parser agree. Consequence: `drt schema` output reflects the plugins installed when it runs. I think that's correct — a static file can't describe types that arrive by `pip install` — but it's a real behaviour change and the piece I'd most like a second opinion on.

## Verification

Against pristine upstream `75719cf` with the repo's own commands:

| check | result |
|---|---|
| `pytest` (full) | **28 failed, 3154 passed** — byte-identical failure set to pristine upstream (28 pre-existing). **Zero new, zero fixed.** |
| `ruff check drt tests` | **All checks passed** |
| `mypy drt` | **1 error**, unchanged — `credentials.py:207`, pre-existing on pristine upstream (from #1001) and deliberately not touched |

21 new tests in `tests/unit/test_plugin_config_extensibility.py`, most of them guarding the negative space: built-ins still parse to concrete classes; invalid built-in still errors at `('destination','postgres','upsert_key')`; typo'd built-in still `union_tag_invalid`; missing `type` still `union_tag_not_found`; unregistered type rejected; registered plugin parses, keeps extras, resolves via `get_destination()`, and supports `describe`/`describe_safe`/`rate_limit_key`/`retry`/`rate_limit`; `model_dump()` round-trips; plugin cannot shadow a built-in; tag set matches the registry exactly (drift guard). Profile side mirrors it.

Three existing tests changed, each a real consequence rather than an accommodation: the retry-capable sweep 19 → 20 (the generic config carries both fields on purpose), and two in `test_plugins.py` that pinned the now-false "not yet usable in sync YAML" label.

## Docs

ADR 0009 gains a **Resolution** section (Context/Decision kept as written); `drt plugins list` no longer labels connectors unusable; `docs/guides/plugins.md` rewritten with a working example and the two caveats; CHANGELOG entry.

## Note on the base

Branched from `upstream/main` at `75719cf`. My fork's `main` was 39 commits behind and predates ADR 0009 entirely, so nothing here is based on it.


Muawiya-contact (commented):
Reviewed

drt/config/base.py
```diff
@@ -353,3 +355,155 @@ class RateLimitConfig(BaseModel):
...
+            # implementations, which import this module back.
+            from drt.connectors.registry import destination_config_class
+
+            config_class = destination_config_class(type_name)
+            if (
+                isinstance(config_class, type)
+                and issubclass(config_class, BaseModel)
+                and config_class is not cls
+            ):
+                return config_class.model_validate(value)
```
masukai, line 466: This is where the "`retry` and `rate_limit` are understood generically" claim (CHANGELOG, and `docs/guides/plugins.md`: "Both are understood generically… the same way they do to a built-in") breaks down for the case it's actually describing.

When a plugin registers a real pydantic `config_class` — the case this branch exists for — `config_class.model_validate(value)` validates against *that* class, not `GenericDestinationConfig`. If the plugin's model doesn't declare `retry`/`rate_limit` itself and doesn't subclass `GenericDestinationConfig` (and it won't — `docs/guides/building-a-destination.md`'s own example just extends `BaseModel`), those fields are silently dropped during validation. Not carried as extras, not raising — just gone:

```python
class AcmeConfig(BaseModel):
    type: str
    instance_url: str

registry.register_destination("acme_premium", AcmeConfig, SomeDestination)

dest = _sync({"type": "acme_premium", "instance_url": "https://x",
              "retry": {"max_attempts": 7}, "rate_limit": {"requests_per_second": 2.5}})
hasattr(dest, "retry")  # False
```

Every built-in destination calls `resolve_retry(config.retry, sync_options)` directly (no `getattr` guard), and the plugin guide tells authors their connector's `retry`/`rate_limit` handling "still works… the same way" — so a plugin author following that guide and doing the same in their own `load()` gets an `AttributeError` the moment someone sets `retry:` on their destination, not a working retry.

I think this needs either (a) requiring/documenting that a registered `config_class` must subclass `GenericDestinationConfig` to get this behavior, or (b) merging `retry`/`rate_limit` back onto the validated instance generically regardless of the plugin's base class. Right now it's neither — only the fallback-only path (no config_class registered) actually carries them.

masukai, line 466: Second issue on the same line — this one on the `SyncConfig` side.

`config_class.model_validate(value)` can return an instance of any pydantic model the registry hands back, not `GenericDestinationConfig` and not one of the 34 built-in `Tag` members either. But `SyncConfig.destination`'s declared type is still exactly that closed `Discriminator`-based union. Once a plugin's own config class parses, `sync.destination` holds an object outside its own field's declared type.

That's invisible as long as you only call `.model_dump()` on the destination object directly — which is all `test_plugin_config_extensibility.py` does (`test_generic_config_round_trips_through_model_dump` dumps `_sync(payload)`, never the containing `SyncConfig`) — but it breaks the moment you dump the *container*, which is what `drt validate --output json` / `drt docs generate` / the MCP tools actually do:

```python
sync = SyncConfig.model_validate({"name": "s", "model": "m",
    "destination": {"type": "acme_premium", "instance_url": "https://x"}})
sync.model_dump()
# UserWarning: Pydantic serializer warnings:
#   PydanticSerializationUnexpectedValue(Expected `RestApiDestinationConfig` -
#   serialized value may not be as expected […], input_type=AcmeConfig])
#   … (repeated per union member pydantic tries)
```

It still produces a dict via a permissive best-effort fallback, but that's a warning-driven pydantic escape hatch, not a supported contract, and it'll only get noisier as the union grows. Same fix shape as the comment above — either `config_class` has to be constrained to something that's actually a union member (a `GenericDestinationConfig` subclass), or the field's declared type needs to widen to admit arbitrary plugin classes, which gives up the mypy-narrowing property the PR is otherwise careful to preserve.

drt/config/credentials.py
```diff
@@ -595,6 +688,19 @@ def save_profile(
             entry["catalog_name"] = profile.catalog_name
         if profile.properties:
             entry["properties"] = profile.properties
+    elif is_dataclass(profile) and not isinstance(profile, type):
```
masukai, line 717: This branch is written for a plugin profile, but it also silently catches `RestApiProfile` — a built-in that never got its own `isinstance` branch above (the same gap the changelog calls out for `load_profile()`; `save_profile()` has always had it too, just less visibly, since it used to hard-fail here instead of at load time).

Before this PR, `save_profile(name, RestApiProfile(...))` raised `ValueError: Unknown profile type`. Now it silently succeeds through this dataclass fallback — and since `RestApiProfile.auth` is `dict[str, Any] | None` (the same free-form dict that lets `BearerAuth`/`ApiKeyAuth` carry a literal, non-`_env` secret), `asdict()` dumps it verbatim:

```python
save_profile("myrest", RestApiProfile(
    type="rest_api", url="https://api.example.com/v1",
    auth={"type": "bearer", "token": "sk-live-SECRET-VALUE-12345"},
))
```
writes straight to `~/.drt/profiles.yml`:
```yaml
profiles:
  myrest:
    auth:
      token: sk-live-SECRET-VALUE-12345
      type: bearer
    type: rest_api
    url: https://api.example.com/v1
```

Every hand-written branch above this one only ever persists `*_env` names precisely so a raw secret never lands in `profiles.yml`. This path doesn't participate in that convention at all — which is a deliberate tradeoff for a plugin's own fields (drt-core can't know its shape), but `rest_api` landing here is accidental, not something #997 meant to relax. I think `rest_api` wants its own explicit branch here, mirroring the one `load_profile()` just got, so it stays on the `*_env`-only path like every other built-in.

drt/config/base.py
```diff
@@ -353,3 +355,155 @@ class RateLimitConfig(BaseModel):
...
+        instantiated: every plugin destination became a
+        ``GenericDestinationConfig``, so the ``assert isinstance(config,
+        MyConfig)`` that ``docs/guides/building-a-destination.md`` tells authors
+        to write failed, and every default declared on their model was lost.
+
+        Falls through to this model whenever the registry has no usable pydantic
+        class for the type, which is also what keeps parsing possible before
+        registration has happened at all.
+        """
+        type_name = (
```
masukai: Nit, non-blocking: this type-extraction (`value.get("type") if isinstance(value, Mapping) else getattr(value, "type", None)`) is duplicated verbatim in `_destination_tag` over in `sync_options.py`. Might be worth one shared helper — there's a comment right above explaining why it has to be `Mapping` and not `dict`, and as two copies that reasoning only has to be remembered in one place to bit-rot in the other.

masukai (changes requested):
This is really thorough work — the callable-discriminator design (registry membership check, not validation, to keep the typo'd-builtin error path intact) is exactly the right shape for what ADR 0009 flagged, and the test suite here is unusually good at pinning the negative space (typo'd builtins, the sentinel tag, non-dict Mappings, the drift guard between `_BUILTIN_DESTINATION_TAGS` and the registry… that's a lot of edge cases nailed down).

Found two real issues in `_use_registered_config_class`, both specific to the "plugin supplies its own `config_class`" branch — the `GenericDestinationConfig` fallback path itself looks solid. Plus one in `save_profile()`'s new branch that looks like collateral damage rather than something #997 meant to change. Left inline comments on each with repros.

All three trace back to the same shape of gap: once a plugin's own class is doing the validating instead of falling back to the generic one, nothing keeps that class inside the contract the rest of the file assumes — union membership for destinations, `*_env`-only for profiles. Also left a small dedup nit, non-blocking.

Nothing here touches the built-in path or the registry-membership routing logic itself (`_destination_tag`, the three-way split) — that part looks right to me.

drt/config/base.py
Muawiya-contact: Confirmed and fixed: delegation now requires the registered `config_class` to subclass `GenericDestinationConfig`, so `retry`/`rate_limit` are inherited rather than silently dropped. A plain `BaseModel` no longer half-delegates — it falls back to `GenericDestinationConfig`, which carries the fields as extras and keeps retry working. Took your option (a); docs, CHANGELOG and `building-a-destination.md`'s example all state the subclassing contract now instead of the "understood generically" claim.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Muawiya-contact: Fixed by the same constraint — a subclass stays inside the declared union, so `SyncConfig.model_dump()` no longer falls through the warning path (asserted: zero serializer warnings). That alone dropped the subclass's own declared fields from the dump, since pydantic serializes via the declared type's schema, so the union member is now `SerializeAsAny[GenericDestinationConfig]`. mypy narrowing is untouched.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

drt/config/credentials.py
Muawiya-contact: Fixed — `rest_api` has its own explicit branch ahead of the dataclass fallback, mirroring `load_profile()`. It persists the shape only, and a literal (non-`*_env`) value in `auth` is now refused with a message pointing at the `*_env` key rather than written to profiles.yml. Agreed this was collateral damage, not something #997 meant to relax; two tests cover the refusal and the env-var form.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

drt/config/base.py
Muawiya-contact: Taken — both copies now call one `destination_type_of()` helper in `base.py`, with the Mapping-vs-dict reasoning in that one place.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Muawiya-contact:
@masukai — one scope question I'd rather have stated than assumed before you re-review.

#997's text and ADR 0009 both explicitly deferred registry-backed validation ("`config_class` is already stored and can be used for future plugin validation, but strict registry-backed validation does not need to be part of this first change"). The `_use_registered_config_class` wrap validator in `drt/config/base.py` **is** that feature, or close enough that I don't want it landing by default.

It wasn't in the original design — it came out of the review pass. Without it, every plugin destination parsed as `GenericDestinationConfig`, which meant:

- the `assert isinstance(config, MyConfig)` that [`building-a-destination.md`](docs/guides/building-a-destination.md) tells authors to open `load()` with **fails**, and
- every default declared on the plugin's own model is silently lost (`config.method` → `AttributeError` rather than `"POST"`).

So the feature made plugin types *nameable* but not *usable* per our own documented contract. Closing that seemed better than shipping a half-working extension point — but it is a scope call, not mine to make quietly.

Your review engaged with the branch's details rather than its existence, and the subclassing constraint you pushed for makes it a much tighter contract than anything ADR 0009 sketched (delegation only to `GenericDestinationConfig` subclasses — no attempt to validate arbitrary classes). I've read that as acceptance, but flagging it explicitly:

**Option A — keep it (what's on the branch).** #997 delivers working plugin destinations end to end. The deferred item is effectively done, but narrowly: only for registered classes that subclass the catch-all. Field-level validation against arbitrary registered classes is still not attempted, so ADR 0009's "second pass" is still open in the general case.

**Option B — drop it.** Revert the wrap validator, ship the nameable-but-generic version, and file delegation as the follow-up the ADR anticipated. That's a small revert plus a caveat in `building-a-destination.md` saying a plugin's `load()` receives a `GenericDestinationConfig`, not its own class — and the guide's `assert isinstance(...)` line would have to go.

Happy to do B if you'd rather keep #997 to its stated scope; it's maybe twenty minutes. I'd just rather not merge A silently and have it discovered later as scope that crept in during review.

ADR 0009's Resolution section currently documents A. If you pick B I'll rewrite it to match.

_🤖 Posted by [Claude Code](https://claude.com/claude-code) on behalf of @Muawiya-contact_

masukai:
Option A — keep it. Agreed the subclassing constraint (delegation only to `GenericDestinationConfig` subclasses, no attempt at arbitrary-class validation) keeps this inside what ADR 0009 sketched rather than quietly delivering its deferred "second pass." Shipping something that contradicts `building-a-destination.md`'s own `assert isinstance(...)` guidance would be worse than the scope creep here. Appreciate you flagging it explicitly rather than letting it ride.

ADR 0009's Resolution section can stay as-is documenting A. Thanks for the fast turnaround on all four review findings, especially the profile secret leak — that one mattered.

masukai (approved):
All three findings fixed and re-verified directly against the branch (not just re-reading the diff):

- `retry`/`rate_limit` inheritance: confirmed a non-`GenericDestinationConfig` config class now falls back to the generic config (preserving both fields) instead of losing them; a subclass delegates correctly and also preserves them.
- Union serialization: `SyncConfig.model_dump()` on a delegated subclass instance now emits zero serializer warnings.
- Profile secret leak: `save_profile()` with a `RestApiProfile` carrying a literal (non-`_env`) `auth` value now raises `ValueError` instead of writing the secret to `profiles.yml`.

Also re-ran `ruff`, `mypy`, and the full test suite locally — clean.

Thanks for the thorough turnaround, especially on the secret-leak fix.

PR #1012, opened 2026-08-26
fix(state): cross-process file locking for local DLQ/state/history stores

### LINKED ISSUE #963: fix(state): local backend has no cross-process write safety — LocalDlqStore/LocalStateManager/LocalHistoryManager
## Problem

All three local state stores (`LocalDlqStore`, `LocalStateManager`, `LocalHistoryManager`) use a `threading.Lock` for their read-modify-write cycle — process-local by construction, and each class's own docstring says so plainly ("does not coordinate across processes", "single-writer-at-a-time is the expected operational model"). Two genuinely separate `drt` CLI invocations (the realistic case — `drt run` and `drt retry`/`drt status` are always separate OS processes, never threads sharing one lock) racing a read-modify-write on the same file is last-writer-wins, full stop, no matter how carefully the in-process side of that cycle is written.

Flagged during review of #962 (`LocalDlqStore.reconcile()`, #955's fix): reconciling by identity against a fresh read closes the race for `ObjectStoreDlqBackend` (real generation/ETag preconditioning), but for `LocalDlqStore` a concurrent process's append between the read and the write is still fully lost — the method has no way to detect the file changed underneath it. #962 documents this explicitly rather than overclaiming a fix; this issue tracks actually closing it.

## Proposal

OS-level file locking (`fcntl.flock` on POSIX, `msvcrt.locking` on Windows, or a small cross-platform wrapper — no existing dependency in this codebase does this today, confirmed via grep) around each store's read-modify-write cycle, so a second process blocks (or a conditional-write pattern analogous to the object-store primitive's generation check, using the file's mtime/size as a cheap local stand-in for a generation token) rather than silently racing.

## Scope

**In:** all three local stores, since they share the identical process-local-lock limitation and a fix to one without the others would be an inconsistent contract across `StateStore`/`HistoryStore`/`DlqBackend`.
**Out:** changing the documented "single-writer-at-a-time" default posture for users who don't hit this — this is about actually enforcing safety when two processes *do* overlap, not changing the recommended deployment model.

## Related

- #955 / #962 — where this was found (DLQ specifically)
- #919 — the equivalent bug already fixed for the GCS-backed watermark store, using the object-store tier's real preconditioned-write primitive; this issue is the local-tier analog
- ADR 0005 — explains why the *remote* tier was built with real concurrency safety from the start (Decision 1's "team-shared" requirement); local was never designed for concurrent writers, which is exactly why this gap exists

## Milestone fit

Not currently milestoned. Worth a look before any messaging that positions the local backend as safe for concurrent `drt run`/`drt retry`/`drt status` invocations — today's actual guarantee is "works fine single-writer, races silently otherwise," which the class docstrings already say but which #955's reconcile() fix doesn't extend to.

Pawansingh3889:
Closes #963.

## Problem

`LocalDlqStore`, `LocalStateManager`, and `LocalHistoryManager` each guarded their read-modify-write cycle with a `threading.Lock`, which only serialises threads inside one process. Two genuinely separate `drt` CLI invocations (`drt run` and `drt retry`/`drt status` are always separate OS processes, never threads) racing the same file was last-writer-wins, with no way to detect it.

Flagged during review of #955/#962: `reconcile()` closed the race for the object-store backends via real generation/ETag preconditioning, but the local file backend had no equivalent primitive, so a concurrent process appending between the read and the write was still silently lost.

## Fix

Adds `drt/state/_file_lock.py`: an OS-level advisory lock (`fcntl.flock` on POSIX, `msvcrt.locking` on Windows) around a dedicated per-resource lock file, with a bounded timeout so a genuine deadlock surfaces as `FileLockTimeout` instead of hanging forever.

Wired into every read-modify-write path:
- `LocalDlqStore.append`/`replace`/`reconcile`
- `LocalStateManager.save_sync`/`reset` (raising `StateContentionError` on timeout, matching the `StateStore` Protocol's published contract from #992)
- `LocalHistoryManager.append`/`prune`

One thing that surfaced while writing the cross-process tests: `LocalHistoryManager.append` needed the lock too, not just `prune`, even though `O_APPEND` already makes its own write atomic against other appends. Without it, an append could still land between `prune`'s read and its replace and be silently discarded, since `flock` only serialises processes that both take it.

## Tests

Real `multiprocessing.Process` workers (spawn context) racing each store, since an in-process mock can't reproduce a cross-process race, per the issue's own note. Verified they actually catch the regression by temporarily stubbing the lock to a no-op: all four failed reliably, including one run that corrupted `state.json` outright. Restored the real implementation and confirmed all pass.

`make test`: 3525 passed (full suite plus the 9 new tests), 40 skipped. One pre-existing, unrelated failure locally (`test_invalid_key_fails_with_actionable_message`, missing the optional `pyrage`/encryption extra) confirmed to fail identically on unmodified `main`.

### DIFF
--- a/drt/state/_file_lock.py
+++ b/drt/state/_file_lock.py
@@ -1,0 +1,109 @@
+"""Cross-process file locking for the local (non-object-store) state backend (#963).
+
+``LocalDlqStore`` / ``LocalStateManager`` / ``LocalHistoryManager`` each guard
+their read-modify-write cycle with a ``threading.Lock``, which only
+serialises threads sharing one process. Two separate ``drt`` CLI invocations
+(``drt run`` and ``drt retry``/``drt status`` are always separate OS
+processes, never threads) racing the same file is last-writer-wins with no
+way to detect it (#955, #962). :func:`cross_process_lock` wraps the same
+critical section in an OS-level advisory lock (``fcntl.flock`` on POSIX,
+``msvcrt.locking`` on Windows), so a second process blocks until the first
+one finishes instead of racing it.
+
+This is a real mutex, not the object-store tier's optimistic
+generation/ETag retry (:mod:`drt.state._objectstore`): a local file has no
+version token to precondition a write against, so callers get exclusion
+instead of conflict detection. Advisory OS locks are held by the kernel
+against the open file description, not the file's content, so a crashed
+holder releases the lock automatically when its process exits: there is no
+stale-lock file to clean up the way a plain "lock file exists" convention
+would need.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+import time
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+
+DEFAULT_TIMEOUT_SECONDS = 30.0
+_POLL_INTERVAL_SECONDS = 0.05
+
+
+class FileLockTimeout(TimeoutError):
+    """Could not acquire the cross-process file lock within the timeout.
+
+    No other ``drt`` process is expected to hold a state/DLQ/history lock
+    for more than a handful of read-modify-write cycles. A caller stuck
+    this long is either genuinely deadlocked behind another process (killed
+    mid-hold on a platform where that doesn't release the lock, or a real
+    hang), or advisory locking isn't actually supported on this filesystem.
+    Both are conditions to surface, not to wait out silently.
+    """
+
+
+if sys.platform == "win32":
+    import msvcrt
+
+    def _try_lock(fd: int) -> bool:
+        try:
+            msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
+            return True
+        except OSError:
+            return False
+
+    def _unlock(fd: int) -> None:
+        os.lseek(fd, 0, 0)
+        msvcrt.locking(fd, msvcrt.LK_UNLCK, 1)
+
+else:
+    import fcntl
+
+    def _try_lock(fd: int) -> bool:
+        try:
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            return True
+        except OSError:
+            return False
+
+    def _unlock(fd: int) -> None:
+        fcntl.flock(fd, fcntl.LOCK_UN)
+
+
+@contextmanager
+def cross_process_lock(
+    lock_path: Path, *, timeout: float = DEFAULT_TIMEOUT_SECONDS
+) -> Iterator[None]:
+    """Hold an exclusive OS-level lock on ``lock_path`` for the block's duration.
+
+    ``lock_path`` is a dedicated lock file, created if missing and never
+    removed (deleting it after use would let a concurrent locker acquire a
+    lock on a since-recreated inode while this process still holds the
+    original one, defeating the exclusion). Callers pick a lock path scoped
+    to the resource they're protecting (e.g. ``state.json.lock`` for the
+    whole state file, ``<sync>.jsonl.lock`` per DLQ/history file) so
+    unrelated resources never contend on the same lock.
+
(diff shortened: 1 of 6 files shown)


masukai (changes requested):
This is genuinely solid work — the lock-nesting logic is correct throughout (process-local lock outside, OS-level lock inside, matching each store's documented error contract: `StateContentionError` for state, best-effort swallow for history, propagate for DLQ), and I did my own regression check: stubbing `cross_process_lock` to a no-op made all four new cross-process tests fail exactly as the description claims, which is exactly the kind of evidence I want to see for a fix like this.

One real design issue before merge, though: `LocalHistoryManager`/`LocalDlqStore._lock` is a single `threading.Lock` shared across every sync name in the instance, not scoped per sync. Under `drt run --threads N`, if a separate process is holding sync A's file lock, A's thread now blocks up to 30s *while holding that global lock* — which stalls every other thread working on unrelated syncs B–Z too, even though their own per-file locks are free. Before this PR the global lock only ever guarded a fast in-memory read-modify-write, so holding it was cheap; now it can hold for the full timeout. Not a correctness bug (nothing's lost or corrupted), but it can turn one contended sync into a whole-run stall under `--threads`. Worth moving to a per-sync lock (or not holding the global mutex while waiting on the OS lock) before this ships — same pattern in both `LocalHistoryManager` and `LocalDlqStore`.

Also needs a CHANGELOG entry — I don't see one under `[Unreleased]` for #963.

Pawansingh3889:
Closing this, superseded by #1031.
I see #963 was closed on 30 August by your own implementation, with #1032 and #1035 following on. Having read through it, _filelock.py covers more ground than this PR did: LocalWatermarkStorage and LocalStateManager as well as the DLQ and history stores.

The branch now conflicts with main in all three files it touches, which is the same fact from another angle, since #1031 rewrote the same read-modify-write paths. Rebasing would mean dropping this implementation in favour of the one already on main, so closing is the cleaner outcome.

Thank you for the review here, and for doing your own regression check by stubbing out the lock. That was a useful thing to see.

One note on the point you raised about self._lock being shared across sync names rather than scoped per sync. Your implementation avoids that by construction: taking advisory_lock(path) on the outside and self._lock on the inside means a thread waiting on one sync's file lock never holds the shared mutex, so unrelated syncs are not stalled. This PR had the nesting the other way round, which is what made the shared lock a problem in the first place. Might be worth a comment somewhere that the ordering is load bearing, in case it gets refactored later.

The one difference left is that advisory_lock blocks indefinitely, where this PR bounded the wait at 30 seconds and raised FileLockTimeout. Indefinite blocking is probably the right default for a local advisory lock, since timing out risks discarding the write it was protecting. If a bounded wait is ever wanted for drt serve or a CI runner, I am happy to open a separate issue rather than reopening this.

masukai:
Thank you for closing this so gracefully, and for the careful read-through of #1031 — that's exactly the kind of review I'd want on my own work too.

Apologies for the duplicated effort here — I should have checked for open PRs touching `drt/state/` before starting on #963 and didn't, so I ended up re-discovering and re-implementing something you'd already largely built. That's on me, not a reflection on the quality of what you submitted.

Your read on the lock ordering is exactly right, and it genuinely wasn't something I'd thought to write down as a deliberate invariant rather than an implementation detail. I'll add a comment in `_filelock.py`/the three stores making explicit that `advisory_lock(path)` outside, `self._lock` inside is load-bearing — a future refactor swapping that order would silently reintroduce the whole-run stall your review of your own PR caught.

On the bounded-wait variant: please do open that issue if you'd like to build it — a `FileLockTimeout` option for `drt serve`/CI-runner use cases sounds like a genuinely useful addition on top of what's on `main` now, and you clearly already have the design worked out. Happy to review whenever it's ready.

PR #1013, opened 2026-08-26
Cross-process rate-limit coordination: pluggable extension point (#921)

### LINKED ISSUE #921: feat(engine): cross-process rate-limit coordination for orchestrator-launched sync processes

masukai:
## Summary

Closes #921 at its own stated decision/documented-workaround bar. Implements [ADR 0012](docs/adr/0012-cross-process-rate-limit-coordination.md):

- **Default behavior is unchanged.** The local `RateLimiter`, pacing algorithm, `rate_limit` precedence, vendor ceilings, burst semantics, and min-wins tightening are untouched.
- `RateLimiter`'s implicit shape becomes the public `RateLimiterBackend` Protocol; a single-active-factory registry (`register_rate_limiter_backend`/`get_rate_limiter_backend`) lets a third-party package supply cross-process coordination backed by infrastructure the operator already runs, via a new `drt.rate_limiter_backends` entry-point group.
- **drt-core ships no Redis client or reference backend** — this is a contract to implement against, per ADR 0011.
- The zero-code default mitigation (batch changed syncs into one process via `--select state:modified`, #772) is documented in new `docs/guides/rate-limiting.md`.

## Review history

Went through four Codex review rounds; every finding independently verified with a live repro (not just re-reading the diff) before being fixed:

1. **`LimiterFactory` needed a `key` parameter** — without it, a distributed backend has no stable identity to derive a shared bucket name from, making the cross-process contract unimplementable as originally specced.
2. **17 of 20 rate-limited destinations already pass `limiter_factory=RateLimiter`** (for the test-patch seam), which silently shadowed any registered backend. Fixed by treating the bare, unpatched class as "no override" — a test's patched mock or a genuinely different factory still wins.
3. **Plugin discovery only ran in the CLI's root callback** — `drt.rate_limiter_backends` entry points stayed unregistered under dagster-drt/Airflow/Prefect/MCP, the exact orchestrator-launched scenario #921 is about. `get_rate_limiter_backend()` now triggers discovery itself, mirroring `drt/connectors/registry.py`'s existing fix for the same class of gap.
4. **That fix introduced a same-thread reentrancy deadlock** (a registration callback reading the active backend re-enters `load_plugins()` while the outer call holds its non-reentrant lock) — reproduced as a genuine indefinite hang, fixed with a `threading.local()` guard. Also fixed: the guide's example backend wrote `key` verbatim into a Redis key, contradicting `rate_limit_key()`'s existing "do not log or serialize" contract.

## Test plan

- [x] `mypy drt` clean
- [x] Full `pytest tests/unit/` — clean except 2 pre-existing, unrelated `test_cli_version.py` failures caused by this worktree's long absolute path wrapping the `--version` output (confirmed the same signature on multiple unrelated branches this session)
- [x] Live-reproduced the deadlock fix (killed a genuinely hung process before fixing it) and the plugin-discovery gap (built-in destination parsing does not trigger `load_plugins()`, confirmed empirically)
- [x] New regression tests for every fix, each confirmed to fail without its corresponding fix


masukai:
**Codex review** (4 rounds, summarized — full narrative in the PR description):

**Round 1** — [P1] The extension point can't work as specced: `LimiterFactory` never receives the endpoint key, so a distributed backend has no stable identity to derive a shared bucket from. [P1] Most destinations already pass `limiter_factory=RateLimiter` explicitly (for the test-patch seam), which shadows any registered backend for 17 of 20 rate-limited destinations. [P2] The new `docs/guides/rate-limiting.md` was untracked in the diff.

**Round 2** — [P1] `resolve_rate_limiter()`'s callers used through dagster-drt/Airflow/Prefect/MCP never trigger plugin discovery, since that historically only happened in the CLI's root callback — the exact orchestrator-launched scenario #921 targets.

**Round 3** — [P2] The fix for round 2's finding introduced a same-thread reentrancy deadlock: a registration callback reading the active backend re-enters `load_plugins()` while the outer call still holds its non-reentrant lock. [P2] The guide's example backend wrote the endpoint `key` verbatim into a Redis key, contradicting `rate_limit_key()`'s "do not log or serialize" contract.

**Round 4** — clean, no findings.

All six findings verified against the live code before fixing (not just re-reading the diff) — including reproducing the deadlock as a genuine indefinite hang (killed the hung process, then fixed it) and confirming empirically that parsing a built-in destination type does not itself trigger `load_plugins()`. Details and fixes are in the commit message and PR description.

PR #1014, opened 2026-08-27
feat(cli): Stripe timestamped-signature verification for drt serve (#969)

### LINKED ISSUE #969: feat(cli): Stripe timestamped-signature verification for drt serve
## Background

#854 (shipped v0.9.0) added pluggable `drt serve` auth (`--auth none|bearer|hmac`) with HMAC-SHA256 body-signature verification covering GitHub (`sha256=<hex>`), Shopify (base64), and bare-hex senders. The CHANGELOG entry for #854 explicitly named what it does *not* cover:

> Stripe's timestamped scheme and Pub/Sub's OIDC JWT are *not* covered — OIDC is a filed follow-up (#903 — it needs a JWT dependency decision), and until it lands Pub/Sub push still needs a verifying proxy.

OIDC got tracked as #903. Stripe's scheme didn't get its own issue — this is that follow-up, filed for parity with #903 rather than left as a dangling CHANGELOG mention.

## Why Stripe needs a different code path

Stripe's `Stripe-Signature` header is not a plain HMAC-over-body like GitHub/Shopify: it's `t=<timestamp>,v1=<hex>[,v0=<hex>]`, and the signed payload is `<timestamp>.<body>`, not the body alone. Verification also requires a replay-window check (Stripe's own libraries default to 5 minutes) comparing `t` against wall-clock time — a check the existing GitHub/Shopify/bare-hex schemes have no equivalent of, since none of them sign a timestamp.

## Scope

- New `--hmac-scheme stripe` (or equivalent) alongside the existing header-name/encoding options in `drt/cli/commands/serve.py`'s HMAC auth path
- Parse `Stripe-Signature`, verify `v1` against `HMAC-SHA256(secret, f"{t}.{body}")`
- Reject on stale timestamp (configurable tolerance, default matching Stripe's own 5-minute recommendation)
- Docs: `docs/guides/using-webhook-trigger.md`'s per-scheme auth setup table gains a Stripe row alongside GitHub/Shopify/bare-hex

## Out of scope

- Pub/Sub OIDC JWT (#903, separate follow-up, separate dependency decision)
- Any Stripe-specific *source* (this is only about verifying an inbound webhook trigger, not reading Stripe data)

## References

- #854 (parent — `drt serve` delivery contract + pluggable auth)
- #903 (sibling follow-up — OIDC JWT for Pub/Sub, same "named gap, filed separately" pattern)
- `docs/guides/using-webhook-trigger.md` — the contract + per-scheme auth setup doc to extend

Muawiya-contact:
Closes #969.

## Why it's a separate verifier

Stripe doesn't sign the body. `Stripe-Signature` is `t=<ts>,v1=<hex>[,v0=...]` and the signed payload is `<ts>.<body>` — the timestamp sits *inside* the signature, which is exactly what makes a replay window enforceable. That's a different construction, not a fourth encoding of the same digest, so it's `_verify_stripe_hmac()` beside `_verify_hmac()` rather than another candidate in the existing tuple.

`AuthConfig`'s docstring already anticipated this ("schemes are dispatched rather than hardcoded so a later verifier is an addition, not a rewrite"), and that held — nothing about the generic path changed.

## Surface

- `--hmac-scheme generic|stripe`, default `generic` = **#854's behaviour byte for byte**, so this is additive
- `--hmac-tolerance` (default 300, matching Stripe's libraries)
- `--hmac-header` now defaults per scheme (`Stripe-Signature` under `stripe`); an explicit value still wins

## Three details I checked against Stripe's spec rather than assumed

Each one silently breaks the scheme if you get it wrong, and the issue text didn't name the first two:

**`v0` is ignored.** Stripe attaches a deliberately fake `v0` to test events, and its docs are explicit: *"to prevent downgrade attacks, ignore all schemes that aren't v1."* An implementation that loops over "any `v*` that matches" is exploitable. A v0-only header is rejected.

**Multiple `v1` signatures are accepted.** Rolling an endpoint secret leaves the old one live for up to 24h and Stripe sends *one signature per active secret*. Stopping at the first mismatch drops deliveries through every rotation — and it wouldn't show up in testing.

**The replay window is symmetric.** Stripe's own libraries reject only *stale* timestamps, which leaves a future-dated signature valid indefinitely. Clock skew runs both ways; bounding both costs nothing. This is a deliberate, documented divergence from their reference behaviour.

Also: `--hmac-tolerance 0` is rejected at both the CLI and `AuthConfig` — Stripe's docs call it out as *disabling* the recency check rather than tightening it. Comparison is constant-time and deliberately doesn't short-circuit on first match, so the number of candidates tried doesn't leak.

## Scope call worth your review

**`--hmac-scheme stripe` covers POSTs only.** Stripe never sends a GET, so there's no `Stripe-Signature` on one and `GET /runs/<id>` answers 401 under this scheme.

I documented that with two workarounds (poll from the process that got the 202, or run a second listener on `--auth bearer`) rather than inventing a drt-side convention like signing `t.path` — that would be a signature Stripe would never produce, and `--hmac-scheme stripe` would stop meaning "Stripe-compatible". Happy to revisit if you'd rather it be uniform across methods.

## Interaction with #998

[#998](https://github.com/drt-hub/drt/pull/998) changes `_verify_hmac`'s signature (`+ path: str = ""`) and its call site. This branch is off `main` and touches the same `_check_auth` block, so **whoever merges second rebases** — it's a small conflict either way, since Stripe POSTs always have a body and #998 only alters the empty-body case.

Two things I noticed reading #998, neither in scope here: it carries ~51 lines of unrelated CLAUDE.md ASCII churn (`—`→`--`, `→`→`->`), and `path.encode() + b"\x00"` means a POST whose body is literally `/runs/xyz\0` collides with a GET signature for `/runs/xyz`.

## Verification

23 new tests: valid signature, body tampering, stale timestamp, symmetric window (with an injected clock, no sleeping), `v0` alongside, `v0`-only rejection, secret rotation with two `v1`s, whitespace after commas, missing/garbage/duplicate `t`, cross-scheme rejection in both directions, and CLI validation for the new flags.

| check | result |
|---|---|
| full suite | **28 failed, 3185 passed** — identical failure set to pristine `upstream/main` (28 pre-existing). Zero new. |
| `ruff check drt tests` | clean |
| `mypy drt` | 1 error, unchanged and pre-existing (`credentials.py:209`) |

Docs gain a Stripe section, including the raw-body / TLS-proxy caveat — `drt serve` speaks plain HTTP and Stripe requires HTTPS for live endpoints, so a terminating proxy is mandatory, and a proxy that re-serializes JSON is the most common cause of Stripe verification failures.


masukai (changes requested):
This is excellent work — the three Stripe-specific details you called out (v0 downgrade rejection, multi-v1 secret rotation, symmetric replay window) are all real and correctly handled, and I verified all three directly against `_verify_stripe_hmac`. The scope call on GET-under-stripe-scheme also makes sense to me — inventing a drt-side `t.path` convention would make `--hmac-scheme stripe` stop meaning "what Stripe actually sends," so I'd rather keep the documented workaround than uniformity for its own sake.

One real issue found and reproduced directly, both confirmed live:

**Unauthenticated input can raise out of `_verify_stripe_hmac` instead of returning `False`.**

1. A `v1` candidate with non-ASCII bytes (e.g. `v1=é`) makes `secrets.compare_digest(expected, candidate)` raise `TypeError` — that function only accepts ASCII-only `str` operands (or bytes, which has no such restriction). `_verify_hmac` above already avoids this by comparing as bytes (`header_value.encode()` vs `candidate.encode()`); `_verify_stripe_hmac` compares as `str` and doesn't.
2. A `t` with a few hundred digits parses fine via `int(timestamp)` (Python ints are arbitrary precision) but raises `OverflowError` the moment it's used in `abs(current - signed_at)` — converting a too-large int to `float` overflows.

Repro:
```python
>>> _verify_stripe_hmac(body, "t=1000000,v1=é", secret, 300, now=1000000.0)
TypeError: comparing strings with non-ASCII characters is not supported
>>> _verify_stripe_hmac(body, f"t={'9'*400},v1=deadbeef", secret, 300, now=1000000.0)
OverflowError: int too large to convert to float
```

Since `_check_auth` doesn't catch these, an unauthenticated client can trigger them on demand — connection reset plus a traceback in the server logs, on every route this scheme guards.

Suggested fix (verified — both regression cases now return `False`, existing tests + 2 new parametrized cases all pass, and reverting the fix makes the new cases fail as expected):

```python
    try:
        signed_at = int(timestamp)
        current = time.time() if now is None else now
        if abs(current - signed_at) > tolerance:
            return False
    except (ValueError, OverflowError):
        return False

    expected = hmac.new(
        secret.encode(), f"{timestamp}.".encode() + body, hashlib.sha256
    ).hexdigest().encode()
    return (
        sum(
            secrets.compare_digest(expected, candidate.encode())
            for candidate in signatures
        )
        > 0
    )
```

(Comparing as bytes throughout, and widening the `except` to also catch `OverflowError` around the arithmetic, not just the `int()` parse.)

masukai:
Just noting for visibility: with #998 merged, this branch now shows a merge conflict in the `_check_auth` block (as you anticipated in the PR description). Whenever you get to the CHANGES_REQUESTED items, a rebase onto current `main` will be needed alongside the fix.

No rush on my end — flagging so it doesn't surprise you when you pick this back up.

Muawiya-contact:
@masukai don't worry, I'll handle it. Thanks for your patience.

Muawiya-contact:
@masukai Thanks, both confirmed and fixed.

Compared as bytes throughout now, matching `_verify_hmac`, and the window
arithmetic moved inside the `try` with `OverflowError` added, so the huge-`t`
case is caught where it actually raises rather than at the parse. Added your
two cases to `test_stripe_header_parsing_edge_cases`; both fail on the old code.

Also rebased onto #998 since it landed first. One thing that came out of that:
a GET under `--hmac-scheme stripe` is now an explicit 401 rather than falling
through to a check against the empty body. Same outcome as before, but for a
stated reason. A Stripe signature covers `<t>.<body>`, which on a bodyless GET
carries no path, so honouring one would have handed back exactly the replay you
closed in #998. Documented and tested.

Small correction while I was in there: the "23 new tests" in my description was
an overcount. It's 19 collected (12 functions, one parametrized over 8), measured
against a baseline rather than counted by hand.

Muawiya-contact:
@masukai, Please feel free to assign me more tasks; I'm happy to help out.

Muawiya-contact (commented):
LGTM

masukai (approved):
Confirmed — both regression cases now return \`False\` cleanly instead of raising, verified independently against the fixed \`_verify_stripe_hmac\`:

\`\`\`
>>> _verify_stripe_hmac(body, "t=1000000,v1=é", secret, 300, now=1000000.0)
False
>>> _verify_stripe_hmac(body, f"t={'9'*400},v1=deadbeef", secret, 300, now=1000000.0)
False
\`\`\`

Bytes comparison throughout matching \`_verify_hmac\`, and the \`except (ValueError, OverflowError)\` correctly wraps the arithmetic too, not just the parse. New parametrized test cases (\`non_ascii_signature\`, \`oversized_timestamp\`) cover exactly this. Full test file (62 tests) + ruff + mypy all clean locally.

Thanks for the quick turnaround on this, and for offering to pick up more — appreciated as always 🙏

PR #1015, opened 2026-08-27
dwh-smoke: explicitly start the Databricks warehouse before connecting

masukai:
## Summary

The smoke warehouse (Serverless Starter Warehouse) auto-stops after idle time. `databricks-sql-connector`'s implicit auto-start-on-connect proved unreliable against it — observed 2026-08-27: three attempts over ~10h all failed identically at session-open with `BAD_REQUEST: Cannot create the resource, please try again later`. Manually clicking Start in the Databricks UI succeeded immediately, and the very next scheduled run passed once the warehouse was already running.

- Adds a step to `dwh-smoke.yml`'s Databricks job that starts the warehouse via the SQL Warehouses REST API (`POST .../start`) and polls `GET .../warehouses/{id}` until `state == RUNNING` (5 minute timeout) before handing off to pytest.
- Uses the existing `SMOKE_DATABRICKS_HOST`/`HTTP_PATH`/`TOKEN` secrets — no new secrets needed. The warehouse ID is parsed from the existing `SMOKE_DATABRICKS_HTTP_PATH` secret's path.
- Updated `tests/integration/dwh/README.md`'s prerequisites note (previously said the warehouse just needs to be "running" — now the workflow ensures that itself).

[skip changelog] — CI/workflow reliability fix only, no user-facing drt-core behavior change.

## Test plan

- [x] YAML validated with `python3 -c "import yaml; yaml.safe_load(...)"`
- [x] Dry-run tested the bash logic locally against a mock `curl` for both the success path (STOPPED → STARTING → RUNNING) and the timeout-failure path — both behave correctly
- [x] Root-caused live against the actual failing workflow run: manually started the warehouse in the Databricks UI, re-ran the smoke job while it was warm, confirmed it passed

### DIFF
--- a/.github/workflows/dwh-smoke.yml
+++ b/.github/workflows/dwh-smoke.yml
@@ -115,6 +115,45 @@ jobs:
       - name: Install drt-core[dev,duckdb,databricks]
         if: steps.creds.outputs.has_creds == 'true'
         run: uv pip install --system -e ".[dev,duckdb,databricks]"
+      # The smoke warehouse auto-stops after idle time, and the databricks-sql-
+      # connector's implicit auto-start-on-connect proved unreliable against it
+      # (observed 2026-08-27: repeated `BAD_REQUEST: Cannot create the resource,
+      # please try again later` at session-open, three attempts over ~10h, while
+      # a manual Start from the Databricks UI succeeded immediately and let the
+      # very next run pass). Start it explicitly via the REST API and wait for
+      # RUNNING before handing off to pytest, rather than relying on the SQL
+      # client to trigger and outlast a cold start on its own.
+      - name: Start Databricks SQL warehouse and wait until running
+        if: steps.creds.outputs.has_creds == 'true'
+        run: |
+          set -euo pipefail
+          warehouse_id="${DRT_SMOKE_DATABRICKS_HTTP_PATH##*/}"
+          api="https://${DRT_SMOKE_DATABRICKS_HOST}/api/2.0/sql/warehouses/${warehouse_id}"
+          auth=(-H "Authorization: Bearer ${DRT_SMOKE_DATABRICKS_TOKEN}")
+
+          state=$(curl -sf "${auth[@]}" "$api" | jq -r '.state')
+          echo "Warehouse ${warehouse_id} current state: ${state}"
+
+          if [ "$state" != "RUNNING" ]; then
+            echo "Starting warehouse..."
+            curl -sf -X POST "${auth[@]}" "${api}/start" >/dev/null
+
+            deadline=$((SECONDS + 300))
+            while [ "$SECONDS" -lt "$deadline" ]; do
+              state=$(curl -sf "${auth[@]}" "$api" | jq -r '.state')
+              echo "  state=${state} (${SECONDS}s elapsed)"
+              if [ "$state" = "RUNNING" ]; then
+                break
+              fi
+              sleep 10
+            done
+
+            if [ "$state" != "RUNNING" ]; then
+              echo "::error::Warehouse did not reach RUNNING within 300s (last state: ${state})"
+              exit 1
+            fi
+          fi
+          echo "Warehouse is running."
       - name: Run Databricks smoke
         if: steps.creds.outputs.has_creds == 'true'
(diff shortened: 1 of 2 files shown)


masukai:
**Codex review:**

The warehouse-start logic is syntactically valid, but a single transient failure while polling aborts the workflow rather than honoring its intended startup timeout. This can leave the Databricks smoke job flaky.

- [P2] Keep transient status failures inside the polling window — `.github/workflows/dwh-smoke.yml:143`
  If a status request during warehouse startup encounters a transient 429/5xx response or network error, `curl -f` returns nonzero and `set -euo pipefail` aborts the step immediately instead of using the remaining 300-second polling budget. Handle or retry failed status requests until the deadline so the new preflight does not introduce flaky smoke failures.

Verified and fixed in 8fe4fbb: removed the global `set -e` in favor of explicit handling at each failure point — a bounded 3-attempt retry for the two one-shot calls (initial status check, start request), and tolerant retry-until-deadline for the polling loop's own checks. Dry-run tested locally against a mock `curl` for the exact transient-failure-mid-poll scenario (now survives and reaches `RUNNING`), plus the success, permanent-initial-failure, and stuck-timeout paths.

masukai:
**Codex review (round 2):**

The startup logic works for a fully configured secret set, but it breaks the workflow's credential-agnostic contract when Databricks secrets are only partially configured.

- [P2] Gate startup on every required API credential — `.github/workflows/dwh-smoke.yml:127`
  When `SMOKE_DATABRICKS_HOST` is configured but `SMOKE_DATABRICKS_HTTP_PATH` or `SMOKE_DATABRICKS_TOKEN` is still absent, this condition is true because the credential step checks only the host. The new step then calls the API with an empty warehouse ID or token and fails the job, whereas the existing `require_env` gate previously skipped the tests and preserved the workflow's documented partial-configuration no-op behavior.

Verified against the actual "Check credentials" step (it only checked `DRT_SMOKE_DATABRICKS_HOST`) and fixed in c1b5199: the gate now requires `HOST`, `HTTP_PATH`, and `TOKEN` together — all three are what the new REST API step actually needs. Tested all three states locally (all set / host-only / none set) — only the fully-configured case now produces `has_creds=true`.

masukai:
**Codex review (round 3):**

The warehouse-start logic is generally sound, but its polling deadline does not actually bound individual network requests. A stalled request can therefore hang the smoke workflow well beyond five minutes.

- [P2] Bound each curl call to the polling deadline — `.github/workflows/dwh-smoke.yml:177`
  If a Databricks status request establishes a connection but then stalls, `curl` has no total timeout and can block indefinitely; the `SECONDS` deadline is checked only between requests, so this step may run far beyond the advertised 300-second budget and potentially occupy the runner for hours.

Confirmed `curl` has no default timeout behavior, and fixed in 5c53900: added `--connect-timeout 10 --max-time 30` to both curl invocations (the retry helper and the polling loop's direct call), bounding any single request well under the overall 300s budget. Dry-run re-verified the normal success path still works with the added flags.

masukai:
**Codex review (round 4):**

The startup step can run against an incompletely configured warehouse while the entire smoke suite subsequently skips, causing unnecessary compute activation and a misleading successful run.

- [P2] Gate startup on all required Databricks secrets — `.github/workflows/dwh-smoke.yml:108`
  When `HOST`, `HTTP_PATH`, and `TOKEN` are configured but `CATALOG` or `SCHEMA` is absent, this condition sets `has_creds=true`, so the new API step wakes the live warehouse; every Databricks test then skips because `require_env(...)` requires both missing values, leaving a green run that exercised nothing and unnecessarily resumed compute.

Verified against `test_databricks_smoke.py`'s own `require_env(HOST_ENV, HTTP_PATH_ENV, TOKEN_ENV, "DRT_SMOKE_DATABRICKS_CATALOG", "DRT_SMOKE_DATABRICKS_SCHEMA")` calls — all five are genuinely required. Fixed in 4f62c84: widened the gate to check all five secrets, matching `require_env`'s actual requirement rather than just what the REST API call itself needs. Directly tested all five partial-configuration combinations — only the fully-configured case now produces `has_creds=true`.

PR #1023, opened 2026-08-29
perf(clickhouse): batch insert instead of one client.insert() call per record

### LINKED ISSUE #961: perf(clickhouse): batch insert instead of one client.insert() call per record

masukai:
## Summary

Closes #961. ClickHouse is a columnar warehouse built for bulk inserts — one-row-at-a-time `client.insert()` calls are the documented anti-pattern its own docs warn against (many small inserts create many parts, triggering merge pressure under sustained load). Every other batch-capable destination already has a batching story (Databricks multi-row `VALUES`, #734; REST API `body_mode: batch`, #770) — this closes the gap the connector's own `# TODO: batch insert with fallback to row-by-row on error` had flagged since its initial commit.

- Batches `client.insert()` calls by `sync.batch_size` (default 100, **now validated as `> 0`** — a `0` or negative value previously either crashed ungracefully or silently inserted nothing while reporting success)
- A batch >1 row is attempted as a multi-row insert. `client.insert()`'s HTTP transport isn't a SQL transaction, so a failure is ambiguous (the request may never have reached ClickHouse, or it may have committed and only the response was lost) — retried once with the same `insert_deduplication_token` (a safe no-op if it already committed) before falling back to row-by-row `RowError` attribution
- **This retry is best-effort, not a guarantee**: `insert_deduplication_token` only takes effect if the table's dedup window is enabled server-side (on by default for `Replicated*MergeTree`, off by default for a plain non-replicated table). drt does not verify or configure this. Operators who need a hard guarantee should enable `non_replicated_deduplication_window` on their table — documented in the module docstring, an inline comment, and the CHANGELOG entry. This was a real finding from a second Codex review round; decided not to chase full correctness here since a complete fix would need an extra round trip and table-level assumptions genuinely out of scope for a batching perf issue — this is a well-known hard distributed-systems problem (idempotent writes over non-transactional HTTP), and the retry is still strictly better than the original unconditional fallback
- `on_error: fail` / `skip` semantics, `RowError` fields, and mirror-mode key/scope accumulation (success-only) are all unchanged
- `mode: replace` + `replace_strategy: swap` gets the same batching (previously still one row at a time)

**250 rows @ batch_size=100: 250 → 3 `client.insert()` calls.**

## Review history

Two independent Codex review rounds, both with real findings, both fixed/addressed:

**Round 1** (3 findings, all fixed):
1. [P1] Naive row-by-row retry on any batch failure could duplicate rows on an ambiguous HTTP failure — fixed with the same-token retry described above
2. [P2] `sync.batch_size` had no validation; `0` crashed, negative silently dropped all data while reporting success — fixed with `Field(gt=0)` at the Pydantic level (global, not ClickHouse-specific)
3. [P2] `replace_strategy: swap` still inserted one row at a time — fixed by extracting a shared `_insert_batched()` helper used by both paths

**Round 2** (1 finding, documented as a known best-effort limitation rather than chased further — see above): the dedup-token retry's effectiveness depends on server-side table configuration drt doesn't control or verify.

## Verification

- [x] Full `pytest tests/unit/ -q` (not just ClickHouse tests, since the `SyncOptions` change is global) — 3470 passed, 3 skipped, 0 unrelated failures (2 known-unrelated `test_cli_version.py` failures are a path-depth artifact of this session's worktree location, confirmed elsewhere this session to pass from a normal-depth checkout)
- [x] `schema-drift` CI check — bundled VS Code JSON schemas regenerated and its own CHANGELOG updated
- [x] New/updated tests assert actual `client.insert()` call counts, arguments, and `settings=` payloads (not just result counts): batch-happy-path, ambiguous-failure-then-safe-retry-succeeds, two-failures-then-row-fallback, `on_error: fail` stopping mid-fallback, `batch_size <= 0` rejected at config validation, swap-replace batching


masukai:
**Codex review (retroactive — this was run during implementation but not posted at the time):**

Two rounds:

- **Round 1** found 3 issues, all fixed before merge-readiness:
  - [P1] Batch retry after an ambiguous transport failure could duplicate an entire committed batch — fixed by narrowing the retry-on-fallback path.
  - [P2] `batch_size <= 0` was silently accepted, producing zero-iteration or invalid ranges — fixed by adding `Field(gt=0)` validation on `SyncOptions.batch_size`.
  - [P2] The `replace_strategy: swap` path still inserted row-by-row, leaving the perf problem unresolved for that mode — fixed by routing it through the same batched insert helper.

- **Round 2** surfaced a deeper, inherent limitation: `insert_deduplication_token` only prevents duplicates if the target table's server-side dedup window is enabled (on by default for `Replicated*MergeTree`, off by default for plain non-replicated `MergeTree`). This is a real ClickHouse distributed-systems constraint, not a bug in this PR — documented explicitly in code comments and the CHANGELOG as "best-effort, not a guarantee" rather than solved, since guaranteeing it would require asserting/enforcing server-side table config, which is out of this issue's scope.

CI green, ruff/mypy clean.

Issue #722, opened 2026-07-07
refactor(destinations): centralize RowError handling, identifier quoting, mirror guards, row-count dispatch


masukai:
## Summary

A cluster of small, low-risk DRY cleanups across the destination layer. Each is independently landable; grouped here as one "SQL/destination cleanup" track.

- [ ] **RowError handling** — the `try/except → result.failed += 1 → append RowError(batch_index=…, record_preview=…[:200], …) → if on_error=='fail': raise` idiom is reproduced ~100 times across 27 `destinations/*.py`, with an inconsistent preview format (SQL dests use `json.dumps(record, default=str)[:200]`; warehouse dests use `str(row)[:200]`). Extract `record_row_error(result, i, record, exc)` (and/or an `iter_rows(...)` generator that centralizes the on_error policy). Compatible with #699.
- [x] **Identifier quoting** — `_quote_ident` is byte-identical in `mysql.py:249-258` and `clickhouse.py:336-345`; `schema.py:78` re-declares its own `_split_qualified` duplicating `postgres.py:98`. Consolidate into `destinations/sql_utils.py`.
- [x] **Mirror-capability guards** — the "mirror.strategy/scope not supported" fail-fast is hand-rolled with the same message template in `snowflake.py:139-151`, `databricks.py:213-225`, `clickhouse.py:107-119`; the `upsert_key`-required check repeats across all five SQL dests. Extract `reject_unsupported_mirror(...)` + `require_upsert_key_for_mirror(...)`.
- [x] **Row-count dispatch** — `sql_utils.get_row_count_for_destination:18-41` dispatches by `isinstance(config, …)` and silently omits Snowflake/Databricks — a third place (besides registry + union) to edit per SQL destination. Model as an optional `RowCountable` Protocol (like `ConnectionTestable`), discovered not enumerated.

**Effort/risk:** S–M / LOW each. No collision with #699/#707/#402 (the identifier + row-count items are Postgres/MySQL/ClickHouse). Good incremental / good-first-issue-friendly chunks.

masukai:
Progress: **identifier quoting**, **mirror-capability guards/messages**, and the **`RowCountable` row-count dispatch** shipped in #726 + #730 (which also shared `record_preview`, so the preview-format inconsistency is gone). Remaining scope is the **RowError-handling extraction** (~100 sites across 27 `destinations/*.py`) — the biggest and most mechanical piece; it can now build on the shared `record_preview` helper.

PR #1025, opened 2026-08-29
refactor(destinations): Snowflake/Databricks join BaseSqlDestination, phase 1 (#720)

masukai:
## Summary

Phase 1 of #720, following the same phased-PR pattern #719 used for Postgres/MySQL. Both `SnowflakeDestination` and `DatabricksDestination` now subclass `BaseSqlDestination`, implementing only the `_dialect_connect`/`_qualify_ident` hooks this phase needs. `load()` and `finalize_sync()` stay each dialect's own concrete override, unchanged — no sync-path behavior change.

Two design decisions were researched and confirmed with the maintainer before implementation, documented in detail on [the issue](https://github.com/drt-hub/drt/issues/720#issuecomment-5465190487):

1. **Databricks does not adopt the `_build_mirror_delete` hook.** Its staged-DELETE/MERGE sequence (driven by #707's 255-marker paramstyle limit and #908's Delta multi-column `IN` rejection) can't fit the `(stmt, params)`-returning contract Postgres/MySQL/Snowflake share. Databricks keeps overriding `_finalize_mirror`/`_finalize_mirror_tracked` wholesale — that pull-up is out of scope, deferred to a later phase.
2. **`finalize_sync`'s swap-commit state-reset ordering stays dialect-specific**, not silently normalized: Snowflake resets in-memory swap state only after a successful `SWAP`, leaving a failed swap's shadow table recoverable for retry; Databricks resets unconditionally in a `finally` block (no such recovery path exists for `INSERT OVERWRITE`).

## What changed

- `_dialect_connect`'s signature widened to accept `query_tags: dict[str, str] | None = None` — both warehouses need `sync_options._query_tags` at connect time; Postgres/MySQL accept-and-ignore the new parameter (additive, no behavior change there).
- Renamed `_load_replace_swap` → `_load_replace_swap_{snowflake,databricks}` and Snowflake's `_build_mirror_delete` → `_build_snowflake_mirror_delete`, so neither class accidentally satisfies `BaseSqlDestination`'s still-abstract hooks with a shape/behavior not yet vetted for a later phase. Unimplemented hooks stay `NotImplementedError` stubs and are never reached this phase.
- Snowflake's internally-triplicated inline `RowError(...)` block (3 call sites) and Databricks' own copy now go through the shared `record_row_error()` helper (#722), preserving each dialect's existing `str(row)[:200]` preview basis exactly.
- **Found and fixed a real, previously-silent gap**: both classes now inherit `BaseSqlDestination.reset_tracked_state`, which depends on `_state_table_ident` (a later-phase hook, not implemented here) — calling it would raise `NotImplementedError`. `drt/cli/commands/state.py` and `drt/mcp/tools/state.py`'s `drt state reset --tracked-mirror` previously used `getattr(dest, "reset_tracked_state", None) is None` to detect "unsupported" and print a clean message; that check alone is no longer sufficient once the method is inherited-but-unimplemented. Both call sites now also identity-check whether `reset_tracked_state`/`_state_table_ident` are still the inherited `BaseSqlDestination` stubs and degrade the same clean way. Postgres/MySQL are unaffected (their hooks are real implementations).

## Verification

- `pytest tests/unit/ -q` — 3476 passed, 3 skipped, plus 2 known-unrelated `test_cli_version.py` failures (this session's deep-scratchpad-path artifact, confirmed elsewhere this session to pass from a normal-depth checkout)

## Test plan

- [x] Codex review pass


masukai:
**Codex review:** No actionable correctness issues found — the refactor preserves existing destination behavior while guarding the newly-inherited unsupported tracked-state reset path.

CI green across all checks (test 3.10–3.13, dagster-drt-test, CodeQL, Analyze).

PR #1026, opened 2026-08-30
refactor(destinations): Snowflake/Databricks join BaseSqlDestination, phase 2 (#720)

masukai:
## Summary

Phase 2 of #720 — pulls `load()` up into the shared `BaseSqlDestination` template for both `SnowflakeDestination` and `DatabricksDestination`, replacing each dialect's own full `load()` override. Same pattern #719's phase 2a established for Postgres/MySQL.

Scoping for this phase (and the reasoning behind splitting it differently from #719) was researched and posted to [the issue](https://github.com/drt-hub/drt/issues/720#issuecomment-5467603519) before implementation:

- `load()` never calls `_build_mirror_delete` (that hook is only used inside `_finalize_mirror`/`_finalize_mirror_tracked`) — so despite Databricks never adopting `_build_mirror_delete` (issue Decision 1), it can still share the `load()` template.
- `finalize_sync()`'s swap-commit reset ordering differs between the two dialects (issue Decision 2: Snowflake resets only after a successful `SWAP`, Databricks resets unconditionally) — so unlike #719's phase 2a, this phase does **not** touch `finalize_sync()` at all, for either dialect.

## What changed

- `_load_replace`, `_load_replace_swap`, and `_load_upsert` are now real per-dialect hooks matching the base's contract (`(conn, cur, records, columns, table, sync_options, config) -> SyncResult`). The former inline `config.mode` (`insert`/`merge`, mirror forces `merge`) dispatch moved inside each `_load_upsert` — a black box to the base's `load()`, which only cares about the binary replace-vs-not-replace split.
- The mirror-key/scope accumulation duplicated inline in both files' old `load()` bodies is gone, in favor of the base's existing `_accumulate_mirror_state`.
- Each dialect overrides `_validate_mirror_scope` to keep calling its existing `check_mirror_supported(...)` check, which is stricter than the base's default validator (also checks `upsert_key` truthiness and dialect scope-tracking support) — avoids a behavior regression a straight swap to the base validator would have caused.
- `finalize_sync()`, `_finalize_mirror()`, `_finalize_mirror_tracked()`, and the mirror-delete machinery are untouched — phase 3 (Snowflake only) and permanently out of scope for Databricks.

## Verification

- **Pure code motion**: no SQL statement, parameter binding, chunk-sizing (Snowflake's MERGE chunk budget, Databricks' native 255-marker INSERT chunking), or `on_error`/`RowError` behavior changed. Every dialect-specific test asserting exact statement/param shape passes **unmodified** — only mock-cursor fixture setup changed (to match the base's cursor lifecycle, same as postgres.py/mysql.py's fixtures).
- `pytest tests/unit/ -q` — 3480 passed, 3 skipped, plus 2 known-unrelated `test_cli_version.py` failures (this session's deep-scratchpad-path artifact)

## Test plan

- [x] Codex review pass


masukai:
**Codex review:** found one real issue — [P2] the shared \`load()\` template closes the connection but never the cursor, unlike Snowflake's previous \`load()\` override (\`with tagged_cursor(conn.cursor(), sync_options) as cur:\`), which relied on the cursor's own \`__exit__\`. Verified independently via direct introspection: \`SnowflakeConnection.close()\` never references cursors at all, so this left result buffers/connector resources to nondeterministic GC across multi-batch or long-running syncs.

Fixed in d9b60cb: \`BaseSqlDestination.load()\` now closes \`cur\` in the same \`finally\` block that closes \`conn\` (guarded against \`conn.cursor()\` itself raising before \`cur\` is assigned). Benefits Postgres/MySQL's cursor hygiene too, not just Snowflake. Two test fixtures gained a no-op \`close()\` on their fake cursor classes to match.

PR #1027, opened 2026-08-30
refactor(destinations): Snowflake joins mirror-finalize template, phase 3 (#720)

masukai:
## What does this PR do?

Phase 3 of #720 — Snowflake now inherits `BaseSqlDestination._finalize_mirror()`/`_finalize_mirror_tracked()` instead of carrying its own wholesale copies. Mirrors #719's phase 2b for Postgres/MySQL. **Stacked on #1026** (phase 2) — this PR's diff is phase 3's changes only; the base of this branch already includes phase 2.

Databricks is not touched at all (`git diff -- drt/destinations/databricks.py` is empty) — per the design decision recorded on issue #720, its staged-key mirror deletion is structurally incompatible with the base's single-statement `_build_mirror_delete` contract and permanently keeps its own wholesale overrides.

## What changed

- `_build_snowflake_mirror_delete` renamed back to `_build_mirror_delete` — already shaped exactly as the base hook contract requires, a straight rename.
- Six state-table hooks (`_state_table_ident`, `_state_table_exists`, `_create_state_table`, `_state_scope_columns_exist`, `_add_state_scope_columns`, `_state_sql`) extracted from Snowflake's former inline `_finalize_mirror_tracked`, preserving the exact `SHOW TABLES`/`information_schema.columns`/`CREATE TABLE`/`ALTER TABLE` statement text.
- Three new narrow base-class hooks, each defaulting to Postgres/MySQL's previous exact inline behavior (verified — `postgres.py`/`mysql.py` are untouched, don't override any of them):
  - `_mirror_table_ident(config) -> str` — defaults to `str(config.table)`; Snowflake overrides to its `database.schema.table` FQN.
  - `_commit_mirror(conn)` — defaults to `conn.commit()`; Snowflake overrides to a no-op (autocommits).
  - `_state_params(*values)` — defaults to `tuple(values)`; Snowflake overrides to `list(values)`, matching its connector's existing list-binding convention.
  - Plus `_try_add_state_scope_columns`/`_stage_mirror_keys`, moving the existing SAVEPOINT-guarded (PG/MySQL) vs try/except-only (Snowflake, no savepoints) logic into overridable methods with byte-identical default bodies.
- `finalize_sync()` itself stays completely untouched — still Snowflake's own override for the atomic `SWAP WITH` path, out of scope per the same design decision.
- The Phase-1 `reset_tracked_state` guard (added so `drt state reset --tracked-mirror` degrades cleanly for a class that inherits the method but not its hooks) now correctly resolves for Snowflake — it genuinely implements the hooks now, so the guard no longer flags it and the reset actually runs. Databricks still hits the guard.

## Verification

- Wire-SQL text checked character-for-character against the pre-refactor version (`SHOW TABLES` probe, `CREATE TABLE` DDL, `information_schema` query, `ALTER TABLE`, temp-table staging, DELETE builder — all identical).
- `pytest tests/unit/ -q` — 3479 passed, 3 skipped, plus 2 known-unrelated `test_cli_version.py` failures (this session's deep-scratchpad-path artifact)
- New test confirms Snowflake's tracked-state reset now runs the real DELETE and asserts `conn.commit` is NOT called (autocommit behavior preserved)

## Related Issue

Part of #720

## Checklist

- [x] Tests pass
- [x] Linter passes


masukai:
**Codex review:** The refactor preserves Snowflake's fully qualified identifiers, parameter bindings, autocommit behavior, and tracked-mirror staging semantics through dialect hooks. Targeted Snowflake and shared SQL mirror tests passed, with no actionable regressions identified.

CI green across all checks (test 3.10–3.13, dagster-drt-test, CodeQL, Analyze, codecov/patch).

masukai:
Superseded by #1029 (same commit, re-targeted to `main` directly — GitHub's native stacked-PR merge flow doesn't support admin bypass of required review, which this repo's branch protection needs). Closed automatically when the phase-2 base branch was deleted; merged via #1029 instead.

PR #1031, opened 2026-08-30
fix(state): OS-level file locking for cross-process local state safety (#963)

masukai:
## Summary

`LocalStateManager`, `LocalHistoryManager`, and `LocalDlqStore` each used only an instance-local `threading.Lock` around their whole-file read-modify-write cycles — meaningful within one process, not across two separate `drt` invocations (e.g. `drt run` + `drt retry`, the issue's motivating example), each of which gets its own fresh, unlocked lock. Two overlapping processes racing a read-modify-write was last-writer-wins with no detection — the cross-process gap #955/#962 named and deferred to this issue.

Design decision (researched and [posted to the issue](https://github.com/drt-hub/drt/issues/963#issuecomment-5468773760) before implementing): rejected an mtime/size conditional-write pattern mirroring the object-store tier's generation/ETag preconditioning — it can't actually close the race, since GCS/S3's `if_generation_match`/`IfMatch` are single indivisible server-side compare-and-swap operations, while a local mtime/size check is two separate client-side syscalls with an inherent TOCTOU gap (worsened by 1-second mtime granularity on some filesystems). Adopted OS-level file locking instead.

## What changed

- New `drt/state/_filelock.py`: `advisory_lock(path)` context manager wrapping the entire read-modify-write span — `fcntl.flock(LOCK_EX)` on POSIX, a one-byte `msvcrt.locking()` range on Windows. No new dependency. Locks a persistent sidecar `<path>.lock`, not the data file itself, so a not-yet-created file on first run is still lockable.
- Wired into every read-modify-write span: `LocalStateManager.save_sync()`/`.reset()`, `LocalHistoryManager.prune()`, `LocalDlqStore.append()`/`.replace()`/`.reconcile()`. The existing `threading.Lock` is retained everywhere (still coordinates threads sharing one instance) — the file lock is additive, coordinating separate processes.
- **`LocalHistoryManager.append()` also joins the lock**, despite already being POSIX-`O_APPEND`-atomic on its own — found during review of the initial implementation (which, per the issue's scoping, correctly excluded `append()` on the reasoning that `O_APPEND` makes concurrent appends safe against *each other*). That reasoning doesn't cover the real risk: `O_APPEND` says nothing about safety against a concurrent `prune()`, whose read-then-`tmp.replace()` rewrite silently drops any append landing in that window. `drt/engine/sync.py:399`/`414` calls `append()` immediately followed by `prune()` for the same sync within one `run_sync()` — so two overlapping processes running the same sync (the issue's literal scenario) can hit exactly this interleaving. Proven with a new test, verified failing with the lock removed from `append()`, passing with it restored.

## Tests

`tests/unit/test_local_state_concurrency.py` uses real `multiprocessing.Process` workers (not threads — the whole point is kernel-level cross-process lock contention shared Python state can't exercise), each constructing its own separate store instance so the file lock — not any in-process lock — is what's proven to coordinate them. Includes a capability canary (mirroring the pattern in `tests/integration/object_store/conftest.py`) proving a second process genuinely blocks before trusting any concurrency assertion on top.

## Verification

- `pytest tests/unit/ -q` — 3484 passed, 3 skipped, plus 2 known-unrelated `test_cli_version.py` failures (deep-scratchpad-path artifact)
- All 5 new concurrency tests pass; negative-control proof (fail-without-fix, pass-with-fix) independently re-verified for the append-vs-prune case

## Test plan

- [x] Codex review pass


masukai:
**Codex review:** found one real issue — [P2] Windows lock acquisition wasn't truly blocking. \`msvcrt.locking(LK_LOCK)\` doesn't block indefinitely like POSIX \`flock(LOCK_EX)\` — the Windows CRT's underlying \`_locking\` function retries internally for ~10 seconds and then raises \`OSError\` if the region is still held. A lock outliving that window (a large DLQ/history rewrite, a slow disk, a suspended holder process) would surface as a dropped write or an unhandled crash, contradicting the blocking cross-process guarantee this fix exists to provide.

Fixed in 0f8a931: \`_acquire_windows()\` now retries \`msvcrt.locking(LK_LOCK)\` itself across that ~10-second ceiling, catching \`OSError\` and polling with a short delay until acquisition succeeds — matching \`flock\`'s actually-blocking semantics.

I don't have a Windows machine to verify the fix runs there, but the retry-on-OSError logic matches the documented CRT behavior. Flagging for extra scrutiny if a Windows CI run is ever added.

Issue #1034, opened 2026-08-31
design: Source.extract()'s frozen ProfileConfig type is dishonest for plugin-registered sources


masukai:
## Problem

Found during a v1.0-readiness audit (Codex + Claude cross-check), independently verified.

`Source.extract()` and `IncrementalSource.extract_incremental()` (`drt/sources/base.py`) both accept `config: ProfileConfig` — already frozen (`Stability: Stable (frozen at v1.0, see ADR 0007...)`). `ProfileConfig` is a closed pydantic union enumerating only the built-in profile types (`drt/config/profiles.py:261`).

But the plugin loading path (`drt/config/credentials.py`, around the `source_profile_class`/registry lookup added by #997) constructs an arbitrary third-party profile class for a registered plugin source, then does `return cast("ProfileConfig", profile)` (`credentials.py:512`) to satisfy the type checker — a pure type-checker silencer, not a structural guarantee. At runtime, `config` passed into a plugin `Source.extract()` can be any registered profile class, not a member of the closed union. `docs/guides/plugins.md:93` explicitly promises arbitrary profile classes to plugin authors, so this isn't accidental — the registry deliberately does this, and the type contract just doesn't reflect it honestly.

The plugin extensibility test doesn't catch this: `_PluginSource` in `tests/unit/test_plugin_config_extensibility.py` has no methods, so it only proves construction/registry lookup work, not that a real plugin `Source.extract(config)` call structurally matches its declared type.

## Why this matters for v1.0 specifically

`Source.extract`'s signature is one of the 21 Protocols ADR 0007 freezes. Changing a frozen Protocol's parameter type after the freeze is explicitly a breaking change. Right now the declared contract (`config: ProfileConfig`, a closed union) doesn't match what plugin sources actually receive — freezing it as-is locks in a signature that's already inaccurate for a documented, supported use case (third-party source plugins).

## What needs a decision (not just a fix)

Genuinely a design question, not a mechanical PR — pick one before v1.0:

1. **Structural/Protocol-based profile type** instead of a closed union — e.g. a `ProfileConfigLike` Protocol (`describe() -> str` plus whatever `Source.extract` implementations actually rely on) that both built-in and plugin profiles satisfy structurally. Closest to "honest about what's actually guaranteed."
2. **Generic/parameterized `Source[TConfig]`** — lets each source (built-in or plugin) declare its own concrete profile type, at the cost of a more complex Protocol shape.
[...]

PR #1036, opened 2026-08-31
refactor(sources): honest structural ProfileConfigLike for frozen Source Protocols (#1034)

masukai:
## Summary

Found during a v1.0-readiness audit — `Source.extract()`/`Source.test_connection()` and `IncrementalSource.extract_incremental()` (frozen Protocols per ADR 0007) type their `config` parameter as the closed 13-member `ProfileConfig` union, but #997/#1011 (contributed by @Muawiya-contact) already made plugin-registered profiles loadable at runtime via `cast("ProfileConfig", profile)` in `load_profile()` — a type-checker silencer, not a real guarantee. Fixing this after the v1.0 freeze would be a breaking Protocol change, so it needed to land before.

Design research (posted to [the issue](https://github.com/drt-hub/drt/issues/1034#issuecomment-5472333767) before implementing) found #997's own fix (`GenericDestinationConfig` catch-all) doesn't transfer directly to the source side — `ProfileConfig` isn't pydantic-validated, `load_profile()` already returns the plugin's real class rather than an extras-bag, and there's no JSON-Schema `oneOf` knock-on to solve.

## What changed

- New `ProfileConfigLike` (`drt/config/profiles.py`) — a runtime-checkable structural Protocol with only the two members drt-core actually reads off a profile without an `isinstance` narrow anywhere: `type` (as a read-only Protocol **property**, so each built-in's narrower `Literal[...]` field satisfies it covariantly instead of tripping Protocol attribute invariance) and `describe() -> str`. All 13 built-in profile dataclasses satisfy it unchanged.
- `Source`/`IncrementalSource`'s three frozen methods now type `config` as `ProfileConfigLike`. Every built-in source's own concrete signatures updated to match (for consistency, not correctness) — their existing `assert isinstance(config, ConcreteProfile)` first-line narrowing is unchanged.
- `load_profile()` now returns `ProfileConfigLike` and the registry-fallback branch replaces the cast with a real `isinstance(profile, ProfileConfigLike)` check — raising a clear error if a plugin profile class doesn't actually satisfy the contract. **This is a genuine strengthening, not just an annotation fix**: removing the cast surfaced 12 real, previously-suppressed mypy errors (widening the type carried through the registry/resolver/engine/CLI dispatch path), all resolved without another cast.
- New `docs/guides/building-a-source.md` documents the convention that makes this safe for plugin authors, explicitly calling out the contravariance gotcha (annotating your method directly with your own concrete profile class does **not** structurally satisfy `Source`).
- `tests/unit/test_plugin_config_extensibility.py`'s `_PluginSource` previously had no methods — the existing test only proved construction/registry lookup. It now implements `extract()`/`test_connection()` per the documented convention, and the test calls `source.extract(...)` end-to-end through a loaded plugin profile.

## Verification

- `mypy drt` — clean, 200 files, zero `cast()` remaining on this path
- `pytest tests/unit/ -q` — 3486 passed, 3 skipped, plus 2 known-unrelated `test_cli_version.py` failures (deep-scratchpad-path artifact)
- All 13 built-in source test suites pass unmodified in their assertions

## Test plan

- [x] Unit suite green
- [x] Codex review pass


masukai:
**Codex review:** found one real issue — [P2] widening \`load_profile()\`'s return type to \`ProfileConfigLike\` broke the static type of the natural round-trip \`profile = load_profile(name); save_profile(name, profile)\`, since \`save_profile(profile: ProfileConfig)\` kept the narrower closed-union parameter.

Investigated what \`save_profile()\` actually does at runtime: its isinstance chain over the 13 built-ins already ends in \`elif is_dataclass(profile) and not isinstance(profile, type):\` (added by #997) — a generic fallback that already persists any plugin profile dataclass correctly. So \`save_profile()\` already supported \`ProfileConfigLike\`-satisfying plugin profiles at runtime; only its declared parameter type was narrower than what it actually handles — the identical annotation-honesty gap this PR fixes on the read side, just on the write side.

Fixed in 2e2206f: widened \`profile: ProfileConfig\` → \`profile: ProfileConfigLike\`. No runtime behavior change. Verified the round-trip now type-checks with a standalone mypy check, plus the full ruff/mypy/pytest suite stays green.

masukai:
Codecov flag on `drt/config/credentials.py` (97.83%, 1 line) was real — unlike this session's earlier false-alarm cases, this was a genuinely new, uncovered defensive branch (the `isinstance(profile, ProfileConfigLike)` rejection this PR adds), not a pre-existing line shifted by a refactor. Added a test in e6a044e covering it: a plugin profile that accepts `type=` in its constructor (passes signature binding) but stores it privately rather than exposing a readable `type` attribute, so it fails the structural check specifically rather than the earlier `describe()`-callable one.

Issue #1037, opened 2026-08-31
fix(cli): drt retry crashes on staged destinations (salesforce_bulk, staged_upload)


masukai:
## Problem

Found during a v1.0-readiness audit (Codex flagged the underlying type-honesty issue as a "registry typing" finding; investigating it surfaced an actual crash bug, more severe than the original framing).

`drt/connectors/registry.py`'s \`get_destination(config: DestinationConfig) -> Destination\` (and its re-export at \`drt/cli/_helpers.py\`) declares a return type of \`Destination\` — but 2 of the 34 registered destination types (\`salesforce_bulk\`, \`staged_upload\`) only implement \`StagedDestination\`'s \`stage()\`/\`finalize()\`, not \`Destination.load()\`. \`salesforce_bulk\` is a real, first-class, user-configurable destination type — not an internal-only class.

\`drt/engine/sync.py\` already knows this: \`run_sync()\`'s own \`destination\` parameter is correctly typed \`Destination | StagedDestination\`, and the engine \`isinstance\`-checks before calling \`load\` vs. \`stage\`/\`finalize\`. Every CLI command that calls \`get_destination()\` and might call \`.load()\` directly does the same check — **except one**:

**\`drt/cli/commands/retry.py\`'s \`replay_dead_letters()\` (shared by both the \`drt retry\` CLI command and the MCP \`drt_retry\` tool) calls \`dest.load(records, sync.destination, sync.sync)\` unconditionally at line ~99, with no \`isinstance(dest, StagedDestination)\` check anywhere in the file.**

Confirmed this is reachable, not theoretical: \`drt/engine/sync.py\` (~line 707) persists staged-destination \`stage()\`/\`finalize()\` failures to the DLQ specifically \"so a DlqObserver can persist them for \`drt retry\`\" — i.e. staged destinations are explicitly meant to support \`drt retry\`. Any sync configured with \`destination: {type: salesforce_bulk}\` (or \`staged_upload\`) that accumulates DLQ entries and is then retried via \`drt retry <sync>\` will crash with \`AttributeError: '<class>' object has no attribute 'load'\`.

No existing test covers this — \`tests/unit/test_cli_retry.py\`'s fake destinations all implement \`load()\`.

## Proposal

Fix \`replay_dead_letters()\` in \`drt/cli/commands/retry.py\` to branch on \`isinstance(dest, StagedDestination)\`:
[...]

PR #1038, opened 2026-08-31
fix(cli): drt retry stages/finalizes StagedDestination instead of calling missing load()

masukai:
## Summary

Found during a v1.0-readiness audit — `drt/cli/commands/retry.py`'s `replay_dead_letters()` (shared core for `drt retry` and the MCP `drt_retry` tool) called `dest.load(...)` unconditionally, crashing with `AttributeError` for `StagedDestination`-only destinations (`salesforce_bulk`, `staged_upload`), which have no `load()` method.

Verified the reachability precisely (and corrected my own initial overstatement on [the issue](https://github.com/drt-hub/drt/issues/1037#issuecomment-5472881679) before implementing): staged destinations don't currently populate the DLQ through normal sync execution (confirmed via a full-repo grep for `observer.on_records_failed` callers — there's exactly one, exclusively in the non-staged path), so the crash requires a sync whose destination is/becomes a staged type while its DLQ already holds entries from some other source — real and unguarded, just narrower than "any `salesforce_bulk` sync's normal failures."

## What changed

- Refactored the per-chunk `load()`/correlate loop into a `retry_groups` generator unifying both paths: `load()` still yields one `(chunk, result)` per chunk (unchanged behavior); the staged path stages every chunk, then calls `finalize()` **once** (authoritative per `StagedDestination`'s own Protocol docstring).
- A raised `finalize()` (documented as `SalesforceBulkDestination`'s real failure shape, vs. the generic `staged_upload` destination which returns a failed `SyncResult`) keeps every retried entry queued and reports a clear `"failed"` status instead of crashing.
- Staged correlation operates on `RowError.batch_index` against the **full accumulated** `to_retry` list (one `finalize()` spans every staged chunk) — same "trust only when fully attributable, else keep everything conservatively queued" principle the existing `load()` path already used.
- **Found and fixed a second, more subtle bug while verifying this independently**: `SalesforceBulkDestination.finalize()` hardcodes `RowError.batch_index=0` for every failure (its failed-results CSV parser has no column carrying the original record's position — verified in `drt/destinations/salesforce_bulk.py`). The generic "distinct-index-count == failed-count → trust it" correlation would, for exactly one Salesforce failure, wrongly conclude only the record at index 0 failed and **silently drop every other still-failing entry from the DLQ**. `retry.py` now special-cases `sync.destination.type == "salesforce_bulk"` to never trust `batch_index` attribution.
- Widened `get_destination()`'s return type (`drt/connectors/registry.py`, `drt/cli/_helpers.py`, `drt/cli/main.py`'s back-compat shim) from `Destination` to `Destination | StagedDestination`, matching `run_sync()`'s own already-correct signature — this is what would have caught the `retry.py` bug at type-check time from the start.

## Tests

New tests in `tests/unit/test_cli_retry.py`: staged success, staged partial-failure with correct global-index correlation, `finalize()` raising, and — added after independently reviewing the initial implementation, since it wasn't covered — a dedicated `salesforce_bulk`-specific test proving the `batch_index=0` guard actually fires. **Verified the guard is load-bearing**: temporarily removed it and confirmed the test fails exactly as predicted (2 of 3 still-failing records wrongly marked succeeded and dropped), then restored it and confirmed all tests pass.

Also added the optional `RestApiSource` coverage to `tests/unit/test_source_contract.py` noted on the issue.

## Verification

- `pytest tests/unit/ -q` — 3494 passed, 3 skipped, plus 2 known-unrelated `test_cli_version.py` failures (deep-scratchpad-path artifact)

## Test plan

- [x] Unit suite green
- [x] Codex review pass


masukai:
**Codex review:** found one real issue — [P1] the staged retry path could silently misattribute a failure across multiple \`stage()\` chunks. \`StagedDestination\`'s Protocol never defines whether \`RowError.batch_index\` from \`finalize()\` is local to the individual \`stage()\` call it came from or global across the whole accumulated set. My original fix assumed global indexing unconditionally — for a third-party \`StagedDestination\` reporting chunk-local indices (the more natural convention, matching \`Destination.load()\`'s own always-chunk-local \`batch_index\`), this would silently drop the real failure from the DLQ while re-queuing an already-succeeded record.

Verified neither currently-shipped \`StagedDestination\` actually triggers this today (\`salesforce_bulk\` already had its own \`batch_index=0\` guard; \`staged_upload\` never reports \`row_errors\` at all) — the gap was real but latent, hitting only a hypothetical plugin author following the more natural convention, with zero warning and no test coverage.

Fixed in 26f7c01: only trust \`batch_index\`-based attribution when the whole retry set fit in a single \`stage()\` call (chunk-local and global indexing coincide when there's only one chunk); multiple chunks always fall back to the existing conservative "keep everything queued" behavior. Added a purpose-built \`_FakeChunkLocalStagedDestination\` test reproducing the exact scenario, verified load-bearing by temporarily disabling the guard and confirming the test fails exactly as predicted before restoring it.

masukai:
Codecov flag on \`drt/cli/commands/retry.py\` (94.59%, 2 lines) was real, not a false alarm — both were genuinely new branches this PR added:

- Line 103 (\`retry_groups = iter([])\`): reachable via \`drt retry <sync> --limit 0\` against a staged destination — only negative limits are rejected, 0 is valid, so a non-empty DLQ with an empty requested retry slice is a real case. Added a test confirming \`stage()\`/\`finalize()\` are never called and the queue is untouched.
- Line 203 (the \`salesforce_bulk\`-specific guard): the existing test used 3 entries against \`batch_size=2\` (2 chunks), so it actually exercised the more general multi-chunk guard from the previous fix, not this line — both guards happened to produce the same observable outcome, masking that the intended line was never reached. Reduced to 2 entries (a single chunk) so the salesforce-specific check actually runs.

Fixed in 414b4df — \`drt/cli/commands/retry.py\` is now 100% covered.

PR #1041, opened 2026-08-31
chore(state): remove StateStore.now(), an unimplemented-by-callers Protocol method

masukai:
## Summary

Found during a v1.0-readiness audit (Codex, "Medium-low," explicitly noted it wouldn't block v1.0 either way). `StateStore.now()` was part of the frozen Protocol with concrete implementations in both `LocalStateManager` and `ObjectStoreStateStore`, but no production code ever called it — `save_sync()` takes a caller-supplied `SyncState` whose timestamp the engine already computes itself (verified via a full-repo grep: zero non-test, non-definition call sites).

The only callers were two object-store integration test assertions checking the return value was a timezone-aware ISO string — unrelated to the method's own purpose, tacked onto the end of an otherwise-unrelated conformance test.

## Why now, not later

Removing an unused method from a Protocol is free before the v1.0 freeze (ADR 0007) and a breaking change after it — every third-party `StateStore` backend would otherwise have had to implement something drt-core itself never calls.

## What changed

- Removed `now()` from `StateStore`, `LocalStateManager`, `ObjectStoreStateStore`.
- Removed the two now-orphaned assertions in the S3/GCS integration test conformance checks.
- Updated `test_state.py`'s `_STATE_STORE_METHODS` drift-detection literal to match.

No behavior change to `save_sync`/`get_last_sync`/`get_all`/`reset`.

## Verification

- `pytest tests/unit/ -q` — 3499 passed, 3 skipped, plus 2 known-unrelated `test_cli_version.py` failures (deep-scratchpad-path artifact)

## Test plan

- [x] Unit suite green


masukai:
Closing without merging. Codex review correctly flagged that this repo's VERSIONING.md requires the standard two-minor-release deprecation cycle for public Protocol methods even pre-1.0 (`StateStore`/`StateManager` are both explicitly named in the policy) — a bare removal like this one doesn't satisfy that, regardless of how confident the zero-callers grep is. The process cost of a formal deprecation cycle isn't worth it for one genuinely unused method, so leaving `now()` in place is the simpler, policy-respecting outcome.

PR #1057, opened 2026-08-31
fix(engine): sync.mode: replace/mirror fail fast on unsupported destinations (#1042)

### LINKED ISSUE #1042: bug(engine): sync.mode: mirror/replace succeeds silently on destinations that don't implement it
## Problem

`SyncOptions.mode` (`drt/config/sync_options.py:264`) accepts `mirror`/`replace` for every destination type — there is no capability check anywhere in the write path.

Verified directly:
- `drt/engine/sync.py` only checks `MatchPolicyCapable` (around line 130); grepping the whole file for `sync_options.mode`/`options.mode`/`.mode ==` finds exactly one hit, an unrelated `cursor_field` branch (line 461). No mirror/replace capability check exists.
- `drt/destinations/slack.py`'s `load()` never reads `sync_options.mode` at all — it just posts every record as a normal message regardless of configured mode.
- `drt/cli/commands/validate.py` has no mode-vs-destination-type check either.
- BigQuery's own destination `mode` field is `Literal["insert", "merge"]` (`drt/config/destinations_sql.py:94`) — no `replace`/`mirror` option — so a sync-level `mode: mirror` against BigQuery has nowhere to land either.

## Why it matters

A user who sets `sync.mode: mirror` against Slack, most SaaS destinations, or BigQuery gets a normal successful run — no error at config-validation time, no error at runtime — while the mirror DELETE pass silently never happens. `mode: replace` degrades the same way (behaves like append/upsert instead of a full replace). The requested data contract is silently weakened with no signal to the operator.

## Suggested direction

A structural capability check (e.g. `MirrorCapable`/`ReplaceCapable` Protocols, mirroring the existing `MatchPolicyCapable` pattern at `drt/engine/sync.py:130`) that fails fast — either at `drt validate` time or at the start of `drt run` — when `sync.mode` requests a capability the target destination doesn't implement.

**Effort/risk:** Small–medium to add the fail-fast check itself; larger only if the goal becomes implementing mirror/replace on destinations that currently lack it (out of scope for this issue — fail loudly first).

masukai:
Closes #1042.

## Problem

No capability check existed anywhere in the write path. `sync.mode: mirror` against Slack, HubSpot, BigQuery, or any other non-SQL/ClickHouse destination reported success while never deleting stale destination data. `mode: replace` silently behaved as a normal upsert/append. `MatchPolicyCapable`'s own docstring already claimed "the same fail-fast philosophy as unsupported `mirror` configs" — that check didn't actually exist.

## Fix

New optional `ModeCapable` Protocol (`drt/destinations/base.py`), mirroring the existing `MatchPolicyCapable` pattern from #757 exactly:

```python
@runtime_checkable
class ModeCapable(Protocol):
    def supported_modes(self) -> frozenset[str]: ...
```

`full`/`incremental`/`upsert` need no declaration (always safe — normal load path). Only `replace`/`mirror` require it. `BaseSqlDestination` (Postgres/MySQL/Snowflake/Databricks) and `ClickHouseDestination` (standalone, not a `BaseSqlDestination` subclass) both declare `frozenset({"replace", "mirror"})` — the only five destinations with real replace/mirror machinery today.

New `_check_mode_supported()` in `drt/engine/sync.py`, placed and worded to match the existing `_check_match_policy_supported()` exactly, called from the same call site in `_run_sync_body()` right alongside it — before cursor/watermark resolution, before any extraction or write.

This is a new, additive Protocol — no existing frozen Protocol signature changed (ADR 0007).

## Verification

- New `tests/unit/test_mode_capability.py`: `run_sync()` end-to-end rejects `mirror`/`replace` against an incapable destination **before any I/O** (source `extract()` never called, destination `load()` never called) — the actual regression test for the bug. All 5 currently-capable destinations (Postgres, MySQL, Snowflake, Databricks, ClickHouse) confirmed unaffected. `full`/`incremental`/`upsert` confirmed to need no capability. BigQuery specifically confirmed to now fail loudly instead of silently mishandling `mirror`.
- `pytest tests/unit/test_mode_capability.py tests/unit/test_engine.py tests/unit/test_match_policy.py tests/unit/test_sql_base.py tests/unit/test_clickhouse_destination.py tests/unit/test_clickhouse_mirror_mode.py tests/unit/test_postgres_destination.py tests/unit/test_mysql_destination.py tests/unit/test_snowflake_destination.py tests/unit/test_snowflake_mirror_mode.py tests/unit/test_databricks_destination.py` — 451 passed.
- Full suite: 3,651 passed, 39 skipped, 2 pre-existing unrelated failures (`test_cli_version.py`, a known artifact of this checkout's directory-depth, not this change).
- `docs/guides/building-a-destination.md` updated with a new "Optional advanced sync modes" section documenting `ModeCapable` for connector authors.


masukai:
Codex review caught a real gap (high severity): `BaseSqlDestination.supported_modes()` unconditionally claimed `{"replace", "mirror"}` even though the base class's own hooks (`_load_replace_swap`, `_build_mirror_delete`, etc.) are abstract `NotImplementedError` stubs — a subclass implementing only some of them (e.g. plain upsert) would silently inherit a capability it can't actually serve, letting this PR's own fail-fast guard wave a `mirror`/`replace` sync through and crash mid-run instead of before any I/O.

Fixed: moved `supported_modes()` off the base and onto Postgres/MySQL/Snowflake/Databricks individually (matching how ClickHouse already declared it on its own concrete class). Added a regression test with a deliberately incomplete `BaseSqlDestination` subclass proving it's correctly rejected before I/O now.

Re-verified: 452 tests pass, ruff/mypy clean.

masukai:
Second Codex review pass found two more real gaps:

1. **S3/GCS/Azure Blob would have broken on upgrade.** Their connector docs (`docs/connectors/{s3,gcs,azure-blob}.md`) explicitly document `sync.mode` (including `replace`/`mirror`) as a documented no-op — one object per run, mode never changes behavior. The new fail-fast guard would have turned that documented, pre-existing contract into a breaking failure for anyone currently relying on it. Fixed: all three now declare `ModeCapable` returning `{"replace", "mirror"}` as accepted — runtime behavior is unchanged, only the capability declaration is new. Added regression tests.
2. **`ModeCapable` was missing from ADR 0007's Protocol inventory.** Added it to the freeze-scope table, bumped the Protocol count (21 → 22) everywhere it's cited in ADR 0007 and the v1.0 migration guide, and noted the retroactive addition following the same pattern already established there for `QueryableDestination`/`RateLimiterBackend`/etc.

Re-verified: 85 targeted tests + full suite (3,658 passed, 39 skipped, same 2 pre-existing unrelated failures as before), ruff/mypy clean.

masukai:
Third Codex review pass found one more real gap: `run_unit_test()`'s `CaptureDestination` stand-in (#780) substitutes for the real destination when replaying a fixture through `run_sync()`, but had no `supported_modes()` — so #1042's new guard rejected every `replace`/`mirror` unit-test fixture as a false failure, even for a sync whose real destination (Postgres, Snowflake, etc.) fully supports the mode in production.

Fixed: `CaptureDestination` now accepts every mode unconditionally — a unit test verifies the transform pipeline (row shape/count), not destination write mechanics, so mode-gating it was never the right check to begin with. New regression test proves a `replace`/`mirror` fixture against a mode-capable real destination now passes.

Re-verified: 41 targeted tests pass, full suite 3,660 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

masukai:
Fourth review pass raised one more point I looked at closely but am **not** acting on, with reasoning:

**Finding: a third-party plugin destination that already implements replace/mirror via duck-typed `load()`/`finalize_sync()` (no `ModeCapable`) would start failing after this upgrade, with no staged migration window.**

This is a real, correctly-identified mechanical effect. But it's also exactly the pattern ADR 0007 already classifies, in its own compatibility table, as **not breaking**: "Add a new, `@runtime_checkable`, optional-capability Protocol (`FooCapable`) checked via `isinstance()`, never touching an existing Protocol's method set → No [not breaking] — this is the sanctioned extension path." This is the *only* mechanism this codebase has for adding new capability post-freeze without a major version bump, and it's the same mechanism `MatchPolicyCapable` (#757), `StagedDestination`, `OrphanCleanup`, and `QueryableDestination` (#469) all already shipped through — none of them got a staged migration window or deprecation notice either. `match_policy`'s CHANGELOG entry explicitly says "No breaking changes" for the release that introduced `MatchPolicyCapable` with the exact same mechanical effect on any hypothetical plugin duck-typing that behavior.

Inventing a staged-migration mechanism here, that no other optional-capability Protocol in this codebase has ever used, would be inconsistent with the project's own written policy rather than a fix. The mitigation this PR already provides is the same one every prior optional-capability Protocol relied on: it's a 3-line, purely-additive method (`supported_modes() -> frozenset[str]`) for any plugin author who already has working replace/mirror machinery — `docs/guides/building-a-destination.md`'s "Optional advanced sync modes" section (added by this PR) documents it.

Not blocking on this — flagging the reasoning for the record in case it's revisited.

PR #1058, opened 2026-08-31
fix: three small independent fixes (#1049, #1051, #1043)

### LINKED ISSUE #1043: bug(docs): drt docs serve crashes unconditionally with a stale v0.8.x message; docs/llm/CONTEXT.md is stale
## Problem

`drt docs serve` is registered as a CLI command but unconditionally raises `NotImplementedError`:

```
drt/cli/commands/docs.py:154-159
@docs_app.command(name="serve")
def docs_serve() -> None:
    """Live Web UI for the sync catalog (scheduled for v0.8.x — epic #499)."""
    raise NotImplementedError(
        "`drt docs serve` is scheduled for v0.8.x (Phase 4 of epic #499). "
        "Use `drt docs generate --format mermaid` in the meantime."
    )
```

The message still says "v0.8.x" — we're past v0.10 now and the docs epic (#499) is otherwise considered complete. ADR 0001 named this Phase 4 (`docs/adr/0001-drt-docs-manifest.md:9`).

Separately, `docs/llm/CONTEXT.md` (the LLM-facing SSoT) is also stale: it still says v0.9.0 (`CONTEXT.md:17`) and claims tracked/scoped mirror is Postgres/MySQL-only and non-combinable (`CONTEXT.md:251`) — both wrong since v0.8.4 shipped tracked+scoped mirror on Snowflake/ClickHouse/Databricks too.

## Why it matters

One advertised CLI command crashes on every invocation with an outdated message. The LLM context doc can produce actively wrong advice about current capabilities to anyone (human or AI agent) relying on it as source of truth.

## Suggested direction

Either implement `drt docs serve` (a real live server, matching what `render_html --inline`/`generate` already produce) or remove/hide the stub and its `NotImplementedError` in favor of pointing at `drt docs generate`. Separately, refresh `docs/llm/CONTEXT.md`'s stale version number and mirror-support claims.

**Effort:** Small to remove/deprecate the stub and refresh CONTEXT.md; medium to actually implement live serving.

### LINKED ISSUE #1049: bug(cli): validate --check-connection hardcodes 4 SQL types, ignoring ConnectionTestable destinations that already implement it
## Problem

`drt/cli/commands/validate.py:274` hardcodes four SQL config classes before it even checks the `ConnectionTestable` Protocol. BigQuery and Databricks already implement `test_connection()` (`drt/destinations/bigquery.py:190`, `drt/destinations/databricks.py:1076`), and Airtable/Klaviyo implement it as well.

## Why it matters

A BigQuery or Databricks sync's connection test reports as skipped even though the connector can perform it — the hardcoded gate defeats the whole point of `ConnectionTestable` being an optional structural Protocol, and will keep silently drifting out of sync as new connectors implement the capability.

## Suggested direction

Replace (or precede) the hardcoded type list with an `isinstance(dest, ConnectionTestable)` check, matching the pattern already used elsewhere in the codebase (e.g. `QueryableDestination` dispatch in `drt/destinations/query.py`, from #469).

**Effort:** Small.

### LINKED ISSUE #1051: chore(integrations): SQL Server Dagster sensor cadence asymmetry (ADR 0004) was never filed
## Problem

ADR 0004 (streaming and event-triggered syncs) notes that SQL Server opens a fresh connection per sensor tick but, unlike Snowflake, doesn't require `minimum_interval_seconds` (`docs/adr/0004-streaming-and-event-triggered-syncs.md:82`). The Dagster sensor implementation repeats the same observation almost verbatim, calling it a "real asymmetry" (`integrations/dagster-drt/dagster_drt/sensors.py:58`), but no issue was ever filed for it.

## Why it matters

Without an enforced `minimum_interval_seconds` for the SQL Server sensor, an aggressively-scheduled Dagster sensor falls back to Dagster's own tick cadence and can open unnecessary connections / trigger unnecessary auth load on every tick — a cost Snowflake's sensor is already protected against by its own `minimum_interval_seconds` floor.

## Suggested direction

Give the SQL Server sensor the same `minimum_interval_seconds` floor Snowflake's sensor already has, closing the asymmetry the ADR and the code both already flagged.

**Effort:** Small.

masukai:
Closes #1049. Closes #1051. Closes #1043.

Three independent, small fixes bundled into one PR — each is its own commit, reviewable separately.

## #1049 — `drt validate --check-connection` hardcoded 4 SQL types
`_run_connection_test()` gated on a 4-type isinstance tuple before checking `ConnectionTestable` at all. BigQuery/Databricks/Airtable/Klaviyo (which implement `test_connection()`) were reported as skipped instead of tested. Now dispatches solely on `isinstance(dest, ConnectionTestable)`.

## #1051 — SQL Server Dagster sensor missing `minimum_interval_seconds` requirement
Snowflake's change sensor already requires `minimum_interval_seconds=` (each poll opens a fresh, full-auth-handshake connection). SQL Server has the identical cost shape (`pymssql.connect()` per poll) but wasn't gated the same way — a documented, self-acknowledged asymmetry in this module's own docstring, never actually enforced. Now it is.

## #1043 — `drt docs serve` stale crash + stale mirror-support docs
`docs_serve()` raised a raw `NotImplementedError` claiming "scheduled for v0.8.x" (we're past v0.10). Now a clean CLI error pointing at `drt docs generate`. `docs/llm/CONTEXT.md`'s mirror section still said tracked+scope was Postgres/MySQL-only — stale since v0.8.4 added Snowflake/ClickHouse/Databricks. (Note: this issue's original scope also claimed a stale version number in CONTEXT.md — verified that part was already correct on `main`, not touched here; see the issue's correction comment.)

## Verification

- `pytest tests/unit/test_cli_validate_connection.py tests/unit/test_mcp.py tests/unit/test_docs_mermaid.py` — 95 passed.
- `integrations/dagster-drt` suite — 71 passed.
- Full suite: 3,661 passed, 39 skipped, 2 pre-existing unrelated failures (`test_cli_version.py`, directory-depth artifact of this checkout, not this change).
- `ruff check drt tests integrations/dagster-drt` / `mypy drt` clean.


masukai:
Codex review caught a real gap: #1051's SQL Server \`minimum_interval_seconds\` requirement is a breaking change to \`dagster-drt\`'s already-public API, but I'd only updated \`sensors.py\`'s own docstrings — \`docs/guides/event-driven-syncs.md\` and the package README still said SQL Server "isn't gated the same way" (now false), and the CHANGELOG entry was filed under drt-core's \`[Unreleased]\` instead of a \`dagster-drt\` version bump, where a separately-versioned-package breaking change actually belongs.

Fixed:
- Bumped \`dagster-drt\` to 0.5.0, moved the CHANGELOG entry there with an explicit ⚠️ breaking-change note and upgrade instructions
- Updated \`event-driven-syncs.md\` and the README to drop the stale claim, add an explicit SQL Server example, and note the upgrade path
- Added a dated amendment to ADR 0004 closing the asymmetry it had documented but left open

Re-verified: dagster-drt suite 71 passed, root ruff/mypy clean.

masukai:
Second Codex review pass caught a real one, confirmed by checking \`load()\` in both connectors (pure-write, zero GET calls outside the probe itself) against each connector's own documented minimal credential:

- Klaviyo: \`test_connection()\` probes \`GET /accounts/\`, needs \`accounts:read\`. \`docs/connectors/klaviyo.md:31\` documents "profile + list write access" as sufficient.
- Airtable: \`test_connection()\` probes a record GET, needs \`data.records:read\`. \`docs/connectors/airtable.md:29\` documents \`data.records:write\` as sufficient.

Wiring these into #1049's structural dispatch would have reported a false connection-test failure for anyone following each connector's own least-privilege setup instructions. Fixed by renaming both to \`_test_connection_needs_broader_scope()\` so they no longer satisfy \`ConnectionTestable\` — behavior is byte-identical to before this PR (reported as skipped). Filed #1059 to actually fix the probe (or the docs) properly, since I can't verify the right fix without live API access.

Also caught: an existing contract test (\`tests/unit/test_destination_contract.py\`) explicitly asserted Airtable/Klaviyo *did* implement \`ConnectionTestable\` — updated it to the new, correct expectation.

Re-verified: 182 targeted tests pass, full suite 3,663 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

masukai:
Third review pass, two findings:

**Fixed — BigQuery has the identical Klaviyo/Airtable problem.** \`SELECT 1\` is a query job requiring \`bigquery.jobs.create\`; \`docs/connectors/bigquery.md:59\` documents append-mode's minimal credential as \`bigquery.tables.updateData\` only (\`jobs.create\` is only needed for the merge-path temp table, which insert-mode \`load()\` never touches). Same fix, same reasoning: renamed to \`_test_connection_needs_broader_scope()\`, folded into #1059 alongside Klaviyo/Airtable.

**Not acting on — "renaming test_connection breaks existing callers/API."** Verified this isn't accurate for the actual state of the code: I grepped every \`.test_connection(\` call site across \`drt/\` before making this change (posted in an earlier comment on this PR) and found exactly one production consumer of \`Destination.test_connection()\` — \`drt/cli/commands/validate.py\`, which is the file I'm editing in this same PR. Since the OLD \`is_sql\` gate already excluded BigQuery/Airtable/Klaviyo from ever being reached by that call site in every released version of drt to date, these three \`test_connection()\` methods have never been callable from any shipped CLI or MCP path — there is no released version where "existing callers" of this specific capability on these three connectors could exist. The only other references were this PR's own unit/integration tests, which I updated alongside the rename. \`ConnectionTestable\` itself (the Protocol) is correctly frozen per ADR 0007; which *classes* choose to implement it isn't a compatibility guarantee in either direction — see the identical, already-established precedent from #1042 hours earlier (\`BaseSqlDestination\` deliberately not declaring \`supported_modes()\` it can't serve).

Re-verified: 169 targeted tests pass, full suite 3,664 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

masukai:
Fourth review pass — this one I'm reversing my own earlier decision on. Codex re-raised the compatibility concern, this time citing \`VERSIONING.md\` directly ("Any public class or method imported from \`drt.*\`" is listed under Public API Breaking Changes). I checked the actual text and it's accurate — that clause isn't scoped to frozen Protocols the way ADR 0007 is, it's a blanket rule for any public class/method importable from the package, and it doesn't carve out an exception for "no callers currently exist in this repo." That's the identical principle behind the \`StateStore.now()\` decision from earlier today, which I got right; this is the same call, and I got it wrong on the previous pass by treating "no observed callers" as sufficient justification when the written policy explicitly doesn't condition on that.

Reverted the rename. \`AirtableDestination\`, \`BigQueryDestination\`, and \`KlaviyoDestination\` all keep their normal, public, unchanged \`test_connection()\` and genuinely satisfy \`ConnectionTestable\` again (contract tests confirm this). The exclusion now lives entirely in \`drt/cli/commands/validate.py\`'s \`_run_connection_test()\` — an explicit \`isinstance(dest, (AirtableDestination, BigQueryDestination, KlaviyoDestination))\` check before the generic dispatch. Same practical outcome (\`drt validate --check-connection\` reports them as skipped), zero public API surface change.

Re-verified: 204 targeted tests pass, full suite 3,665 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

PR #1060, opened 2026-08-31
fix(engine): complete dry-run mirror delete previews on Snowflake/ClickHouse/Databricks (#1044)

### LINKED ISSUE #1044: feat(destinations): dry-run mirror DELETE preview is incomplete on Snowflake/ClickHouse/Databricks
## Problem

Since v0.8.4, Snowflake/ClickHouse/Databricks all support `sync.mode: mirror` with `strategy: tracked` + `scope`, but the `--dry-run --diff` preview path (`drt/engine/diff.py`) can't show their destructive pass on all of them:

- `fetch_tracked_state()` (`drt/destinations/query.py`, around line 446 as of this writing) only has Postgres/MySQL branches — no Snowflake/ClickHouse/Databricks case.
- `fetch_all_keys()` (around line 584) dispatches Postgres/MySQL/ClickHouse/Snowflake but has no Databricks branch.
- Both preview paths degrade to "no deletions" on an unsupported type (`drt/engine/diff.py:144`, `diff.py:180`) rather than an explicit "preview unavailable" — so `supported=True` with an empty delete list is indistinguishable from "nothing will be deleted."

## Why it matters

This conflicts with the v0.8.3 changelog claim that mirror DELETEs are previewed under dry-run — for `strategy: tracked` mirrors on Snowflake/ClickHouse, and any mirror mode on Databricks, they currently aren't. An operator trusting `--dry-run --diff` before a mirror run on these dialects could be shown zero pending deletions when deletions will in fact happen.

## Suggested direction

Extend `fetch_tracked_state()`/`fetch_all_keys()` with the missing dialect branches, and make the "can't preview" and "genuinely nothing to delete" cases distinguishable in the diff output rather than both collapsing to an empty list.

**Effort:** Medium.

masukai:
Closes #1044.

## Problem

`drt run --dry-run --diff`'s mirror DELETE preview (`fetch_tracked_state()`/`fetch_all_keys()` in `drt/destinations/query.py`) only had Postgres/MySQL branches, despite all five SQL destinations (Postgres, MySQL, Snowflake, ClickHouse, Databricks) supporting both mirror strategies since v0.8.4. A failed/unsupported read also silently collapsed to the same empty list as "genuinely nothing to delete" — indistinguishable to the operator.

## Fix

- `fetch_tracked_state()` and `fetch_all_keys()` now have real Snowflake/ClickHouse/Databricks branches — each is a plain read-only `SELECT`, independent of how each dialect's actual DELETE/mirror-finalize statement is phrased (ClickHouse's `ALTER TABLE ... DELETE` mutation, Databricks' staged-key anti-join). The real mirror finalize/DELETE code paths are untouched.
- Databricks gains `QueryableDestination` (`get_table_name`/`execute_test_query`) — needed to reach true-diff mode instead of sample mode, and a genuine pre-existing capability gap closed as a side effect (this also means `drt test`'s custom SQL query feature now works on Databricks for the first time).
- `DiffResult.delete_preview_unavailable_reason` (new field) distinguishes a failed/unsupported read from a successful zero-delete read. Text output: `- Deleted (mirror DELETE): preview unavailable` + the error. JSON: `null` on success (including zero deletions), an error string otherwise.

## Verification

- `pytest tests/unit/test_destination_query.py tests/unit/test_diff.py tests/unit/test_cli_diff.py` — 109 passed.
- Full suite: 3,677 passed, 39 skipped, 2 pre-existing unrelated failures (`test_cli_version.py`, directory-depth artifact of this checkout).
- Docs (`docs/guides/dry-run-and-diff.md`) updated; stale "Snowflake needs #468" / "#469 hardcoded types" follow-up notes removed since both are now resolved.


masukai:
Review found 3 issues, all verified:

**Fixed:**
1. **Databricks 255-param-marker limit** — \`_fetch_all_keys_databricks\` built one unchunked query; now chunks via \`_rows_per_chunk()\`, same math every other Databricks write path already uses. New test proves 300 scope values split into 2 correctly-bounded queries.
2. **ClickHouse key-type mismatch** — the real \`_build_mirror_delete\` compares keys as strings on both sides specifically because a typed column (e.g. UUID) won't compare equal to the source's plain value. The preview's SELECT now wraps key columns in \`toString()\` to match, and the diff.py comparison itself normalizes both dest/source keys to string form (safe for every dialect — a no-op when both sides already round-trip the same Python type). New tests cover both the SQL shape and the comparison layer.

**Filed as a follow-up, not fixed here — #1061:** the tracked-mirror preview (\`_preview_tracked_mirror_deletes\`) ignores \`mirror.scope\` entirely. Verified this exists on \`origin/main\` today, unchanged by this PR — it's been live since #693 shipped tracked+scope, affecting Postgres/MySQL already. This PR only changed *how* \`fetch_tracked_state()\` is populated (adding 3 dialects); it didn't touch the scope-filtering logic, which was already absent. Fixing a pre-existing, dialect-agnostic bug is out of scope for a PR about adding dialect coverage.

Re-verified: 112 targeted tests pass, full suite 3,680 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

masukai:
Third review pass found three more, all verified and fixed:

1. **\`_fetch_rows_by_keys_databricks\` (the primary add/update keyed lookup — a different function from the mirror-scope reader already fixed) also exceeded 255 params**, using the generic default \`batch_size=1000\` unbounded. Capped via \`_rows_per_chunk\`, matching the earlier fix.
2. **Enabling Databricks' \`QueryableDestination\` surfaced a real, pre-existing gap**: \`compute_diff()\` passes \`columns=[]\` for the replace-mode full scan, and every dialect's \`_fetch_rows_*\` does \`dict(zip(columns, row))\` — with \`columns=[]\` that's always \`{}\`, silently collapsing every destination row and understating what a real \`replace\` run would delete. Confirmed this is live on Postgres/MySQL/Snowflake **today**, unrelated to this PR — #1060 just made it newly reachable for Databricks (previously it hit a clean \`TypeError\`/sample-mode fallback instead). Fixed for Databricks specifically by deriving columns from \`cursor.description\`; filed [#1062](https://github.com/drt-hub/drt/issues/1062) for the other four dialects, out of scope here.
3. **My own comparison fix from the last round was too broad.** Coercing every dialect's keys to strings (not just ClickHouse's) can turn two *natively equal* values into a false mismatch — e.g. a destination value and source value that already compare correctly under their native Python types could disagree once stringified. Narrowed to ClickHouse only, where the coercion is actually needed to match its real DELETE's own \`toString()\` comparison. New test proves non-ClickHouse dialects are unaffected.

Also re-confirmed the fourth finding (tracked-mirror preview ignoring \`mirror.scope\`) is the same pre-existing issue already filed as #1061 — not re-fixing here for the reasons already given.

Re-verified: 115 targeted tests pass, full suite 3,683 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

PR #1063, opened 2026-08-31
fix: dry-run diff preview correctness — mirror.scope + columns=[] collapse (#1061, #1062)

### LINKED ISSUE #1061: bug(engine): dry-run tracked-mirror delete preview ignores mirror.scope (pre-existing, affects Postgres/MySQL too)
## Problem

Found in Codex review of PR #1060 (dry-run mirror preview completion for Snowflake/ClickHouse/Databricks). Confirmed this is **pre-existing on \`origin/main\` today**, affecting Postgres/MySQL since #693 shipped — not something #1060 introduced, just newly reachable on 3 more dialects since it previously short-circuited to "nothing to preview" for them.

\`drt/engine/diff.py\`'s \`_preview_tracked_mirror_deletes()\` reads the full tracked-mirror state (\`fetch_tracked_state()\`) and computes \`previous - current\` via \`diff_keys()\` — with **no scope filtering at all**, even when \`sync_options.mirror.scope\` is set. The real tracked finalizers (\`BaseSqlDestination._finalize_mirror_tracked\` and each dialect's own) filter stale keys to the scopes actually observed in the current run before deleting.

## Why it matters

A scoped tracked-mirror sync that only touches one parent's rows this run would preview deletions for **every** parent's stale keys, not just the touched one — false positives that also consume the diff's row limit and can hide genuine, in-scope deletions the operator actually needs to see.

## Suggested direction

Derive observed scopes from the source records (same helper \`_observed_scopes()\` the destination-strategy preview already uses), and filter \`previous\`'s keys to those scopes before diffing — matching what \`_finalize_mirror_tracked\` does at run time. Needs prior-state fixtures spanning both observed and unobserved scopes to catch a regression.

**Effort:** Small–medium.

### LINKED ISSUE #1062: bug(engine): fetch_rows(columns=[]) collapses every destination row to {} in replace-mode dry-run diff (Postgres/MySQL/Snowflake/ClickHouse)
## Problem

Found in Codex review of PR #1060 (dry-run mirror preview completion). Confirmed **pre-existing on \`origin/main\` today** for Postgres/MySQL/Snowflake — not introduced by #1060, which only made Databricks newly reachable by this same pre-existing bug (see below).

\`drt/engine/diff.py\`'s \`compute_diff()\` calls \`fetch_rows(config, select_query, columns=[])\` for the replace-mode / no-keyed-fetch full-table scan (line ~315). Every dialect's \`_fetch_rows_<dialect>()\` (\`drt/destinations/query.py\`) builds result dicts via \`dict(zip(columns, row))\` — with \`columns=[]\`, \`zip([], row)\` is always empty, so **every fetched row becomes \`{}\`**. \`compute_diff()\` then keys destination rows via \`row.get(c) for c in upsert_key\` (all \`None\`), collapsing every real destination row into a single dict entry keyed \`(None, ...)\`.

## Why it matters

This is the **replace-mode dry-run diff preview** — \`drt run --dry-run --diff\` on \`mode: replace\`. The bug can make the preview report every source row as "added" while collapsing many real deletions into one keyless deletion entry, severely understating what a real \`replace\` run would actually delete. Confirmed live for Postgres/MySQL/Snowflake today; #1060 additionally made this reachable for Databricks (previously Databricks wasn't \`QueryableDestination\` at all, so this path degraded safely to sample mode instead — #1060 fixed that specific regression for Databricks by deriving columns from cursor metadata there; the same fix is needed for the other four dialects).

## Suggested direction

In each \`_fetch_rows_<dialect>()\`, when \`columns\` is empty, derive real column names from the cursor's own metadata (\`cursor.description\` for DB-API 2.0 drivers; ClickHouse's \`result.column_names\` equivalent) instead of using the caller-supplied (possibly empty) list. Add a regression test per dialect proving a \`columns=[]\` call returns real dicts, not \`{}\`.

**Effort:** Small per dialect, straightforward once the pattern is set on one.

masukai:
Closes #1061. Closes #1062.

Both are correctness bugs in `drt run --dry-run --diff`'s read-only preview, pre-existing on `origin/main` before #1044/#1060 (which made them newly reachable for Snowflake/ClickHouse/Databricks, but didn't cause them — Postgres/MySQL already had both live).

## #1061 — tracked-mirror preview ignored `mirror.scope`

`_preview_tracked_mirror_deletes()` diffed the entire tracked-mirror state against source keys with no scope filtering, unlike the real `_finalize_mirror_tracked()`. A run touching one tenant/parent could preview another tenant's stale keys as deletions — the exact blast-radius mistake `mirror.scope` exists to prevent.

Fixed Python-side (matching what `_finalize_mirror_tracked`'s own docstring confirms is equivalent-correctness): decode each prior key, extract scope-column values by position within `upsert_key`, filter to the scopes this run's source records actually observed, before diffing. Unscoped tracked mirror is unaffected. Tested across all five SQL dialects.

## #1062 — `fetch_rows(columns=[])` collapsed every row to `{}`

`compute_diff()`'s replace-mode full-table scan passes `columns=[]`; `dict(zip([], row))` is always `{}`, silently collapsing every destination row into one keyless entry and severely understating what a real `replace` run would delete. Already fixed for Databricks in #1044/#1060 — this applies the identical pattern (derive column names from query metadata when `columns` is empty) to Postgres, MySQL, Snowflake, and ClickHouse.

## Verification

- `pytest tests/unit/test_destination_query.py tests/unit/test_diff.py tests/unit/test_cli_diff.py` — 125 passed.
- Full suite: 3,693 passed, 39 skipped, 2 pre-existing unrelated failures (`test_cli_version.py`, directory-depth artifact).


masukai:
Review caught a real gap: the scope-filtering/\`decode_key()\`/\`diff_keys()\` logic I added for #1061 ran **outside** the \`try/except\` that already guards \`fetch_tracked_state()\`. Malformed/legacy \`key_json\`, or a \`mirror.scope\` column that isn't actually a member of \`upsert_key\`, would raise uncaught (\`JSONDecodeError\`/\`ValueError\`/\`IndexError\`) and crash the whole \`--dry-run --diff\` command instead of degrading the delete preview to "unavailable" — the established contract every other preview path in this file already follows (an add/update comparison that's perfectly fine shouldn't die because the delete-preview half hit bad state).

Fixed: moved scope filtering and the final \`diff_keys()\` call inside the same guard. New test proves malformed tracked state now degrades cleanly — \`result.supported\` stays \`True\`, \`delete_preview_unavailable_reason\` is set, \`deleted == []\`.

Re-verified: full suite 3,694 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

masukai:
Second review pass found another real one: Snowflake reports unquoted columns in their stored UPPERCASE form (\`cursor.description\` → \`"ID"\`, \`"NAME"\`, ...), but \`compute_diff()\` looks up rows via \`row.get(c) for c in upsert_key\`, where \`upsert_key\` is conventionally lowercase. Unlowered, every lookup returned \`None\`, collapsing every Snowflake replace-mode destination row into one keyless entry — the exact same bug #1062 fixed for the literal \`columns=[]\` case, reappearing via a casing mismatch specific to the one dialect that folds identifiers.

Fixed by lowercasing in the fallback, matching the same convention \`_value_clause\` already uses elsewhere in this codebase for Layer 3 JSON-column matching against Snowflake's \`INFORMATION_SCHEMA\`. Also added the end-to-end test the review specifically asked for — the first version only unit-tested \`fetch_rows()\` in isolation; the new one mocks the real Snowflake cursor and runs the full \`compute_diff()\` pipeline, proving no collapse.

Re-verified: 127 targeted tests pass, full suite 3,695 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

masukai:
Third review pass, correctly caught: the previous fix's blanket lowercase corrupted the opposite case — a sync genuinely configured with an uppercase \`upsert_key\` (itself Snowflake-native, since unquoted DDL stores it that way) needs \`"ID"\` preserved, not folded to \`"id"\`, or the same collapse-into-one-keyless-entry bug reappears from the other direction.

Fixed properly this time: \`fetch_rows()\`/\`_fetch_rows_snowflake()\` now accept an optional \`key_hint\` (the caller's \`upsert_key\`), reconciled case-insensitively per column via a new \`_reconcile_column_case()\` — a column matching a hint entry keeps the hint's casing, everything else falls back to lowercase (the common no-hint case). \`compute_diff()\` passes \`upsert_key\` as \`key_hint\` at its \`columns=[]\` call site.

Also caught my own gap while implementing this: \`compute_diff()\` has **two** \`columns=[]\` call sites (the primary full-scan branch and a ClickHouse-\`NotImplementedError\` fallback), and my initial edit only updated one of them — found and fixed while writing the new test, before it could ship half-broken.

New tests: helper-level \`key_hint\` reconciliation, and an end-to-end \`compute_diff()\` test with an uppercase-configured \`upsert_key\` proving the full pipeline no longer collapses.

Re-verified: 129 targeted tests pass, full suite 3,697 passed / 39 skipped (same 2 pre-existing unrelated failures), ruff/mypy clean.

masukai:
Fourth review pass raised a real but genuinely broader concern than #1062's original scope: casing reconciliation should, in principle, cover every source field (not just \`upsert_key\`) and every dialect whose driver doesn't round-trip configured casing (claimed for MySQL/Databricks too, not just Snowflake) — otherwise a non-key field casing mismatch could misreport unchanged rows as "updated" in the preview.

Not fixing that here. Two reasons:
1. It's a materially larger scope than #1062's original bug (the literal \`columns=[]\` → \`{}\` collapse, which is fully fixed and tested across all 5 dialects) — general per-field, per-dialect case reconciliation is its own design/implementation effort, not a follow-up patch on this PR.
2. The MySQL/Databricks casing claim needs live-platform verification (MySQL's actual case-folding depends on server config like \`lower_case_table_names\` — not something I can confirm from a mock or by reading code, matching this repo's own established "mocks can't prove SQL validity" lesson). Guessing at a fix for unverified platform behavior would be worse than not fixing it yet.

Filed as [#1064](https://github.com/drt-hub/drt/issues/1064) with the reasoning above and a suggested direction (verify against real MySQL/Databricks via \`local_sql_smoke\`/\`dwh_smoke\` first, then extend \`_reconcile_column_case()\` — the helper this PR introduced — to cover all source fields, not just \`upsert_key\`).

This PR's actual scope (#1061's mirror.scope fix, #1062's literal empty-columns collapse fix across 5 dialects, plus the Snowflake upsert_key casing follow-up) is complete and verified — 129 targeted tests, full suite 3,697 passed, ruff/mypy clean.

PR #1065, opened 2026-08-31
fix(destinations): least-privilege connection-test probes for BigQuery/Airtable/Klaviyo (#1059)

### LINKED ISSUE #1059: bug(destinations): BigQuery/Klaviyo/Airtable test_connection() probes need broader scope than documented write-only credentials
## Problem

Found while implementing #1049 (structural `ConnectionTestable` dispatch in `drt validate --check-connection`, caught in Codex review before shipping — the three connectors below were disabled from that dispatch rather than shipped with this gap).

All three probe a broader scope than what's documented as sufficient for their actual write operation:

- `drt/destinations/klaviyo.py`: probes `GET /accounts/` — Klaviyo's granular API-key scoping requires `accounts:read` for this specifically. `docs/connectors/klaviyo.md:31` documents the minimal credential as "profile + list write access" only.
- `drt/destinations/airtable.py`: probes a record `GET` (`maxRecords: 1`) — Airtable's personal access token scoping requires `data.records:read` for this. `docs/connectors/airtable.md:29` documents the minimal credential as `data.records:write` only.
- `drt/destinations/bigquery.py`: probes `SELECT 1` (a BigQuery query job) — requires `bigquery.jobs.create`. `docs/connectors/bigquery.md:59` documents append-mode's minimal credential as `bigquery.tables.updateData` only, with `bigquery.jobs.create` needed *only* for the merge-path temp table — an append-only principal doesn't have it.

For Klaviyo/Airtable, `load()` is verified pure-write (`client.post`/`client.patch`/`client.request` with a write method — no `GET` anywhere outside `test_connection()` itself). For BigQuery, insert-mode `load()` doesn't run a query job at all — only the merge path does. A user following each connector's own documented least-privilege setup would get a **false connection-test failure** despite a fully working sync.

## Current state

All three `test_connection()` methods were renamed to not satisfy the `ConnectionTestable` Protocol (so `drt validate --check-connection` reports them as `skipped`, matching pre-#1049 behavior) rather than shipping the false-failure risk. The implementations are otherwise untouched and easy to reinstate once fixed.

## Suggested direction

Either find a genuinely lower-privilege connectivity probe for each (a lightweight "verify this credential is valid at all" endpoint/call, if one exists — needs checking against real API docs/behavior, not guessed — e.g. BigQuery might have a metadata-only call like listing the dataset that doesn't need `jobs.create`), or explicitly document that these connectors' minimal-credential recommendation needs to include the broader scope if there's no cheaper alternative. Re-enable `ConnectionTestable` (rename back) once resolved, per connector as each is fixed — they don't need to land together.

**Effort:** Small–medium per connector, mostly needs live verification against each API rather than repo-only research.

masukai:
Closes #1059.

All three connectors were excluded from #1049's structural `ConnectionTestable` dispatch because their probes needed broader API scope than each connector's own documented minimal write-only credential — verified live via each vendor's current documentation (not guessed) before implementing.

## BigQuery
`SELECT 1` (a query job, needing the separate project-level `bigquery.jobs.create`) → `get_table()` (needs only `bigquery.tables.get`). Verified via Google's own IAM docs that `roles/bigquery.dataEditor` — the role that already grants `bigquery.tables.updateData`, the documented minimal append-mode credential — also includes `bigquery.tables.get`/`tables.list`/`datasets.get`. Zero additional grant needed; also a more meaningful check since it confirms the actual configured target table specifically.

## Airtable
A record read (`data.records:read`) → `GET /v0/meta/whoami`. Verified via Airtable's official API reference that this endpoint explicitly requires "no scopes" and works with any valid Personal Access Token.

## Klaviyo
No scope-free identity endpoint exists here (confirmed via Klaviyo's developer docs/community — there's no supported key-scope-lookup endpoint). Still probes `GET /accounts/`, but now also accepts Klaviyo's own documented `403` + `code: "permission_denied"` response as proof the key is valid (just more narrowly scoped than `accounts:read`) rather than treating every non-2xx as failure. Any other error (401, other 403 codes, network failure) still raises normally.

## Cleanup
Removed the `_PROBE_NEEDS_BROADER_SCOPE` exclusion tuple in `drt/cli/commands/validate.py` — all three genuinely satisfy `ConnectionTestable` again, so the normal structural dispatch #1049 established is sufficient.

## Verification

- `pytest tests/unit/test_klaviyo_destination.py tests/unit/test_airtable_destination.py tests/unit/test_bigquery_destination.py tests/unit/test_cli_validate_connection.py tests/unit/test_destination_contract.py tests/unit/test_mcp.py` — 212 passed.
- Full suite: 3,705 passed, 39 skipped, 2 pre-existing unrelated failures (`test_cli_version.py`, directory-depth artifact).


masukai:
Reviewed this carefully — it's a real, thoughtful observation, but I'm not acting on it, because it asks BigQuery/Airtable/Klaviyo to meet a bar the existing \`ConnectionTestable\` contract never promises, for **any** destination in this codebase, not just these three.

\`ConnectionTestable\`'s own docstring (\`drt/destinations/base.py:112\`, frozen per ADR 0007): "Optional destination capability for **validating external connectivity**." The method contract: "Raise an exception if the destination **cannot be reached**." Not "can perform the configured write."

Checked what the already-shipped SQL destinations actually do: \`BaseSqlDestination.test_connection()\` (shared by Postgres/MySQL/Snowflake/Databricks) is literally \`SELECT 1\` — a pure connectivity check. A destination user with only \`SELECT\` and zero \`INSERT\`/\`UPDATE\`/\`DELETE\` grants passes this today, on every SQL dialect, and would still fail on the first real \`load()\`. This has been the established, shipped behavior since before #1059 — it's not something this PR introduces or could regress, and it's consistent across every dialect.

Making BigQuery/Airtable/Klaviyo verify actual write-capability for the configured \`sync.mode\` would mean holding these three to a stricter, different contract than every other destination — inconsistent, and outside what #1059 was ever about. #1059's scope was specifically the false-negative bug: a correctly-configured, minimally-scoped, **working** credential reporting failure. That's fixed and verified. Genuinely verifying write-capability (for any destination, not just these three) would be a real, separate feature — worth its own issue and design discussion, not a should-block-this-PR fix, since it changes what \`ConnectionTestable\` promises across the whole codebase.

CI green, ready to merge from my side.

Issue #1093, opened 2026-09-04
bug(config): profile-level fetch_size is documented as configurable in profiles.yml but load_profile() never reads it


masukai:
## Summary

Six profile dataclasses declare a \`fetch_size\` field (DuckDB, SQLite, Postgres, Redshift, Snowflake, SQL Server), added in #765/v0.8.4 as a streaming batch-size knob consumed directly off the profile object at connect time (e.g. \`drt/sources/postgres.py:138\`, \`cur.itersize = config.fetch_size\`).

\`docs/connectors/postgres.md:289\` and \`docs/connectors/snowflake.md:242\` both show \`fetch_size: 10000\` as a \`profiles.yml\` example field. But none of \`load_profile()\`'s six matching branches in \`drt/config/credentials.py\` read \`fetch_size\` from the raw YAML mapping — every profile silently gets the dataclass default (10000) regardless of what's configured in \`profiles.yml\`. \`save_profile()\` has the same gap on write.

Verified: \`grep -n "fetch_size" drt/config/credentials.py\` returns zero matches, and \`config.fetch_size\` is read only off the profile dataclass in each source implementation (\`grep -rn fetch_size drt/sources/\`) — there's no other path that could supply the configured value.

## Impact

Any documented \`fetch_size:\` override in a Postgres/Redshift/Snowflake/SQL Server/DuckDB/SQLite profile is silently ignored — the memory/round-trip tuning knob #765 shipped doesn't work via the documented config surface.

## Suggested fix

Add \`fetch_size=int(raw.get("fetch_size", DEFAULT_FETCH_SIZE))\` to each of the six \`load_profile\` branches and the matching conditional write to \`save_profile\` (write only when it differs from the default, matching the terse-file convention already used for other optional fields).

Found alongside #1092 while auditing \`load_profile()\`/\`save_profile()\` for #402's registry-based refactor.

ump45nose:
I'd like to work on this — the fix is small and contained (read `fetch_size` in `load_profile()` for the six affected profile types and write it back in `save_profile()` when it differs from `DEFAULT_FETCH_SIZE`). Opening a PR shortly.

masukai:
Thanks for picking this up so quickly, and for the thorough test coverage (all 6 profile types, both the load-default and non-default-write paths)! Left a couple of review notes on #1094.

PR #1094, opened 2026-09-04
fix: read and write fetch_size in profiles.yml

ump45nose:
## Summary

`fetch_size` has been a first-class field on the six streaming SQL profile dataclasses (`DuckDBProfile`, `SQLiteProfile`, `PostgresProfile`, `RedshiftProfile`, `SnowflakeProfile`, `SQLServerProfile`) since the #765 streaming work, and the source connectors honor it via `arraysize` / `itersize` / `fetchmany`. But `load_profile()` never read the key from `profiles.yml`, and `save_profile()` never wrote it — so a user who set `fetch_size: 2500` in their profile got silently ignored, with no way to discover the drift by round-tripping the profile.

## Changes

- `load_profile()` now reads `fetch_size` for all six profile types, falling back to `DEFAULT_FETCH_SIZE` (matching the dataclass default) when absent.
- `save_profile()` now writes `fetch_size` back **only when it differs from `DEFAULT_FETCH_SIZE`**, preserving the existing terse-file convention used by the other optional fields.
- `DEFAULT_FETCH_SIZE` is imported from `drt.config.profiles`.

No behavior change for existing profiles: an absent key still resolves to 10000, and a default-valued profile still round-trips without an explicit `fetch_size` line.

## Tests

Added to `tests/unit/test_config.py`:

- `test_load_profile_fetch_size_from_yaml` — parametrized over all six profile types, asserts `fetch_size: 2500` is loaded.
- `test_load_profile_fetch_size_default` — parametrized over all six types, asserts the default is used when the key is absent.
- `test_save_profile_writes_fetch_size_only_when_non_default` — asserts the key is written at 2500 and omitted at the default.
- `test_save_and_load_profile_fetch_size_roundtrip` — full save→load round trip at a non-default value.

## Validation

- `ruff check drt tests` — passes.
- `ruff format --check drt tests` — passes.
- `pytest tests/unit/test_config.py` — 174 passed.
- `mypy drt` — no new errors (the single `credentials.py:183` `no-any-return` diagnostic is pre-existing on `main`, in `load_snowflake_private_key()`, untouched by this PR).

## Notes

- AI-assisted development disclosure: this change was implemented with the assistance of an AI coding agent, then reviewed and verified by a human.


ump45nose:

masukai (dismissed):
Thanks for the fast, well-scoped fix — this is exactly the shape I'd hoped for after filing #1093 (add `fetch_size` to `load_profile()`/`save_profile()` for the 6 affected types, no behavior change for existing profiles, terse-file convention preserved on write).

Verified locally on this branch:
- `ruff check` / `ruff format --check` clean
- `mypy drt` clean (the correct CI scope — `mypy drt tests/unit/test_config.py` reports pre-existing unrelated errors in that file that aren't part of `make lint`'s target)
- New tests: 14/14 pass
- Full unit suite: 3906 passed, 16 skipped

One heads-up, not a blocker: this branch was opened against `main` before #1095 merged (a different fix in the same file, for `SnowflakeProfile.private_key_env`), so there's now a small conflict — both PRs add a `### Fixed` entry to `CHANGELOG.md` in the same spot. `drt/config/credentials.py` and `tests/unit/test_config.py` merge cleanly with no actual overlap. Whenever it's convenient, a rebase onto current `main` (`git fetch origin && git rebase origin/main`) should resolve the CHANGELOG.md conflict by just keeping both entries.

Nit for a future pass, not asking for a change here: the two new parametrized tests take a `profile` fixture but only ever read `.type` off it inside an f-string — a plain list of type-name strings would read a touch more directly, and the parameter's missing its `: ProfileConfig` type hint (the only tests in this file without one). Not worth a revision cycle for.

masukai:
\`codecov/patch\` is failing (33% patch coverage, target 80%) — I dug into why, since the new tests looked thorough at a glance.

The gap: \`test_save_profile_writes_fetch_size_only_when_non_default\` and \`test_save_and_load_profile_fetch_size_roundtrip\` only exercise \`save_profile()\`'s new \`fetch_size\` write for \`PostgresProfile\`/\`DuckDBProfile\`. The \`load_profile()\` side is fully parametrized across all 6 types, but the \`save_profile()\` write branches for \`SQLiteProfile\`, \`RedshiftProfile\`, \`SnowflakeProfile\`, and \`SQLServerProfile\` (the \`if profile.fetch_size != DEFAULT_FETCH_SIZE: entry["fetch_size"] = ...\` lines you added to each) never run under any test, which is the uncovered-patch-lines codecov is flagging.

Easiest fix: parametrize \`test_save_profile_writes_fetch_size_only_when_non_default\` (or add a sibling test) across all 6 profile types the same way \`test_load_profile_fetch_size_from_yaml\` already is, asserting the non-default write for each. That should close the gap without needing new test logic.

masukai:
Heads up — I just re-ran CI (it had been sitting on a pending approval) and it's now failing lint, separately from the codecov gap above:

\`\`\`
F811 Redefinition of unused \`SnowflakeProfile\` from line 11
  --> tests/unit/test_config.py:32:5
   |
30 |     PostgresProfile,
31 |     RedshiftProfile,
32 |     SnowflakeProfile,
   |     ^^^^^^^^^^^^^^^^ \`SnowflakeProfile\` redefined here
   |
help: Remove definition: \`SnowflakeProfile\`
\`\`\`

The new \`from drt.config.profiles import (...)\` block you added imports \`SnowflakeProfile\` again — it's already imported at the top of the file from \`drt.config.credentials\` (line 11), which re-exports it. Easiest fix: drop \`SnowflakeProfile\` from the new import block and just use the one already in scope.

So there are two things to fix in the same push: this lint error, plus the codecov coverage gap from the comment above (parametrizing \`save_profile()\`'s fetch_size write test across all 6 profile types). No rush — take your time.

ump45nose:
Thanks for the detailed review — both points are addressed in `3b25499`:

1. **F811**: dropped the duplicate `SnowflakeProfile` from the new `drt.config.profiles` import block (it stays in scope via the `drt.config.credentials` re-export at line 11).
2. **codecov**: parametrized `test_save_profile_writes_fetch_size_only_when_non_default` across all six streaming SQL profile types (DuckDB, SQLite, Postgres, Redshift, Snowflake, SQLServer), so each `save_profile()` `fetch_size` write branch now runs under both the non-default and default paths.

Local validation:
- `ruff check drt tests` → All checks passed
- `ruff format --check drt tests` → formatted
- `pytest tests/unit/test_config.py` → 181 passed (the parametrized save test is 6/6)

The MySQL unit-test failures on a bare local run are only `ModuleNotFoundError: No module named 'pymysql'` (optional dependency), unrelated to this change.

masukai (approved):
Both review points addressed cleanly — thanks for the quick turnaround! LGTM.

masukai:
Merged — thanks for the clean fix and for working through the review rounds! This is your first contribution to drt, so welcome 🎉 If you find the project useful, a ⭐ on the repo would mean a lot: https://github.com/drt-hub/drt

Looking forward to more contributions if anything else catches your eye.

Issue #1096, opened 2026-09-04
bug(config): save_profile() never writes BigQuery profile's location back to profiles.yml


masukai:
## Summary

\`load_profile()\`'s bigquery branch correctly reads \`location\` (\`raw.get("location", "US")\`, see \`test_load_profile_bigquery_location\` / \`test_load_profile_bigquery_location_default\` in \`tests/unit/test_config.py\`), but \`save_profile()\`'s bigquery branch never writes it back:

\`\`\`python
elif isinstance(profile, BigQueryProfile):
    entry: dict[str, Any] = {
        "type": "bigquery",
        "project": profile.project,
        "dataset": profile.dataset,
        "method": profile.method,
    }
    if profile.keyfile:
        entry["keyfile"] = profile.keyfile
\`\`\`

So a \`BigQueryProfile\` with a non-default \`location\` (e.g. \`"asia-northeast1"\`) silently loses it on \`save_profile()\` → \`load_profile()\` round trip — the reloaded profile always comes back with \`location: "US"\` regardless of what was originally set.

## Reproduction

\`\`\`python
from drt.config.credentials import BigQueryProfile, load_profile, save_profile

p = BigQueryProfile(type="bigquery", project="x", dataset="y", location="asia-northeast1")
save_profile("dev", p, config_dir=some_tmp_dir)
loaded = load_profile("dev", config_dir=some_tmp_dir)
assert loaded.location == "US"  # not "asia-northeast1"
\`\`\`

## Suggested fix

Add the same guarded-write pattern used for \`keyfile\`:

\`\`\`python
if profile.location != "US":
    entry["location"] = profile.location
\`\`\`

## Scope note

Found while building a round-trip regression baseline for #402's registry-based refactor of \`load_profile()\`/\`save_profile()\`. Intentionally not fixed as part of that PR to keep it a pure refactor with no behavior change — filed separately here.

Issue #1092, opened 2026-09-04
bug(config): load_profile()/save_profile() never read/write Snowflake private_key_env, breaking documented key-pair auth


masukai:
## Summary

\`load_profile()\`'s hand-written \`snowflake\` branch in \`drt/config/credentials.py\` never reads \`private_key_env\` / \`private_key_passphrase_env\` from the raw YAML mapping, even though \`docs/connectors/snowflake.md:67\` documents both as applying to the source profile in \`~/.drt/profiles.yml\`, and #737 shipped key-pair auth specifically because new Snowflake accounts enforce MFA on password sign-in.

\`drt profile add\`'s raw-dict write path (\`write_raw_profile()\`) persists the fields to disk correctly. Only the typed load path used at actual sync time silently drops them — so a profile configured exactly as documented reproducibly connects with \`private_key_pem=None\` instead of the configured key pair (verified live).

\`save_profile()\` has the matching gap on write: a profile round-tripped through the typed path (\`save_profile()\` then \`load_profile()\`) also loses both fields.

## Fix

Being fixed directly — small, mechanical, two-line addition to each of \`load_profile\`'s and \`save_profile\`'s snowflake branches, matching the existing \`password_env\` handling. Found while auditing this module for #402.

PR #1097, opened 2026-09-04
refactor(config): registry-based load_profile()/save_profile()

### LINKED ISSUE #402: refactor: registry-based load_profile() / save_profile()

masukai:
## Summary

Closes #402.

`load_profile()`/`save_profile()` were each a 13-branch hand-written `if`/`elif` chain — one branch per built-in source type, mixing dispatch with each type's construction/serialization quirks (per-field defaults, `int()` port coercion, required-field checks, terse-write conditionals). Adding a new built-in type meant editing both chains.

Replaced with a per-type spec table: each type gets a small `_load_<type>`/`_dump_<type>` function pair (all per-type logic preserved verbatim — no behavior consolidation), registered in `_PROFILE_LOADERS: dict[str, loader]` / `_PROFILE_DUMPERS: dict[type, dumper]`. `load_profile()`/`save_profile()` become a lookup + call. Adding a new built-in type now means one table entry plus its two functions.

`_DISPATCHED_SOURCE_TYPES` (used for the "unsupported source type" error message and the plugin-fallback boundary) is now derived from `_PROFILE_LOADERS.keys()` instead of being a separately hand-maintained literal set — closes the drift surface `test_every_registered_source_is_dispatched_or_reachable` was already guarding against by hand.

## What didn't change

- `#997`'s plugin-registry fallback in `load_profile()` and the `is_dataclass` fallback in `save_profile()` — unchanged, still run only after every built-in is checked in the table, preserving "a plugin cannot shadow a built-in."
- Every per-type default, required-field error message, and optional-field write condition — copied verbatim into its own function.
- Dispatch order and semantics — dict lookup by `type:` string (load) / exact `type(profile)` (save, mirroring the original `isinstance` chain; no built-in profile dataclass is subclassed, so this is equivalent).

## Verification

Written as a genuine regression baseline, not an after-the-fact check: `test_save_profile_load_profile_roundtrip_all_builtin_types` (parametrized over all 13 built-in profile types, `save_profile()` → `load_profile()` → equality) was written and confirmed **passing against pre-refactor `main`** before any refactor code was touched.

- Full unit suite: 3907 passed, 16 skipped
- `ruff check` / `ruff format --check` / `mypy drt` — all clean

## Bugs found while building the baseline (deliberately not fixed here)

Building the round-trip baseline surfaced two real, pre-existing bugs unrelated to this refactor's scope — kept out to keep this a behavior-neutral PR:

- **#1096** (filed): `save_profile()` never writes a non-default BigQuery `location` back to `profiles.yml`.
- **#1092** (already fixed, merged separately): Snowflake `private_key_env`/`private_key_passphrase_env` were dropped by both `load_profile()` and `save_profile()`.

## Test plan

- [x] New round-trip regression test covers all 13 built-in types, confirmed it fails appropriately before the fix would apply (i.e. passes on `main` as a true baseline, not a change-dependent assertion)
- [x] Full unit suite green


masukai:
Codex review: no findings

Reviewed the diff against main (per-type `_load_<type>`/`_dump_<type>` refactor of `load_profile()`/`save_profile()`). Verdict: "The refactor preserves the existing per-profile loading and serialization behavior while retaining built-in precedence and subclass-aware save dispatch. No functional regressions were identified."

Consistent with this PR's own round-trip regression baseline (confirmed passing on pre-refactor main before the refactor was written) and the full green unit suite already reported in the PR description. Nothing further to flag from this pass.

masukai:
Rebased onto current main to resolve a merge conflict — this branch predated #920/#1106's `managed_schema` and #765's `fetch_size` additions to `load_profile()`/`save_profile()`'s old if/elif chain (both fields missing from `_load_postgres`/`_load_snowflake`/etc., not just the two flagged conflict markers). Ported all of it into the corresponding `_load_<type>`/`_dump_<type>` functions, removed the now-redundant old if/elif chain entirely (kept the lookup-table dispatch), fixed one duplicate-import lint error the merge surfaced in `test_config.py`. Full unit suite (3991 passed, 3 skipped), ruff, and mypy all clean after the merge. `mergeable` is now `MERGEABLE`.

Issue #1100, opened 2026-09-05
feat(state): compliance audit trail — queryable record-to-destination delivery log, with built-in retention/purge


masukai:
## Problem

Reverse ETL has a liability shape data-loading tools don't: it actively pushes data **out** to third-party destinations (CRMs, ad platforms, messaging APIs). "Which record went where, and when" is a question a GDPR/compliance request can ask, and today drt has no answer beyond whatever's in local sync history (unstructured, not queryable, not retained with any compliance framing).

## Competitive context

Verified via research (2026-09): Hightouch's `hightouch_audit.Changelog` table is the closest existing analog — it records field-level before/after changes per sync. But Hightouch's own docs frame this as a **liability the customer must manage**, not a feature: PII sits in `Changelog` indefinitely with **no automatic purge**, and Hightouch's docs explicitly warn EU customers to keep syncs running so old PII gets refreshed/cleared manually — the retention burden is pushed onto the customer entirely. No vendor researched (Hightouch, Segment, RudderStack, Census, Polytomic, Rivery) ships a retention-managed, purge-aware audit trail as a first-class product feature.

This is the differentiation opportunity: the underlying mechanism (a delivery-log table) is not novel, but *owning the retention/purge contract instead of leaving it to the customer* is unclaimed.

## Proposal

```yaml
# drt_project.yml
state:
  backend: warehouse
  audit_trail:
    enabled: true
    retain_days: 30              # required when enabled — no unbounded-retention default
    fields: [email, user_id]     # which source fields to log per delivered record (not the full payload)
```

- `_drt_audit_log(sync_name, run_id, record_key, logged_fields_json, destination_type, delivered_at)` table, managed via #960's primitive.
- **Deliberately does not log full record payloads by default** — only the configured identifying fields (mirrors the existing `sync.mask` (#427) posture of being conservative with PII by default; logging everything by default would turn the audit table into the exact liability Hightouch's is).
- **`retain_days` is a required field when `audit_trail.enabled: true`**, not an optional one with a silent "forever" default — forces an explicit retention decision at config time, which is precisely the gap in Hightouch's design this issue is responding to.
- Auto-purge: a housekeeping pass (triggered at sync start, or a separate `drt state prune` extension — needs a design decision, not prescribed here) deletes rows past `retain_days`. This is the feature Hightouch's docs ask the *customer* to build by hand; here it's drt's job.
[...]

Issue #1099, opened 2026-09-05
feat(state): warehouse-backed idempotency ledger for fire-and-forget destinations


masukai:
## Problem

#897 already scoped the "pass a native idempotency mechanism through to destinations that have one" case (Stripe-style header, dedup ID field) — that stays a small, self-contained feature and is unaffected by this issue.

#897's own "Out of scope" section names the gap this issue fills: for a destination with **no native dedup mechanism** (most webhook-style / fire-and-forget destinations), the only way to give drt's own retry a strong guarantee against double-sending is a dedup ledger drt maintains itself. #897 explicitly deferred this as "a much larger, stateful feature" — this issue is that feature, now that #960 gives it somewhere durable and transactional to live.

## Why this needs the warehouse (not local/object-storage state)

A dedup check-and-mark has to be atomic against concurrent writers (two overlapping `drt run` processes, or a retry racing the original attempt) or it doesn't actually prevent the duplicate it exists to prevent:

- **Local JSON state**: no cross-process atomicity beyond the file-lock work already shipped (#963) — a lock serializes access but a JSON read-modify-write is not the same guarantee as a database-level atomic upsert, and doesn't scale to a high-cardinality per-row ledger anyway (one row per delivered record, potentially millions).
- **Object storage (GCS/S3, #756)**: same shape — a per-record ledger entry would mean one object (or one growing object) per record, either way the wrong primitive for row-level dedup at scale.
- **Warehouse (this issue, built on #960)**: a `MERGE ... WHEN NOT MATCHED THEN INSERT` (or dialect equivalent) against a `_drt_idempotency` table is a single atomic statement — check-and-mark in one round trip, with the warehouse's own transactional guarantees doing the work drt would otherwise have to build from scratch.

## Competitive context

Verified via research (2026-09): **no reverse-ETL vendor researched (Hightouch, Segment, RudderStack, Census, Polytomic, Rivery) offers this today.** All of them solve diff-based incremental and SQL-queryable history with a warehouse-managed schema, but none extend that same pattern to a durable dedup ledger for at-least-once delivery. This is currently unclaimed territory in the category, not a catch-up feature — see the "Assessment" section of the research (also referenced from ADR 0005's follow-up).

## Proposal

```yaml
# drt_project.yml
state:
  backend: warehouse
  idempotency: true              # opt-in on top of the warehouse backend, not automatic
```

- `_drt_idempotency(sync_name, idempotency_key, delivered_at)` table, managed via #960's primitive.
[...]

Issue #1064, opened 2026-08-31
bug(engine): dry-run diff preview needs general case-insensitive column reconciliation, not just upsert_key


masukai:
## Problem

Found across multiple Codex review passes on PR #1063 (which fixes the narrower, literal \`fetch_rows(columns=[])\` → \`{}\` collapse bug, #1062, and the Snowflake upsert_key casing follow-up). The scope kept broadening beyond what #1062 originally covered, so splitting it out here rather than chasing it indefinitely in that PR.

The fix shipped in #1063 (\`key_hint\` reconciliation in \`_fetch_rows_snowflake\`) only reconciles casing for \`upsert_key\` columns, and only for Snowflake. Two related gaps remain, per the last review pass:

1. **Non-key field casing isn't reconciled at all, on any dialect.** \`compute_diff()\`'s \`changed_fields()\` comparison (the "updated" detection) compares source records against destination rows field-by-field. If a destination's non-key column casing differs from the source's field naming — e.g. Snowflake's unquoted \`SCORE\` falling back to lowercase \`score\` while the source record's field is actually \`Score\` or some other non-lowercase casing — every row could be misreported as changed even when nothing changed. Not yet confirmed as a live bug (needs a source connector that actually produces non-lowercase field names in practice — TBD which dialects can), but the mechanism is real.
2. **MySQL/Databricks column-name case sensitivity was flagged but not verified.** Unlike Snowflake (which reliably uppercase-folds unquoted identifiers), MySQL's and Databricks' actual case-folding behavior depends on server/session configuration (e.g. MySQL's \`lower_case_table_names\`, collation) that can't be verified via static code reading or mocks — this needs a live \`local_sql_smoke\`-style test against a real MySQL instance with a mixed-case or uppercase-configured \`upsert_key\` to confirm whether the same collapse bug #1062/#1063 fixed for Snowflake actually reproduces there, or whether MySQL's behavior already matches source casing without intervention (unlike Snowflake's forced uppercase-folding).

## Suggested direction

- Verify the MySQL/Databricks claim against a real instance (\`local_sql_smoke\` for MySQL; \`dwh_smoke\` for Databricks) before assuming it needs the same fix — don't guess platform SQL-identifier-folding behavior from a mock.
[...]

PR #1103, opened 2026-09-05
Add warehouse-managed table primitive, Postgres-first (#960)

masukai:
## Summary

New optional `ManagedTableCapable` source Protocol (`drt/sources/base.py`, same pattern as `IncrementalSource`/`QueryableDestination` per ADR 0007) with `ensure_managed_schema`/`managed_table_exists`/`drop_managed_table`, plus a new `PostgresProfile.managed_schema` field (default `_drt`, never `public`) naming where drt's own bookkeeping tables live.

This is pure infrastructure — no feature in this release reads or writes the managed schema yet. It exists so #755, #920, #1099, and #1100 have a shared, already-verified foundation to build their own tables on, instead of each reinventing create-if-absent / escape-hatch / reversibility handling independently.

- Deliberately **no caller-supplied column definitions**: each consuming feature owns its own table's DDL, hardcoded per dialect — matching tracked mirror's `_create_state_table` hook rather than interpolating a caller-controlled column list into SQL text (the exact class of bug #1064/#1090 found and fixed on the diff-preview read path).
- **Escape hatch and reversibility** (ADR 0005 Decision 4) both live-verified against a real Postgres via testcontainers: a role with `CREATE` explicitly revoked plus a pre-created schema succeeds without the `CREATE SCHEMA` statement ever being issued; create-a-table-then-drop-it leaves the shared schema untouched and a second drop is a safe no-op.
- Snowflake/Databricks/BigQuery deferred — no `DRT_SMOKE_*` credentials in this environment to live-verify them, and BigQuery has no existing mirror-mode precedent to model the implementation on at all. Postgres lands alone rather than shipping three more dialects backed only by mock-cursor tests.

**Follow-up fix included in this branch:** the initial `managed_table_exists()` used `to_regclass('schema.table')` as a formatted string — `to_regclass` parses its argument as an identifier and silently lowercases an unquoted mixed-case schema name, so it never found a table in a schema that `ensure_managed_schema()` had genuinely created case-sensitively via `Identifier()`. Live-reproduced against a real Postgres, then fixed by binding schema and table as separate bound parameters to a plain `information_schema.tables` lookup instead (no identifier parsing), matching MySQL's `_state_table_exists` for the same reason. A live regression test with a mixed-case schema now guards this. Also removed `drt/sources/_managed_state.py` (an unused constant nothing imported) — its naming-convention note is folded into `ManagedTableCapable`'s docstring instead.

## Test plan

- [x] `ruff format --check` clean on all changed files
- [x] `mypy drt` clean
- [x] `pytest -q -m "not dwh_smoke"` — 3912 passed, 16 skipped
- [x] `pytest tests/integration/local_sql/test_managed_table_primitive_smoke.py -m local_sql_smoke` — 3 passed against a real Postgres via testcontainers (escape hatch, reversibility, mixed-case schema regression)
- [x] Escape hatch and the mixed-case bug were both verified by deliberately breaking the implementation first and confirming the smoke test actually catches the failure, before fixing and re-confirming green

Closes part of #960 (Postgres leg only — Snowflake/Databricks/BigQuery tracked separately, see summary above).


masukai:
**Codex review:** adversarial review flagged two issues, both addressed in 25d7f8d.

1. **`managed_table_exists` accepted the wrong PostgreSQL object type** — `information_schema.tables` lists views/foreign tables alongside base tables; the probe didn't filter `table_type`, so a same-named view would read back as "exists" and a later `drop_managed_table` would fail (DROP TABLE errors on a view). Fixed with the same `table_type = 'BASE TABLE'` filter the destination catalog query already uses, live-verified against a real Postgres with a same-named view, plus a permanent regression test.
2. **New public Protocol (`ManagedTableCapable`) wasn't in ADR 0007's compatibility inventory** — added it to the ADR's Protocol count/list and freeze-scope table, explicitly marked "not yet frozen" (one implementer, no consumer yet).

Both were live-verified: broke the fix intentionally first, confirmed the new tests catch it, then re-fixed and re-confirmed green.

masukai:
**Self-review finding (Codex hit its usage limit mid-review on #920, the first real consumer of this primitive, while specifically checking whether table bootstrap is safe under simultaneous first use):** \`ensure_managed_schema()\`'s \`CREATE SCHEMA IF NOT EXISTS\` is not atomic across Postgres sessions. Two sessions can both pass the existence probe (schema doesn't exist yet) and both attempt the \`CREATE\` — the loser gets a catalog \`UniqueViolation\` (\`pg_namespace_nspname_index\`), not a graceful no-op.

Reproduced live: 8 threads calling \`ensure_managed_schema()\` for the first time on a fresh database raised on 3 of 8 threads, every run. This is a realistic scenario, not contrived — \`drt run --threads N\` running multiple syncs that each touch the managed schema for the first time hits exactly this path.

Fixed in 2e39f7e by catching the race and re-confirming the schema exists before swallowing the error. New live regression test with 8 concurrent callers.

masukai:
**Codex review:**

Target: branch diff against `main`

> The new Postgres managed-schema capability uses parameter binding and identifier quoting appropriately, handles concurrent schema creation, and preserves existing source behavior. No actionable correctness regressions were identified.

PR #1104, opened 2026-09-05
Add warehouse-backed state, history, and DLQ, Postgres-first (#920)

masukai:
## Summary

Stacked on #1103 (the Postgres managed-table primitive this backend builds on, #960). ADR 0005 step 4.

**This PR was restructured after two rounds of Codex review** — see the "Design correction" note below.

- New `state.backend: warehouse` reuses an existing connection profile (`state.connection_profile` in `profiles.yml`) instead of a bucket, so `StateStore`/`HistoryStore`/`DlqBackend` (#756's Protocols) get a second, SQL-queryable implementation alongside `local`/`gcs`/`s3`.
- Three tables, one per Protocol, DDL hardcoded in `drt/state/warehouse.py` (the managed-table primitive deliberately owns only schema/table *existence*, not column definitions): `_drt_runs` (one row per sync, upserted), `_drt_history` (append-only), `_drt_dlq` (one row per dead-letter entry, keyed by id) — all under the connection's managed schema (`PostgresProfile.managed_schema`, #960).
- Every write is a single `INSERT ... ON CONFLICT DO UPDATE` / plain `DELETE` — atomic per row via Postgres's own row lock. Unlike the object-storage backends (`drt/state/_objectstore.py`), this one never needs a retry loop and never raises `StateContentionError`: the failure class #919 exists to prevent (a client-side read-modify-write race) structurally cannot happen when the database does the read-modify-write atomically server-side.
- `state.connection_profile`, not `state.profile` — `ProjectConfig` already has a top-level `profile` field meaning "the project's default connection profile", a different concept this avoids colliding with.
- A non-Postgres profile fails loudly with `NotImplementedError` rather than silently misbehaving. Snowflake/BigQuery/Databricks are **next up as immediate follow-up issues, not indefinitely deferred backlog** — this environment has no `DRT_SMOKE_*` credentials to live-verify them today (provisioning scripts exist under `tests/integration/dwh/provisioning/`, but need real cloud accounts).
- `_drt_watermarks` was in #920's original proposal but is out of scope — watermarks live behind the separate `WatermarkStorage` Protocol (`sync.watermark.storage`), not `StateBundle`. Recorded in [the issue's comments](https://github.com/drt-hub/drt/issues/920#issuecomment-5555166189).

**Design correction recorded during review:** an earlier version of this stack split config and implementation across two PRs, and the first (config-only) PR briefly widened `state.backend`'s `Literal` to include `"warehouse"` ahead of any implementation. Codex review caught this as a schema-valid configuration that would deterministically crash `run`/`status`/`retry`/`serve`/MCP state operations at runtime. Checked against #756's own git history: `"gcs"` was added to this same `Literal` in the *same PR* as the GCS backend implementation (`d58a9c4`), not in the config-only PR before it — so this PR folds config and implementation back into one, matching that precedent.

Next in the stack: docs/CHANGELOG/ADR updates (ADR 0005's Implementation ordering table, `docs/guides/remote-state.md`).

## Test plan

- [x] Unit tests: config validation (`state.connection_profile` required/rejected), factory dispatch (Postgres profile → warehouse bundle, non-Postgres profile → `NotImplementedError`), `warehouse.py` module (UndefinedTable fallback paths, row reconstruction) — `tests/unit/test_state_factory.py`, `tests/unit/test_state_warehouse.py`
- [x] Live smoke tests against a real Postgres via testcontainers (`tests/integration/local_sql/test_warehouse_state_backend_smoke.py`, `local_sql_smoke` marker): state upsert/read/reset round-trip, cross-sync history merge ordering + retention pruning, DLQ FIFO capping + `reconcile()`'s remove/update-by-identity semantics, shared managed schema across stores
- [x] FIFO-cap ordering deliberately broken and confirmed the smoke test catches it (`ORDER BY seq ASC` instead of `DESC`), then fixed and re-confirmed green
- [x] `pytest -q -m "not dwh_smoke"` — 3941 passed, 16 skipped
- [x] `pytest tests/integration/local_sql/ -m local_sql_smoke` — 12 passed (includes #960's own primitive tests)


masukai:
**Codex review:** first pass flagged a real, high-severity issue — I had widened \`StateConfig.backend\`'s \`Literal\` to include \`"warehouse"\` in this PR, ahead of the actual backend implementation. Since \`build_state_bundle()\` unconditionally raises \`NotImplementedError\` for it, that made a schema-valid config deterministically crash \`run\`/\`status\`/\`retry\`/\`serve\`/MCP state operations at runtime — not just a theoretical risk, since PRs merge to \`main\` individually as they land.

Checked #756's own git history to find the actual precedent I should have followed: \`"gcs"\` was added to this same \`Literal\` in the *same PR* as the GCS backend implementation (\`d58a9c4\`), not in the config-only PR before it (\`1656244\`, which kept \`backend: Literal["local"]\` while still adding the then-unusable \`bucket\`/\`prefix\` fields). Force-pushed a corrected version that mirrors that shape: \`backend\` stays \`Literal["local", "gcs", "s3"]\`, only the new (currently-unusable) \`state.profile\` field is added, and the \`Literal\` widens to \`"warehouse"\` together with the Postgres backend in the next PR of this stack.

masukai:
**Codex review:** two rounds. First flagged a real, high-severity issue — I had widened \`StateConfig.backend\`'s \`Literal\` to include \`"warehouse"\` in a config-only PR, ahead of the backend implementation, which would let a schema-valid config deterministically crash at runtime. Second round flagged a related issue in my fix — the added \`state.profile\` field was schema-valid but unconditionally rejected by every real backend, a "tooling accepts, runtime rejects" gap.

After checking #756's own git history for the actual precedent (\`"gcs"\` was added to the \`Literal\` in the same PR as the GCS backend itself, not before it — \`d58a9c4\`), I folded config and implementation back into a single PR, matching that shape exactly. Also caught in this pass: \`state.profile\` collides conceptually with \`ProjectConfig.profile\` (a different, pre-existing field) — renamed to \`state.connection_profile\` before anything shipped.

masukai:
**Codex review, round 3:** three findings.

1. **[critical→resolved by existing precedent] No enforced per-project namespace** — rows key only by \`sync_name\`, so two projects sharing one \`connection_profile\`/\`managed_schema\` collide. Checked against \`gcs\`/\`s3\`: they have the identical property today (\`state.prefix\` is an optional operator convention, not an enforced tenant key — see \`docs/guides/remote-state.md\`'s own examples). This backend's equivalent knob is \`PostgresProfile.managed_schema\` (already settable per profile since #960). Documented in the module docstring rather than adding an enforced namespace column, to stay consistent with the rest of the state-backend surface.
2. **[high→resolved by existing precedent] Concurrent runs of the same sync can move a cursor backward** — true, but not novel: \`drt/engine/observer.py\`'s own \`on_sync_completed\` comment already documents this exact race as an accepted, cross-backend gap needing a StateStore Protocol-level CAS primitive neither local, gcs, nor s3 have either. Documented in the module docstring with the exact code reference.
3. **[high→fixed] Pre-provisioned tables still triggered \`CREATE TABLE\`** — genuine gap, real regression against #960's own escape-hatch discipline (probe via \`managed_table_exists()\` before every \`CREATE\`, same as the schema check). Fixed in 14b9169, plus a new live test with a role that has \`CREATE\` fully revoked and all three tables pre-provisioned by an admin.

That live test then caught two more real bugs on its own, before I'd even finished reading the review: \`_drt_history\`'s \`id BIGSERIAL\` and \`_drt_dlq\`'s \`seq BIGSERIAL\` columns each have an implicit sequence requiring its own \`GRANT USAGE\`, which the escape hatch's plain table-privilege grants don't cover — \`permission denied for sequence\` on the very first live run with a locked-down role. Removed both columns entirely (history's \`id\` was unused by any query; DLQ orders by \`(ts, id)\` instead of \`seq\`) rather than documenting a second required grant.

masukai:
**Codex review, round 4:** three findings — two repeats of round 3's namespace/monotonicity findings (already addressed in the module docstring with existing-precedent evidence, see the previous comment; not re-litigated since nothing new was raised against that reasoning), and one genuinely new, real bug.

**[high→fixed] \`DlqBackend.replace()\` wasn't atomic** — it deleted the old queue on one connection, committed, closed it, then called \`append()\` which opens a **second** connection for the replacement insert. A connection failure between the two would permanently erase the queue with nothing written back. Fixed in e0f11a4: one connection, one transaction, delete and every replacement insert together, one commit at the end. Added a live test that forces a mid-transaction failure (a duplicate id genuinely violates the PRIMARY KEY here, since \`replace()\`'s plain INSERT has no \`ON CONFLICT\` the way \`append()\`'s upsert does) and confirms the original queue survives completely untouched — Postgres's own uncommitted-transaction rollback does the rest.

masukai:
**Self-review (Codex hit its usage limit mid-round-5; I picked up the two areas it was checking when it ran out):** "whether table bootstrap is safe under simultaneous first use."

Reproduced a real, previously-unknown Postgres concurrency gotcha, one layer deeper than expected. \`CREATE TABLE IF NOT EXISTS\` (and, it turns out, \`CREATE SCHEMA IF NOT EXISTS\` in #1103's \`ensure_managed_schema()\`) is **not atomic across sessions**: two sessions can both pass the existence probe and both attempt the \`CREATE\` — the loser gets a catalog \`UniqueViolation\`, not a graceful no-op.

Reproduced live with 8 threads calling \`save_sync()\`/\`ensure_managed_schema()\` for the first time on a fresh database — 3+ of 8 threads raised, every single run. This is a realistic scenario, not contrived: \`drt run --threads N\` running multiple syncs that each touch the managed schema/tables for the first time hits exactly this path.

Two fixes, on two branches:
- **#1103** (\`2e39f7e\`): \`ensure_managed_schema()\` now catches the race and re-confirms the schema exists before swallowing the error.
- **This PR** (\`718f4eb\`): all three \`_ensure_table()\` bodies extracted into one shared \`_ensure_table_exists()\` helper using the identical catch-and-reconfirm pattern, since the same race existed independently at the table level for \`_drt_runs\`/\`_drt_history\`/\`_drt_dlq\`.

Both fixes have live regression tests with 8 concurrent first-time callers. Rebased this branch onto #1103's fix so both races are closed together.

The other two round-4/5 findings (no enforced per-project namespace; concurrent-cursor-backward under last-writer-wins) are unchanged from the earlier reply — checked against existing \`gcs\`/\`s3\` precedent and documented in the module docstring rather than fixed, since neither is novel to this backend.

Issue #1106, opened 2026-09-06
feat(state): Snowflake leg of the managed-table primitive + warehouse state backend (#960/#920)


masukai:
## Problem

#960 (`ManagedTableCapable` — schema/table create-if-absent primitive) and #920 (`state.backend: warehouse` — SQL-queryable state/history/DLQ) both shipped Postgres-only. Snowflake was explicitly deferred in both, in order of decreasing available prior art — this environment had no `DRT_SMOKE_SNOWFLAKE_*` credentials to live-verify against a real account, and this repo's established discipline (#908, #1064) is not to ship SQL-dialect-specific behavior backed only by mock-cursor tests.

This is filed as an immediate next-priority follow-up, not indefinitely deferred backlog — Snowflake is a materially more common drt deployment target than Postgres.

## Scope

**In:**
- `ManagedTableCapable` on the Snowflake source: `ensure_managed_schema`/`managed_table_exists`/`drop_managed_table`, same escape-hatch (pre-provisioned schema, no `CREATE` grant) and reversibility discipline as the Postgres implementation (`drt/sources/postgres.py`).
- `SnowflakeProfile.managed_schema` field (default `_drt`, matching Postgres's own default and the cross-dialect naming convention documented on `ManagedTableCapable`'s docstring).
- `PostgresWarehouseStateStore`/`HistoryStore`/`DlqBackend`-equivalent Snowflake implementations in `drt/state/warehouse.py`, dispatched from `drt/state/factory.py`'s `warehouse` branch alongside the existing Postgres dispatch.
- `_drt_runs`/`_drt_history`/`_drt_dlq` table DDL adapted to Snowflake's SQL dialect (e.g. `MERGE` instead of `INSERT ... ON CONFLICT` for the upsert-shaped `_drt_runs` write — Snowflake has no native `ON CONFLICT`).
- Live verification against a real Snowflake account (testcontainers has no Snowflake emulator — this needs `DRT_SMOKE_SNOWFLAKE_*` credentials via the existing `dwh_smoke` gating, see `tests/integration/dwh/conftest.py`).

**Out:** BigQuery/Databricks legs (tracked separately), any change to the Postgres implementation.

## Blocked on

`DRT_SMOKE_SNOWFLAKE_*` credentials being available to live-verify against — provisioning script exists at `tests/integration/dwh/provisioning/snowflake.sql`, needs a real Snowflake account set up and credentials wired into CI/this environment.

## Related

#960, #920, ADR 0005

[...]

PR #1109, opened 2026-09-06
Docs for warehouse-backed state (#920)

masukai:
## Summary

Third and final PR in the #920 stack (stacked on #1104, which is stacked on #1103).

- New [`docs/guides/warehouse-state.md`](https://github.com/drt-hub/drt/blob/main/docs/guides/warehouse-state.md): quick start, config reference, the three managed tables and their exact escape-hatch DDL/grants, concurrency behavior (including the two accepted cross-backend limitations already documented in the module docstring), and the watermark-independence note mirrored from `remote-state.md`.
- `docs/connectors/postgres.md`'s #960 section updated — "none shipped yet" was no longer true, links to the new guide, and documents the concurrent-first-use escape hatch fix live-verified this session.
- `docs/guides/remote-state.md`'s "What this isn't" bullet updated to link to the new guide instead of describing #920 as a hypothetical future tier.
- ADR 0005's Implementation ordering table gains a status column tracking steps 1–4's actual shipped/PR state, plus a note that Snowflake/BigQuery/Databricks (#1106/#1107/#1108) are tracked as immediate follow-ups blocked on credentials, not deferred backlog.

This completes the #920 stack.

## Test plan

- [x] All internal doc links point to real anchors/files


masukai:
**Codex review:**

Target: branch diff against `feat/920-warehouse-state-backend`

> The new guide contains an unsafe backend-switching instruction and omits a required installation dependency. It also leaves contradictory backend and implementation-status documentation in authoritative references.

- [P1] Warn before switching state backends without migration — `docs/guides/warehouse-state.md:123-126`
  When an incremental sync uses the `StateStore` cursor fallback (the default without explicit `sync.watermark.storage`), switching away from `warehouse` selects a separate store and does not copy `_drt_runs`; the next run can therefore replay records or use stale state. Qualify "at any time" and document migration or cursor reseeding before resuming, as the remote-state guide does.

- [P2] Add the required Postgres extra to the quick start — `docs/guides/warehouse-state.md:17-24`
  For users with a base `drt-core` installation, this quick start fails when the backend imports `psycopg2`, because that driver is only provided by the optional `postgres` dependency.

- [P2] Update the advertised state backend list — `docs/guides/remote-state.md:307-310`
  Its configuration table still says the only valid values are `local`, `gcs`, and `s3`, while `docs/llm/API_REFERENCE.md` repeats that stale list and omits `connection_profile`.

- [P2] Remove the obsolete factory status paragraph — `docs/adr/0005-state-location-and-write-grants.md:203-204`
  Lines 217–226 still said the factory half was open with a placeholder in `manager.py`, contradicting the table just above marking it shipped in v0.9.0.

---

All 4 verified as real and fixed in 9de6b56: qualified the backend-switch claim with the cursor-fallback caveat + reseed instructions, added the `pip install 'drt-core[postgres]'` step, corrected both stale backend lists, and rewrote the ADR paragraph to reflect the factory having shipped.

Issue #1110, opened 2026-09-06
feat(engine): mode: mirror consumes diff's removed_keys directly (#755 follow-up)


masukai:
## Context

#755 (diff-based incremental, shipped Postgres-first) computes an exact `removed_keys` stream as part of every warehouse-side snapshot diff — the primary keys of rows present in the previous snapshot but absent from this run's. Today that stream is only surfaced for observability (`SyncResult.diff_removed_keys`); nothing deletes those rows from the destination.

`mode: mirror` already has its own, independently-computed delete pass (`mirror.strategy: destination` scans the whole destination table; `mirror.strategy: tracked` reads a `_drt_synced_keys` state table, #686). Both work today and are unaffected by this issue — this is a *third*, more precise option for sources that already pay for the diff computation.

## Proposal

A new `mirror.strategy: diff` (alongside `destination`/`tracked`) that deletes exactly the keys `SnapshotDiffSource.extract_snapshot_diff()` classified as `removed`, instead of either scanning the destination or maintaining separate tracked-key state. Requires `sync.incremental_strategy: diff` — reject the config combination otherwise (cheapest, most precise mirror delete when both features are already in use; meaningless without the diff computation actually running).

## Why not shipped with #755

The existing mirror delete pass runs inside `destination.finalize_sync()` — a duck-typed hook implemented per-dialect (Postgres/MySQL/Snowflake/ClickHouse; Databricks has its own permanently-separate implementation, see #720's design notes). Wiring `removed_keys` through to that call site means either:

- extending `finalize_sync(sync.destination, sync.sync)`'s signature to optionally carry the removed-key list (touches every dialect's call site even if only one dialect consumes the new argument), or
- a new, narrower Protocol method that only diff+mirror destinations implement.

Either way the blast radius (every `mode: mirror`-supporting dialect, not just Postgres) is materially larger than #755's own scope, and deserves review on its own rather than riding along. `SnapshotDiffResult.removed_keys` was deliberately shaped as an iterator of full key dicts (not just a count) specifically so this follow-up only has to plumb the value through, not change what #755 already ships (see `drt/sources/base.py`'s `SnapshotDiffResult` docstring).

## Scope

- In: `mirror.strategy: diff` wiring for Postgres (matching #755's own Postgres-first scope).
- Out: other dialects until `mirror.strategy: diff` itself is verified — same phased rollout #960/#920 already established.

Related: #755, #686, #694, ADR 0005.

PR #1111, opened 2026-09-06
feat(engine): diff-based incremental extraction, Postgres-first (#755)

### LINKED ISSUE #755: feat(engine): diff-based incremental — warehouse-side snapshot diff, no cursor column required

masukai:
## Summary

Closes #755 (ADR 0005 step 5), Postgres-first — same phased-rollout pattern already established by #960/#920 in this stack. Stacked on #1103 (this PR's base branch) because it needs `ManagedTableCapable` (#960) at compile time, not just runtime; does **not** depend on #1104/#1109 (#920's warehouse state backend — a related but independent feature sharing only the #960 primitive).

New `sync.incremental_strategy: diff` (default stays `cursor`, zero behavior change for every existing config) for models with no reliable cursor column — a curated mart or aggregation with no `updated_at`, and structurally no way for cursor-based incremental to ever detect a deleted row.

## How it works

Each run snapshots the full model result into a `#960`-managed table (`_drt_snapshot_<sync_name>`) and diffs it against the previous run's committed snapshot via a server-side SQL JOIN on `destination.upsert_key`, classifying every row as **added** / **changed** / **removed**:

- `added` + `changed` chain into one flat iterator and feed the engine's existing upsert write path completely unchanged — no change to batching, transforms, or `destination.load()`.
- `removed` (key columns only) is exposed as `SyncResult.diff_removed_keys` for observability. **Nothing deletes those rows yet** — `mode: mirror`'s existing delete pass (`destination`/`tracked` strategy) is untouched and keeps working exactly as before. Wiring a destination to consume `removed_keys` directly is a separate follow-up, #1110 — it touches the per-dialect `finalize_sync()` hook across every mirror-supporting destination (Postgres/MySQL/Snowflake/ClickHouse, Databricks separately), a materially larger blast radius than this PR's Postgres-only source-side change. `SnapshotDiffResult.removed_keys` is shaped as an iterator of full key dicts (not just a count) specifically so that follow-up only has to plumb the value through, not change what ships here.

`incremental_strategy: diff` requires `mode: upsert` or `mode: mirror` (rejects `full`/`incremental`/`replace` — it replaces cursor filtering rather than composing with it) and rejects `cursor_field` (meaningless here). `sync.diff.hash_columns` (default `"all"`) selects which non-key columns determine "changed"; an explicit list is validated against the model's actual output columns **at run time** and raises loudly on a typo rather than silently narrowing the hash and hiding real changes as unchanged rows.

## Design decisions worth flagging in review

- **No `sync.diff.schema` or `sync.diff.state` config field.** The snapshot always lives in the source profile's own `managed_schema` (#960) — a second, independently-settable schema knob could disagree with it, same reasoning #920 already settled for its own state tables. There's also nowhere else for the snapshot to live today (a destination-side managed-table primitive doesn't exist), so a `state` field would have exactly one legal value. See the design-correction comments on #755.
- **NULL-safe row hash, not `ROW(...)::text`.** Postgres's `ROW(...)::text` cast has a documented ambiguity between `NULL` and `''` for some column types. The hash here is a `coalesce(col::text, sentinel) || separator || ...` concatenation instead — live-verified: a column transitioning from `NULL` to `''` is correctly classified as `changed` (`test_null_to_empty_string_transition_is_detected_as_changed`). A mock cursor cannot catch this class of bug (see `feedback_mocks_cannot_prove_sql_validity`); it needed the live Postgres smoke suite.
- **Commit-only-on-zero-failures.** The snapshot is promoted to be the next diff's baseline only after a run completes with `total_result.failed == 0` and is not a dry run. On partial failure the baseline is deliberately left stale, so the same rows are reclassified as added/changed again next run rather than risking a row that never reached the destination being silently treated as delivered. Same conservative posture as #920's and #955's reconcile-on-next-run fixes.
- **Table swap uses the established `_complete_swap` idiom**, not a bare `DROP; CREATE`: rename current→current_old, scratch→current (one transaction), then drop current_old in a separate transaction — matching `drt/destinations/postgres.py`'s existing `replace_strategy: swap` pattern exactly, so a failure dropping the old table never unwinds an already-completed swap.

## Testing

- `tests/unit/test_config.py::TestDiffIncrementalStrategy` — config validation (mode/cursor_field rejection, hash_columns defaults/validation).
- `tests/unit/test_engine_diff_incremental.py` — engine wiring: added+changed routed to the load path unchanged, `diff_removed_keys` surfaced correctly, commit called only on success/not-dry-run, `NotImplementedError`/`ValueError` fail-fast paths for unsupported sources / missing `upsert_key`.
- `tests/integration/local_sql/test_diff_incremental_smoke.py` (`local_sql_smoke`, live Postgres via testcontainers) — full round trip (first run / edit / no-op run), the NULL-vs-empty-string hash case, explicit `hash_columns` scoping, typo/missing-key error paths, and the uncommitted-run-rediffs-against-the-same-baseline guarantee the commit-gating design depends on.
- `ruff check` / `mypy` clean; full existing suite (3940 tests, excluding `dwh_smoke`) green with no regressions.

## Scope

Snowflake/BigQuery/Databricks deferred the same way #960/#920 deferred theirs, pending `DRT_SMOKE_*` credentials for live verification — will file follow-up issues once that's possible.


masukai:
**Claude self-review** (Codex's review plugin hit a reproducible, upstream-acknowledged hang bug on this branch — [openai/codex-plugin-cc#391](https://github.com/openai/codex-plugin-cc/issues/391) — after 6 attempts, so this pass ran without it, same discipline as this branch's original self-review).

Found and fixed 2 real defects, pushed in 64eb16c:

1. **P1 — `commit_snapshot_diff` could wedge every future run after a crash.** The two-commit rename/drop swap left `<current>_old` behind if a process died between the two commits. The *next* run's own rename then failed with `duplicate_table` forever, since nothing dropped the orphan first — live-reproduced (confirmed the exact `psycopg2.errors.DuplicateTable` by reverting the fix and re-running), then fixed with the same `DROP TABLE IF EXISTS` crash-recovery idiom `extract_snapshot_diff`'s own scratch-table rebuild already uses. New live test: `test_commit_recovers_from_a_crash_left_old_table`.

2. **P2 — `SyncResult.diff_removed_keys` bypassed `sync.mask`.** The main `added`/`changed` record batch is masked before reaching the destination/observers, but `diff_removed_keys` wasn't — a masked `upsert_key` column's raw value would leak through observers/`--output json`/persisting consumers. Fixed by masking removed keys where they're extracted. New unit test: `test_diff_removed_keys_are_masked_like_the_main_record_batch`.

Also rebased #1115 (`feat/1110-mirror-diff-strategy`) on top of this fix and removed its own now-redundant `apply_mask` call on the same list (it had independently found and fixed the exact same masking gap for its own consumption of `diff_removed_keys` — now applied once, upstream).

Full verification: `ruff check`/`ruff format --check`/`mypy drt` clean, `pytest tests/unit` (3810 passed), `pytest tests/integration/local_sql -m local_sql_smoke` (19 passed, includes the new crash-recovery test) — all green on both this branch and the rebased #1115.

PR #1115, opened 2026-09-06
feat(engine): mirror.strategy: diff — exact removed-key delete (#1110)

### LINKED ISSUE #1110: feat(engine): mode: mirror consumes diff's removed_keys directly (#755 follow-up)

masukai:
## Summary

Closes #1110, the follow-up flagged in #1111's own description: wiring `mode: mirror` to consume `SnapshotDiffResult.removed_keys` (#755) directly, instead of leaving it as observability-only. Stacked on #1111 (this PR's base branch) — needs `SyncOptions.incremental_strategy`/`diff` and the engine's `diff_removed_keys` computation to exist first.

## What this adds

A third `mirror.strategy` value, `diff`, alongside the existing `destination` (whole-destination-table anti-join) and `tracked` (state-table anti-join): deletes **exactly** the keys `incremental_strategy: diff`'s source-side snapshot diff classified as removed, computed once, source-side, with no extra destination-side scan or state table at all.

- `sync.mirror.strategy: diff` requires `sync.incremental_strategy: diff` (rejected otherwise — meaningless without it) and rejects `sync.mirror.scope` (the removed-key list is already exact, row-level output of the source-side diff; a scope restriction on top has nothing left to narrow).
- Dispatch happens in `BaseSqlDestination._finalize_mirror` **before** the existing `self._mirror_keys` empty-guard — a run that adds/changes nothing this cycle can still have rows to delete under this strategy, and vice versa, so the two checks are independent.

## Why this was almost free

`_build_mirror_delete(..., negate=False)` already exists — it's the exact "delete these specific keys" form `mirror.strategy: tracked` uses today (`negate=True` is the "delete keys NOT in this set" form `destination` uses). `mirror.strategy: diff` is really just `tracked`'s DELETE shape fed from a different key source, so this PR adds **zero new dialect-specific SQL** — every dialect that already implements `_build_mirror_delete` (Postgres, MySQL, Snowflake — all three via the shared `BaseSqlDestination._finalize_mirror` template from #719/#720) gets this strategy automatically. ClickHouse and Databricks each carry their own separate `finalize_sync`/mirror-delete implementation outside this shared template and are out of scope here, same split #720's design notes already established.

## The one real design decision: how removed_keys crosses from source to destination

`removed_keys` is computed source-side, during extraction (`drt/engine/sync.py`), but the mirror DELETE fires destination-side, inside `finalize_sync()` — a duck-typed hook (`getattr(destination, "finalize_sync", None)`, not a Protocol) called with a fixed `(config, sync_options)` signature. Two ways to bridge that:

1. Widen `finalize_sync()`'s signature to optionally carry the removed-key list — touches every dialect's implementation (Postgres/MySQL/Snowflake/ClickHouse/Databricks/file.py) even though only the SQL-family ones sharing `_finalize_mirror` would ever use it.
2. Smuggle it through `SyncOptions`, the object both sides already receive — exactly the pattern `_sync_name` (tracked mirror, #686) and `_query_tags` (#768) already use for engine-computed-at-runtime values that aren't YAML config.

Went with (2): new `SyncOptions._diff_removed_keys` private attribute, set once by the engine right after `extract_snapshot_diff()` runs, read via `getattr(sync_options, "_diff_removed_keys", None)` in the new `_finalize_mirror_diff` — so `finalize_sync()`'s signature, and every dialect that doesn't care about this feature, is completely untouched.

## Testing

- `tests/unit/test_config.py::TestMirrorDiffStrategy` — config validation (requires `incremental_strategy: diff`, rejects `scope`, doesn't affect other strategies).
- `tests/unit/test_sql_base.py` — `_finalize_mirror` dispatch (fires independent of `_mirror_keys`, no-op with no removed keys including the attribute being entirely absent, multi-column key ordering preserved).
- `tests/integration/local_sql/test_mirror_diff_strategy_smoke.py` (`local_sql_smoke`, live Postgres) — full `run_sync()` end to end against a real Postgres source **and** destination: first diff run (nothing to delete), a row deleted source-side, second run deletes exactly that row destination-side and leaves the others untouched. This is the test that actually exercises the private-attr smuggling across the source/destination boundary, not just each side in isolation.
- `ruff check`/`ruff format --check`/`mypy` clean; full suite (3949 tests, excluding `dwh_smoke`) green with no regressions.
- Bundled VS Code JSON schemas regenerated for `mirror.strategy`'s new enum value (`vscode-schema-drift` CI check).


masukai:
**Codex review:**

Target: branch diff against `feat/755-diff-incremental-postgres`

> The new strategy can silently perform the wrong deletion algorithm on supported destinations, and it mishandles transformed keys and dry-run previews. These defects can either delete unchanged rows or permanently retain stale rows.

- [P1] Reject diff mirror on destinations bypassing the base finalizer — `drt/config/sync_options.py:210`
  When a Postgres diff source targets ClickHouse or Databricks, this globally valid option reaches their overriding `_finalize_mirror()` implementations, which only recognize `tracked`. A run with any added/changed row therefore treats the delta keys as the complete source and executes the `destination`-strategy `NOT IN` delete, removing unchanged destination rows; a removal-only run skips deletion entirely.

- [P1] Transform removed keys before deleting destination rows — `drt/destinations/sql_base.py:270-271`
  The removed-key dictionaries contain source snapshot values, but destination keys are written after lookups, `computed_fields`, `field_mappings`, masking, and metadata injection. A valid `upsert_key: [email]` with `mask: {email: hash}` writes a hash, while this code deletes using the raw email; the stale hashed row remains and the snapshot is then committed, so it is never retried.

- [P2] Include removed keys in dry-run diff previews — `drt/engine/sync.py:644-650`
  With `dry_run=True` and `compute_diff=True`, the `compute_diff()` logic recognizes only `tracked` and `destination` mirror strategies. A `mirror.strategy: diff` preview reports no deleted rows even when the removed-key list is non-empty, contradicting what the real run will delete.

---

All 3 verified as real and fixed in adf8014: `check_mirror_supported()` gains a `supports_diff_strategy` flag (matching the existing `supports_tracked_scope` pattern) so ClickHouse/Databricks now fail fast at `load()` instead of silently reaching their incompatible finalizer — only Snowflake opts in, since it's the one dialect besides Postgres/MySQL that inherits the shared finalizer. Removed keys are now masked the same way the main record batch is, before being smuggled onto `SyncOptions._diff_removed_keys`. `compute_diff()` gained a `mirror.strategy: diff` branch that reads the already-computed removed-key list directly.

Issue #897, opened 2026-08-03
feat(destinations): idempotency keys for fire-and-forget destinations (retry-safe writes)


masukai:
## Problem

drt's write-path retry (#277, per-destination `retry:` override) assumes a failed call had no side effect and is therefore safe to resend. That's true for a connection refused or a timeout before the request left the client — but for a **fire-and-forget / non-transactional destination** (webhook, Slack/Discord message, GitHub Actions `workflow_dispatch`, an SMS via Twilio, a payment-adjacent HubSpot workflow trigger), a request can succeed *server-side* and still look like a failure to drt: the response times out, the connection drops after the server processed it, or a retryable 5xx is returned from a load balancer after the origin already committed. Retrying in that case doesn't recover a lost write — it duplicates a side effect that already happened once.

This is a different gap from #825 (reconciliation sync for fire-and-forget destinations): #825 detects and recovers rows that were **silently dropped** (never sent, or sent and lost). This issue is about rows that were **sent twice** because drt's own retry resent something the destination had already accepted — #825's own text names this as an assumption ("idempotency lives downstream... the destination's dedup absorbs it") rather than something drt provides.

Some destinations already dedup on their own — Mixpanel's `people_set`/`import_events` legs use a deterministic `$insert_id` (shipped in v0.7.8, PR #608), and Salesforce Bulk API's own upsert semantics are naturally idempotent on `upsert_key`. Most of the 19 HTTP-style destinations have no such mechanism, and drt doesn't provide one generically.

## Prior art

- Stripe's `Idempotency-Key` header — client-generated, deterministic per logical operation, server dedups on it for a rolling window.
- dlt / Airbyte destinations that support upsert-by-key get this for free from the destination's own semantics; the gap is specifically the **non-upsertable, fire-and-forget** class.

## Proposal

A destination-level opt-in, not a new sync mode:

```yaml
destination:
  type: slack
  idempotency_key: "{{ sync.name }}:{{ row.id }}:{{ run_id }}"   # Jinja, same templating drt already uses elsewhere
```

- Where the destination's API has a native idempotency mechanism (a header, a client-supplied dedup ID field), drt computes the key and passes it through — same shape as query tagging (#768) discovering and wiring a native mechanism per dialect rather than inventing one.
- Where there's no native mechanism, this is a no-op with a `drt validate` warning ("`idempotency_key` set on a destination with no native support — has no effect"), so configuring it never silently does nothing without a signal.
[...]

PR #1116, opened 2026-09-06
feat(state): warehouse-backed idempotency ledger, Postgres-first (#1099)

### LINKED ISSUE #1099: feat(state): warehouse-backed idempotency ledger for fire-and-forget destinations

masukai:
Closes #1099.

## Summary

Opt-in warehouse-backed idempotency ledger (`state.idempotency: true`, on top of `state.backend: warehouse`) so fire-and-forget destinations with no native dedup mechanism (Slack, webhooks, SMS, most REST-style connectors) get protection against `drt retry`/rerun-caused duplicate sends.

- New `_drt_idempotency(sync_name, idempotency_key, delivered_at)` table via #960's `ManagedTableCapable` primitive, `PostgresWarehouseIdempotencyLedger` in `drt/state/warehouse.py`.
- New `IdempotencyLedger` Protocol (`drt/state/idempotency.py`) as a fourth optional member of `StateBundle`.
- New `sync.idempotency_key` Jinja template field, defaulting to the destination's `upsert_key` when set.
- New `SyncResult.skipped_duplicate` (subset of `skipped`, same shape as #757's `skipped_no_match`).

## Two design corrections made before writing any code

Both posted on #1099 as a comment before implementation started:

1. **Claim-before-send → check-then-send-then-mark.** The issue's original proposal reserves the key before calling the destination. If the destination call then fails (timeout, 5xx), the key is already marked delivered — so every future retry of that exact record silently no-ops forever. Implemented instead as: read-only existence check before the send (filters the batch), write only after `destination.load()` reports success (row-level, via `RowError.batch_index` — a partial-batch failure marks only the successful rows).
2. **Default key template must not include `run_id`.** The issue's suggested default borrows #897's `{{ sync.name }}:{{ row[upsert_key] }}:{{ run_id }}` shape. A warehouse-persisted ledger exists specifically to catch a duplicate sent by a *different* `drt run` invocation — folding `run_id` into the default key would make every separate run's rows look new and never actually deduplicate anything. Shipped default is `{{ row[upsert_key] }}` (`sync_name` is its own ledger column, not folded into the key text).

## Other scope notes

- `sync.idempotency_key` lives on `sync:`, not `destination:` (unlike #897's own field) — this check runs entirely inside the engine and never touches the destination. Noted on the issue as the same reusable surface #897 can build its native-mechanism pass-through on when it lands.
- Staged destinations are out of scope: their "success" is determined at `finalize()`, not per `stage()` call, which doesn't fit this per-record check/mark shape.
- Retention is a hardcoded 7-day TTL for this first cut (pruned once per sync, mirroring `HistoryStore.prune`), not a second config knob — same scope-narrowing call made on #755.
- `mark_delivered` is a direct call in `engine/sync.py`'s batch loop, not routed through a new observer capability — confirmed against actual precedent: `history_manager.append`/`prune` are already plain direct calls in the outer `finally` block, documented as accepted exceptions to the `SyncObserver` I/O boundary.

## Testing

- `tests/unit/test_state_warehouse.py` — `PostgresWarehouseIdempotencyLedger` mock-based tests (undefined-table fallback, ensure-schema dispatch, best-effort mark failure).
- `tests/unit/test_state_factory.py` — `StateBundle.ledger` wiring, cache-key differentiation on `state.idempotency`.
- `tests/unit/test_config.py` — `sync.idempotency_key` / `state.idempotency` validation.
- `tests/unit/test_engine_idempotency.py` — the two correctness properties above, directly: `mark_delivered` never called before `already_delivered`, never called when `load()` raises, excludes rows `load()` reported as failed; key-template resolution (explicit / upsert_key default / unresolvable); `run_id` exclusion regression (same row, two different run_ids, second one is recognized as a duplicate); dry-run and staged-destination no-ops; prune timing.
- **`tests/integration/local_sql/test_idempotency_ledger_smoke.py`** — live Postgres via testcontainers, added after the first pass since a mock cursor can't prove `= ANY(%s)` binds a Python list correctly or that concurrent `mark_delivered` calls for the same key coalesce under `ON CONFLICT DO NOTHING` instead of raising: round-trip check-then-mark, a repeated mark is a genuine no-op against the real PRIMARY KEY, retention pruning keeps only recent rows, and two concurrency races (first-use table creation — the same #920/#960 class; concurrent marks of the identical key) both survive with no errors.
- Full `pytest tests/unit -m "not dwh_smoke"` (3829 passed), `ruff check`/`ruff format --check`, `mypy drt` all clean.


masukai:
Codex's review of the stacked #1117 (#1100) diff found 2 real correctness bugs that were actually already present here too, via the shared `_successful_indices` computation both PRs use:

1. A row skipped without a `RowError` (e.g. `match_policy`'s `skipped_no_match`, #757 — a bare counter with no per-row index) was being wrongly treated as delivered and marked in the ledger.
2. `mode: replace, replace_strategy: swap` reports batch success against a shadow table; the real cutover happens later in a separate `finalize_sync()` rename, so a failed rename could leave rows marked delivered that never reached the real target.

Both fixed in 635ab15, along with `drt build` never having been wired to pass `state_bundle.ledger` through (it builds its own `_RunContext`, separate from `drt run`'s). Full details on #1117's review comment.

PR #1117, opened 2026-09-07
feat(state): compliance audit trail, Postgres-first (#1100)

### LINKED ISSUE #1100: feat(state): compliance audit trail — queryable record-to-destination delivery log, with built-in retention/purge

masukai:
Closes #1100. Stacked on #1116 (#1099's idempotency ledger) — this PR's base branch is `feat/1099-idempotency-ledger-postgres`.

## Summary

Opt-in, **project-wide** `state.audit_trail: {enabled, retain_days, fields}` on top of `state.backend: warehouse` — a queryable "which record went where, and when" delivery log for GDPR/compliance requests.

- New `_drt_audit_log(sync_name, run_id, sync_run_id, record_key, logged_fields_json, destination_type, delivered_at)` append-only table via #960's `ManagedTableCapable` primitive, `PostgresComplianceAuditTrail` in `drt/state/warehouse.py`.
- New `ComplianceAuditTrail` Protocol (`drt/state/audit_trail.py`) as a new optional member of `StateBundle`.
- Differentiated from the closest prior art (Hightouch's `hightouch_audit.Changelog`, which per this issue's own competitive research leaves unbounded PII retention entirely to the customer): `retain_days` is **required** the moment `enabled: true` — no silent forever-retention default.

## Two design decisions settled before implementation

Both posted on #1100 as a comment before writing any code:

1. **Not the same thing as `drt.observability.audit.AuditLogger`.** That Protocol (ADR 0008/#299) covers two enterprise config/secret-access events (`config_changed`/`secret_accessed`) and has nothing to do with record delivery. Named `ComplianceAuditTrail` specifically to avoid the collision.
2. **Fields are captured after `sync.mask` runs, not before.** The issue cites `sync.mask`'s conservative-by-default posture as this feature's precedent — but capturing the pre-mask value would actually contradict it: the destination never received that value, so it isn't part of "what went where," and logging it anyway recreates the exact PII liability `sync.mask` exists to prevent, in a second table nobody designed to protect the same way. (Symmetrically: a sync with no `sync.mask` configured logs whatever raw values `fields` names, for the full retention window — an explicit operator choice, not an oversight, documented in `AuditTrailConfig`'s docstring.)

Also settled without a separate issue comment (small enough to just implement): `record_key` derives from the same `fields` list already required when enabled (`":"`-joined present values), rather than inventing a third per-record key-template surface after #897's `destination.idempotency_key` and #1099's `sync.idempotency_key`. A field absent from a given sync's schema (different syncs, one shared project-wide field list) is simply omitted from its logged JSON rather than raising.

## Engine wiring, and its one real interaction with #1099

Same shape as #1099's `mark_delivered`: a direct call in the batch loop after a successful `destination.load()`, gated on `RowError.batch_index` exclusion — now factored into a shared `_successful_indices(record_batch, result)` helper since both features need the identical "which of these did the destination confirm" computation (two callers, not a hypothetical one).

The one place the two features touch: a record #1099's idempotency ledger already filtered out as `skipped_duplicate` this run never reaches `destination.load()` at all, so it correctly never gets a new audit row either — that filter shrinks `record_batch` in place *before* this point runs, so no special-casing was needed, just confirming the audit write reads the post-filter batch.

Purge reuses #1099's own precedent: `ComplianceAuditTrail.prune(sync_name, retain_days)` called once per non-dry-run sync in `run_sync()`'s outer `finally` block, answering the issue's own open question about where purge is triggered from.

## Testing

- `tests/unit/test_state_warehouse.py` — `PostgresComplianceAuditTrail` mock-based tests.
- `tests/unit/test_state_factory.py` — `StateBundle.audit_trail` wiring, cache-key differentiation on the nested `audit_trail` config.
- `tests/unit/test_config.py` — `AuditTrailConfig` validation (required `retain_days`/`fields` when enabled, backend gating).
- `tests/unit/test_engine_audit_trail.py` — the mask-ordering correction directly (a masked field logs the masked value, not the raw one), missing-field omission, `record_key` join order, partial-batch success filtering, the #1099 interaction, staged/dry-run no-ops, prune timing, `run_id`/`sync_run_id` passthrough including the `None` case.
- `tests/unit/test_engine_observer.py` — audit log prune-failure routes through `observer.on_warning` without failing the sync.
- **`tests/integration/local_sql/test_audit_trail_smoke.py`** — live Postgres via testcontainers: JSONB round-trip through `executemany`, null `run_id`/`sync_run_id` handling, retention pruning, and the same first-use table-creation race #920/#960/#1099 already found and fixed for their own tables.
- Full `pytest tests/unit -m "not dwh_smoke"` (3864 passed), `ruff check`/`ruff format --check`, `mypy drt` all clean. Re-measured coverage on every new/touched module directly (not just codecov's patch summary, per this session's own established practice) — `drt/state/audit_trail.py` and `drt/config/base.py` 100%, `drt/engine/sync.py`'s new code 100%, `drt/state/warehouse.py` 96% with the live suite (remaining misses are the same "re-raise a non-UndefinedTable exception" pattern already unremarkable across every sibling store class in that file).


masukai:
**Codex review:**

Target: branch diff against `feat/1099-idempotency-ledger-postgres`

> The audit trail can record deliveries that did not occur, omit successful deliveries for common value types and execution paths, and retain stale audit data indefinitely. These defects undermine the feature's core compliance guarantees.

- **[P1] Correlate audit entries with actual successful rows** — `drt/engine/sync.py`
  When a destination skips rows without `RowError`s, such as HubSpot/Postgres `match_policy` skips, or returns early after an `on_error=fail` failure with trailing rows unattempted, those indices are absent from `row_errors`. This complement therefore classifies them as delivered and writes false compliance records; use explicit successful-row correlation or conservatively avoid auditing rows whose success cannot be mapped.

- **[P1] Defer swap-mode audit writes until finalization** — `drt/engine/sync.py`
  With `mode: replace` and `replace_strategy: swap`, SQL destinations report batch success after writing only to the shadow table, while the target is updated later by `finalize_sync()`. These audit rows are committed before that finalizer, so a failed rename/swap leaves records claiming delivery even though the target remained unchanged; buffer this path's entries until finalization succeeds.

- **[P1] Normalize warehouse values before JSON serialization** — `drt/state/warehouse.py`
  If any configured audit field contains a normal warehouse value such as `datetime`, `date`, `Decimal`, or `UUID`, this plain `json.dumps` raises `TypeError`. The surrounding best-effort handler then swallows the failure and drops the entire batch's audit records despite successful delivery; serialize using the project's JSON-safe conversion used elsewhere.

- **[P1] Wire audit policy through every delivery entry point** — `drt/engine/sync.py`
  Because these parameters silently default to disabled, supported execution paths not updated by this patch bypass an enabled project-wide policy: `drt build` constructs `_RunContext` without the audit fields, both Dagster execution paths call `run_sync()` without them, and ordinary `replay_dead_letters()` deliveries call the destination directly. Successful records from those paths therefore never reach `_drt_audit_log`; each production entry point needs the configured bundle capability wired in.

- **[P2] Purge audit rows for inactive sync names** — `drt/engine/sync.py`
  Pruning only the currently executing `sync.name` means rows for a renamed, removed, or permanently disabled sync are never considered again, even while other project syncs continue running. Those potentially sensitive rows can remain indefinitely despite their required `retain_days`; retention cleanup needs a backend-wide cutoff or another path that also covers inactive sync names.

---

Verified against the code. Findings 1–3 are real, confirmed bugs (the first also affects #1099's `mark_delivered` via the same shared helper, not just this PR's audit log) — fixing now. Finding 4: `drt build` is a same-package oversight and will be fixed; `drt retry`/dagster-drt are separate execution paths this PR doesn't reach — documenting as explicit known scope with a follow-up issue rather than expanding this PR further (dagster-drt trailing drt-core state features by one cycle is an existing, already-accepted pattern in this repo). Finding 5 (P2) matches `HistoryStore.prune`'s identical per-sync-name scoping, an existing accepted tradeoff — noting it rather than fixing now.

masukai:
All 4 P1 findings addressed:

1. Skipped-no-match false positives → fixed via `_successful_indices` failing closed on any unattributed skip (635ab15, shared with #1099 which had the identical bug independently — fixed there too and cross-posted).
2. Swap-mode timing → excluded `mode: replace, replace_strategy: swap` from both the ledger and the audit trail (635ab15 for #1099, 2864962 for #1100), same exclusion shape as the existing staged-destination gate.
3. JSON serialization → added `_json_safe_audit_value` normalization before values reach `AuditEntry.fields` (2864962).
4. Missing entry points → `drt build` fixed (its own separate `_RunContext`, both commits). `drt retry`/dagster-drt filed as #1118 rather than expanding this PR further — consistent with this repo's existing pattern of dagster-drt trailing drt-core state features by a release.

P2 (pruning inactive sync names) — left as-is, matches `HistoryStore.prune`'s identical existing scoping.

New regression tests for all of the above; full suite + live Postgres smoke tests green.

PR #1120, opened 2026-09-07
feat(state): Snowflake leg of the warehouse state/history/DLQ backend (#1106)

masukai:
## Summary

Second half of #1106 — ports #920's warehouse-backed `StateStore`/`HistoryStore`/`DlqBackend` to Snowflake, stacked on top of #1119 (this stack's #960 Snowflake `ManagedTableCapable` leg). New `drt/state/warehouse_snowflake.py` (its own module, mirroring how `drt/sources/postgres.py`/`drt/sources/snowflake.py` already split one file per dialect), dispatched from `factory.py` by profile type (`state.backend: warehouse` now supports both Postgres and Snowflake connection profiles).

## Dialect-specific translations from the Postgres implementation

1. **`MERGE` replaces `INSERT ... ON CONFLICT DO UPDATE`** — Snowflake has no `ON CONFLICT`. `_drt_runs` merges on `sync_name`, `_drt_dlq` merges on `id`, matching the MERGE shape `drt/destinations/snowflake.py`'s own mirror write already establishes for this connector.
2. **`errors`/`record` are `VARIANT`, not `JSONB`** — written via `INSERT ... SELECT ... PARSE_JSON(%s)` (Snowflake disallows function calls inside an `INSERT ... VALUES` literal list — same `PARSE_JSON` wrap site already established for JSON-category columns elsewhere in this connector). Read back defensively (`json.loads` if the connector hands back a string, used as-is if already parsed) since whether the connector auto-deserializes VARIANT has no prior art in this codebase to confirm ahead of a live test — the smoke test proves which branch actually fires.
3. **Explicit transactions only where more than one DML statement must land together.** This connector autocommits every statement individually by default (confirmed: no `.commit()` call anywhere in the existing Snowflake destination code), unlike psycopg2's one-transaction-per-connection default the Postgres implementation relies on. `DlqBackend.append` (N upserts + a prune `DELETE`), `.replace()` (`DELETE` + N inserts), and `.reconcile()` (N removes/updates) each wrap their connection with `conn.autocommit(False)` / explicit `commit()`-or-`rollback()`. `.replace()` specifically closes the **#955 failure class** — an unwrapped delete-then-insert would permanently erase the DLQ on a crash between the two. Live-verified by forcing a real mid-transaction failure: Snowflake enforces `NOT NULL` but — unlike Postgres — does **not** enforce `PRIMARY KEY`, so the Postgres smoke test's duplicate-id trick wouldn't reproduce a real constraint violation here; a deliberately-null `error_message` does.
4. **No `_is_undefined_table`-style exception classification.** Postgres catches `psycopg2.errors.UndefinedTable` to make a read against a never-written store return empty rather than raise. Snowflake's equivalent driver exception shape has no precedent in this codebase to build on with confidence, so every read here (`get_last_sync`/`get_all`/`read`/`depth`/`all_depths`) probes `managed_table_exists()` first instead — one extra round trip per read (not a hot path) in exchange for not guessing at unverified driver internals.
5. **No `conn.rollback()` inside the table-creation race guard** — same reasoning as #1119's `ensure_managed_schema()`: Snowflake DDL autocommits with no savepoints.

## Tests

- `tests/unit/test_state_warehouse_snowflake.py` — 33 mock-based tests, 99% coverage of the new module (directly measured, not just `codecov/patch`'s percentage — the one remaining miss is a trivial one-line `_connect` delegate, matching the same gap in the Postgres test suite).
- `tests/unit/test_state_factory.py` — updated dispatch tests (a Snowflake profile now builds a Snowflake bundle instead of raising; a still-unsupported dialect, e.g. MySQL, still raises `NotImplementedError` naming both supported dialects).
- `tests/integration/dwh/test_snowflake_warehouse_state_smoke.py` — 6 live tests (`dwh_smoke`), mirroring the Postgres smoke suite's coverage (upsert/read/reset round-trip, history append/read/prune, DLQ FIFO/reconcile/clear, the `.replace()` atomicity proof, the escape hatch, and the concurrent-first-write race). All skip cleanly without `DRT_SMOKE_SNOWFLAKE_*` credentials; the concurrent-race test additionally needs the `DRT_SMOKE_SNOWFLAKE_HAS_CREATE_SCHEMA_GRANT` opt-in from #1119.

## Validation

- `ruff check drt tests` / `ruff format --check drt tests` — clean.
- `mypy drt` — clean (204 files).
- `pytest tests/unit -m "not dwh_smoke"` — 3847 passed.
- `scripts/check_drift.sh` — 0 new drift.


masukai:
**Codex review:**

> The persistent Snowflake smoke suite is not repeatable and also contains a date-dependent assertion that will inevitably fail. The DLQ implementation additionally scales warehouse queries linearly with every queued record.

- **[P1] Fixed literal DLQ entry ids not repeatable across runs against the persistent smoke account** — the DLQ `MERGE`s globally on `id` and doesn't update `sync_name` on a match, so a leftover row from a prior run would silently update under its stale sync_name and go invisible to the current run's assertions. **Fixed**: per-run-unique ids + explicit `dlq.clear()` cleanup in every DLQ smoke test.
- **[P2] Hard-coded "recent" history timestamp would eventually age past `retention_days=30`** and get pruned alongside the deliberately-old 2020 entry. **Fixed**: derived from the test's own current UTC time.
- **[P2] Per-entry `MERGE`/`INSERT`/`UPDATE` instead of batched writes** — verified against `drt/state/warehouse.py`'s Postgres implementation: it has the identical one-statement-per-entry loop today, so this isn't a Snowflake-specific regression. Recorded as a documented follow-up covering both dialects (#1121) rather than fixed for one dialect alone, which would leave them with different write-volume characteristics for no principled reason.

Fixed in 1bca25b.

Issue #1108, opened 2026-09-06
feat(state): Databricks leg of the managed-table primitive + warehouse state backend (#960/#920)


masukai:
## Problem

#960 (`ManagedTableCapable`) and #920 (`state.backend: warehouse`) both shipped Postgres-only. Databricks was explicitly deferred in both — this environment had no `DRT_SMOKE_DATABRICKS_*` credentials to live-verify against.

Filed as an immediate next-priority follow-up, not indefinitely deferred backlog.

## Scope

**In:**
- `ManagedTableCapable` on the Databricks source: schema create-if-absent, table existence probe, drop — following Databricks' own existing tracked-mirror precedent (`_create_state_table` hook, `destinations/databricks.py`) for connection/cursor handling and Delta-specific DDL quirks (Databricks' native `?` paramstyle migration from #706/#707 applies here too).
- `DatabricksProfile.managed_schema` field.
- Warehouse-backed `_drt_runs`/`_drt_history`/`_drt_dlq` via Databricks' multi-row `VALUES` batching pattern (#734) for the upsert-shaped table, using `MERGE INTO` (Delta's native upsert).
- Live verification against a real Databricks workspace (`DRT_SMOKE_DATABRICKS_*`, provisioning script at `tests/integration/dwh/provisioning/databricks.sql`).

**Out:** Snowflake/BigQuery legs (tracked separately).

## Blocked on

`DRT_SMOKE_DATABRICKS_*` credentials — provisioning script exists, needs a real Databricks workspace set up and credentials wired into CI/this environment.

## Related

#960, #920, ADR 0005

masukai:
## Design decisions (before implementation)

Following the same design-first discipline as #1106 (Snowflake leg). Orientation research + an advisor review surfaced one materially harder constraint than Snowflake, plus a few dialect-specific choices worth recording before writing code.

### The hard constraint: Delta Lake has *no* multi-statement transactions at all

Snowflake's autocommit is a **driver default** — `conn.autocommit(False)` gets you a real explicit transaction, which is exactly how `warehouse_snowflake.py` closed the #955 failure class for `DlqBackend.replace()` (DELETE-then-INSERT, wrapped so a crash between the two can't permanently erase the queue).

Databricks/Delta has **no equivalent at all** — this is already documented in this codebase's own `_finalize_mirror_tracked` docstring as a platform limitation, not something a method can fix on its own. So the Snowflake fix doesn't port directly.

[...]

PR #1122, opened 2026-09-08
feat(sources): Databricks leg of the managed-table primitive (#1108)

masukai:
## Summary

- Databricks leg of #960 (ADR 0005 step 3), following the same pattern as the Snowflake leg (#1106): `DatabricksSource` implements `ManagedTableCapable` (`ensure_managed_schema`/`managed_table_exists`/`drop_managed_table`).
- New `DatabricksProfile.managed_schema` field (default `_drt`), living inside the existing `catalog` field (Unity Catalog's three-level namespace) rather than a new managed-catalog concept.
- `catalog` is now required (raises `ValueError`) for every `ManagedTableCapable` operation — the profile field itself stays optional for plain extraction queries.
- Design decisions posted to #1108 before implementation: https://github.com/drt-hub/drt/issues/1108#issuecomment-5577292996

## Design highlights

- **No case-folding needed**, unlike the Snowflake leg: Unity Catalog is case-preserving for unquoted identifiers, so probes compare plain strings directly (no `UPPER()`).
- Probes use `SHOW SCHEMAS`/`SHOW TABLES ... LIKE`, matching this connector's existing `_target_exists` convention exactly (not `information_schema`).
- No `conn.rollback()` in the race guard — Delta Lake has no multi-statement transactions or savepoints at all (stronger constraint than Snowflake's autocommit-by-default).
- `provisioning/databricks.sql` gains a `GRANT CREATE SCHEMA ON CATALOG`, needed only by the positive-create-path and concurrent-race smoke tests — gated behind `DRT_SMOKE_DATABRICKS_HAS_CREATE_SCHEMA_GRANT` so they skip cleanly until an operator grants it.

## Follow-up (this PR does not include it)

The Databricks warehouse state/history/DLQ backend (`drt/state/warehouse_databricks.py` + factory dispatch) lands as a second PR on top of this one, mirroring how #1106 shipped in two PRs.

## Test plan

- [x] `ruff check drt tests`
- [x] `mypy drt`
- [x] `pytest tests/unit/` (3994 passed)
- [ ] `dwh_smoke` — deferred to the nightly `dwh-smoke` workflow (no `DRT_SMOKE_DATABRICKS_*` credentials in this environment)


masukai:
**Codex review:** ran a review pass on this PR's own diff (against `origin/main`) and found one item:

> [P2] Run the new Databricks smoke suite in CI — `tests/integration/dwh/test_databricks_managed_table_smoke.py:34`
> The scheduled Databricks job still invokes only `tests/integration/dwh/test_databricks_smoke.py`, so pytest never collects this new module; the job also does not export `DRT_SMOKE_DATABRICKS_HAS_CREATE_SCHEMA_GRANT`, preventing the grant-gated cases from being enabled.

This is already fixed in #1123 (`.github/workflows/dwh-smoke.yml` now names this file explicitly and wires the opt-in env var), which is stacked on top of this PR. No separate fix needed here.

masukai:
**Codex review:** a follow-up pass (reviewing this PR's diff in isolation against `origin/main`) found:

> [P2] Add the new smoke suite to the nightly workflow — the scheduled Databricks job still runs only `test_databricks_smoke.py`, and doesn't expose `DRT_SMOKE_DATABRICKS_HAS_CREATE_SCHEMA_GRANT`.

Same as before — already fixed in #1123 (stacked on top of this PR), which wires this file into `.github/workflows/dwh-smoke.yml` and maps the opt-in env var.

> [P3] Remove the internal type suppression — `test_databricks_managed_table_smoke.py`'s `_profile()` helper used `# type: ignore[arg-type]` on a project-owned `DatabricksProfile` construction, which `AGENTS.md` reserves for external library issues.

Fixed directly in this PR (a414287) — restructured through a plain `dict.update()` instead.

Issue #1107, opened 2026-09-06
feat(state): BigQuery leg of the managed-table primitive + warehouse state backend (#960/#920)


masukai:
## Problem

#960 (`ManagedTableCapable`) and #920 (`state.backend: warehouse`) both shipped Postgres-only. BigQuery was explicitly deferred in both — this environment had no `DRT_SMOKE_BIGQUERY_*` credentials to live-verify against, and BigQuery additionally has **no existing mirror-mode precedent** to model the implementation on at all (unlike Snowflake/Databricks, which both already implement `sync.mode: mirror`'s tracked-state pattern this primitive follows).

Filed as an immediate next-priority follow-up, not indefinitely deferred backlog — BigQuery is drt's original, most-established source connector.

## Scope

**In:**
- `ManagedTableCapable` on the BigQuery source. BigQuery has real differences from Postgres/Snowflake worth resolving explicitly rather than copying the pattern blind: schemas are called "datasets", `CREATE SCHEMA IF NOT EXISTS` maps to `CREATE SCHEMA` on a dataset reference, and there is no session-level `information_schema.schemata` probe identical to Postgres's — BigQuery's own `INFORMATION_SCHEMA.SCHEMATA` is project-scoped and needs the right qualification.
- `BigQueryProfile.managed_schema` (or a BigQuery-appropriate name if "schema" reads wrong against "dataset" terminology — worth a naming check before implementing, since #960's docstring convention says every dialect uses `managed_schema` specifically to avoid clashing with an existing `schema`-meaning field, and BigQuery may not have that collision at all).
- Warehouse-backed `_drt_runs`/`_drt_history`/`_drt_dlq` via BigQuery's `MERGE` statement for the upsert-shaped table.
- Live verification against a real BigQuery project (`DRT_SMOKE_BIGQUERY_*`, provisioning script at `tests/integration/dwh/provisioning/bigquery.sh`).

**Out:** Snowflake/Databricks legs (tracked separately).

## Blocked on

`DRT_SMOKE_BIGQUERY_*` credentials — provisioning script exists, needs a real GCP project set up and credentials wired into CI/this environment.

## Related

#960, #920, ADR 0005

masukai:
[...]

PR #1124, opened 2026-09-08
feat(sources): ManagedTableCapable on BigQuery (#1107/#960)

masukai:
## Summary

- `ManagedTableCapable` on `BigQuerySource` (#1107, ADR 0005 step 3), following the same probe-before-create discipline as the Postgres (#920)/Snowflake (#1106)/Databricks (#1108) legs.
- New `BigQueryProfile.managed_schema` field (default `_drt`, lives inside `project`), per the shared naming convention documented on `ManagedTableCapable` itself.

## Design notes

Unlike every other dialect's implementation so far, this one is entirely `google-cloud-bigquery` **client-API** based (`get_dataset`/`create_dataset`/`get_table`/`delete_table`), not SQL DDL:

- Mirrors `destinations/bigquery.py`'s existing `test_connection()`, which deliberately uses `get_table()` instead of a query job specifically to avoid requiring the project-level `bigquery.jobs.create` permission just to check something exists.
- A single `client.get_table()` call already raises the same `NotFound` whether the *table* or the *dataset* is missing — unlike Databricks, there's no separate schema-existence probe needed before checking table existence.
- `ensure_managed_schema()` follows the same two-part discipline as the other dialects: probe first (escape hatch — never call `create_dataset` if the dataset already exists), and on any exception from `create_dataset`, re-probe rather than assuming failure (handles a concurrent first-use race, surfaced by BigQuery as `Conflict`).

## Scope

**In:** `ManagedTableCapable` (schema/table existence + create/drop) on the BigQuery source.

**Out (follow-up on #1107):** the `state.backend: warehouse` leg (`_drt_runs`/`_drt_history`/`_drt_dlq` via BigQuery `MERGE`). Orientation research and two open design questions were posted to #1107 before this PR: (1) BigQuery's dataset-create permission is project-level, not scoped like Snowflake/Databricks' `HAS_CREATE_SCHEMA_GRANT` pattern — broadening the smoke service account needs an explicit IAM decision; (2) this repo has zero bound-parameter (`QueryJobConfig(query_parameters=...)`) precedent for BigQuery anywhere, and state-table writes carry genuinely user-controlled strings (`error_message`, `json.dumps(record)`) — needs live verification once `DRT_SMOKE_BIGQUERY_*` credentials exist. This repo currently has no such credentials (this issue's own "Blocked on" section), so no live smoke test is included in this PR — mocked unit tests only, matching how the other three dialects' `ManagedTableCapable` legs shipped before their own smoke tests came online.

## Test plan

- [x] `pytest tests/unit/test_bigquery_source.py -q` (15 tests, including escape hatch, concurrent-create-race, and Protocol-conformance cases)
- [x] `ruff check` / `ruff format --check` clean
- [x] `mypy drt` clean
- [x] Full unit suite (4047 passed) green, no regressions


masukai:
Codex review: Round 1

Two P2 findings, both real:

1. **[Fixed]** `managed_table_exists()`/`drop_managed_table()` only checked `get_table()`'s `NotFound`, not `table_type` — a view/materialized view/snapshot/external table sharing the managed table's name was misreported as an existing managed table (a consumer would skip its own `CREATE TABLE` and fail later on writes), and `drop_managed_table()` could delete a resource this capability doesn't own. Fixed in a43c423: both now require `table.table_type == "TABLE"` before reporting existence or deleting. Added 3 new unit tests covering the view case for both methods plus the absent case.

2. **[Deferred, tracked]** `load_profile()`/`save_profile()` in `drt/config/credentials.py` don't read/write `managed_schema` for BigQuery — an operator configuring a non-default managed dataset would have it silently replaced with `_drt` on every load. Confirmed correct, but **deliberately not fixed in this PR**: `drt/config/credentials.py`'s `load_profile()`/`save_profile()` are being restructured wholesale from an if/elif chain to a per-type function-table pattern in #1097 (open, pending review). Wiring `managed_schema` into the current if/elif shape now would mean rewriting it again the moment #1097 merges — the same reasoning already applied to holding off Databricks/Snowflake's equivalent PRs' rebase. Will add BigQuery's `managed_schema` load/save handling directly in the new post-#1097 shape once that PR lands, alongside the equivalent fix for Databricks/Snowflake if needed. Tracking here so it isn't lost.

masukai:
Rebased onto main (now includes #1097's per-type loader/dumper refactor and #1122's Databricks leg). Also resolved the Round 1 Codex-review item deferred pending #1097: `_load_bigquery`/`_dump_bigquery` now read/write `managed_schema`, matching Postgres/Snowflake/Databricks. Full unit suite (4068 passed, 3 skipped), ruff, and mypy clean.

PR #1125, opened 2026-09-09
docs(state): add Databricks/Snowflake DDL to warehouse-state escape hatch

masukai:
## Summary

Follow-up from Codex review Round 4 on #1123 (Databricks warehouse-state PR), tracked as a known item rather than blocking that PR:

- `docs/guides/warehouse-state.md`'s escape-hatch SQL block was Postgres-only despite the guide's title/body covering three dialects — added dialect-specific DDL for Snowflake (`VARIANT`, uppercase-folding identifiers, `GRANT ... TO ROLE`) and Databricks (`VARIANT`, case-preserving identifiers, `USING DELTA`, `MODIFY` grant in place of a combined `INSERT`/`UPDATE`/`DELETE`).
- Documented a second, narrower Databricks concurrency case also flagged in that review: a concurrent `drt state reset` between `save_sync()`'s existence probe and its `UPDATE` makes the write silently affect zero rows — `save_sync()` returns as if it persisted state while the row (and cursor) is gone, rather than raising. Distinct from the already-documented duplicate-row race (this one is a silent no-op, not an extra row).

Also adds a CHANGELOG note for `BigQueryProfile.managed_schema` being wired into `load_profile()`/`save_profile()` — that code fix already landed with #1124 (deferred there pending #402's refactor), this just records it in the Unreleased notes.

## Test plan

- [x] Markdown code-fence balance sanity check
- [x] `ruff check` / `ruff format --check` / `mypy drt` clean (docs-only + CHANGELOG changes; no code touched by this PR)


masukai:
Codex review: Round 1

**[Fixed]** The Snowflake escape-hatch `GRANT SELECT, INSERT, UPDATE, DELETE ON <t1>, <t2>, <t3>` used Postgres-style comma-separated table list syntax, which Snowflake's `GRANT` doesn't accept (it takes exactly one `ON TABLE <name>` or the whole-schema `ON ALL TABLES IN SCHEMA` form) — an admin copy-pasting this block would hit a syntax error on the exact grant statement giving the sync role table access. Fixed in 9cc54ef using `ON ALL TABLES IN SCHEMA`, equivalent here since `_drt` holds only drt's own three managed tables.

PR #1126, opened 2026-09-10
feat(state): wire idempotency_ledger/audit_trail through retry + dagster-drt (#1118)

### LINKED ISSUE #1118: Wire #1099/#1100's idempotency ledger + audit trail through drt retry and dagster-drt
## Problem

#1099 (warehouse-backed idempotency ledger) and #1100 (compliance audit trail) hook into `engine/sync.py`'s batch loop, reached from `run_sync()`. Codex review on #1100 (PR #1117) found that two production delivery paths never reach `run_sync()` at all, so records delivered through them bypass both features entirely, even when a project has `state.idempotency: true` / `state.audit_trail.enabled: true`:

- **`drt retry`** (`drt/cli/commands/retry.py`'s `replay_dead_letters()`) calls `destination.load()` directly to replay dead-lettered records — it never calls `run_sync()`. This is arguably the most important gap for the compliance audit trail specifically: a retried record is exactly the kind of delivery an auditor would ask about ("you said this failed — did it eventually go out?").
- **dagster-drt** (`integrations/dagster-drt/dagster_drt/resource.py` and `assets.py`) is a separately-versioned package with its own two `run_sync()` call sites, neither passing `idempotency_ledger`/`audit_trail`/`audit_fields`/`audit_retain_days` through. This is the same shape of gap CHANGELOG history already documents for #756/#920/#762/#763 needing separate dagster-drt follow-up fixes after the drt-core feature ships (e.g. "`DagsterDrtResource.run()` and the legacy `drt_assets_legacy()` path ignored `state.backend`" fixed after #855's research, "`DagsterDrtResource.run()` never passed `history_manager=` to `run_sync()`" — same class of bug for `history_manager`, now presumably fixed but worth re-auditing for the newer `idempotency_ledger`/`audit_trail` params too).

## Scope

**In:** wire `idempotency_ledger`/`audit_trail`/`audit_fields`/`audit_retain_days` through `retry.py`'s replay path (needs its own check-then-mark / log_delivered logic since it doesn't share `run_sync()`'s batch loop or `_successful_indices` helper — retry's own `RowError.batch_index` is documented as chunk-local, a different correlation shape) and through dagster-drt's two `run_sync()` call sites (straightforward — same shape as #920/#762/#763's fixes for `history_manager`).

**Out:** any new feature — this is purely closing an integration gap the two features already have everywhere else.

## Related

- #1099, #1100 — the two features this completes wiring for.
- #920, #762, #763 — prior examples of the identical "dagster-drt didn't get the drt-core `run_sync()` parameter" gap.

## Milestone fit

Unscheduled backlog, gated on #1099/#1100 landing.

masukai:
## Summary

Closes #1118. `#1099`/`#1100` (idempotency ledger, compliance audit trail) hook into `run_sync()`'s batch loop — but two production delivery paths never reach `run_sync()` at all, so records delivered through them silently bypassed both features even with `state.idempotency: true` / `state.audit_trail.enabled: true`:

- **`drt retry`** (`replay_dead_letters()`) calls `destination.load()`/`stage()+finalize()` directly.
- **dagster-drt** has its own two `run_sync()` call sites (`resource.py`'s op/asset path, `assets.py`'s legacy path), neither passing the two params through — the same class of gap already found and fixed for `history_manager` (#980).

## Design (reviewed via advisor, posted to #1118 before implementation)

**Gating.** `run_sync()` excludes idempotency/audit for staged destinations and `mode: replace, replace_strategy: swap` (`sync.py:891-902`) — a per-`load()`-call "success" isn't a trustworthy delivery signal until `finalize()`/the shadow-table cutover. `replay_dead_letters()` mirrors the identical exclusion.

**A ledger hit does NOT delete the DLQ entry.** The ledger keys on record identity (`upsert_key`), not this exact payload — a hit means "some version of this record was delivered," not "this queued payload was." Deleting on that basis risks silently discarding a payload that has since drifted, the same blast radius #955 exists to prevent. Instead: skip resending, report under a new `skipped_duplicate` count, leave it queued.

**Marking scope.** Only entries retry's own correlation logic (`remove_ids`) confirms delivered get `mark_delivered()`/`log_delivered()` — pre-filtered duplicates never enter that computation, so no double-marking.

**`log_delivered(run_id=None, sync_run_id=None, ...)`.** Retry has no run identity of its own; attributing to the *original failing* run would misattribute the delivery for a compliance audit.

**Shared helpers moved out of `drt/engine/sync.py`.** `CLAUDE.md` names it the Rust-core migration candidate, kept pure — importing its internals into a CLI command is the wrong direction. `resolve_idempotency_key_template`/`compute_idempotency_key` → `drt/state/idempotency.py`; `json_safe_audit_value` → `drt/state/audit_trail.py`. Both now public. Pure refactor, no behavior change for `run_sync()`'s existing call sites.

## Test plan

- [x] New unit tests in `test_cli_retry.py`: successful mark+log, duplicate-skip-and-keep-queued, staged-destination exclusion, swap-mode exclusion, unaffected-when-disabled regression
- [x] New unit tests in dagster-drt's `test_assets.py`: both call sites (resource.py, legacy assets.py) pass the four new kwargs through
- [x] Full drt-core unit suite (4073 passed, 3 skipped) + dagster-drt suite (73 passed) green


masukai:
**Codex review: Round 1**

Codex reviewed this PR and found four issues. Triage and resolution below.

1. **[Real bug, fixed]** dagster-drt's `pyproject.toml` still declares `drt-core>=0.9.1`, which doesn't actually guarantee `StateBundle.ledger`/`.audit_trail` or `run_sync()`'s new `idempotency_ledger`/`audit_trail` kwargs exist — those are #1099/#1100, still `[Unreleased]` (no tagged drt-core contains them as of this PR, confirmed via `git log v0.10.0..main`). Under a floor-satisfying install, every dagster-drt sync would crash before executing, even with the features disabled.
   - Since no already-released drt-core version actually contains these APIs, there's no honest floor number to bump *to* yet. Added an explicit comment in `integrations/dagster-drt/pyproject.toml` documenting the gap and stating dagster-drt must not be published against the current floor until a drt-core release containing #1099/#1100 ships.

2. **[Real bug, fixed]** `retry.py`'s new ledger/audit marking reused `remove_ids` (the pre-existing "drop from DLQ" set) as its delivery signal. A `match_policy` skip (`skipped_no_match`, #757 — no `RowError`, no `batch_index`) would land in `remove_ids` via the existing "not explicitly failed" logic and then get wrongly marked delivered / audit-logged — exactly the false-positive class `_successful_indices()` was built to prevent for `run_sync()` itself (caught in Codex review on #1100).
   - Extracted `_successful_indices()` out of `drt/engine/sync.py` into a shared, public `successful_indices()` in `drt/state/idempotency.py` (both `run_sync()` and `retry.py` now import it — same relocation rationale as the other #1118 helper moves, keeping `engine/sync.py` free of anything `retry.py` would otherwise have to reach into). `retry.py` now tracks a `delivered_ids` set distinct from `remove_ids`, computed per retry chunk via `successful_indices()` — fails closed on any skip in that chunk. `remove_ids`/DLQ-removal semantics are unchanged; only the ledger mark / audit log write is now correctly withheld. New regression test: `test_retry_excludes_unattributed_skip_from_ledger_and_audit`.

3. **[Real, pre-existing gap — not fixed here]** For a non-staged, multi-chunk retry, a later chunk's `dest.load()` exception aborts the whole function before the post-loop marking/audit/reconcile block runs, losing earlier chunks' confirmed delivery (no DLQ removal, and now no ledger mark either) — so those records get resent on the next retry.
   - Confirmed this predates #1118: `store.reconcile()` was already a single post-loop call, so this exact class of loss already existed for DLQ removal before this PR. Fixing it properly means restructuring `retry.py`'s loop into per-chunk persistence (mirroring how `run_sync()` already persists idempotency/audit writes per batch, not deferred to sync end) — a real behavior change beyond this PR's stated scope ("wire two params through"), on a path with no existing exception-case test coverage. Filed as [#1127](https://github.com/drt-hub/drt/issues/1127), linked from #1118.

4. **[Held, re-affirmed]** `replay_dead_letters()` never calls `idempotency_ledger.prune()`/`audit_trail.prune()`, unlike every non-dry `run_sync()`.
   - This was an explicit pre-implementation design decision (posted on #1118, validated by advisor consultation before writing any code): `run_sync()` already prunes both on every non-dry run, so for any project that runs on a normal schedule, retention is maintained regardless of whether `drt retry` ever prunes. Threading `audit_retain_days` into `retry.py` just for a project that *only* ever calls `drt retry` and never `drt run` — an unusual usage pattern — was judged scope creep for this PR. Documented in CHANGELOG.md; not implemented.

Verification: `ruff check`, `mypy drt`, full `tests/unit` (4074 passed), and dagster-drt's own suite (73 passed) all green after the fix.

masukai:
**Codex review: Round 2**

Ran a second Codex pass on the Round 1 fix. It correctly pushed back on one thing and found one new, real issue.

1. **[Correction to Round 1]** Codex re-flagged the dagster-drt floor as P1, and it was right to: a comment in `pyproject.toml` doesn't constrain the dependency resolver — it does nothing to stop a real `pip install` of a future dagster-drt release against drt-core 0.9.1–0.10.0, which would still crash (`StateBundle` has no `ledger`/`.audit_trail`, `run_sync()` rejects the new keywords). Bumped the floor to `drt-core>=0.11.0`. This isn't a confirmed release number — no drt-core tag contains #1099/#1100 yet, and `ROADMAP.md` doesn't name a next version — but it's deliberately picked too high rather than too low: an over-high floor just fails to resolve at install time (loud, and dagster-drt isn't published against this code yet, so today's blast radius is zero), where the old floor let a real install through that crashes on the first sync.
   - This repo's own CI installs dagster-drt against an *editable local* drt-core checkout, whose `pyproject.toml` version field trails its actual floor until a real release is tagged (matches this repo's convention of not pre-bumping the version between releases). Raising the floor to `0.11.0` would otherwise break `ci.yml`'s `dagster-drt-test` job and `publish-dagster-drt.yml`'s `verify` job right now, since the local checkout reports itself as `0.10.0`. Fixed both by installing dagster-drt with `--no-deps` plus an explicit `dagster` install — trusting what's actually checked out rather than its version string. `publish-dagster-drt.yml`'s real `publish` job is deliberately **not** touched, so an actual tagged release still can't ship until drt-core's floor is genuinely satisfiable from PyPI. Verified the full install→lint→mypy→pytest sequence end-to-end in a scratch venv for both the old (broken) and new (fixed) install order.

2. **[Real, fixed]** `tests/unit/test_cli_retry.py`'s new fakes (`_FakeLedger`, `_FakeAuditTrail`, `_FakeDestination.load`, and my own `_FakeSkippingDestination.load` from Round 1's fix) used `# type: ignore[no-untyped-def]`, which `AGENTS.md` reserves for external-library issues only — these are project-owned test fakes. Replaced all of them with real Protocol-matching type annotations. Also fixed a stale docstring cross-reference in `drt/state/audit_trail.py` (`engine/sync.py`'s `_successful_indices` → `drt.state.idempotency`'s `successful_indices`, missed in Round 1's relocation).

3. **[Expected repeat, no new action]** Codex re-raised the per-chunk-persistence finding from Round 1. It reviews the diff, not this PR's comments, so it has no way to know [#1127](https://github.com/drt-hub/drt/issues/1127) already tracks this — not new information.

Also added a regression test for the fail-closed guard's *partial*-failure branch specifically (a chunk with one real failure + one unattributed skip + one genuine success — the genuine success's ledger/audit protection is correctly withheld too, same conservative trade-off as the all-succeed case Round 1's test covered).

Verification: `ruff check`, `mypy drt`, full `tests/unit` (4075 passed), dagster-drt's own suite (73 passed), and a scratch-venv simulation of both CI install sequences all green.

PR #1128, opened 2026-09-11
fix(retry): persist per-chunk ledger/audit confirmation immediately

### LINKED ISSUE #1127: drt retry: a mid-loop dest.load() exception loses earlier chunks' DLQ removal (and now ledger/audit) confirmation
## Context

Caught by Codex review on PR #1126 (#1118's idempotency ledger / audit trail wiring for `drt retry`), but the underlying gap in `replay_dead_letters()` (`drt/cli/commands/retry.py`) pre-dates #1118 — #1118's new ledger/audit code just inherits and compounds it.

For a non-staged destination, `retry_groups` is a generator that calls `dest.load()` lazily, once per chunk, as the `for retry_group, result in retry_groups:` loop iterates:

```python
retry_groups = (
    (chunk, dest.load([e.record for e in chunk], sync.destination, sync.sync))
    for chunk in _chunks(to_retry, sync.sync.batch_size)
)
```

`remove_ids`/`updates`/`delivered_ids` accumulate as local variables inside the loop. `store.reconcile(...)` (the only call that actually persists DLQ removal) — and, since #1118, the idempotency ledger's `mark_delivered()` / the audit trail's `log_delivered()` — all run only **after** this whole loop completes.

If a later chunk's `dest.load()` call raises an exception (a permitted batch-level error, not a `RowError`), the exception propagates out of `replay_dead_letters()` before any of that post-loop persistence runs. Earlier chunks that already succeeded:

- never get removed from the DLQ (`store.reconcile()` never runs) — so the next retry resends them, **and**
- never get `mark_delivered()`/`log_delivered()` calls — so the idempotency ledger has no record that they were ever sent, defeating the dedup protection it exists to provide for exactly this resend.

## Why this is pre-existing, not new in #1118

Before #1118, only DLQ removal was at stake (already a real bug: a crashed later chunk loses an earlier chunk's successful removal, matching the #955 failure class this same function's `reconcile()` docstring already discusses). #1118 didn't introduce the gap — it added a second, higher-stakes casualty (ledger/audit confirmation) to the same pre-existing exception path.

## Suggested direction (not decided here)

Persist confirmed deliveries (DLQ removal + ledger mark + audit log) incrementally, per chunk, rather than only once after the full loop — mirroring how `run_sync()`'s own per-batch loop (`drt/engine/sync.py`) already persists idempotency/audit writes per batch rather than deferring to sync end, specifically to keep the crash-loses-data window scoped to one batch. The exact mechanism (try/except around each chunk with per-chunk reconcile, vs. restructuring the loop) is an open design question — this issue is scoped to *raising* the gap, not prescribing the fix.

## Not fixed in #1126

Deliberately deferred out of #1126's scope (mechanical `idempotency_ledger`/`audit_trail` wiring) per discussion during that PR's review — flagged here for separate follow-up.

masukai:
## Summary

- Follow-up to #1118/#1127 (found during PR #1126's Codex review, deliberately deferred out of that PR's scope).
- For a non-staged, multi-chunk `drt retry`, a later chunk's `dest.load()` raising (the documented, uncaught `Destination.load()` contract) used to lose an earlier chunk's already-confirmed idempotency-ledger mark and audit-trail log entirely, since both were only persisted once after the whole retry_groups loop completed. The next `drt retry` would then resend the earlier chunk, duplicating whatever fire-and-forget action had already succeeded.
- Restructured `replay_dead_letters()` so each chunk's ledger mark + audit log happen immediately after that chunk's own result is processed, mirroring `run_sync()`'s own per-batch persistence (`drt/engine/sync.py`). No new try/except — the exception still propagates exactly as before; only what's already durable by the time it does has changed.
- **DLQ removal itself stays a single `reconcile()` call at the end, unchanged.** An earlier version of this PR also persisted DLQ removal per chunk, but Codex review correctly caught that as making retry cost quadratic — `reconcile()` re-reads/rewrites the entire remaining queue on every call, and `DlqBackend` has no cheaper mutation-only primitive. Since the ledger mark alone is what protects against a resend (the next retry's duplicate check catches it), deferring DLQ removal to one final call closes the duplicate-delivery risk for ledger-enabled projects without the quadratic cost.
- **Scope note:** this closes #1127 only for projects with `state.idempotency: true`. A project without it has no ledger-based backstop, so a crashed later chunk can still lose an earlier chunk's DLQ removal and cause a plain resend — that half of #1127 remains open (would need a new, cheaper mutation-only primitive on `DlqBackend`).
- Behavior change (documented in CHANGELOG): `log_delivered()`'s `delivered_at` is now per-chunk instead of one timestamp for the whole retry.

Design decision posted to #1127, with a follow-up comment correcting the original plan after Codex's Round 1 review on this PR.

## Test plan

- [x] `test_retry_persists_ledger_and_audit_for_earlier_chunk_before_a_later_chunk_raises`: two dead letters forced into separate chunks (`batch_size: 1`), a fake destination succeeds on the first `load()` call and raises on the second. Asserts chunk 1 is ledger-marked and audit-logged (both entries stay queued — DLQ removal is deferred), then proves the fix is load-bearing: a second `replay_dead_letters()` call against a fresh, non-raising destination skips chunk 1 as a duplicate and only actually sends chunk 2.
- [x] All pre-existing `drt retry` tests pass unchanged.
- [x] `ruff check`, `ruff format --check`, `mypy drt`, full `tests/unit` (4076 passed) all green.


masukai:
**Codex review: Round 1 → Round 2**

**Round 1** found two real problems with the original design (per-chunk `store.reconcile()` for DLQ removal):

1. **[P1] Quadratic reconcile cost** — `reconcile()` re-reads and rewrites the entire remaining queue on every call; calling it once per chunk makes a large, small-`batch_size` queue's retry cost O(n²) (a 10,000-record queue with `batch_size: 1` would do ~10,000 full reconciliations).
2. **[P2] Duplicate legacy-entry collision** — two byte-identical pre-#955 JSONL lines get the same SHA-256 identity (`decode_dead_letter_line()`); reconciling one chunk's removal could silently wipe out a different, not-yet-processed chunk's still-queued duplicate-ID entry.

Both fixed by reverting DLQ removal to a **single `reconcile()` call at the end** (exactly like before this PR), while keeping the idempotency-ledger mark and audit-trail log persisted **per chunk** — those are cheap, targeted key writes, not full-collection reads, so there's no quadratic cost, and the legacy-duplicate-ID bug is structurally impossible again (it only existed because `reconcile()` ran more than once against a mutating ID-keyed collection).

This means the fix now closes #1127 only for projects with `state.idempotency: true` — the ledger mark from an earlier chunk is what stops the *next* retry from resending it, even though the DLQ itself may still show it as queued after a crash. A project without the ledger enabled has no such backstop and keeps the original gap (documented in CHANGELOG.md and in a follow-up comment on #1127, which stays open for that remaining half). Rewrote the regression test to prove this is load-bearing: after a simulated crash, a second `replay_dead_letters()` call against a fresh destination must skip the ledger-marked chunk as a duplicate rather than resending it.

**Round 2** re-reviewed the corrected diff and found no further issues: "The patch correctly persists idempotency-ledger and audit-trail records after each confirmed chunk while retaining the existing single DLQ reconciliation. No functional regressions were identified."

One thing worth noting: Codex's own sandbox run of `pytest tests/unit/test_cli_retry.py` exited 1 during its Round 2 pass, but it didn't surface this as a finding. I re-ran the same file directly (25/25 passed) and confirmed the real CI run on this PR also passed `Test (pytest)` cleanly across Python 3.10–3.13 plus `dagster-drt-test` — so this looks like an environment/dependency setup gap in Codex's own ephemeral sandbox, not a real regression, but flagging it here rather than silently discounting it.

Verification: `ruff check`, `ruff format --check`, `mypy drt`, full `tests/unit` (4076 passed) all green; CI all green.

PR #1129, opened 2026-09-11
feat(state): batch Postgres DlqBackend writes into multi-row statements

masukai:
## Summary

- Part of #1121 (Postgres leg only) — raised in Codex review on #1120, confirmed as a shared Postgres+Snowflake issue, scope updated on #1121 before starting since the Databricks leg landed after the issue was filed and is now itself partially batched.
- `PostgresWarehouseDlqBackend.append()`/`.replace()` issued one `INSERT`/`INSERT ... ON CONFLICT` per dead letter; `.reconcile()`'s `updates` issued one `UPDATE` per changed entry. A sync producing hundreds/thousands of dead letters (or a `drt retry` reconciling that many) meant hundreds/thousands of warehouse round trips.
- Chunked into multi-row `INSERT ... VALUES (...), (...), ... ON CONFLICT (id) DO UPDATE SET ...` (`append`), a multi-row `INSERT ... VALUES` (`replace`), and a multi-row `UPDATE ... FROM (VALUES ...) AS v(...)` (`reconcile`'s `updates`), at a 2000-bound-parameter budget per statement — the same conservative constant `destinations/snowflake.py`'s `_MERGE_PARAM_BUDGET` already uses, kept file-local rather than imported across the `state`/`destinations` boundary.
- `append()` dedupes entries sharing an id within one call (last wins) — a multi-row `INSERT ... ON CONFLICT DO UPDATE` raises "command cannot affect row a second time" if two `VALUES` rows target the same id, a real (not hypothetical) case for legacy dead letters whose id is a content hash.
- No `DlqBackend` Protocol change — it's frozen per ADR 0007, and this is purely internal SQL construction.
- Out of scope, tracked as follow-ups on #1121: the Snowflake leg (identical per-entry pattern), a Databricks parity fix (`append()`/`reconcile()`'s `updates` aren't batched there either, though `replace()` already is), and the idempotency-ledger/audit-trail's `executemany()` calls (a related but hidden per-row-over-the-wire pattern).

## Test plan

- [x] New unit tests mirror `test_state_warehouse_databricks.py`'s existing chunking-test pattern (`test_append_chunks_many_entries_across_multiple_insert_statements`, `test_replace_chunks_many_entries_across_multiple_insert_statements`, `test_reconcile_chunks_many_updates_across_multiple_statements`, plus a dedup test and single-entry correctness tests).
- [x] Real-Postgres integration smoke (`tests/integration/local_sql/test_warehouse_state_backend_smoke.py`, testcontainers) passes 7/7 against a live Postgres, exercising the new multi-row `INSERT`/`UPDATE ... FROM (VALUES ...)` SQL end-to-end — a mock cursor can't validate the SQL itself (#908's lesson).
- [x] All pre-existing tests in both files pass unchanged.
- [x] `ruff check`, `ruff format --check`, `mypy drt`, full `tests/unit` (4082 passed) all green.


masukai:
**Codex review:** clean pass — "The batched statements preserve existing DLQ behavior while respecting the parameter budget." Unit tests, ruff, and mypy all passed in its sandbox; the real-Postgres integration smoke was skipped there because Docker wasn't available in that environment. I already ran it locally against a live testcontainers Postgres (7/7 passing, including the multi-row `INSERT`/`UPDATE ... FROM (VALUES ...)` SQL) before opening this PR — see the PR description's test plan.

CI green across all jobs.

PR #1130, opened 2026-09-12
feat(state): batch Databricks DlqBackend append()/reconcile() writes

masukai:
## Summary

- Part of #1121 (Databricks leg) — the Postgres leg landed in #1129; Snowflake remains as the last dialect on the same issue.
- `DatabricksWarehouseDlqBackend.append()` was a per-entry probe-then-`UPDATE`-or-`INSERT` loop (`SELECT 1 ... LIMIT` existence check, then the matching statement); `.reconcile()`'s `updates` issued one plain `UPDATE` per changed entry. `.replace()` already batched its own upsert into a chunked `MERGE INTO ... USING (VALUES ...) AS s(...)` when the Databricks leg shipped (#1108), but `append()`/`reconcile()`'s `updates` were explicitly deferred to #1121 rather than fixed then — this PR closes that gap.
- Extracted `.replace()`'s upsert-by-id `MERGE` into a shared `_upsert_dlq_entries()` helper, now used by both `append()` and `.replace()` — both need the identical write shape. `.reconcile()`'s `updates` gets a sibling, update-only `_update_dlq_entries()` helper (no `WHEN NOT MATCHED` branch, since these touch existing rows only).
- `append()` dedupes entries sharing an id within one call (keeping the last) — a multi-row `MERGE` raises if two `VALUES` rows target the same id, the same collision class already guarded against on the Postgres leg (#1129).
- Side effect, not the point of the change: switching `append()` to `MERGE` closes its old check-then-act race for two concurrent writers touching the same `sync_name`/`id` (both could pass the existence probe and both `INSERT`, producing a duplicate row).
- No `DlqBackend` Protocol change — frozen per ADR 0007, purely internal SQL construction.
- Out of scope, tracked as the remaining follow-up on #1121: the Snowflake leg (`warehouse_snowflake.py`) still has the fully per-entry pattern across `append()`/`replace()`/`reconcile()`'s `updates`.

## Test plan

- [x] Rewrote the two now-stale `append()` tests (they asserted the old probe-then-act SQL shape) and the `reconcile()`-updates test for the new `MERGE` shape.
- [x] New tests: `append()`'s `sync_name`-preserved-on-match precedent, id-dedup within one call, multi-chunk (40 entries → 2 `MERGE` statements); `reconcile()`'s `updates` multi-chunk (300 entries → 9 `MERGE` statements at 36 rows/chunk).
- [x] All pre-existing tests in the file pass unchanged (44/44 total).
- [x] `ruff check`, `ruff format --check`, `mypy drt`, full `tests/unit` (4085 passed) all green.
- [ ] Real-Databricks integration smoke (`tests/integration/dwh/test_databricks_warehouse_state_smoke.py`) — blocked on `DRT_SMOKE_DATABRICKS_*` credentials in this environment, same pre-existing blocker as before this change (8 tests skip, not fail).


masukai:
**Codex review:** clean pass — "The Databricks DLQ batching preserves existing append and reconcile semantics while respecting the connector's parameter limit. The changed unit tests pass, and Ruff and mypy report no issues." No findings.

CI green across all jobs.

PR #1133, opened 2026-09-12
feat(state): batch Snowflake DlqBackend writes, closing #1121

### LINKED ISSUE #1121: feat(state): batch DlqBackend per-entry writes instead of one MERGE/INSERT per dead letter

masukai:
## Summary

- Closes #1121 — the last of the three warehouse dialects. Postgres and Databricks landed in #1129/#1130.
- `SnowflakeWarehouseDlqBackend.append()`/`.replace()` issued one `MERGE`/`INSERT` per dead letter; `.reconcile()`'s `updates` issued one `UPDATE` per changed entry.
- `append()` and `replace()` now share one upsert-by-id `MERGE` helper (`_upsert_dlq_entries`) — `replace()` upserts via the same helper after its `DELETE` rather than a plain per-entry `INSERT`, since every id is guaranteed new immediately after that delete.
- `reconcile()`'s `updates` gets a sibling, update-only `MERGE` helper (`_update_dlq_entries`, no `WHEN NOT MATCHED` branch, since these touch existing rows only).
- `PARSE_JSON` still can't appear inside a `VALUES` literal list on this connector, so both helpers use the same generic-alias-then-outer-`SELECT` technique `destinations/snowflake.py`'s own mirror `MERGE` already established, via a file-local `_MERGE_PARAM_BUDGET`/`_rows_per_chunk` — mirroring that module's own constant, not importing the private destination-layer function across the `state`/`destinations` boundary.
- `append()` dedupes entries sharing an id within one call (keeping the last) — Snowflake's `MERGE` doesn't support updating the same target row twice in one statement, the same collision class already guarded against on the Postgres/Databricks legs.
- **Also fixes a related, Snowflake-only gap found while in this code:** `reconcile()`'s `remove_ids` `DELETE` is now chunked too — it previously built one `%s` placeholder per id with no bound on the statement's total parameter count, unlike Postgres's single `ANY(%s)` array param or Databricks' already-chunked equivalent.
- No `DlqBackend` Protocol change — frozen per ADR 0007, purely internal SQL construction.

## Test plan

- [x] Rewrote the three now-stale tests (`PARSE_JSON(%s)` on a direct placeholder → `PARSE_JSON(v2)`/`PARSE_JSON(v1)` on a VALUES-derived alias; per-entry `INSERT`/`UPDATE` → `MERGE`).
- [x] New tests: `append()`'s `sync_name`-preserved-on-match precedent, id-dedup within one call, multi-chunk (600 entries → 3 `MERGE` statements); `reconcile()`'s `updates` multi-chunk (600 → 3) and `remove_ids` multi-chunk (2500 → 2 `DELETE` statements).
- [x] All pre-existing tests in the file pass unchanged (38/38 total).
- [x] `ruff check`, `ruff format --check`, `mypy drt`, full `tests/unit` (4090 passed) all green.
- [ ] Real-Snowflake integration smoke — blocked on `DRT_SMOKE_SNOWFLAKE_*` credentials in this environment, unchanged by this PR (6 tests skip, not fail).


masukai:
Codex review: **[P2] Scope replacement merges to the requested queue** — fixed in 44ad62b.

The finding was correct: `replace()`'s reuse of `_upsert_dlq_entries()` (an id-matching `MERGE`, shared with `append()`) could silently corrupt a different sync's row. Since Snowflake enforces no id uniqueness at all (`id TEXT PRIMARY KEY` is purely informational), a replacement entry whose id already existed under a *different* sync_name (possible for legacy content-hash ids) would match that `MERGE` and update the wrong sync's payload while leaving its `sync_name` untouched — the requested queue would end up missing the entry, and the foreign queue's data would be silently overwritten.

Fix: `replace()` no longer shares `_upsert_dlq_entries()` with `append()`. It now uses a new, dedicated batched helper, `_insert_dlq_entries()` — a plain, chunked multi-row `INSERT`, not a `MERGE`. This matches `replace()`'s pre-batching per-entry behavior exactly: since every entry is genuinely new for *this* sync_name immediately after `replace()`'s own `DELETE`, a plain `INSERT` always creates a new row for the requested sync and structurally cannot match/update a pre-existing row belonging to another sync, regardless of any id collision.

Added a regression test (`test_replace_does_not_touch_another_syncs_row_sharing_an_id`) asserting `replace()`'s SQL is a plain `INSERT` with no `MERGE`/`WHEN MATCHED`/id-match clause, plus updated the pre-existing `test_replace_wraps_delete_and_inserts_in_one_transaction` to assert the same. `append()`'s own `_upsert_dlq_entries()` is unaffected — its "match globally on id, never touch sync_name" behavior is deliberate and pre-existing (see that function's docstring), not part of this bug.

Full unit suite (4222 tests) + ruff + mypy all green after the fix.

Issue #1134, opened 2026-09-12
bug(destinations): several non-SQL destinations also derive columns/headers from records[0] alone


masukai:
## Summary

While fixing #1091 (`BaseSqlDestination.load()`/ClickHouse deriving a batch's write columns from `records[0]` alone, silently dropping a field that first appears in a later record), the same root-cause pattern was found independently in several other destinations that don't share that code path:

- `drt/destinations/bigquery.py:121` — `columns = list(records[0].keys())`
- `drt/destinations/file.py:100` — `columns = tuple(records[0].keys())`
- `drt/destinations/google_sheets.py:62` — `headers = list(records[0].keys())`
- `drt/destinations/staged_upload.py:177` — `csv.DictWriter(buf, fieldnames=self._records[0].keys())`
- `drt/destinations/salesforce_bulk.py:90` — `fieldnames = list(self._records[0].keys())`
- `drt/destinations/_blob_serializer.py:72` — `columns = list(records[0].keys())`

Deliberately kept out of #1091's scope (same move as splitting #1121 into per-dialect PRs) — each of these has its own write mechanics and needs its own fix, not a shared one.

## Two distinct failure modes

- **BigQuery and `_blob_serializer`** likely share #1091's silent-drop failure mode (need to confirm per file how the column list is actually used downstream).
- **The three `csv.DictWriter`-based ones** (`file.py`, `staged_upload.py`, `salesforce_bulk.py`) fail differently: `DictWriter`'s default `extrasaction='raise'` means a record with a key *not* in `fieldnames` raises `ValueError` outright, rather than silently dropping data. Still a real bug (a heterogeneous batch shouldn't fail the whole write), but the fix needs to decide between the same union/group-by-signature approaches #1091 used, or `extrasaction='ignore'` (which would silently drop the extra field again — probably not the right choice) plus explicit backfill of missing fields via `restval`.
- **`google_sheets.py`** needs its own analysis: a spreadsheet's columns are positional, so widening `headers` to the cross-record union has different implications (existing rows already written under a narrower header set) than a SQL `INSERT`.

## Suggested approach

Handle one destination (or one failure-mode group) at a time, following #1091's design-first pattern: confirm the actual downstream column usage before picking union vs. grouping vs. something destination-specific, and land as separate PRs.

masukai:
Investigated all six sites before starting. The failure-mode grouping in the issue description needs two corrections, and one of the six needs a materially different plan than "fix it like #1091":

[...]

PR #1135, opened 2026-09-12
fix(destinations): stop deriving a batch's write columns from records[0] alone

### LINKED ISSUE #1091: bug(destinations): BaseSqlDestination.load() derives batch columns from records[0] alone, silently dropping fields that first appear later in the same batch
## Summary

`BaseSqlDestination.load()` (`drt/destinations/sql_base.py:87`) builds each
write batch's column list from `records[0].keys()` only:

```python
columns = list(records[0].keys())
```

If a later record in the *same batch* has a field the first record doesn't
(a legitimately heterogeneous/optional-field source — REST API sources in
particular, but any source that doesn't guarantee uniform keys per row),
that field is silently never written for the whole batch, not just for that
one row. This is a live data-loss path in the real write, not just a
preview-accuracy issue.

## Where this was found

Surfaced during Codex adversarial review of PR #1090 (#1064's dry-run diff
case-reconciliation fix). An earlier version of that PR made the dry-run
diff preview fetch destination columns as the *union* of every record's
keys across the whole run, on the theory that this was more complete. Codex
review correctly flagged that the union diverges from what the real write
actually does — `records[0].keys()` per batch — so the preview could report
a phantom "changed" (or worse, silently miss a real update) on a field the
real write never touches. PR #1090 was reverted to intentionally mirror
`records[0].keys()` on the keyed-fetch diff path specifically *because* that
matches current write behavior — see the comment left at
`drt/engine/diff.py` around the keyed-fetch `columns` assignment.

That revert makes the diff preview honest about *today's* write behavior,
but doesn't fix the write behavior itself, which has the same root
limitation.

## Suggested fix directions (not decided)

- Normalize each batch to the union of its own records' keys before
  building `columns` (filling absent keys with `None`/appropriate default)
  so a batch always writes a rectangular column set.
- Or: validate/require homogeneous keys per batch and fail loudly
  (`RowError`) rather than silently dropping a field — safer default if
  silent partial writes are considered worse than a hard stop.
- Whichever direction is chosen, `drt/engine/diff.py`'s dry-run preview
  should be updated to match the corrected write semantics.

## Scope note

This predates PR #1090 and is not something that PR introduces — it's an
existing property of `BaseSqlDestination.load()` for every dialect built on
that base class (Postgres, MySQL, ClickHouse, Snowflake, Databricks).

masukai:
## Summary

- `BaseSqlDestination.load()` (and ClickHouse's own `load()`) derived each batch's write column list from `records[0].keys()` only — a field absent from the first record but present in a later one (a legitimately heterogeneous/optional-field source, e.g. REST API sources) was silently dropped for the **entire** batch, not just for the records lacking it.
- The obvious-looking fix ("widen `columns` to the batch-wide union") was considered and **rejected**: on an upsert-style write (`ON CONFLICT DO UPDATE` / `MERGE`), every column in the statement gets written for every record — a record lacking a field would write `NULL` via the existing `record.get(c)` convention, silently clobbering a pre-existing destination value. Worse than the original bug for an enrichment-style write.
- Fix: `replace()`/`replace_swap()` (no pre-existing row to clobber) and ClickHouse's always-append `insert()` now use the cross-record union directly. The upsert path instead groups a batch's records by their exact key signature and dispatches one `_load_upsert()` call per group — no dialect's `_load_upsert` signature changes, and a homogeneous batch (the common case) is exactly one group, dispatching exactly as before this fix.
- Also fixes the same `records[0]`-only pattern in `_validate_mirror_scope` (and its Snowflake/Databricks/ClickHouse overrides) — a `mirror.scope` column first appearing in a later record used to raise a spurious "columns missing" error.
- Updates `drt/engine/diff.py`'s `--dry-run --diff` keyed-fetch preview to match the corrected write semantics.

Design decision (including the rejected alternative and its NULL-clobber consequence) posted to #1091 before implementation. Out-of-scope destinations sharing the same root-cause pattern (`bigquery.py`, `file.py`, `google_sheets.py`, `staged_upload.py`, `salesforce_bulk.py`, `_blob_serializer.py`) filed as #1134.

Closes #1091

## Test plan

- [x] New unit tests per dialect (Postgres/MySQL/Snowflake/Databricks/ClickHouse) proving a late-appearing field lands without null-clobbering the group that lacks it
- [x] Orchestration-level tests (`test_sql_base.py`) for key-signature grouping, `SyncResult`/`RowError.batch_index` remapping, and `on_error: fail` stop-after-failing-group semantics
- [x] Mirror-scope regression tests (Snowflake/Databricks/ClickHouse) proving a late-appearing scope column no longer raises
- [x] New `local_sql_smoke` test (real Postgres via testcontainers) proving the end-to-end round trip: a late-appearing field lands, and a pre-existing destination value in a column the other group's records don't send is never clobbered — a mock cursor can't validate the SQL itself (#908 lesson)
- [x] Full unit suite (4278 passed, 4 skipped) + `ruff check` + `ruff format --check` + `mypy` all green


masukai:
Codex review: all three P1 findings were real and fixed in e6b9176. Full write-up of the design correction is on #1091 and in the updated CHANGELOG entry; summary here:

1. **Reordering** — the original design grouped records by key signature *globally* (dict-keyed), which could dispatch writes to a repeated upsert key out of their original relative order. Fixed: `BaseSqlDestination._contiguous_signature_runs()` now splits only at a signature *change*, preserving original order even when the same key reappears later under a different signature.

2. **Broken atomicity** — dispatching one `_load_upsert`/`_load_replace` call per group meant each call committed independently on Postgres/MySQL, so an earlier group's work could already be permanent by the time a later group failed under `on_error: fail`. Fixed: each dialect's write hook is now called **exactly once per `load()`**, same as before #1091 — the transaction boundary, commit/rollback behavior, and `on_error` semantics are completely untouched. Per-run column variance is pushed down into each dialect's own per-record (Postgres/MySQL) or per-chunk (Snowflake/Databricks) statement construction instead.

3. **NULL overriding column DEFAULTs** — the `_union_columns` "safe for replace()" reasoning only considered UPDATE-clobber risk and missed that an explicit `NULL` bind also overrides a column's `DEFAULT` on a plain `INSERT`. Fixed: `replace()`/`replace_swap()` now get the same per-run treatment as the upsert path (via the same `_contiguous_signature_runs()`), not the union. Databricks' `MERGE` path builds its staging-table `INSERT` per run too (staging shares the target's schema/`DEFAULT`s), while the single final `MERGE` still projects the full column union, since staging is by then correctly populated either way.

While investigating the atomicity question, found and separately filed (not fixed here, confirmed unmodified by this PR's diff) a pre-existing Postgres bug where `on_error: skip` can silently roll back an earlier successful row in the same batch when a later row fails — #1136.

New tests: `_contiguous_signature_runs()` unit tests for the run-splitting/ordering behavior; per-dialect unit tests proving a late-appearing field lands without null-clobbering or overriding a DEFAULT; and three new real-Postgres `local_sql_smoke` tests specifically targeting the three findings above (DEFAULT applies instead of an explicit NULL, `on_error: fail` still rolls back every run in the batch, a repeated upsert key across signatures still resolves to its original last-write value).

Full suite (4280 tests, including `local_sql_smoke`) + ruff + mypy all green.

masukai:
Codex review (round 2): all three findings fixed in d80a79d.

1. **[P1] Databricks MERGE still clobbering omitted fields** — real bug. Staging was per-run but the final MERGE still ran once over the batch-wide column union, so `WHEN MATCHED THEN UPDATE SET note = source.note` fired for every matched row regardless of which run it came from — a run whose records never sent `note` would overwrite an existing destination row's `note` with staging's `DEFAULT`-filled value. Fixed: both the staging INSERT *and* the MERGE now run once per contiguous key-signature run, so a run's own `update_cols` only ever mentions columns that run's records actually sent.

2. **[P2] `diff.py` phantom diff** — real bug. `DiffResult.changed_fields()` compared `set(old) | set(new)`; since the dry-run preview fetches `old` using the batch-wide union, a record lacking a field another record has would report a phantom `field: value → None`. Fixed: compares only `new`'s own keys now, matching what the corrected per-run write actually touches.

3. **[P2] Mirror-scope validation still permits per-record omission** — real bug, and it's a stricter fix than my first pass: requiring a scope column to appear *anywhere* in the batch (the union) still let a record genuinely missing it through, and that record's scope would record as `None`, which a delete predicate's `IN (...)` can't represent correctly. Fixed: every record must now have every scope column, not just "some record" or "the union."

Updated tests for all three (per-dialect MERGE-clobber checks, a `diff.py` phantom-diff regression test, and mirror-scope tests proving a per-record — not just first-record or any-record — omission rejects). Full suite (4242 unit + 40 `local_sql_smoke`) + ruff + mypy all green.

masukai:
Codex review (round 3): the P2 finding was real, fixed in c054e33.

The previous round's fix for the phantom-diff bug (restricting `DiffResult.changed_fields()` to only compare the new record's own keys) was correct for upsert/mirror but wrong for `sync.mode: replace`: a replace-mode write rebuilds each row from the new record alone, so a column the destination has that the new record omits genuinely resets to its `DEFAULT`/`NULL` — that *is* a real change, and hiding it from `--dry-run --diff` would make the preview disagree with the actual write.

Fixed: `changed_fields()` now takes `include_removed: bool = False`. `compute_diff()` passes `True` when `sync_options.mode == "replace"`; the two `drt/cli/output.py` renderers (which only see the already-computed `DiffResult`, not `sync_options`) pass `True` when `diff.delete_reason == "replace"` — the field is set to exactly `"replace"` for this case already.

New tests: `compute_diff` end-to-end (replace mode surfaces an omitted column as a real update), plus renderer-level tests for both `print_diff_table` and `diff_to_dict` showing the column is included for replace and excluded for upsert/mirror. Full suite (4247 unit + 40 `local_sql_smoke`) + ruff + mypy all green.

masukai:
Codex review (round 4): both P2 findings were real, fixed in a376ea7.

1. **Renderer keyed off the wrong signal** — real bug. `diff.delete_reason == "replace"` is unconditionally suppressed to `None` whenever nothing is actually deleted, which is exactly the common case here (same key set, nothing to delete, but a column still needs to show as changed) — so both `drt/cli/output.py` renderers were silently hiding the exact change round 3 was supposed to surface. Fixed: `DiffResult` now carries an explicit `writes_full_row: bool` field, set once in `compute_diff()` and read directly by both renderers, decoupled from delete provenance entirely.

2. **Gating only on `sync.mode == "replace"`** — also real. Snowflake/Databricks *default* to `destination.mode: insert` (a plain INSERT with no upsert-by-key semantics — only `mode: merge`, or `sync.mode: mirror` which forces merge internally, does a genuine partial UPDATE), and ClickHouse has no upsert-by-key write at all. Both reset an omitted column the same way `replace` mode does, not just leave it alone. New `_writes_full_row(config, sync_options)` helper covers all three cases.

New tests: unit tests for `_writes_full_row` across Postgres/ClickHouse/Snowflake (default-insert vs explicit-merge vs mirror-forces-merge), `compute_diff`-level tests proving the signal survives when nothing is deleted and correctly distinguishes insert-only from merge-mode Snowflake configs. Full suite (4255 unit + 40 `local_sql_smoke`) + ruff + mypy all green.

masukai:
Codex review (round 5): all three findings addressed in 0d03083.

1. **[P1] Missing upsert_key column** — real bug, and a good catch: the whole `_contiguous_signature_runs()` design partitions purely by which keys a record has, with no notion that some keys are structurally special. A record missing a configured `upsert_key` column would build a run whose write statement omits that key — on Postgres/MySQL, `ON CONFLICT (id)` referencing a column absent from the `INSERT` lets the table's identity/default fill in an *unintended new id* instead of matching the intended row (silently creating an extra row — the pre-#1091 code bound an explicit `NULL` and failed loudly instead). In mirror mode, the same gap could record a `(None,)` key that poisons the `NOT IN` delete predicate. Fixed with `_validate_upsert_keys_present()`, mirroring the same "every record, not just some" precedent `_validate_mirror_scope` already established.

2. **[P2] Append-only writes misclassified** — also real, and it's a correction of my own round-4 fix: an append-only write (ClickHouse always, Snowflake/Databricks' default `mode: insert`) never touches an existing same-key row at all — it inserts a new physical row alongside it, or a uniqueness constraint rejects it. The destination's existing value is untouched, not reset, so reporting it as `Updated: field: value → None` was wrong. Reverted `_writes_full_row` to `sync.mode: replace` only — the one case #1091 is actually about. Properly modeling the append-only case (e.g. as a new row rather than a same-key update) is a separate, larger question than #1091 set out to fix.

3. **[P3] Internal `type: ignore`** — correct, AGENTS.md forbids these for non-external issues. Removed all six: the smoke test now uses `unittest.mock.patch.object` instead of direct instance-attribute assignment (matching this codebase's own established mocking convention), and the two dict-payload tests use `typing.cast` instead.

New tests for all three. Full suite (4259 unit + 40 `local_sql_smoke`) + ruff + mypy (on `drt/`, matching this repo's `make lint` scope) all green.

masukai:
Codex review (round 6): both P2 findings were real, fixed in cd030b5.

1. **Databricks MERGE atomicity gap** — a genuine regression from round 2's own fix: running the MERGE per contiguous key-signature run (needed to fix the clobber bug) meant each run's MERGE autocommits immediately (Databricks has no multi-statement transaction here), so under `on_error: fail`, an earlier run's MERGE could already be committed to the target by the time a later run's staging insert failed — the original single-staging-then-one-MERGE design never allowed this partial application. Fixed by splitting into two passes: stage every run first (each into its own uniquely-suffixed staging table so one run's MERGE can't re-match another's already-merged rows), and only MERGE once all staging has succeeded. Restores the original all-or-nothing guarantee while keeping each run's own scoped MERGE for the earlier clobber fix.

2. **Replace-mode default value** — also real. `changed_fields()`'s round-3 fix represented a removed column's new value as a plain `None`, but the real reset target is the destination's own `DEFAULT` clause, which `compute_diff()` has no schema access to know and which isn't necessarily `NULL`. Reporting `None` was wrong for a non-null default, and — the sharper case — an `old` value that already happened to be `NULL` was incorrectly reported as *unchanged*, even though the write still resets it. Fixed with a `RESET_TO_DESTINATION_DEFAULT` sentinel (renders as `<default>`), unconditionally reported for a removed column rather than compared against `old`'s value (there's nothing real to compare against).

New tests: a Databricks unit test proving `on_error: fail` on a later run's staging failure stops before *any* MERGE reaches the target, and diff tests proving both the sentinel's value and its rendering (`<default>`, not `None`) — including updating a pre-existing test whose old expectation (`old=NULL` → "unchanged") was itself an instance of this exact bug. Full suite (4260 unit + 40 `local_sql_smoke`) + ruff + mypy all green.

PR #1146, opened 2026-09-13
fix(cli): drt retry reconciles DLQ removal before re-raising a mid-loop exception (#1127)

masukai:
## Summary

Closes the remaining half of #1127 (the idempotency-ledger half was closed separately by #1128).

`replay_dead_letters()` calls `store.reconcile()` exactly once, at the very end of the function, to remove/update DLQ entries for everything processed across all chunks. If a later chunk's `dest.load()` raised (the documented, uncaught `Destination.load()` contract), that final `reconcile()` call was skipped entirely — losing DLQ-removal confirmation for every earlier chunk that had already been successfully delivered in that same `drt retry` invocation. For a project without `state.idempotency: true` enabled, this meant a plain resend of an already-delivered record on the next retry.

An earlier draft of this fix (see the design comment on #1127) assumed a genuinely cheaper mutation-only `DlqBackend` primitive was needed across Local/GCS/S3, on the theory that `reconcile()` is too expensive to call more than once. Re-reading all five `DlqBackend` implementations found that assumption wrong: the three warehouse-backed ones (Postgres/Snowflake/Databricks) already do targeted, cheap `DELETE ... WHERE id = ANY(...)` / chunked `UPDATE ... FROM (VALUES ...)` mutations — not a full read-then-rewrite — so the "expensive reconcile()" concern only ever applied to `LocalDlqStore`/`ObjectStoreDlqBackend`, and even for those, calling `reconcile()` **one additional time only on the exception path** doesn't reintroduce the quadratic-cost problem #1128 fixed (still O(1) calls per invocation either way).

## Changes

- Wrapped the per-chunk retry loop in `try`/`except Exception`. On an exception, `reconcile()` is called once more with whatever `remove_ids`/`updates` had already accumulated from fully-processed earlier chunks, before re-raising the original exception.
- The chunk that raised gets no `updates` entry and no `attempts` bump — it was never attempted, unlike a chunk that was sent and failed again.
- The exception-path `reconcile()` call is itself wrapped in a nested `try`/`except` that logs a warning and lets the original exception propagate, so a DLQ-backend hiccup during this cleanup (e.g. `ObjectStoreDlqBackend` exhausting its write-precondition retries) never masks the real destination failure that aborted the retry.
- Updated `test_retry_persists_ledger_and_audit_for_earlier_chunk_before_a_later_chunk_raises` — its assertions described the pre-fix behavior (both entries left queued); now id=1 is confirmed gone from the DLQ, matching the new reconcile-on-exception behavior.
- New test `test_retry_persists_dlq_removal_for_earlier_chunk_before_a_later_chunk_raises_without_ledger` exercises the fix specifically **without** `state.idempotency` enabled, proving it independent of #1128's ledger backstop.

## Scope

This closes the *exception* path only. A hard process kill (SIGKILL/OOM) between an earlier chunk's success and this handler running still loses that window — the same residual gap `run_sync()`'s own per-batch persistence has between batches.

## Test plan

- [x] `pytest tests/unit/test_cli_retry.py tests/unit/test_dlq_store.py` — 53 passed
- [x] Full `pytest -m "not dwh_smoke and not local_sql_smoke and not object_store_smoke"` — 4276 passed, 4 skipped
- [x] `ruff check` / `mypy drt/cli/commands/retry.py` clean


masukai:
### Codex review: legacy-duplicate-id collision in the exception-path reconcile()

Codex's review flagged (P2, on `retry.py:405-406`): if a pre-#955 legacy DLQ holds two byte-identical entries sharing the same content-derived SHA-256 id, and one is confirmed by an earlier chunk while its twin sits untouched in a later chunk that never runs (because that chunk's `dest.load()` is what raised), the exception-path `reconcile()` added by this PR removes **both** physical entries — not just the confirmed one.

This is correct, but it's a **pre-existing** hazard, not one this PR introduces. `LocalDlqStore.reconcile()` (`drt/state/dlq.py:276`) filters by id-set membership against a fresh read:

```python
result = [updates.get(entry.id, entry) for entry in current if entry.id not in remove_ids]
```

This removes every physical entry whose id is in `remove_ids`, regardless of how many share that id. The exact same shape is already reachable on `main` today via `drt retry --limit`: if one of two identical legacy duplicates is retried (and succeeds) while its twin is left in `untouched` beyond the limit, the single end-of-function `reconcile()` call already removes both. This PR's exception-path `reconcile()` call is a second trigger for that same pre-existing mechanism, not a new one.

Fixing it properly needs either a legacy-id migration (give pre-#955 entries real per-line identity) or making `reconcile()` occurrence-aware (remove at most N physical matches per named id) — neither belongs in a PR specifically about persisting DLQ-removal confirmation before a mid-loop exception re-raises. Filed as #1147, scoped to `reconcile()`/`decode_dead_letter_line()`, noting it affects both the pre-existing `--limit` path and this PR's new exception path.

Also tightened the except-block's own comment (066422d) — it previously claimed a not-yet-attempted duplicate "can never be named in remove_ids," which is exactly what Codex's finding disproves. Now it correctly describes the id-set-vs-occurrence-count mechanism and cites #1147 instead of asserting the risk away.

No code change to the reconcile-on-exception logic itself in this PR — the finding doesn't change in kind from the risk this fix already inherits from the pre-existing single end-of-function reconcile() call.

masukai:
### Update: Codex's default-path point was right — fixed in 62c79cd

On reflection (and after re-reading Codex's round-2 wording more carefully), my previous reply undersold the finding. I'd argued this was purely a pre-existing hazard already reachable via `--limit`, so no code change was needed here. That's incomplete: **without `--limit`, the pre-fix code never reached this hazard at all** — every entry ends up in `to_retry`, every duplicate copy gets processed, and `untouched` is empty. This PR's exception-path `reconcile()` call is what newly makes it reachable on a plain, default `drt retry <sync>` — purely as a function of chunk ordering and destination failure timing, not an explicit opt-in flag. That's a materially worse exposure than what already shipped, and this fix exists specifically to prevent DLQ data loss, so shipping a narrower silent-delete vector inside it would be the wrong trade.

Fixed in 62c79cd: the loop now tracks `processed_count` (how many of `to_retry`'s entries have a fully-processed chunk behind them, incremented only after a chunk's result is fully handled — not merely fetched, so a chunk whose own processing raises partway through isn't counted). The exception handler computes `unprocessed_ids = {e.id for e in to_retry[processed_count:]}` and reconciles with `remove_ids - unprocessed_ids` instead of the raw `remove_ids` — an id with a still-unprocessed occurrence is simply left queued rather than removed, costing one more retry cycle rather than silently losing a record.

New test `test_retry_exception_path_does_not_delete_an_untouched_legacy_duplicate` writes two byte-identical legacy JSONL lines directly (no `id` field), confirms they really do decode to the same SHA-256 fallback id before relying on that, and verifies both entries survive the exception path. Confirmed this test fails (deleting both entries) against the pre-guard commit, so it's actually load-bearing.

**Left deliberately unfixed, retitled/rescoped on #1147:** the *original*, pre-existing `--limit` manifestation (the always-run end-of-function `reconcile()` call has no equivalent guard against `--limit`'s `untouched` list) and a narrower `updates`-cross-contamination case (a legacy id's untouched twin can pick up the same bumped `attempts`/content without being deleted). Neither is touched by this PR's diff, and fixing them needs the same broader design decision (legacy-id migration or an occurrence-aware `reconcile()`) that doesn't belong in this fix.

masukai:
### Update: round 3 found a real gap in the guard itself — fixed in 601d09c

Round 3 correctly caught that the round-2 guard was incomplete: it only excluded ids from `to_retry`'s own unprocessed suffix (`to_retry[processed_count:]`), but missed `untouched` — entries `--limit` excludes from `to_retry` entirely. Repro: legacy entries `[A, B, A2]` (A/A2 byte-identical, same content-derived id; B distinct), `--limit 2` + `batch_size: 1`. `to_retry = [A, B]`, `untouched = [A2]`. A succeeds, B raises. The round-2 guard's `unprocessed_ids` only contained B's id (the unprocessed suffix of `to_retry`), so A's id stayed in `remove_ids` and the exception-path `reconcile()` would still delete both physical A/A2 entries — exactly the failure mode the guard exists to prevent, just reached through `--limit` instead of a plain multi-chunk exception.

Fixed in 601d09c: `unprocessed_ids` is now `{e.id for e in to_retry[processed_count:]} | {e.id for e in untouched}` — both sources of "not confirmed this call" are excluded from the exception-path `remove_ids` before reconciling.

New test `test_retry_exception_path_does_not_delete_an_untouched_legacy_duplicate_beyond_limit` reproduces the exact `[A, B, A2]` + `--limit 2` scenario above and confirms it fails against the round-2-only guard (deletes both A entries, leaving only B) before the fix.

To be precise about what's now covered vs. still open: this exception-path guard (the code this PR adds) now correctly accounts for both ways an entry can go untouched this invocation. The **pre-existing**, always-run end-of-function `reconcile()` call — reached on a normal, no-exception completion — still has no equivalent `untouched` guard; that's unchanged behavior outside this PR's diff and stays tracked on #1147 along with the narrower `updates` cross-contamination case.

masukai:
### Codex review round 4: remove_ids beats updates for a shared id — pre-existing, deferred to #1147

Codex flagged: if byte-identical pre-#955 duplicates share a content-derived id, and one is confirmed (added to \`remove_ids\`) while its twin fails again in the same invocation (added to \`updates\`), the exception-path \`reconcile()\` call deletes both — including the one that should have stayed queued with a bumped \`attempts\`. Root cause, \`drt/state/dlq.py:276\`:

\`\`\`python
result = [
    updates.get(entry.id, entry) for entry in current if entry.id not in remove_ids
]
\`\`\`

An id in \`remove_ids\` is dropped by the \`if\` clause before \`updates.get(...)\` is ever reached — \`remove_ids\` unconditionally wins.

Unlike rounds 2 and 3, this is **not a new exposure from this PR's except block**. Applying round 2's discriminating test (does this code path reach something the pre-fix code couldn't?): here it doesn't. Two byte-identical legacy duplicates landing in the *same* chunk under the default \`batch_size: 100\` is sufficient — one succeeds (\`remove_ids\`), the other row-fails (\`updates\`) — and the single, always-run, unconditional \`reconcile()\` call at the end of a completely normal \`drt retry\` invocation (no \`--limit\`, no exception, nothing from this PR) already deletes both today on \`main\`. This actually needs *fewer* preconditions than either of the previous two findings.

Fixing it correctly means changing what both \`reconcile()\` call sites pass — including the always-run final call, which is already-shipped behavior outside this PR's diff. Patching only the exception-path call would make the same queue behave inconsistently depending on whether an unrelated later chunk happened to raise, which is worse than the current, at least consistently-broken behavior. Added as a third manifestation on #1147 (https://github.com/drt-hub/drt/issues/1147#issuecomment-5653975955) — actually the most severe of the three, since it needs no \`--limit\` and no exception at all, so it should probably drive whatever fix #1147 eventually lands.

No code change in this PR for this finding.

Issue #277, opened 2026-04-13
feat: retry policy configuration in sync YAML


masukai:
## Summary

Allow retry behavior to be configured per-sync in YAML, instead of relying on global defaults.

## Scope

- Add \`retry:\` block to sync YAML schema
- Fields: max_retries, backoff_strategy (linear/exponential), initial_delay_ms, max_delay_ms
- Per-destination override (some APIs need different retry strategies)
- Default values that work for most cases

## Example

```yaml
sync:
  name: hubspot_contacts
  retry:
    max_retries: 3
    backoff: exponential
    initial_delay_ms: 1000
```

PR #1150, opened 2026-09-14
feat(config): destination.native_idempotency_key config surface (#897)

masukai:
## Summary

First step of #897 (idempotency keys for fire-and-forget destinations): the config field and a `drt validate` warning, with no destination wiring yet. Full design comment on the issue: https://github.com/drt-hub/drt/issues/897#issuecomment-5671248093

drt's per-destination retry (#277) assumes a failed call had no side effect — true for a connection refused before the request left the client, false for a fire-and-forget HTTP destination where the request can succeed server-side and still look like a failure client-side (timeout, dropped connection, a retryable 5xx after the origin already committed). Retrying then duplicates the side effect instead of recovering a lost write.

## Survey (answering the issue's own open question)

Surveyed every HTTP-style destination against its code and official API docs before adding anything:

- **Already safe** (confirmed in code, not inferred from the CHANGELOG): `hubspot`, `klaviyo` (event leg), `meta_conversions`, `mixpanel`, `salesforce_bulk`, `zendesk`.
- **Safe only when an optional key field is configured** (a different, lower-priority gap, not folded into this issue): `airtable`, `elasticsearch`.
- **No native mechanism on the endpoint drt actually calls**, confirmed against official docs: `slack`, `discord`, `teams`, `twilio` (its `I-Twilio-Idempotency-Token` exists but is documented only for newer resources, not the classic Messages endpoint this destination uses), `sendgrid`, `email_smtp`, `github_actions`, `notion`, `jira`, `linear`.
- **Real, not-yet-wired gap**: `google_ads` — `ClickConversion` has a documented `order_id` dedup field the payload never sets.
- **Highest-value target**: `rest_api.py` — the generic escape hatch for any Stripe-style `Idempotency-Key` API.

## The `rest_api.py` design problem (why this needed its own PR first)

`body_mode: batch` sends one HTTP request per chunk of records, not one per record — the issue's own proposed default template (`{{ sync_name }}:{{ row[upsert_key] }}`) has no single row to bind to there. A batch-mode key needs a genuine per-*run* component or it's actively harmful: a key that's stable across every run's first chunk would make the destination's own dedup silently swallow every day's new data after the first.

Added `SyncOptions._sync_run_id` — a third instance of the existing `_sync_name`/`_query_tags` engine-smuggling pattern, but set **unconditionally** in `run_sync()` rather than gated on an opt-in feature. Reusing `_query_tags`'s embedded run_id was rejected: it's `None` whenever `query_tagging.enabled: false`, which would make batch-mode retry-safety silently depend on an unrelated, independently-configurable feature staying on.

## Naming

Deliberately `native_idempotency_key`, not `idempotency_key` — that name is already taken by `sync.idempotency_key` (the existing warehouse-backed cross-invocation dedup ledger from #1099/#1100/#1118). Same word, different scope, different mechanism (drt's own ledger vs. a destination's native HTTP-level dedup) — reusing the name would have been a real footgun in the same YAML file.

## `drt validate` warning

Warns (not an error, deliberately excluded from `--strict`'s warning-promotion — a missed retry-safety opportunity is a different severity than the hardcoded-secret scan `--strict` already promotes) when `native_idempotency_key` is set on a destination type with no wiring. Implemented via a `_NATIVE_IDEMPOTENCY_WIRED_TYPES` allowlist that starts **empty** (nothing is wired at this PR) and gains an entry in the same change that actually wires each destination — kept exactly "types where the field does something," never aspirational.

## Scope

No behavior change to any existing sync. `rest_api.py`'s wiring (record mode + the batch-mode design above) and `google_ads.py`'s `order_id` wiring (gated on confirming re-upload semantics from Google's docs first — a wrong answer risks silently dropping real conversion data) are tracked as separate follow-up PRs, matching the #1134/#1140/#1141/#1142 one-PR-per-destination pattern.

## Test plan

- [x] New tests: `tests/unit/test_cli_validate.py` (warning fires/doesn't fire, `--strict` doesn't promote it, JSON output), `tests/unit/test_engine.py` (`_sync_run_id` smuggled and matches `result.sync_run_id`, survives `query_tagging.enabled: false`, differs across separate `run_sync()` calls on a reused `SyncConfig`)
- [x] Full `pytest -m "not dwh_smoke and not local_sql_smoke and not object_store_smoke"` — 4295 passed, 4 skipped
- [x] `ruff check`, `ruff format --check`, `mypy drt` clean


masukai:
**Codex review response** (commit 58b0226)

Fixed 3 of the 4 findings:

1. **Stale generated schemas** — regenerated `integrations/vscode-drt/schemas/sync.schema.json` via `./scripts/regenerate-schemas.sh` and noted the new fields in the extension's own `CHANGELOG.md`. Confirmed a second regen run now produces a clean diff (no drift).

3. **MCP validate parity** — `drt/mcp/tools/validate.py` now calls the same `_find_ineffective_native_idempotency_keys` the CLI uses, surfaced as a new `idempotency_warnings` key on `drt_validate`'s response, explicitly excluded from `strict` promotion (same severity stance as the CLI). Added coverage in `tests/unit/test_mcp.py`.

4. **Plugin extensibility** — replaced the hardcoded `_NATIVE_IDEMPOTENCY_WIRED_TYPES: frozenset[str]` with a new `NativeIdempotencyCapable` capability Protocol in `drt/destinations/base.py`, following the exact precedent `ModeCapable`/`MatchPolicyCapable`/`QueryableDestination` already set for "can this destination do X" questions — a plugin destination (#297 entry points) can now implement `supports_native_idempotency_key()` and correctly opt out of the warning without touching core code. Construction failure (missing credentials, unresolved env var) is treated as "can't determine, don't warn" rather than crashing `drt validate`, mirroring `_run_connection_test`'s own try/except shape. Added unit coverage for the capable, partially-capable, and construction-failure cases.

**Finding 2 (concurrent-reuse race on `_sync_run_id`) — not fixed here, and I don't think it should be.** I checked: `sync.sync._query_tags = query_tags` at `drt/engine/sync.py:740` mutates the same shared `SyncOptions` object the same way, so a caller invoking `run_sync()` concurrently on one reused `SyncConfig` already has this exposure today, independent of this PR. More importantly, nothing in this diff reads `_sync_run_id` yet — no destination is wired to consume it, so there is no reachable consequence from this PR as it stands. I'll revisit this properly in the `rest_api.py` wiring PR, where a real consumer appears and "concurrent `run_sync()` on one reused `SyncConfig`" becomes a decision worth making deliberately (document as unsupported, or pass an invocation-local copy instead of mutating shared state).

Full suite + ruff + mypy clean.

masukai:
**Codex review response, round 2** (commit bcfd24e)

Both findings verified and fixed:

- **Secret-scanner false positive** — confirmed: \`{{row.customer_uuid}}\` is 21 chars, no whitespace, Shannon entropy 3.78 (threshold is 3.5/length≥20), so it tripped the existing high-entropy heuristic and \`drt validate --strict\` would fail on a config using exactly the kind of compact per-record template this feature is meant to encourage. Fixed by exempting \`{{\` templates in \`_secret_reason\`, mirroring the existing \`${ENV_VAR}\` exemption. Worth noting: this was a latent bug in the pre-existing \`sync.idempotency_key\` field too (same \`_key\` suffix, same template shape) — the fix covers both, not just the new field.
- **Capability probe missing config** — agreed, this was a real gap between what the Protocol's docstring promised ("can answer precisely instead of all-or-nothing") and what the call site actually passed (nothing). \`supports_native_idempotency_key\` now takes the destination config; added a test with a fake destination that answers differently for \`body_mode: record\` vs \`batch\` on two syncs to prove the per-config answer actually threads through.

Full suite (4302 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 3** (commit 361e769)

Confirmed and fixed: the \`{{ }}\` exemption from the previous round returned early before the known-secret-pattern regex scan ran, so \`token: "sk_live_XXXX{{ var('suffix') }}"\` would have silently passed \`drt validate --strict\` — a real regression, and it applied symmetrically to the pre-existing \`\${ENV_VAR}\` exemption too. Reordered \`_secret_reason\` so the known-pattern scan always runs first; the template exemption now only suppresses the entropy fallback (the only check that can't tell "template" from "random-looking literal"). Added a regression test asserting a literal Stripe-key pattern with a Jinja suffix appended is still flagged.

Full suite (4303 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 4** (commit b7d59d4)

**\`_sync_run_id\` race (re-flagged P1, was P2 in round 1):** removed it from this PR entirely rather than defending it again. \`--scope branch\` reviews the whole diff fresh each round with no memory of a prior reply, so that line was always going to get flagged regardless of explanation — the only way to actually resolve it is to take it out of the diff. It had zero consumers in this PR (nothing wired to \`rest_api.py\` yet); it existed solely to serve a future batch-mode template. Removed \`SyncOptions._sync_run_id\`, its engine assignment, and the three tests that only existed to prove it worked. The concurrent-reuse decision (invocation-local \`SyncOptions\` copy vs. documenting concurrent \`run_sync()\` reuse as unsupported) now belongs to the \`rest_api.py\` wiring PR, where a real consumer will exist and the decision can be made deliberately instead of pre-built ahead of need.

**Secret-scanner entropy gap:** confirmed — round 3's fix exempted the entropy check whenever \`{{\`/\`\${\` appeared anywhere in the value, which missed a generic high-entropy secret (no known-provider pattern) with a template suffix appended; that case was caught by entropy before this PR and silently passed after round 3. \`_secret_reason\` now strips template expressions before the entropy check runs rather than bypassing it outright — a pure template collapses to nothing (no finding), a literal secret next to one still gets scanned on what's left over. Added a regression test for the generic-entropy case specifically.

Also cleaned up a smaller consistency issue an advisor pass caught: \`NativeIdempotencyCapable\`'s docstring had invented a \"Provisional\" stability tier that isn't part of ADR 0007's actual policy — relabeled to match the existing \"Public, not yet frozen\" precedent (\`ManagedTableCapable\`) and added its row to the ADR's freeze-scope table.

Full suite (4301 passed) + ruff + mypy clean.

masukai:
**Codex review, round 5: clean.** No actionable issues found — config schema, CLI/MCP validation paths, the capability Protocol, the bundled VS Code schema, and regression tests all reviewed as consistent.

Summary of the 4 review rounds this PR went through:
1. Stale VS Code schemas, hardcoded `_NATIVE_IDEMPOTENCY_WIRED_TYPES` allowlist, missing MCP parity — all fixed.
2. Secret-scanner false positive on compact Jinja templates, capability probe missing the destination config — both fixed.
3. Secret-scanner fix itself had a gap (known-pattern secrets next to a template bypassed entirely) — fixed by reordering.
4. `_sync_run_id` concurrency concern re-flagged at P1 — resolved by removing it from this PR's scope entirely (zero consumers in this diff; belongs in the `rest_api.py` wiring PR where a real consumer exists), plus a second secret-scanner gap (generic high-entropy secrets next to a template) — fixed by stripping template expressions instead of bypassing the check.

Ready for merge review whenever you have a chance to look.

PR #1153, opened 2026-09-15
fix(google_ads): correctly parse partialFailureError (#1152)

### LINKED ISSUE #1152: google_ads: partialFailureError parsing undercounts failures and can't distinguish real errors from OrderIdAlreadyInUse
## Summary

`GoogleAdsDestination.load()`'s partial-failure handling (`drt/destinations/google_ads.py:135-141`) parses the Google Ads API's `partialFailureError` response with a shape that does not match the real API — it undercounts failures whenever more than one conversion fails in a batch, and never surfaces the actual per-conversion error code or message.

## Current (incorrect) code

```python
partial_errors = resp_data.get("partialFailureError")
if partial_errors:
    error_details = partial_errors.get("details", [])
    result.failed += len(error_details)
    result.success += len(conversions) - len(error_details)
    for detail in error_details[:10]:
        result.errors.append(str(detail.get("message", "")))
```

This treats `partialFailureError.details`'s **array length** as the number of failed conversions, and expects each entry to have a flat `message` field.

## What the real API actually returns

Per the [Partial Failure guide](https://developers.google.com/google-ads/api/docs/best-practices/partial-failures) and [error handling docs](https://developers.google.com/google-ads/api/docs/get-started/handle-errors):

- `partialFailureError.details` contains **exactly one** `Any`-packed `GoogleAdsFailure` object, **regardless of how many conversions failed** — not one entry per failure.
- That single `GoogleAdsFailure` has its own `errors: []` array — one `GoogleAdsError` per actual failure.
- Each `GoogleAdsError` has `errorCode` (a oneof, e.g. `{"conversionUploadError": "ORDER_ID_ALREADY_IN_USE"}`), a `message`, and `location.fieldPathElements[].index` — the `index` identifies which conversion (by position in the request's `conversions[]` array) the error belongs to.

So today: if 2 conversions in one batch both fail, `details` has length 1 (one `GoogleAdsFailure` wrapping 2 `errors`), and the current code reports `failed=1, success=len(conversions)-1` — silently miscounting one of the two failures as a success, and `detail.get("message", "")` reads a field that doesn't exist on the outer `GoogleAdsFailure` object at all (always empty).

## Why this wasn't caught

`tests/unit/test_google_ads.py::test_partial_failure` mocks `partialFailureError.details` as `[{"message": "Invalid gclid"}]` — a shape that matches the current (wrong) code, not the real API. This is the same "mock can't prove response-shape validity" failure class as the SQL-mock issues found in `#908`/`#920`'s history — the test is internally consistent but never checked against Google's actual documented (or a captured real) response.

## Impact

- Any `google_ads` sync writing more than one failing conversion in a single `load()` call under-reports `result.failed`, which feeds `drt run`'s row-error counts, `--output json`'s `rows_failed`, and DLQ/retry behavior downstream.
- Per-row error messages are always empty for partial failures today (`detail.get("message", "")` never finds anything).
- This blocks correctly wiring `order_id`-based native idempotency ([#897](https://github.com/drt-hub/drt/issues/897)) — distinguishing a genuine validation failure from an `OrderIdAlreadyInUse` "already delivered" response (which should count as **success**, not failure) requires parsing the real per-conversion error list, which the current code structurally cannot do.

## Fix sketch

Parse `details[0].errors[]`, map each error's `location.fieldPathElements[].index` back to the conversion it refers to, and count/report failures per actual `GoogleAdsError` rather than per `details` array entry. Rewrite `test_partial_failure` (and add a multi-failure case) using the real documented response shape rather than the current invented one.

## Not fixed here

Filing as tracking/investigation — the actual fix lands together with (or just ahead of) #897's `order_id` wiring PR, since that PR needs this parsing to be correct anyway.

masukai:
## Summary

Closes #1152, found while confirming Google's documented `order_id` re-upload/dedup semantics before wiring `native_idempotency_key` (#897 PR3).

`partialFailureError.details` always contains exactly **one** `Any`-packed `GoogleAdsFailure` object regardless of how many conversions failed — not one entry per failure. The previous code treated `details`'s array length as the failure count (`result.failed += len(error_details)`), so a batch with 2+ failing conversions in one upload silently miscounted one of them as a success, and `detail.get("message", "")` read a field that doesn't exist on the outer object at all (always empty).

The real per-conversion errors live in `details[0].errors[]`, each carrying `location.fieldPathElements[].index` back to the conversion it belongs to (per the [Partial Failure guide](https://developers.google.com/google-ads/api/docs/best-practices/partial-failures)). `load()` now parses that nested structure and attributes each failure to its correct original record via `record_row_error` (structured `row_errors`, matching every other destination) instead of the flat `errors` string list. An unparseable response shape degrades to conservatively failing the whole upload rather than guessing.

This is a prerequisite for #897 PR3 (`order_id` wiring): distinguishing a genuine validation failure from an `OrderIdAlreadyInUse` "already delivered" response (which should count as success, not failure) requires correctly parsing the per-conversion error list, which the old flat-count code structurally could not do.

`tests/unit/test_google_ads.py::test_partial_failure` had mocked the response as a flat `{"message": ...}` shape that doesn't match the real API — same "mock can't prove response-shape validity" failure class as the SQL-mock issues found in #908/#920's history.

## Test plan

- [x] Rewrote `test_partial_failure` using the real documented response shape.
- [x] New `test_partial_failure_with_multiple_failed_conversions` — 4 conversions, 2 fail, proving the previous undercount (confirmed it fails against the pre-fix code by inspection: old code would report `failed=1, success=3` instead of `failed=2, success=2`).
- [x] New `test_partial_failure_unparseable_shape_fails_whole_batch` — malformed response degrades safely instead of guessing.
- [x] Full suite (4303 passed) + `ruff check`/`ruff format --check` + `mypy drt` clean.
- [x] VS Code schema regeneration produces no diff (no Pydantic field changes).


masukai:
**Codex review response** (commit f9dff63)

Both findings verified and fixed:

- **Duplicate-error overcounting** — confirmed: Google can return more than one \`GoogleAdsError\` for the same conversion index, and the previous code called \`record_row_error\` once per error rather than once per conversion, inflating \`failed\` past the batch size and risking a duplicate DLQ enqueue for the same record. Fixed by grouping \`mapped\` by conversion index first, joining multiple messages for the same conversion, and emitting exactly one \`RowError\` per distinct failing conversion.
- **Unvalidated nested \`.get()\` calls** — confirmed: a non-dict \`details[0]\`/\`location\`/field-path element would raise \`AttributeError\` instead of returning \`None\`, which fell through to the *outer* generic \`except Exception\` handler — that path records \`failed\` but populates no \`row_errors\`, silently omitting those records from the DLQ. \`_parse_conversion_upload_errors\` now validates every nested mapping's type before dereferencing it, and additionally treats an empty \`errors\` list as unparseable (a non-null \`partialFailureError\` with nothing inside it isn't a legitimate all-clear).

Added 3 regression tests: multiple errors on one conversion counted once, a malformed non-dict \`location\` degrading safely (not crashing) with a proper \`row_errors\` entry, and an empty-\`errors\`-list response correctly treated as unparseable rather than a silent success.

Full suite (4306 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 2** (commit 8e85f58)

Confirmed and fixed: \`details\` is documented to always contain exactly one packed \`GoogleAdsFailure\`, but the parser only checked "non-empty" and read \`details[0]\` unconditionally — silently ignoring any further entries in an undocumented multi-entry response, which would drop real errors and wrongly credit those conversions as successfully delivered. Now requires \`len(details) == 1\`, degrading to the existing conservative whole-batch fallback otherwise. Added a regression test with two detail entries proving it now fails the whole batch instead of crediting the second one as a silent success.

Full suite (4307 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 3** (commit 13cca8f)

Confirmed and fixed both sub-cases:

- \`if partial_errors:\` used truthiness — a present but empty \`partialFailureError: {}\` is falsey in Python, so it took the success branch and credited every conversion as delivered, contradicting the conservative-fallback philosophy the rest of this parser follows. Changed to \`is not None\`.
- A truthy but non-dict \`partialFailureError\` (e.g. a bare string) would have been handed straight to \`_parse_conversion_upload_errors\`'s \`.get()\` call and raised, falling through to the outer generic \`except Exception\` — which records \`failed\` but no \`row_errors\`. Now explicitly checked before parsing, so any non-dict payload goes straight to the conservative whole-batch fallback with proper \`row_errors\` attribution.

Added regression tests for both: \`partialFailureError: {}\` and a non-dict string payload.

Full suite (4309 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 4** (commit 8ec0239)

Both findings verified and fixed:

- **Explicit \`null\` vs absent field** — \`resp_data.get("partialFailureError")\` returns the same \`None\` for both, so a malformed/intermediary response sending an explicit \`"partialFailureError": null\` took the absent-field success branch and credited every conversion as delivered. Now checks field presence (\`"partialFailureError" in resp_data\`) separately from the value, so an explicit null routes through the same conservative fallback as \`{}\` or a non-dict payload.
- **Boolean conversion indices** — confirmed: \`bool\` is an \`int\` subclass in Python, so \`isinstance(index, int)\` accepted \`"index": true/false\` as index \`1\`/\`0\`, which would misattribute the error to an arbitrary record instead of hitting the whole-batch fallback. Now explicitly rejects \`bool\` before accepting an index.

Added regression tests for both.

Full suite (4311 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 5** — pushing back on this one with evidence, not implementing it.

The claim was that \`FieldPathElement.index\` is an \`int64\`, so REST/ProtoJSON would encode it as a decimal string (e.g. \`"index": "1"\`) rather than a JSON number, and the current \`isinstance(index, int)\` check would reject every real response.

I checked the actual current proto source (\`google/ads/googleads/v22/errors/errors.proto\` in [googleapis/googleapis](https://github.com/googleapis/googleapis/blob/master/google/ads/googleads/v22/errors/errors.proto)):

\`\`\`proto
message FieldPathElement {
  string field_name = 1;
  optional int32 index = 3;
}
\`\`\`

\`index\` is \`int32\`, not \`int64\`. Per the [protobuf JSON canonical mapping](https://protobuf.dev/programming-guides/proto3/#json), only 64-bit integer types (\`int64\`/\`uint64\`/\`sint64\`/\`fixed64\`/\`sfixed64\`) are string-encoded in JSON — \`int32\` and friends are encoded as plain JSON numbers, specifically because the string-encoding exists to work around JavaScript's \`Number\` precision limit for 64-bit values, which doesn't apply to 32-bit ones. So the existing numeric \`"index": 1\` in the test fixtures (and the \`isinstance(index, int)\` check) matches the real, current wire format — not implementing this change.

**Separate, unrelated, and more serious finding from this verification pass:** \`_API_VERSION = "v17"\` in \`drt/destinations/google_ads.py\` targets a Google Ads API version that [sunset on 2025-06-04](https://ads-developers.googleblog.com/2025/04/google-ads-api-v17-sunset-reminder.html) — over a year ago. Every \`uploadClickConversions\` call this destination makes today would fail outright. Filing this separately; it's out of scope for this parsing-correctness PR and needs its own investigation (current supported version, any breaking changes since v17 affecting this destination's payload shape).

Issue #1068, opened 2026-09-01
chore(destinations): staged_upload's poll requests share the trigger-host rate-limit bucket, not their own host


masukai:
## Context

Found during adversarial review of #1048's fix (PR #1067), which wired `resolve_rate_limiter()` into `StagedUploadDestination._poll()`. Not fixed there — this is a pre-existing design choice from #769, not a regression #1048 introduces, and the common case isn't affected.

## Problem

`StagedUploadDestinationConfig.rate_limit_key()` (`drt/config/destinations_saas.py`) is deliberately keyed by the **trigger** phase's host:

```python
def rate_limit_key(self) -> str:
    """Per trigger host (#769). A staged upload talks to several URLs, but
    the stage step is usually a one-shot presigned URL on a storage host
    while the trigger is the vendor API that actually publishes a quota..."""
    return f"{self.type}:{urlparse(self.trigger.url).netloc}"
```

This was a reasonable simplification when only `stage`/`trigger` existed. #1048 (PR #1067) now also routes `poll` requests through this same limiter. In the common case `poll.url` shares the trigger's host (checking status on the same vendor API that created the job), so this is correct. But `poll.url` is independently configurable — if a destination polls a *different* host than it triggered against, two problems follow:

- Two syncs polling the same quota-limited poll host but with different trigger hosts get separate limiters and can jointly exceed that host's real rate limit.
- A poll host sharing a trigger host with an unrelated destination's *different* poll host gets incorrectly coupled to that destination's quota.

## Suggested direction

Key polling limiter identity from the rendered `poll.url`'s host when a `poll` phase is configured, rather than always from `trigger.url` — or add an explicit `rate_limit_key` override field on `StagedUploadDestinationConfig` for the rare cross-host case. Needs a decision on backward compatibility for the common same-host case (should probably resolve to the same key it does today, so no behavior change there).

**Effort:** Small-medium.

PR #1155, opened 2026-09-15
fix(google_ads): bump API version v17 -> v25, relax developer-token (#1154)

### LINKED ISSUE #1154: google_ads destination is completely broken: targets Google Ads API v17, sunset 2025-06-04
## Summary

`drt/destinations/google_ads.py:50` hardcodes `_API_VERSION = "v17"`. Google Ads API v17 [sunset on 2025-06-04](https://ads-developers.googleblog.com/2025/04/google-ads-api-v17-sunset-reminder.html) — over a year before today. Every request this destination makes (`POST /v17/customers/{id}:uploadClickConversions`) fails outright with Google's standard "this API version is no longer available" error.

**Every `google_ads` sync has been completely non-functional in production since June 2025**, silently — nothing in drt's own test suite catches this, since all tests mock `_BASE_URL`/`httpx.Client` and never hit the real API version-routing layer.

Found incidentally while verifying a Codex review claim about `FieldPathElement.index`'s wire type for #1153/#897 (confirmed against the real proto source that `google/ads/googleads/v22/errors/errors.proto` still exists on `googleapis/googleapis`, while v17-v21 have already been removed from that repo — consistent with v17's confirmed sunset).

## What needs investigating before fixing

- **Current supported version.** The [versioning docs](https://developers.google.com/google-ads/api/docs/concepts/versioning) reference `v25_0`/`v25_1`-style versions as examples; the [sunset-dates page](https://developers.google.com/google-ads/api/docs/sunset-dates) lists v23/v24/v25 as current/recent. Need to confirm the actual minimum supported version today and whether jumping straight from v17 to it is safe, or whether Google introduced breaking request/response shape changes for `uploadClickConversions` / `ClickConversion` / partial-failure error codes along the way that this destination's payload-building code would need to account for.
- **Interaction with #1153's partial-failure parsing fix.** That PR's `_parse_conversion_upload_errors` was built and tested against the response shape documented for the error-handling guide, which should be stable across versions, but this should be re-confirmed against whatever version this issue ends up targeting.
- **Interaction with #897's planned `order_id` wiring (PR 3).** Should probably land together with or immediately after this fix, since testing `order_id`/`OrderIdAlreadyInUse` semantics against a dead API version would prove nothing.

## Impact

Any user relying on `type: google_ads` has had every sync fail for over a year. This is a correctness/functionality issue for the connector, not a cosmetic one — worth prioritizing above the in-flight #897 PR 3 work.

## Not fixed here

Filing as tracking/investigation only — flagging for the maintainer to decide priority and target version before implementation starts.

### LINKED ISSUE #1156: OAuth2ClientCredentialsAuth token cache keyed by token_url alone — cross-client credential mix-up
## Summary

`drt/destinations/auth.py`'s `_oauth2_cache` was keyed by `token_url` alone:

```python
_oauth2_cache: dict[str, tuple[str, float]] = {}
...
cached = _oauth2_cache.get(auth.token_url)
```

Any two `OAuth2ClientCredentialsAuth` configs sharing one `token_url` — which is common, since it's typically a fixed vendor endpoint shared by every client of that vendor (e.g. `https://oauth2.googleapis.com/token` for any two Google-anything destinations, two different Google Ads customer accounts, a Google Ads + Google Sheets destination pair, etc.) — would silently share one cached access token regardless of having completely different `client_id_env`/`client_secret_env`.

## Impact

Within a token's cache lifetime (up to an hour, minus the 60s safety margin), the **second** distinct OAuth client to authenticate reuses the **first** client's cached access token instead of exchanging its own credentials. Concretely: two `google_ads` destinations for two different customer accounts, authenticated via two different OAuth clients but hitting Google's shared token endpoint, could have the second account's writes silently sent using the first account's access token — either writing to the wrong Google account (if the token happens to have access) or failing with a permission error, neither of which points an operator at the actual cause.

This is a correctness/security bug (credential cross-contamination), not just a rate-limiting inefficiency.

## How it was found

Found via Codex review while fixing #1154 (`google_ads` API version bump): a new tokenless-config rate-limit key override (keyed by OAuth client identity) was flagged as unsound *because* the underlying access-token cache doesn't actually distinguish those same clients — undermining the very identity distinction the rate-limit fix depends on.

## Fix

Fixed alongside #1154 in the same PR (touches shared `drt/destinations/auth.py`, used by every destination configured with `auth: {type: oauth2_client_credentials, ...}`, not just `google_ads`): cache key widened to `(token_url, client_id_env)`, matching the disambiguation convention `drt/config/destinations_saas.py`'s `_auth_identity()` already uses for the identical "which real account is this" question. `client_secret_env` doesn't need to be part of the key — a real `client_id` uniquely determines its own secret.

Regression test added: two `OAuth2ClientCredentialsAuth` configs with different `client_id_env` sharing one `token_url` now correctly make two separate token exchanges and never mix up each other's cached access token.

masukai:
## Summary

Closes #1154. (Originally stacked on #1153; #1153 has since merged and this branch was rebased onto `main`, so the diff now shows only this PR's own changes.)

**The bug:** `_API_VERSION = "v17"` in `drt/destinations/google_ads.py`. Google Ads API v17 [sunset on 2025-06-04](https://ads-developers.googleblog.com/2025/04/google-ads-api-v17-sunset-reminder.html) — over a year ago. Every `uploadClickConversions` call this destination has made since then fails outright with a version-not-supported error. Nothing in the test suite catches this, since every test mocks `_BASE_URL`/`httpx.Client` rather than hitting real version routing.

**The fix:** bumped to `v25` (released 2026-07-22, sunsets ~August 2027 — the longest remaining runway of any current version; v22/v23/v24 each sunset within a year of today). Verified this is a pure version-string bump, not a payload/response migration, by diffing the actual proto sources on `googleapis/googleapis` between v22 and v25: the five `ClickConversion` fields this destination sets (`gclid`, `conversionAction`, `conversionDateTime`, `conversionValue`, `currencyCode`), `orderId` (relevant to the planned #897 `order_id` wiring), and the partial-failure error shape (`GoogleAdsFailure`/`GoogleAdsError`/`FieldPathElement`, including `index`'s `int32` type) are byte-identical across that range.

**Also relaxed the developer-token requirement:** Google's move to a Cloud-project-based access model made the `developer-token` header optional and server-ignored for every caller, not just new adopters. `load()` previously raised `ValueError` when `developer_token_env` resolved to nothing — that's now a purely drt-imposed constraint that would block anyone onboarding under the new access model with no token to give. Still sent when configured (no behavior change for existing setups); omitted rather than raising when absent.

**Rate-limit-key correctness** (surfaced and fixed across 3 Codex review rounds): making the token optional exposed a real gap where every tokenless config shares one rate-limit bucket regardless of actual account. The final fix overrides the limiter key by OAuth-client identity *only* when the token is genuinely absent — `resolve_rate_limiter(..., key_override=...)`, resolved at the one call site (`load()`) that actually knows whether a token resolved, leaving the token-based default untouched (so two OAuth clients sharing one real token still correctly share one bucket).

**Also closes #1156**, a more serious pre-existing bug found while verifying the above: `drt/destinations/auth.py`'s OAuth2 access-token cache was keyed by `token_url` alone, so two distinct OAuth clients sharing one token URL (e.g. any two Google-anything destinations) could silently share one cached access token — a cross-account credential mix-up, not just a rate-limiting inefficiency. Fixed by widening the cache key to `(token_url, client_id_env)`.

**Documented, not fixed** (no client-side workaround exists): since 2026-06-15, Google no longer accepts *new* adopters of offline click-conversion imports on this endpoint at all — a developer token with no prior conversion-import activity gets `CUSTOMER_NOT_ALLOWLISTED_FOR_THIS_FEATURE`. Existing adopters are unaffected during Google's transition to the newer Data Manager API. Noted in `docs/connectors/google-ads.md`.

## Test plan

- [x] All mocked URLs updated `v17` → `v25`.
- [x] Inverted `test_missing_developer_token` → `test_missing_developer_token_no_longer_raises` (proceeds without the header instead of raising).
- [x] New tests: developer token sent when configured; tokenless configs get OAuth-client-based rate-limit key overrides and different clients get different keys; a token-having config's key is left untouched; OAuth2 token cache no longer mixes up two clients sharing one token URL.
- [x] Full suite (4329 passed) + `ruff check`/`ruff format --check` + `mypy drt` clean.
- [x] VS Code schema regeneration produces no diff (no Pydantic field changes).


masukai:
**Codex review response** (commit 0e4851f)

Confirmed and fixed: making the developer token optional means every tokenless config now leaves \`developer_token_env\` at its shared default name, so \`rate_limit_key()\` (keyed solely on that field) would have collapsed unrelated Cloud projects using different OAuth clients onto one shared limiter. This is over-throttling, not a correctness bug — this destination-config method's own established policy (same as the rest-api/staged-upload keying precedent) treats over-sharing as an acceptable safe default — but still narrower than necessary, and worth fixing since it's directly caused by this PR's change.

Fixed by folding in \`_auth_identity(self.auth)\`, the same OAuth-client-disambiguation helper Notion/GitHub Actions/other destinations already use for exactly this "what real account is behind this config" question — no new mechanism, just applying an existing one here too. Two tokenless configs with different OAuth clients are now correctly split; two configs sharing one developer token still share one budget, preserving the original #769 bucketing intent.

Added 2 tests: distinguishes tokenless configs by OAuth client, and confirms same-token configs still share a bucket.

Full suite (4314 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 2** (commit a2179b3) — agreed, and reverted round 1's fix rather than patching it further.

Confirmed: when a developer token *is* configured, Google enforces the quota per token regardless of which OAuth client authenticates the request — that's exactly the original #769 rationale (\"accounts sharing one token share the budget\"). Round 1's fix appended \`_auth_identity(self.auth)\` unconditionally, which splits two configs sharing one real token into separate buckets whenever they use different OAuth clients — letting concurrent syncs multiply the effective request rate against one real, shared ceiling and risk actual \`429\`s. That's the dangerous direction of imprecision (under-sharing a real quota), as opposed to round 1's original finding (over-sharing a limiter across genuinely-unrelated tokenless projects — slower pacing, never a quota error, and the accepted safe default this method's own established policy already tolerates elsewhere, e.g. the \`rest_api\`/\`staged_upload\` host-keying precedent).

There's no way to key correctly for both the token and tokenless cases from config alone — \`rate_limit_key()\` has no runtime credential resolution to tell them apart. Rather than trade one imprecision for a worse one, reverted to the original \`developer_token_env\`-only key and documented why in the method's own docstring, so a future reader doesn't reintroduce the same fix. Replaced the invalidated test with one proving the actually-required invariant: same token + different OAuth clients still share one bucket.

Full suite (4313 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 3** (commit 14d064d) — this is the actual fix.

Confirmed: \`rate_limit_key()\` genuinely has no runtime credential resolution to draw on (that's WHY round 2's revert was correct — it can't tell "tokenless" apart from "has a token I can't see"), but \`GoogleAdsDestination.load()\` *does* know, since it already calls \`resolve_env(None, config.developer_token_env)\` before doing anything else. Moved rate-limiter construction to after that resolution and pass \`resolve_rate_limiter(..., key_override=...)\` — the same escape hatch \`staged_upload\`'s runtime-rendered poll URL already established (#1068) for exactly this "config alone can't answer this, but the call site can" situation — folding in \`_auth_identity(config.auth)\` only for the genuinely-tokenless branch.

This resolves both rounds at once: a config that resolves a real token keeps \`rate_limit_key()\`'s own default untouched (two OAuth clients sharing one token still share one bucket, satisfying round 2's real-quota-holder invariant), while two tokenless configs with different OAuth clients are now correctly split instead of collapsing onto \`developer_token_env\`'s shared default name (round 1's original finding).

Added 3 tests: a tokenless config's override key includes its OAuth client identity, two tokenless configs with different clients get different override keys, and a token-having config passes \`key_override=None\` (falls back to \`rate_limit_key()\` untouched).

Full suite (4316 passed) + ruff + mypy clean.

masukai:
**Codex review response, round 4** (commit 445cf3b)

Confirmed — and this is a real bug more serious than the framing of \"rate-limit bucket splitting\": read \`drt/destinations/auth.py\` and found \`_oauth2_cache\` genuinely is keyed by \`token_url\` alone. Any two distinct \`OAuth2ClientCredentialsAuth\` configs sharing one token URL (very common — e.g. any two Google-anything destinations both hit \`https://oauth2.googleapis.com/token\`) would silently share one cached access token regardless of having completely different \`client_id_env\`/\`client_secret_env\`. That's a cross-account credential mix-up (client B's requests authenticating as client A), not just a quota-coordination inefficiency — and yes, it directly undermines this PR's round-3 fix: if two \"different\" tokenless configs actually share one cached token due to this bug, splitting their rate-limit buckets by OAuth client identity doesn't reflect what's actually happening on the wire.

Filed as #1156 (it affects every destination using \`oauth2_client_credentials\`, not just \`google_ads\`) and fixed in the same PR since it's a direct prerequisite for this PR's own correctness claim. Widened the cache key to \`(token_url, client_id_env)\`, matching \`_auth_identity()\`'s own established disambiguation convention for this exact question. Added a regression test: two configs with different \`client_id_env\` sharing one \`token_url\` now correctly exchange two separate tokens and never return one client's cached token for the other's request.

Full suite (4317 passed) + ruff + mypy clean.

masukai:
Codex review round 5 was correct — verified against Google's own Sept 10, 2026 access-model announcement.

Round 3's fix (splitting the tokenless rate-limit key by `_auth_identity(config.auth)`, i.e. OAuth-client identity) has a real correctness gap: Google now scopes API access/quota to the **Cloud project** that owns the OAuth client's credentials, not to the individual client, and one project can own several OAuth clients. Splitting by client can therefore under-share a quota that's really shared project-wide — the dangerous direction (can cause real 429s), unlike over-sharing (safe, only paces more conservatively).

`OAuth2ClientCredentialsAuth` has no field identifying the underlying Cloud project, so there's no correct finer-grained signal left to split on. Reverted the `key_override`/`_auth_identity` mechanism entirely (0932573) — every `google_ads` config, tokenless or not, now shares one `rate_limit_key()` bucket, per this repo's established over-sharing-is-safe / under-sharing-can-429 policy. Filed #1157 to track adding an explicit Cloud-project-identity config field as the real fix.

Note: the OAuth2 token-cache fix (#1156, `_oauth2_cache` keyed by `(token_url, client_id_env)`) is unaffected and stays — it's a real cross-account credential bug independent of the rate-limit-key question, kept in this PR since it's a direct prerequisite for the destination's own auth correctness.

Full suite (pytest + ruff + mypy) green. Relaunching Codex review.

PR #1158, opened 2026-09-16
feat(google_ads): wire native_idempotency_key as ClickConversion orderId

### LINKED ISSUE #897: feat(destinations): idempotency keys for fire-and-forget destinations (retry-safe writes)

masukai:
## Summary

Last of #897's three-PR split (PR1: #1150 config surface, PR2: #1151 rest_api wiring). Wires `destination.native_idempotency_key` into `google_ads.py`, sending it as ClickConversion's documented `orderId` dedup field so drt's per-destination retry (#277) is safe against a request that succeeded server-side but looked like a failure client-side.

- Rendered per record (`row` + `sync_name` in scope), same content-derived-key convention `rest_api.py`'s `body_mode: record` established in PR2 — the goal is that redelivering the same row dedupes against a previously successful delivery of that exact row.
- Confirmed Google's re-upload semantics directly against the `ConversionUploadErrorEnum` proto source before wiring (per the config-surface PR's own deferred gate): a reused `orderId` comes back as `ORDER_ID_ALREADY_IN_USE`, now counted as a successful delivery rather than a row error.
- Narrower than it could be, on purpose: an index is only treated as dedup when *every* error for it is `ORDER_ID_ALREADY_IN_USE` — a genuine co-occurring error (e.g. `INVALID_GCLID`) still fails the row. `DUPLICATE_ORDER_ID` (two rows colliding on one rendered key within a batch) is a real config problem and is never swept into the dedup path.
- `GoogleAdsDestination` now implements `NativeIdempotencyCapable`, so `drt validate` no longer warns on this destination when the field is set.
- Docs updated with two constraints on `orderId` worth calling out explicitly: it must not contain PII (a natural mistake since `upsert_key` templates elsewhere are often an email), and it isn't permitted at all on externally-attributed conversion actions.

Closes #897.

## Test plan
- [x] `pytest tests/unit/test_google_ads.py tests/unit/test_cli_validate.py` — new coverage for orderId rendering, template errors, dedup success, dedup-with-real-error still fails, `DUPLICATE_ORDER_ID` not treated as dedup, `sync_name` template scope, and the `drt validate` no-warning case
- [x] Full suite (`pytest`), `ruff check`, `ruff format --check`, `mypy drt` all green


masukai:
Codex review: clean on first round — idempotency-key wiring, partial-failure accounting, and capability reporting are internally consistent and covered by focused tests; ruff/mypy pass; no actionable issues found.

Issue #1147, opened 2026-09-13
drt retry --limit: reconcile() can silently drop an unconfirmed legacy dead letter sharing a content-derived ID with a confirmed one


masukai:
## Summary

Found via Codex review on PR #1146 (P2, https://github.com/drt-hub/drt/pull/1146#issuecomment — see the review comment on `drt/cli/commands/retry.py:405-406`). Not introduced by that PR — the hazard is pre-existing in `DlqBackend.reconcile()`'s own id-set semantics and already reachable on `main` today via `drt retry --limit`. PR #1146 adds a second call site with the same shape, which is what surfaced it in review.

## The mechanism

`decode_dead_letter_line()` (`drt/state/dlq.py`) assigns a deterministic `id = hashlib.sha256(raw_line.strip().encode()).hexdigest()` to a legacy JSONL entry that predates #955's explicit `id` field. Two byte-identical legacy lines therefore decode to the **same** id.

`LocalDlqStore.reconcile()` (and `ObjectStoreDlqBackend.reconcile()`, same shape) filters by id-set membership against a fresh read:

```python
result = [
    updates.get(entry.id, entry) for entry in current if entry.id not in remove_ids
]
```

This removes **every** physical entry whose id is in `remove_ids` — not just the one instance that was actually confirmed. If two legacy duplicates share an id, and only one of them was actually retried/confirmed this invocation while the other was never touched (left in `untouched` by `--limit`, or — as of #1146 — left un-attempted because a later chunk raised before reaching it), the untouched twin is silently removed too, even though it was never delivered.

## Reproduction shape

1. A pre-#955 DLQ file contains two byte-identical legacy lines (no `id` field) for the same sync — both decode to the same SHA-256 id.
2. `drt retry <sync> --limit 1` retries only the first; it succeeds and is named in `remove_ids`.
3. `reconcile()` removes **both** physical entries by id, even though the second was never sent.

The same shape now also happens on the exception path PR #1146 added: if one physical duplicate is in an earlier, successfully-processed chunk and its twin is in a later chunk whose `dest.load()` raises before running, the exception-path `reconcile()` removes both.

## Why not fixed in #1146

[...]

PR #1160, opened 2026-09-17
fix(cli): drt retry's normal-completion reconcile() can drop an untouched legacy duplicate

masukai:
## Summary

Part of #1147.

`replay_dead_letters()`'s exception-path `reconcile()` call already guards against a hazard found across three Codex review rounds on #1146: a pre-#955 legacy DLQ entry's id is a SHA-256 hash of its raw line, so two byte-identical legacy lines collide on it, and `reconcile()`'s `remove_ids` is an id-set filter (removes *every* physical entry sharing an id), not an occurrence-count filter. If one physical duplicate is confirmed while its twin is untouched (excluded by `--limit`, or unreached because a later chunk raised), naming that id in `remove_ids` silently deletes both.

The always-run, end-of-function `reconcile()` call — reached when the retry loop completes with **no** exception at all — had no equivalent guard. This PR closes that one remaining gap: `remove_ids - {e.id for e in untouched}` at that call site, mirroring the exception path's existing exclusion.

**Deliberately narrow, not a full fix for #1147**: `updates` (the failed-again attempts-bump path) stays unfiltered everywhere. `reconcile()` has no way to address one occurrence of a shared id without the other — excluding the id from `updates` would mean the entry that *was* actually retried also loses its own attempts bump, which is strictly worse than the existing accepted cosmetic drift (an untouched twin can pick up a bumped `attempts`/error it never earned). A full occurrence-aware `reconcile()`, or giving legacy entries a real per-line identity, would close that residual too — but `DlqBackend` is a Protocol frozen at v1.0 ([ADR 0007](docs/adr/0007-protocol-stability-policy.md)), and this repo is mid-flight toward that freeze, so a Protocol-shape change isn't attempted here.

Scope: only reachable for entries that predate #955 and are byte-identical to another still-queued entry, on `LocalDlqStore`/`ObjectStoreDlqBackend` (the three warehouse-backed `DlqBackend`s key mutations off a real primary-key column, never exposed to this).

## Test plan
- [x] New test verified failing against the pre-fix code (queue `[A, B, A2]`, `--limit 2`, both A and B succeed — pre-fix both physical A copies were deleted; fixed, both survive alongside B's removal)
- [x] `pytest tests/unit/test_cli_retry.py` (29 passed) + full suite, `ruff check`, `ruff format --check`, `mypy drt` all green


masukai:
Codex review: clean on first round — correctly prevents id-set reconciliation from deleting untouched legacy duplicate entries beyond the retry limit; the regression test exercises the normal-completion path without introducing a functional regression.

PR #1161, opened 2026-09-17
feat(config): google_ads.cloud_project_id for correct rate-limit quota scoping

### LINKED ISSUE #1157: google_ads: no config field identifies the Cloud project for correct rate-limit quota scoping
## Background

Google's September 2026 access-model change scopes Google Ads API quota to the **Cloud project** that owns a caller's OAuth client credentials, not to the developer token (now optional/server-ignored) or to the individual OAuth client (a project can own several clients).

PR #1155 (closing #1154, the v17 API-version sunset fix) tried two finer-grained `rate_limit_key()` designs for the now-common tokenless case and reverted both after Codex review:

1. Always folding OAuth-client identity (`_auth_identity(config.auth)`) into the key -- rejected because it splits two clients that share one real token's quota.
2. Folding OAuth-client identity in only for tokenless configs -- rejected because Google's real quota boundary is the *Cloud project*, and `OAuth2ClientCredentialsAuth` has no field identifying that project, so two distinct OAuth clients belonging to one project would incorrectly get independent limiters (able to jointly exceed the real shared quota and trigger 429s).

## Current state

`GoogleAdsDestinationConfig.rate_limit_key()` (`drt/config/destinations_saas.py`) shares one bucket across every `google_ads` config regardless of token or OAuth-client identity -- the safe, conservative default per this repo's over-sharing-is-safe / under-sharing-can-429 policy, but coarser than necessary for setups running multiple genuinely independent Cloud projects concurrently.

## Proposed fix

Add an explicit config field that names the real quota boundary directly, e.g. a `cloud_project_id` (or generic `rate_limit_scope`) on `GoogleAdsDestinationConfig`, and fold it into `rate_limit_key()` when present, falling back to the current shared-bucket behavior when absent.

Out of scope for #1155 since it needs new config surface, not just a keying change.

masukai:
## Summary

Closes #1157.

PR #1155 (closing #1154) tried two finer-than-token `rate_limit_key()` designs for `google_ads` and rejected both after Codex review, because `OAuth2ClientCredentialsAuth` has no field identifying the Google Cloud project that actually owns API quota under Google's Cloud-project-based access model (one project can own several OAuth clients, so per-client keys can under-share a quota that's really shared project-wide).

Rather than keep guessing at a derived signal, `GoogleAdsDestinationConfig.rate_limit_key()` now accepts an explicit, optional `cloud_project_id`:

- When set, it's used directly as the key (`f"{type}:{cloud_project_id}"`), taking precedence over the developer-token fallback regardless of whether a token is also configured.
- When unset, behavior is unchanged from #1155 — every tokenless-or-not `google_ads` config shares one conservative bucket.

Also includes the VS Code extension's regenerated JSON schema (`make`-driven, config-model changes are a hard CI gate) and a bundled-extension CHANGELOG entry, plus ROADMAP.md updates marking #1157 and #1147 (merged in #1160) as done in the v1.0 hardening pass.

## Test plan
- [x] New `rate_limit_key()` tests: same `cloud_project_id` → same bucket across different tokens/OAuth clients; different `cloud_project_id` → different buckets; unset → unchanged fallback behavior
- [x] Full suite (`pytest`), `ruff check`, `ruff format --check`, `mypy drt`, `make check-drift` all green


masukai:
Codex review was correct — verified directly: `rate_limit_key()` for two configs with `developer_token_env="TOKEN_A"` vs `"TOKEN_B"` returns different keys even though both could belong to the same real Cloud project. The docs/CHANGELOG/docstring text overclaimed a universal shared fallback that doesn't exist. Fixed (e35998c): corrected all three call sites (`docs/connectors/google-ads.md`'s table row + Notes bullet, `CHANGELOG.md`, and the `rate_limit_key()` docstring itself) to state the actual guarantee — the fallback shares a bucket only among configs naming the *same* `developer_token_env`; configs naming it differently need an explicit `cloud_project_id` to share correctly even when they're on the same Cloud project. No code-behavior change, docs/comments only. Relaunching Codex review.

masukai:
Codex review round 2: clean — the optional cloud project identifier is propagated through configuration, rate-limit keying, documentation, tests, and the generated VS Code schema while preserving the existing fallback behavior.

PR #1162, opened 2026-09-18
feat(cli): drt serve --auth oidc — OIDC JWT verification for Pub/Sub push

### LINKED ISSUE #903: feat(cli): OIDC JWT verification for drt serve — the Pub/Sub push leg of #854

masukai:
## Summary

Closes #903 (split out of #854, ADR 0004 Tier 3's last residual).

`drt serve` gains a fourth auth scheme, `oidc`, for Google Pub/Sub push subscriptions — which authenticate with an OIDC JWT in the `Authorization: Bearer <token>` header, not a body signature, so this leg previously needed a verifying proxy in front of `drt serve`.

- `--auth oidc` + `--oidc-audience` (required — the URL the push subscription was configured with, checked as the JWT's `aud` claim)
- `--oidc-issuer` (defaults to Google's own `https://accounts.google.com`)
- `--oidc-email` (optional — restrict acceptance to one service account, e.g. the subscription's own push identity)

Verification delegates signature checking (against Google's rotating public keys) to `google-auth`'s `google.oauth2.id_token.verify_oauth2_token`, rather than hand-rolling JWKS fetch + RS256. Per core's no-heavy-deps-outside-extras rule, this lands as its own extra, `drt-core[serve-oidc]` — requesting `--auth oidc` without it logs a clear install hint and fails closed (401), never a 500, matching this module's existing contract that any verification failure is "unauthorized," not a crash.

Unlike `hmac`, a `GET` is verified identically to a `POST` under `oidc` — the JWT signs itself independent of the request body/path, so there's no GET/POST asymmetry to account for the way there is under `hmac`.

Also: `docs/guides/using-webhook-trigger.md` gains a Pub/Sub OIDC section (replacing the old "needs a proxy" caveat), and [ADR 0004](docs/adr/0004-streaming-and-event-triggered-syncs.md) gets a closing amendment noting Tier 3 has no known residual left.

## Test plan
- [x] `_verify_oidc` unit tests via `sys.modules` injection (same pattern `test_deltalake_source.py` uses for its own optional extra, so these run without `drt-core[serve-oidc]` installed): valid token accepted, missing `Bearer` prefix rejected before touching the verifier, verification-exception rejected, wrong `iss`/`email` rejected, unset `iss`/`email` skip those checks, library-not-installed rejected
- [x] Two end-to-end tests through the real HTTP server confirming `Handler._check_auth`'s scheme dispatch
- [x] Fixed 2 pre-existing tests that used `scheme="oidc"` as a stand-in for "not yet supported" (now genuinely supported) — updated to assert the new real validation instead
- [x] Full suite (`pytest`, 4355 passed), `ruff check`, `ruff format --check`, `mypy drt`, `make check-drift` all green


masukai:
Codex review found 3 real gaps, all fixed (5cb7532):

1. **[P1] Require an allowed OIDC caller identity** — correct. A valid signature + right `aud` is not proof of authorization; Google mints ID tokens with any audience for any Google Cloud principal who asks. `--oidc-email` is now required, like `--oidc-audience`.
2. **[P1] Install the transport required by the OIDC extra** — verified: base `google-auth` does not install `requests` (confirmed via package metadata — it's an optional extra on google-auth's own side). Fixed by depending on `google-auth[requests]`.
3. **[P2] Cache Google's signing certificates** — correct, verified `_fetch_certs` does an unconditional fresh HTTP GET on every `verify_oauth2_token` call with no built-in caching. Fixed with a process-wide `cachecontrol`-wrapped session (new `serve-oidc`-only dependency), reused as a singleton across calls so it actually benefits from Google's own `Cache-Control` headers.

Also fixed test isolation: the new singleton needed resetting between tests, and one test's "library not installed" simulation needed adjusting since `sys.modules` injection for a parent package can mask a leaf-module removal via the attribute-lookup path. Full suite green, relaunching review.

masukai:
Codex review round 2: clean — signature, audience, issuer, and caller identity all validated; fails closed on dependency/verification errors; optional deps, CLI wiring, caching, and test coverage all consistent with the implementation. (One exploratory `pytest` invocation in the review sandbox itself failed — likely resolving a different interpreter/venv there, since `python -m pytest tests/unit/test_server.py tests/unit/test_serve_command.py` passes clean here, 79/79 — not flagged as a finding in the final verdict.)

masukai:
CI failed on all 4 Python versions (Linux) despite passing locally on macOS — even with a throwaway venv matching CI's exact resolved typer/click/rich versions. Root cause: the two new oidc CLI tests asserted specific option names appear in Typer/Click/Rich's rendered error panel text, which wraps/truncates based on detected terminal width — apparently different enough on GitHub Actions' Linux (no real TTY) runners to cut the asserted substring out of the captured output. Fixed (a65f7ef) by dropping those fragile output-content assertions — the actual validation messages are already reliably pinned by `test_serve_oidc_scheme_requires_audience`/`_email`, which call `serve()` directly and bypass CLI/Rich rendering entirely, so no real coverage is lost. Full suite green locally; watching this CI run.

masukai:
Codex review round 3: correct, verified directly against the installed library. `google.oauth2.id_token.verify_oauth2_token` already checks `idinfo["iss"] not in _GOOGLE_ISSUERS` internally, where `_GOOGLE_ISSUERS == ["accounts.google.com", "https://accounts.google.com"]` — both spellings are legitimate for a real Google-issued token. Defaulting `--oidc-issuer` to `"https://accounts.google.com"` added a redundant, and actively wrong, exact-match check on top: a legitimate token using the un-prefixed form would have been rejected with a 401 by drt's own extra check, despite already passing google-auth's correct internal validation.

Fixed (0b057b0): `--oidc-issuer`/`AuthConfig.oidc_issuer` now default to unset — `iss` verification is left entirely to `verify_oauth2_token`'s own (correct) internal check, and drt's additional exact-match is applied only when the operator explicitly overrides it (documented use case: a future non-Google IdP reusing this same JWT verification path, which has no such built-in check of its own). Added a regression test proving both accepted spellings pass with the new default. Full suite green, relaunching review.

masukai:
Codex review round 4: correct. `verify_oauth2_token` is Google-specific — it fetches certs only from Google's own hardcoded endpoint (`_GOOGLE_OAUTH2_CERTS_URL`), so a token from any other issuer fails signature verification outright, before any issuer comparison could ever matter. `--oidc-issuer`'s documented purpose ("override for a future non-Google IdP") was a capability the implementation could never actually deliver.

Fixed (cbe209d) by removing `--oidc-issuer`/`AuthConfig.oidc_issuer`/`_verify_oidc`'s `issuer` parameter entirely, rather than keeping dead configuration surface that silently does nothing useful. Issuer verification is left fully to `verify_oauth2_token`'s own correct internal check (which already validates against `_GOOGLE_ISSUERS == ["accounts.google.com", "https://accounts.google.com"]`). This also fully resolves round 3's finding by construction — there's no separate default to get wrong anymore.

This closes the loop on the last 3 rounds: round 2 added the redundant issuer default, round 3 caught that default was actively wrong, round 4 caught that the override concept itself couldn't work at all — the right end state is no issuer parameter here, which is what's shipped now. Full suite green, relaunching review.

masukai:
Codex review round 5: correct, confirmed against Google's own documentation (docs.cloud.google.com/pubsub/docs/authenticate-push-subscriptions), which explicitly checks both `claim["email"]` and `claim["email_verified"]` in its own sample verification code. `email` alone only asserts who the token *claims* to be; `email_verified` is what Google actually vouches for.

Fixed (36f6d42): `_verify_oidc` now rejects when `claims.get("email_verified") is not True` (an explicit `false` and a missing claim are both rejected the same way), alongside the existing email match. Full suite green, relaunching review.

masukai:
Codex review round 6: clean — signature, audience, issuer, expected service-account email, and email verification status all correctly validated, failing closed on dependency/verification errors. CLI wiring, optional deps, docs, and tests all consistent with the implementation.

Issue #1052, opened 2026-08-31
feat(destinations): Klaviyo event tracking (/api/events/) not implemented, no tracking issue


masukai:
## Problem

The Klaviyo destination can only update profiles/lists — event tracking (`/api/events/`) is explicitly called out as not implemented, both in code and docs, but as a bare "follow-up" with no issue reference:

- `drt/destinations/klaviyo.py:27`
- `docs/connectors/klaviyo.md:68`

## Why it matters

This limits Klaviyo to customer-attribute sync rather than behavioral marketing workflows (e.g. syncing a warehouse-computed "abandoned cart" or "upgraded plan" event to trigger a Klaviyo flow) — a core reverse-ETL use case. Comparable destinations already covered by drt (Amplitude, Mixpanel) both support event ingestion, so this is an inconsistency within drt's own existing connector set, not a new category of feature.

## Suggested direction

Add an `events` write mode to the Klaviyo destination targeting `/api/events/`, following the same shape as the existing Amplitude/Mixpanel event-ingestion legs.

**Effort:** Medium.

PR #1163, opened 2026-09-18
feat(destinations): klaviyo.backfill to suppress live flow triggers

### LINKED ISSUE #1075: Klaviyo events destination: expose backfill flag to suppress live flow triggers during historical replay
## Summary

The Klaviyo events destination added in #1052 (PR #1072) sends every event as a live event. Klaviyo's Events API supports a `backfill: true` attribute specifically to suppress triggering live flows/automations when historical data is being replayed (e.g. a first full sync, or a cursor override re-processing old warehouse rows).

Without this, a first full sync or any backfill-style replay of historical rows through the Klaviyo events destination can re-trigger existing customer-facing flows (emails/SMS) for old events, which is surprising and potentially spammy for end customers.

## Suggested scope

- Add a `backfill` option to `KlaviyoDestinationConfig` (event endpoint), emitted in the event `attributes` payload when set.
- Document the safe-historical-replay pattern in `docs/connectors/klaviyo.md`.
- Note: this requires the connector to be on a Klaviyo API revision that supports `backfill` for the Events API — verify current supported revision when implementing (the default revision was bumped in PR #1072, partly for an unrelated near-term retirement-date reason).

## Origin

Surfaced during adversarial review of PR #1072 (#1052). Classified as a new-feature request rather than a defect in "add Klaviyo event tracking" — filed separately per this repo's scope discipline rather than expanding that PR.

masukai:
## Summary

Closes #1075.

The Klaviyo events destination (#1052) sends every event as a live event, so a first full sync or a `--cursor-value` override re-processing old warehouse rows can re-trigger existing Klaviyo flows and send customers duplicate/backdated emails or texts for things that already happened.

`endpoint: event` now accepts `backfill: true`, sent as the event's own `backfill` attribute in the request body — Klaviyo still records the event for metrics/segmentation but skips flow triggering. Rejected at config-validation time when set alongside `endpoint: profile` (profiles have no such concept), matching the class's existing event-only-field validation pattern (`metric_name`/`unique_id_field`).

**Required supporting fix**: Klaviyo's own changelog confirms `backfill` was introduced at API revision `2026-07-15` — newer than the connector's existing default (`2026-01-15`). Bumped the default accordingly, after checking Klaviyo's changelog for every revision in between (`2026-04-15`) and confirming the one breaking change in that range (profiles' `conversation` relationship becoming plural `conversations`) touches nothing this destination reads (it never uses that relationship or `?include=conversation`). No behavior change for existing event configs that don't set the new field.

`docs/connectors/klaviyo.md` gains a "Safe historical replay" section with a worked example.

## Test plan
- [x] Config validation: `backfill=True` rejected on `endpoint: profile`, allowed on `endpoint: event`
- [x] Payload: `backfill` omitted by default (byte-identical to pre-#1075 payloads), present as `true` when set
- [x] Updated existing tests' hardcoded `revision` header expectations to the new default
- [x] Full suite (`pytest`, 4363 passed), `ruff check`, `ruff format --check`, `mypy drt`, `make check-drift` all green


masukai:
Codex review: correct — a config explicitly pinning the pre-#1075 default revision (`2026-01-15`) alongside `backfill: true` passed validation despite Klaviyo not recognizing the field before `2026-07-15`, which would have silently defeated the entire feature's purpose (exactly the customer-facing spam it exists to prevent, with no error to warn the operator).

Fixed (ff82f09): config validation now rejects `backfill: true` combined with any `revision` older than `2026-07-15` (Klaviyo revisions are `YYYY-MM-DD` strings, so lexical comparison against the same minimum the default now uses is a valid date comparison). Full suite green, relaunching review.

PR #1164, opened 2026-09-19
feat(cli): target/run_results.json per-invocation run artifact

### LINKED ISSUE #778: feat(cli): run artifacts — versioned target/run_results.json per invocation

masukai:
## Summary

Closes #778 for `drt run` (see Scope below for what's deliberately deferred).

`drt run --output json` prints a summary to stdout and vanishes with the process — no durable, machine-readable record a CI system can upload as an artifact, or an observability pipeline can ingest after the fact. This mirrors dbt's own `run_results.json` pattern.

`drt run` now also writes `target/run_results.json` on every invocation, **independent of `--output`** (written in the default text mode too — the whole point is a durable record even when JSON wasn't requested on stdout):

```json
{
  "schema_version": 1,
  "invocation": {
    "run_id": "…", "started_at": "…", "completed_at": "…", "duration_seconds": 12.4,
    "argv": [...], "drt_version": "0.10.0",
    "succeeded": 1, "failed": 0, "skipped": 0
  },
  "results": [ /* the exact same per-sync entries --output json's "syncs" array already contains */ ]
}
```

- **One shape, not two**: `results` reuses the exact per-sync entry dicts `--output json` already builds (`run.py`'s existing `_run_one`/`_skipped_entry`) — no second schema to keep in sync.
- **`schema_version`** follows the docs manifest's own versioning convention (a plain int, docstring-documented, additive/omit-when-absent) rather than publishing a separate JSON Schema file — this repo has no precedent for the latter even for the docs manifest, a comparably important artifact.
- **`--target-path`** (default `target/`, mirroring `docs generate --output`'s own default) relocates it.
- **Best-effort**: written unconditionally after the sync-dispatch loop, at the same call site the existing `--output json` stdout print already sits at — already confirmed reached on a graceful-shutdown (#279) exit, so no new signal-handling wiring was needed. Wrapped in `try`/`except OSError` so a permissions/disk-full failure is logged and never changes the run's own exit code.

## Scope

**Deliberately not included:**
- A per-sync watermark before/after pair (the issue's own illustrative example showed one) — the engine never returns the post-run watermark value to the CLI today, only to `StatePersistingObserver`, which persists it directly. Adding it would mean a new `watermark_storage.get()` read per sync purely for this artifact — not justified until a real consumer asks for it.
- `drt test`/`drt build` writing a compatible artifact — `drt test`'s JSON shape (`_SyncTestResult`) is structurally different (no per-sync timing/run_id), a separate design problem.
- drt-action switching its step-summary/outputs to read this file instead of parsing stdout JSON — that's a fully separate repo.

**CLI/MCP parity**: `--target-path` is excluded from `scripts/check_cli_mcp_parity.py`'s `run` → `drt_run_sync` mapping, alongside the existing `--threads`/`--fail-fast`/`--failed` exclusions there — the artifact is inherently multi-sync/one-invocation, and `drt_run_sync` runs exactly one sync per call with no invocation-level concept to place it against.

`docs/guides/ci-cd-integration.md` gains a "Run artifacts" section with a GitHub Actions upload example.

## Test plan
- [x] Written in default text mode (not just `--output json`)
- [x] `results` byte-identical to `--output json`'s `syncs` array for the same invocation
- [x] `--target-path` redirects the write location
- [x] Written even when a sync fails (best-effort, independent of exit code)
- [x] A write failure (blocked target directory) doesn't change the run's exit code or crash
- [x] `scripts/check_cli_mcp_parity.py` still reports zero gaps
- [x] Full suite (`pytest`, 4364 passed), `ruff check`, `ruff format --check`, `mypy drt`, `make check-drift` all green


masukai:
Codex review: all 3 findings from the first round confirmed and fixed (0d66ed9).

**P1 — dbt collision.** `docs/guides/using-with-dbt.md` documents `dbt run && drt run` as the supported co-located workflow, and `dbt run` writes its own `target/run_results.json` there. Default `--target-path` changed from `target/` to `target/drt/` — a drt-owned subdirectory, matching the precedent `docs generate` already set with `target/docs/`.

**P2 — no-op exits skipped the artifact.** Confirmed by reading through `run()`: `_write_run_results` was only called after the full sync-dispatch loop, but several `raise typer.Exit(...)` no-op paths (no syncs found, `--select state:modified` with no changes, `--failed` against a clean prior run, a rejected `--limit`/`--full-refresh` combo) return before reaching it. Fixed by hoisting `run_id`/timing/counters to right after `syncs = load_syncs(...)` resolves, and wrapping everything from there through the end of the function in `try`/`finally`, with the artifact write moved into `finally`. This deliberately does *not* reach further back to cover preflight failures (bad project/profile/vars) before `syncs` is known — mirrors dbt's own `run_results.json`, which isn't written until compilation resolves a node list either.

**P3 — text-mode diff data missing.** `entry["diff"] = diff_to_dict(diff_value)` was gated on `ctx.json_mode`, so a plain-text `--dry-run --diff` run's artifact silently dropped diff data despite the artifact being documented as independent of `--output`. Populating the entry is now unconditional; `print_diff_table(...)` stays a separate, purely-cosmetic text-mode console concern.

Added 2 new tests (no-op exit still writes an artifact; text-mode diff is included) alongside updating the 5 existing tests' paths for `target/drt/`. Full suite (4233 tests), ruff, mypy, and `make check-drift` all pass. Docs/CHANGELOG updated to match.

Relaunching Codex review now.

masukai:
Codex review: round 2 found 3 more real issues, all fixed (9157f03).

**P1 — secrets in the artifact.** `sys.argv` (can carry a `--vars` value verbatim) and per-sync `error` (`str(e)` from a raw connector exception) were persisted unredacted. Unlike a console line, this file is now recommended for CI upload with `if: always()`, meaningfully increasing exposure. Fixed by extracting the docs manifest's existing connector-error redaction (`drt/docs/builder.py`'s `_redact_error_text`, #698 — already applied there because a hosted docs site has the identical problem) into a shared `drt/_redaction.py`, and applying it to both fields.

**P2 — diff sample can crash `json.dumps`.** `diff_to_dict()` can carry warehouse-native `datetime`/`Decimal`/UUID values that `json.dumps` can't serialize by default. This was a pre-existing latent bug in `--output json --dry-run --diff` too, but my own P3 fix from round 1 (populating `entry["diff"]` unconditionally) newly exposed *text-mode* runs to it as well, since now `run_results.json` always attempts to serialize diff data. Fixed with `json.dumps(..., default=str)` in `_write_run_results` — degrades to a readable string rather than turning an otherwise-successful preview into a crash.

**P2 — rejected invocations looked like clean no-ops.** After round 1's fix, every no-op exit writes an artifact with `succeeded`/`failed`/`skipped` all 0 — correct for "nothing to do" (no syncs found/changed/failed), but identical-looking for a genuinely rejected invocation (`--limit 0`, an unmatched selector) that also never attempts a sync. Fixed by wrapping the body in `except typer.Exit as exc: exit_code = exc.exit_code ...` / `except BaseException: exit_code = 1` (both re-raising), and adding `invocation.exit_code` to the envelope — a CI consumer can now tell the two apart without parsing stdout.

Added 4 new tests (exit_code distinguishes rejected-vs-clean, error redaction, argv redaction, diff serialization safety) — 11 total in `test_run_results_artifact.py`. Full suite (4237 tests), ruff, mypy, and `make check-drift` all pass. Docs/CHANGELOG updated.

Relaunching Codex review now.

masukai:
Codex review: round 3 found 1 more real issue, fixed (88ef325).

**P1 — heuristic redaction misses arbitrary --vars key names.** `--vars` lets a project define any variable name it wants (`stripe_key: sk_live_...`), and the free-text keyword sweep only matches a fixed list (`password`/`token`/`api_key`/...) — `stripe_key` isn't `api_key` or `access_key`, so it sailed through unredacted. A finite keyword list can never anticipate every project's own var names, so the fix isn't to extend the list — it's to stop guessing: the entire value following `--vars` (both `--vars value` and `--vars=value` forms) is now redacted outright, key name or not. Everything else in argv still goes through the free-text sweep (e.g. a `--state` path could theoretically embed a connector URL).

Added 1 new test (`stripe_key`-style arbitrary var name, both `--vars` forms) — 12 total. Full suite (4238 tests), ruff, mypy, `make check-drift` all pass.

Also: **PR #1163 (#1075, Klaviyo backfill) merged** — issue closed. This PR is next once clean.

Relaunching Codex review now.

masukai:
Codex review: round 4 found 3 more real issues, all fixed (6430ed7).

**P1 — Authorization header credential survives redaction.** The unquoted-value branch of the KV regex used `\S+`, which stops at the first space — `Authorization: Bearer sk_live_123` only redacted `Bearer`, leaving the actual credential sitting right after it in the persisted artifact. Fixed by letting an unquoted value consume the whole space-separated run instead of just one word, bounded by `,`/`;`/`)` so it still doesn't bleed into unrelated trailing text (`host=x, port=y` stays scoped to `x`). Verified this doesn't regress the existing docs-manifest redaction tests that share this same regex (`test_run_errors_redacted_by_default`, `test_last_error_redacted_by_default`) — full suite confirms.

**P1 — diff-preview failures bypass redaction entirely.** `delete_preview_unavailable_reason`/`fallback_reason` come from `f"{type(error).__name__}: {error}"` around a failed mirror-delete-preview read (`drt/engine/diff.py`) — the identical raw-connector-exception shape `entry["error"]` already redacts, just reached through a different failure path that the round-2 fix never touched. Fixed at the shared `diff_to_dict()` serializer (same place both `--output json` and the artifact already share one shape from), so it can't be missed by a third path either.

**P2 — a stale artifact misrepresents a failed invocation as a stale success.** A preflight failure (bad project/profile/vars, `--diff` without `--dry-run`) never reaches the `try`/`finally` that writes a fresh artifact — deliberate, since syncs aren't known yet. But that meant a *previous* successful run's `run_results.json` was left untouched in a reused workspace, ready for a CI job's `if: always()` step to silently re-upload as if it were this (failed) invocation's result. Fixed by clearing it at the very top of `run()`, before any preflight check.

Added 3 new tests (Authorization-header redaction, diff-error-field redaction, stale-artifact clearing) — 15 total. Full suite (4241 tests), ruff, mypy, `make check-drift` all pass.

masukai:
Codex review: round 5 found 2 more real issues, both fixed (663d005).

**P1 — quoted dict-repr keys bypass redaction entirely.** The KV regex required the separator (`=`/`:`) to sit directly after the keyword, but a Python/JSON dict repr — a very common shape for a `requests`-style exception embedding request headers — quotes its keys: `{'Authorization': 'Bearer sk_live_123'}`. That quote between "Authorization" and ":" broke the match outright, so the credential was persisted verbatim. Fixed by accepting an optional quote right after the keyword. Also addressed in the same regex: the `,`/`;`/`)` boundary I added in round 4 (to avoid bleeding into unrelated trailing text like `host=x, port=y`) under-redacts when a real secret itself contains one of those characters — `password=abc,def` is one value, not `abc` plus unrelated context, and under-redaction is a leak while over-redacting an adjacent field is only cosmetic. Dropped that boundary; bounded instead to 5 extra words and never crosses a newline, so it still can't run on into an unrelated later paragraph of a long exception message.

**P2 — non-atomic write can leave a truncated file for CI to upload.** `write_text()` on the final path directly means a disk-full or interrupted write could leave `run_results.json` truncated/invalid, and the documented `if: always()` step would upload it as-is. Fixed with a standard write-to-temp-sibling + atomic rename (`Path.replace()`) — the final path only ever holds a complete write or its previous (possibly absent) content; the temp file is cleaned up on failure too.

Added 3 new tests (quoted dict-repr credential redaction, comma-embedded secret redaction, atomic-write-no-partial-file) — 18 total. Full suite (4244 tests), ruff, mypy, `make check-drift` all pass.

Relaunching Codex review now.

masukai:
Codex review: round 6 found 1 more redaction bypass and 3 real bugs. Rather than patch the regex a 6th time, I changed the approach (9fc5a10) — flagging that shift explicitly since it changes this PR's documented contract, not just an implementation detail.

**The redaction regex has no fixed point.** Round 6's P1: compound credential identifiers (`client_secret=hunter2`, `refresh_token=...`, `AWS_SECRET_ACCESS_KEY=...`) sail through because `_` is a word character, so `\bsecret`/`\btoken`/`\baccess_key` can't start mid-identifier. That's five rounds, four of them finding a new way through the same regex (unquoted multi-word → quoted dict-repr keys → comma-embedded values → now compound identifiers). A keyword blacklist against arbitrary third-party exception text doesn't converge — there's always another shape.

**Structural fix instead:** stop trying to persist free text safely, don't persist it at all.
- A failed sync's raw `error` (`str(exception)`) is now **dropped from the artifact outright**. `error_type`/`error_stage`/`error_suggestion` — an exception class name, an enum value, a static rule-table string — are bounded vocabulary and stay; that's what CI triage actually needs. `error` itself stays exactly as it always was in `--output json` stdout and the console render (unredacted, unchanged from before #778) — those were never the file this project recommends uploading to CI storage.
- `diff.delete_preview_unavailable_reason` (same raw-exception shape) is replaced with a fixed placeholder in the artifact — not `null`, since `null` already means "the delete read succeeded" and the two outcomes must stay distinguishable. Kept unredacted in stdout.
- `diff.fallback_reason`'s one exception-derived case (`"Could not query destination (...): {e}"`) *is* safe to fix at the source, since existing tests only pin the type-name prefix, not the full message — so it's sanitized in `drt/engine/diff.py` itself (exception class name only) rather than needing the artifact-layer treatment. Its other three (static, always-benign) cases are untouched, so the common non-queryable-destination path loses nothing.
- `drt/_redaction.py`'s KV regex reverts to its original #698 shape. It's demoted to a best-effort, defense-in-depth pass on argv (outside the one option, `--vars`, redacted by whole-value replacement instead of by pattern) and stays doing its original job for the docs manifest's hosted-artifact redaction — it's just not what `run_results.json` relies on for safety anymore.

**Two more real bugs from the same round, fixed the direct way:**
- **P2 — pre-dispatch setup failures escaped `_run_one` uncaught.** A permission-check/destination/watermark/observer-construction failure skipped that sync's entry entirely (and, in a sequential dispatch, silently skipped every sync after it too), while the artifact still recorded `exit_code: 1` with no structured reason. `_run_one`'s `try` now wraps setup, not just `run_sync()` itself — same entry-building code handles both, since it never referenced anything setup-only. Note: this is a real behavior change for a registered `PermissionChecker` (ADR 0008) denial — previously it crashed the whole `drt run` invocation immediately; now it's recorded as a per-sync failure and the run continues to the next sync, same as any other failure mode.
- **P2 — non-atomic write under concurrent invocations.** The temp file used a fixed name, so two `drt run` processes sharing a `--target-path` could collide. Suffixed with the invocation's `run_id` (already a UUID4) instead.

**Not fixed, documented as an accepted limitation:** round 6's other P2 — a Click/Typer argument-parsing failure (`drt run --threads nope`) never reaches the command body at all, so the stale-artifact-clearing from round 4 doesn't run for it. A full fix would mean hooking into Click's own error path ahead of subcommand dispatch, a much larger and riskier change than this issue's scope — and narrower/lower-severity than the cases already covered, since the CI job's own console output already shows the parse error plainly (unlike the round-4 cases, which printed a clean, expected message and exited successfully).

Updated tests: removed 3 now-obsolete tests that asserted the reverted regex's exact behavior, added tests for the drop/placeholder split (stdout keeps raw text, artifact doesn't) and the setup-failure fix — 16 total. Full suite (4242 tests), ruff, mypy, `make check-drift` all pass.

Relaunching Codex review now.

masukai:
Codex review: round 7 clean.

> The run-results artifact implementation is consistently wired through success, failure, no-op, and early-exit paths, with appropriate sanitization and best-effort atomic writes. No actionable correctness regressions were identified in the changed code.

Ready for merge from my side. Summary of the arc: 7 review rounds total — rounds 1-3 fixed the dbt-collision default path, no-op-exit artifact gaps, and text-mode diff data; rounds 4-5 chased a heuristic redaction regex through several real bypasses; round 6 abandoned that approach for a structural fix (drop raw exception text from the artifact instead of trying to pattern-match every way a secret can be shaped) plus two unrelated real bugs (setup-failure swallowing, concurrent-write collision); round 7 confirms clean. One documented, accepted limitation remains: a Click/Typer argument-parsing failure doesn't get the stale-artifact cleanup, since the command body never runs in that case.

PR #1167, opened 2026-09-19
ci(dwh-smoke): surface the real HTTP error on warehouse-start failure

### LINKED ISSUE #1149: dwh-smoke: Databricks leg failing every nightly run since 2026-09-06 ("Failed to request warehouse start")
## Summary

The nightly `dwh-smoke` workflow's Databricks job has failed every single run for 9 consecutive days (2026-09-06 through 2026-09-14, checked as of this filing) — Snowflake and BigQuery legs are unaffected and passing in the same runs.

## Evidence

```
$ gh run list --repo drt-hub/drt --workflow dwh-smoke.yml --limit 20
2026-09-14  databricks=failure
2026-09-13  databricks=failure
2026-09-12  databricks=failure
2026-09-11  databricks=failure
2026-09-10  databricks=failure
2026-09-09  databricks=failure
2026-09-08  databricks=failure
2026-09-07  databricks=failure
2026-09-06  databricks=failure
2026-09-05  databricks=success
2026-09-04  databricks=success
2026-09-03  databricks=success
2026-09-02  databricks=success
```

Clean pass/fail boundary at 2026-09-06 (two earlier isolated failures on 2026-09-01 self-resolved the same day, unrelated pattern).

Failure output (consistent across all 9 failing runs, e.g. [run 34822595505](https://github.com/drt-hub/drt/actions/runs/34822595505/job/103907341356)):

```
##[error]Failed to request warehouse start
##[error]Process completed with exit code 1.
```

## Analysis

This fails at the **infrastructure level** — the workflow's own warehouse-start step (calling the Databricks SQL Warehouses API to bring the smoke warehouse online before running any tests) can't even get the warehouse to start. This is not a code regression in `drt` itself; the failure happens before any drt code runs against the warehouse.

Plausible causes (unconfirmed, needs the Databricks workspace owner to check):
- Workspace-level quota/capacity limit reached
- Billing/subscription issue suspending compute
- `DRT_SMOKE_DATABRICKS_*` credentials expired or rotated without updating the CI secret
- The specific SQL Warehouse referenced by the workflow was deleted/renamed

## Timing note

2026-09-06 is close to (a day or two after) PR #1130 (`feat(state): batch Databricks DlqBackend append()/reconcile() writes`, merged 2026-09-12) and the earlier #1108 Databricks warehouse-state-backend PRs (#1122/#1123, merged 2026-09-09) — but those merge dates don't line up cleanly with the 09-06 failure start, and the error is an infra-level warehouse-start failure, not a test/assertion failure, so a code regression in those PRs is unlikely to be the cause. Worth double-checking once the warehouse itself is confirmed reachable again, but not the first thing to suspect.

## Not fixed here

This needs access to the Databricks workspace/billing console to diagnose, which is outside what this session can act on. Filed for tracking so the failure doesn't keep going unnoticed — flagged by the repo owner as "concerning but no time to address right now."

masukai:
## Summary
- #1149's Databricks smoke job has failed every nightly run since 2026-09-06 (12+ days as of this PR) with only `Failed to request warehouse start` in the log.
- Confirmed by reading a recent run's log: the warehouse is `STOPPED`, and the `POST .../start` request is rejected immediately (not a timeout — all 3 retry attempts fail within ~10s).
- Root cause is invisible in the current log because `curl -sf` swallows the response body on any non-2xx status — an expired token, a deleted/renamed warehouse, and a transient network blip all produce the exact same generic message.
- `curl_retry()` now captures HTTP status + response body on every failed attempt and logs it, so the actual API rejection reason appears in the run log itself.

This does **not** fix #1149 — the underlying cause is credential/infrastructure on the maintainer's own Databricks workspace (a token expiry is the most likely culprit given the clean pass/fail boundary at 2026-09-06), which isn't something visible or actionable from this repo. This just means the next failure (or a manual re-run after merging) will tell us what it actually is instead of requiring a manual repro.

## Test plan
- [x] YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Extracted bash block passes `shellcheck`
- [x] New `curl_retry()` logic verified locally against a live HTTP endpoint (404 correctly detected + logged with body; 200 correctly returns the body for `jq` to parse)
- [ ] Will only be fully exercised against the real Databricks API on the next scheduled/manual `dwh-smoke` run


masukai:
Codex review: clean (reviewed together with #1168, which stacks on this branch).

> The workflow change correctly captures and reports non-2xx response bodies without altering retry behavior.

PR #1168, opened 2026-09-19
fix(state): Databricks warehouse-backed DLQ MERGE rejects column aliases

masukai:
## Summary
Found while re-validating #1149 (dwh-smoke's Databricks leg, dark for 12 days due to a workspace-activation block the maintainer has since cleared): once the warehouse could actually start again, a genuinely different, pre-existing bug surfaced in `drt/state/warehouse_databricks.py`'s DLQ backend.

Databricks' `MERGE` grammar rejects a column-alias list directly on its `USING` source:
```
[COLUMN_ALIASES_NOT_ALLOWED] Column aliases are not allowed in MERGE. SQLSTATE: 42601
```
`_upsert_dlq_entries`/`_update_dlq_entries` built `MERGE INTO t USING (VALUES (...)) AS s(id, sync_name, ...)` — exactly this rejected shape. This has been broken since #1108 merged; the mocked unit suite couldn't catch it (same "mocks can't prove SQL validity" class as #908/#920), and dwh-smoke's Databricks leg going dark for an unrelated reason (#1149) meant it shipped and sat undetected against the real warehouse.

**Fix**: nest the aliased `VALUES` table one level deeper — `USING (SELECT * FROM (VALUES (...)) AS s(...)) AS s` — so the column-alias list sits on the *inner* derived table, and `MERGE`'s `USING` clause sees only a bare `(SELECT ...) AS s` with no alias list of its own.

Stacks on the CI diagnostic fix in #1167 (needed to have seen this error at all) — merge that one first, or alongside this one.

## Test plan
- [x] Updated the 3 unit-test assertions pinning `"USING (VALUES"` to the new `"USING (SELECT * FROM (VALUES"` shape — `tests/unit/test_state_warehouse_databricks.py`, 44/44 pass
- [x] Full suite (4248 tests), ruff, mypy all pass
- [x] **Verified live against the real Databricks warehouse**: triggered `dwh-smoke.yml` via `workflow_dispatch` on this branch — Databricks, Snowflake, and BigQuery legs all green (run https://github.com/drt-hub/drt/actions/runs/35430961908)


masukai:
Codex review: clean (reviewed together with #1167, which this branch stacks on).

> The Databricks MERGE source is safely wrapped to avoid the unsupported direct column-alias list while preserving parameters and column references.

Note: the reviewer's own `pytest tests/unit/test_state_warehouse_databricks.py` run reported a failure during this pass — that was caused by me switching git branches in the shared working directory while the review was still running its shell commands, not a real test failure. Re-ran independently right after on a clean checkout: 44/44 pass.

